PR #31, opened 2015-10-07
Add professor_lab map

treetrunk:
I added a map for the professor's lab. Tested it out and doors/collision seem to work. I also added placeholder events for obtaining your first creatures in the lab. They don't do anything yet though.


treetrunk:
Here is a screen shot of it in Tiled with the collision layer hidden:

Professor's Lab

PR #38, opened 2015-10-10
Add a healing center map and link it to map1

treetrunk:
I threw together a healing/tuxe-center map. The door is linked up and the other two events don't work yet. The healing event does show a dialog box, but it doesn't heal your tuxemon yet.


treetrunk:
Here's what it looks like:
healing_center

ShadowApex:
Looks great so far!

PR #39, opened 2015-10-11
Add boundaries, tuxemart map, fix graphics, and connect a few maps with teleports

treetrunk:
This PR only affects the route1, map1, and BuddhaMountain maps. I basically just tidied up the three maps so that the game is a little more pollished. There shouldn't be any way now for players to walk off the map or teleport somewhere that they can't escape from. 

### Changes Made:
* **map1**
 * Add borders so that you can't walk into the abyss
 * Fix graphical issues (e.g. add tops to signs and trees, fix fence posts)
 * Block right exit from northmost town
* **route1**
 * Add borders so that you can't walk into the abyss
 * Correct and add some collision fields
 * Fix graphical issues (add tops to trees)
 * Create teleport back to map1
* **BuddhaMountain**
 * Add borders so that you can't walk into the abyss
 * Correct some collision fields 
 * Create teleport back to map1


treetrunk:
This PR now also contains a Tuxemart map that is linked up to map1 through teleports. If you press return while at the location where you would normally be talking to the cashier, you currently receive 5 capture devices. I also created an items tileset and added dialogs to the tuxecenter and tuxemart signs.

treetrunk:
^ Last commit. I swear. It just removes the lines that separate the panes of the vertical counters.

ShadowApex:
These maps look great! The only problem I found with them is inside the TuxeMart:

treetrunk:
I'm glad you like them. Nice catch! I just pushed another commit to correct it.

treetrunk:
I figured out that I didn't need 6 layers to make the vertical counter in the tuxemart. So that last commit makes Tuxemart only have 3 tile layers + the above player layer.

Issue #41, opened 2015-10-12
Conditional collision zones


treetrunk:
Currently you can block areas off completely or not at all. I am working making enter and exit properties for collision zones so that you can specify that a tile is blocked unless entering from one or more directions, or exiting from one or more directions. So you could set a tile as a collision field, and then give it the property "enter" with the value "up left". That then would allow entry to the tile from above and left. This will make it possible to implement the ability to jump over ledges with the properties `enter: up` and `exit: down` if it is a south-facing ledge.

dtdannen:
Hi Nathan,

The way I coded up poly-line collisions uses directions. Right now any
vertical poly-line restricts left and right movement, and any horizontal
poly-line restricts up and down movement. So it should be easy to add a
condition field in map files (editable using Tiled) for a poly-line
restricting a single direction and then only generating a poly-line for
that direction. But that just allows movement, you would need to do
something different altogether to get a "jump". Also this is only with
polylines.

dtdannen:
Just in case that helps :)

treetrunk:
Hi,

Cool. Yeah I saw some comments about it in the code but I wasn't sure how far it was implemented yet. Looks pretty cool. I think I might add it for zones as well anyway though because I think it would be useful to set rules for larger groups of tiles rather than drawing individual lines for each side.

dtdannen:
Cool.

PR #42, opened 2015-10-13
Add conditional collision zones

treetrunk:
## Description
This PR closes issue #41. Conditional collision zones work and look just like normal collision zones. The only difference is that you add properties to them. Currently they only take 2 properties: `enter`, and `exit`. Each property can have a value of multiple directions separated by a space. The code only looks for the keyword "enter" or "exit" meaning you can name them enter_from and exit_to to make it easier to understand. 
## Changes Made
* Add logic for enter and exit properties on collision zones
* Implement the feature to make one-way ledges on map1 (note: does not include the jump animation)
* Implement the feature on the stairs in the healing center (can only enter and exit on the right)

## Example
If for some reason I want to have a zone that can only be entered from the tile above and only exited to the left or right, the properties in Tiled would look like this:

**Properties:**
* enter_from: 
 * up
* exit_to: 
 * left right

Note that this is per tile in the zone. So once you are in a conditional collision zone, these rules apply for each tile within it.

### DIFF
--- a/tuxemon/core/components/map.py
+++ b/tuxemon/core/components/map.py
@@ -397,20 +397,20 @@ class Map(object):
             # inside this region.
             for a in range(0, int(width)):
                 for b in range(0, int(height)):
-                  collision_tile = (a + x, b + y) 
-                  collision_map.add(collision_tile)
+                    collision_tile = (a + x, b + y) 
+                    collision_map.add(collision_tile)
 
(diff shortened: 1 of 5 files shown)


ShadowApex:
This is great! I'm thinking that we can make a "jump" event action to compliment this that can be placed on the map that will animate the player jumping and move them *n* tiles in one direction. That way map designers have the option of having both jumping ledges and multi-directional walls.

PR #44, opened 2015-10-14
Making keybindings configurable

ntduffy:
An attempt at making keybindings configurable instead of hardcoded. It works for the most part, except for commands executed through maps (i.e. examining objects) which are still hardcoded to the enter key. As long as the user keeps their action key bound to enter, this shouldn't affect gameplay. Otherwise, the enter key will still be a dedicated examine objects button for now.

### DIFF
--- a/tuxemon/core/components/config.py
+++ b/tuxemon/core/components/config.py
@@ -39,12 +39,12 @@ class Config(object):
         self.config = ConfigParser.ConfigParser()
         self.config.read(file)
 
-	self.key_up = self.config.get("keybindings", "up")
-	self.key_down = self.config.get("keybindings", "down")
-	self.key_left = self.config.get("keybindings", "left")
-	self.key_right = self.config.get("keybindings", "right")
-	self.key_action = self.config.get("keybindings", "action")
-	self.key_menu = self.config.get("keybindings", "menu")
+	self.key_up = eval('pygame.' + self.config.get("keybindings", "up"))
+	self.key_down = eval('pygame.' + self.config.get("keybindings", "down"))
+	self.key_left = eval('pygame.' + self.config.get("keybindings", "left"))
+	self.key_right = eval('pygame.' + self.config.get("keybindings", "right"))
+	self.key_action = eval('pygame.' + self.config.get("keybindings", "action"))
+	self.key_menu = eval('pygame.' + self.config.get("keybindings", "menu"))
         
         self.resolution_x = self.config.get("display", "resolution_x")
         self.resolution_y = self.config.get("display", "resolution_y")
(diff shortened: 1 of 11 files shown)


ShadowApex:
This is really awesome! I definitely think it should be possible to remap the controls any way you'd like. 

Unfortunately if you remapped your controls with this system, it would break gamepad (joystick) and network controller events. The controller overlay for the Android version would also break, but the user wouldn't really have a way to modify the `tuxemon.cfg` config file to actually do that.

Currently the way gamepad, network controllers, and the mobile overlay work is they take input they receive and translate them to `pygame.K_<blah>` events and pass them down to the current state. Take a look at the `joystick_event_loop`, `controller_event_loop`, and `network_event_loop` methods in the `core.tools` module to see how those are currently being handled.

To resolve this, I think there's two approaches that we could take:

1. In the `Control.get_event` method in `core.tools` we could read the configuration file and translate those key presses into the hard coded `pygame.K_RETURN`, `pygame.K_ESC`, etc. objects that the rest of the game states expect. Doing it this way would also resolve the key mapping problem you have for map events (i.e. examining objects), since all input events would come in as the assumed value. 

2. Modify the `joystick_event_loop`, `controller_event_loop`, and `network_event_loop` methods in the `core.tools` module to translate those input events to the keys that were defined in the configuration file. We would additionally need to convert the map events to look at the configuration file for key presses and convert the event to the ones defined in the config.

Thoughts?

ntduffy:
If we were to go with the first approach, would changing the configuration somehow unbind the original keys? Or if the user wanted to, for example, bind the action key to esc, would it also still be the menu key?

ShadowApex:
Whether or not the original keys are unbound depends on how it's implemented. It could be done either way. If the user binded the action key to esc, it wouldn't also be the menu key.

Here's an example of what I was thinking for the first approach:    
```python
if event.key == prepare.CONFIG.key_action:
    self.state.get_event(self.keyboard_events["KEYDOWN"]["enter"])
    continue
...
```

If the user binded `key_action` to "ESC", then the `continue` statement would ensure that only the `pygame.RETURN` event was passed down to the game state.

ntduffy:
If that's the case, then doing that would probably be easier, but I still can't help but feel like the second approach would be safer.

ShadowApex:
I think either approach is valid. We just need a way to pass abstract game events down to the current state (world, combat, etc.). These game events, like "move up" and "activate" could be a simple string or object that gets sent based on raw input events (like joystick events, keyboard events, etc.) or we could send `pygame.K_<blah>` events based on those raw input events. 

By doing this, we're essentially creating an "Input Manager" that translates various inputs (e.g. from keyboard events, joystick events, network events, etc.) into a standard format that the rest of the game understands. As long as we stick to a standard format, it doesn't really matter if we use custom strings, objects, or `pygame.K_<blah>` events.

bitcraft:
@ntduffy Do you plan on getting this change merged?  It would be a nice feature to add, I know I've heard a few requests to use different keys or input devices.  I think the Input Manager approach is valid.  I've also cleanup up some of the input abstractions, so maybe it will be easier to implement now?

PR #46, opened 2015-10-16
Fix bug that broke normal teleports after adding cond_collision

treetrunk:
I realized after the conditional collision zones were added that non-transition teleports weren't working. Looks like I just forgot to add the cond_collision_map variable in one spot. This should fix it though.


ShadowApex:
Good catch! Missed that in the previous PR.

PR #48, opened 2015-10-18
Rework conditional collision so that an extra variable does not need to be passed around

treetrunk:
I found a way to remove the cond_collision_map dictionary byt rewriting the collision_map/set variables. This basically does that and cleans up the code in doing so.


ShadowApex:
Looks good. Using a dictionary instead of a set maintains the uniqueness of the collision tiles.

PR #69, opened 2015-12-10
Update README.md

Airon90:
* Added link to Tuxemon app for Android (add  a badge when available on Google Play, F-Droid or other stores)
* Removed useless empty lines


Airon90:
Thank you!

Issue #72, opened 2015-12-13
Concurrently running States


bitcraft:
As multiplayer/networking was broken with the new state system, I'd like to open discussion for running concurrently operating states.  This would allow for multiplayer to work, and could help simplify menus and such, if that is going to be addressed in the future.

This isn't a regression into the old way of everything running at once, but it will allow for situations where things could run concurrently.  Here's my idea, please fill in comments about API names and such:

* The state at the top of the stack receives input from the player
* States that are still running in the background can be updated and/or rendered, but get no input
* States will have an attribute that lets them know if they are in focus or not
* Rendering will occur from end of the stack to the front, so that they are layered in a natural way

I'd also like to propose changing the following methods' signatures in the State from:
```python
def update(screen, keys, current_time, time_delta):
    ...

def draw():
   ...
```

to the following:
```python
def update(keys, time_delta):
    ...

def draw(surface):
   ...

```
Rationale being that it is better to separate draws and updates at the API level.  Current time is rarely used and doesn't need to be passed each time.  This will also simplify drawing for multiple concurrent states.  The keys argument should stay as it would be a big change, as opposed to incremental.  Eventually though, it would be nice to move into it's own method (imo).

Thoughts?

ShadowApex:
I think concurrently running states would be great. First use case I can think of is the "main menu" that's used in the `WORLD` state. The world will continue to render, but all input will be passed only to the menu. Doing it this way would solve a couple of bugs that still exist where you can still interact with things while some menus are open.

In regards to the state methods, I'm totally on board with removing the `screen` and `current_time`. Those were more legacy parts of the code that I haven't gotten around to removing. I'm not really sure what you're planning with the draw method. Would you maintain a list of surfaces to draw and loop through them or something?

bitcraft:
For drawing, it's very useful to just pass a single surface to be drawn on.
That way you can easily implement display scaling, effects, minimaps, etc.
each state just draws in order from bottom of the stack to the top, so the
world state gets drawn first, then the hud (for the future) then the menu,
if it exists. No need to maintain a list of surfaces.

bitcraft:
I'm currently reworking menus to operate as States.  So far, they are working in my development branch, but there are some bugs that need to be worked out yet.  So, I'd just like to make suggestion that others do not do work on the menus, until this issue is closed.  It will just make my work easier when I merge.  Thanks!

ShadowApex:
[...]

Issue #47, opened 2015-10-16
Implement map pre-loading for non-transition maps.


ShadowApex:
Currently all maps are loaded on-demand, meaning they will be loaded from disk every time a teleport to a new map occurs. This causes a jarring stutter when switching between maps without a transition, which is needed to tie multiple maps together to give the illusion of a single cohesive world.

To resolve this issue, we should pre-load all maps from "teleport" events asynchronously from disk so we can simply switch to them when a teleport occurs. In addition, we should keep a "previous" and "next" map loaded, to make it seamless to switch back and forth between two maps.

bitcraft:
This may be related, but I've fixed the 'fade in being skipped for large map loads' issue on one of my development branches.  The old map fades out, there will be a slightly longish delay, then the new map fades in correctly.  The async map loading may not be needed if the delay is acceptable as long as transistions still look correct.

ShadowApex:
Did some work on this in commit 1b1590b7f1457068c0620f98930810c8ab80ee37

PR #78, opened 2015-12-23
Fix transistion animations

bitcraft:
This PR covers a few transition related changes:

* These transitions are now reusable as states: fade out, battle
* Refactored some drawing/updating as-per issue #72 
* Fixed skipped fade transition described in issue #47 
* Added to repo, my own "animation.py" value tweening utility
* Some light PEP8 and linting as well

In addition, the following changes were made:
* Consolidated the two blocks of level loading code in WORLD: now is method "change_map"
* Removed a few now legacy params to various states
* Updated or removed some docstrings

In general, the game should feel more polished with the transitions working as expected now.  Fade outs are working without code duplication as you can use a new Transition State to implement it.  However, fading in as a state remains problematic as it requires a temp buffer of the next state to be rendered.  It can be addressed in the future, when drawing code is more cleanly implemented (#72).  Currently, it is working as it was before, as a function of the world state.


ShadowApex:
These are some excellent changes. Thanks for the PR!

PR #89, opened 2016-01-05
Start Menu and Save/Load functionality.

superman2k5:


ShadowApex:
I got this error when starting a new game and trying to save:

```python
  File "./tuxemon.py", line 64, in <module>
    main()
  File "/tmp/Tuxemon/tuxemon/core/main.py", line 56, in main
    control.main()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 438, in main
    self.main_loop()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 474, in main_loop
    self.event_loop()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 222, in event_loop
    self.current_state.get_event(event)
  File "/tmp/Tuxemon/tuxemon/core/states/world/__init__.py", line 521, in get_event
    self.save_menu.get_event(event)
  File "/tmp/Tuxemon/tuxemon/core/components/menu/save_menu.py", line 130, in get_event
    1)]
KeyError: 'slot1'
```

tuxemon/core/components/menu/save_menu.py
```diff
@@ -40,12 +40,12 @@ def draw(self):
             slot_name = "slot" + str(slot_num)
 
             # Check to see if a save exists for the current slot
-            if os.path.exists("saves/slot" + str(slot_num) + ".png"):
+            if os.path.exists(".saves/slot" + str(slot_num) + ".png"):
```
ShadowApex: Right now this is going to put save games under the current working directory. We should instead put these save games under the user's home directory. Here's a possible way we could handle this:

https://stackoverflow.com/questions/4028904/how-to-get-the-home-directory-in-python?answertab=votes#tab-top

tuxemon/core/components/menu/save_menu.py
```diff
@@ -119,18 +119,15 @@ def get_event(self, event):
             self.first_run = True
             self.game.main_menu.interactable = True
 
-
-        if event.type == pygame.KEYDOWN and event.key == pygame.K_RETURN and not self.first_run:
+        if event.type == pygame.KEYDOWN and event.key == pygame.K_RETURN:
```
ShadowApex: Probably change this to look for KEYUP events instead?

tuxemon/core/components/save.py
```diff
@@ -58,15 +60,43 @@ def save(player, slot, game):
...
+    try:
+            os.makedirs(path)
+    except OSError:
+            if not os.path.isdir(path):
+                        raise
     # Save a screenshot of the current frame
-    pygame.image.save(game.save_screenshot, 'saves/slot' + str(slot) + '.png')
-
-    saveFile = shelve.open('saves/slot' + str(slot) + '.save')
+    pygame.image.save(game.save_screenshot, '.saves/slot' + str(slot) + '.png')
```
ShadowApex: Instead of defining `'.saves/slot'` multiple times, we should add a variable to `prepare.py` called `SAVE_PATH`. That way if we decide to change the save path, we only need to do it in one place. Then here we can just do:

```python
pygame.image.save(game.save_screenshot, prepare.SAVE_PATH + str(slot) + '.png')
```

tuxemon/core/components/save.py
```diff
@@ -58,15 +60,43 @@ def save(player, slot, game):
...
-    saveFile['current_map'] = game.event_engine.current_map.filename
+    tempinv1 = {}
+    tempinv1 = game.player1.inventory
+    for keysinv, valuesinv in tempinv1.items():
+        for keys2inv, values2inv in valuesinv.items():
+            if keys2inv == 'item':
+                values2inv.surface = None
+    saveFile['inventory'] = tempinv1
+    saveFile['current_map'] = game.event_engine.current_map.filename.split("/")[-1]
+    tempmon1 = []
```
ShadowApex: This should be `tempmon1 = list(game.player1.monsters)` or something similar. The current way you have it here will actually clear the sprites from the `game.player1.monsters` list.

Here's an example of what you're doing right now. Anything you do to `mylist_2` will affect `mylist_1`:
```python
>>> mylist_1 = [1, 2, 3]
>>> mylist_2 = mylist_1
>>> mylist_2.pop()
3
>>> mylist_2
[1, 2]
>>> mylist_1
[1, 2]
```

Doing `mylist_2 = list(mylist_1)` will have this effect:
```python
>>> mylist_1 = [1, 2, 3]
>>> mylist_2 = list(mylist_1)
>>> mylist_2.pop()
3
>>> mylist_2
[1, 2]
>>> mylist_1
[1, 2, 3]
```

superman2k5: This explains my bug I had, which gets fixed with the load function. I can fix that.

tuxemon/core/components/save.py
superman2k5: Looks simple enough. I'll test it out.

tuxemon/core/components/menu/save_menu.py
superman2k5: Yep

tuxemon/core/components/menu/save_menu.py
superman2k5: Makes sense, this was part of my questions I had when I was asking about the folder in the first place.

bitcraft: For what it is worth, I have all the known input bugs sorted out on my
branch. It may not be worth the effort to fix them here.

bitcraft:
https://github.com/Tuxemon/Tuxemon/pull/89/files#diff-9fd4d71feefc93e7f96e05c43dd18e89L222

I see the change to pop_state here.  Its a needed change to pop other states, rather than just the top, but I think that a better change would be this one, that I have implemented on my branch (using a state object, rather than a number):

https://github.com/bitcraft/Tuxemon/blob/menu-state-refactor/tuxemon/core/state.py#L248

States shouldn't know the position of themselves or others on the stack, but having an option to pop a state directly is handy.  I'm currently using this on the main menu...if you go to main menu => tuxemon then exit tuxemon, you will drop back to the world state, instead of main menu (can be changed of course).  This is possible because the main menu will push tuxemon state, then pop itself.

https://github.com/bitcraft/Tuxemon/blob/menu-state-refactor/tuxemon/core/states/world/world_menus.py#L39-L41

superman2k5:
Hey Bitcraft,

First of all thank you for your comments and feedback. I've been asking the question when all the menu refactoring would be done but couldn't get a real timeline. I don't think it makes sense to work on something that is going to be replaced shortly after. I'd rather not create anymore states until what you're doing is complete. I've only been trying to pick easy things to get familiar with the code base and understand it while still providing some value even if it is small. I can change the pop_state function to match, that change was done on the fly, and didn't think it was a long term fix for the situation. 

Some input on the features you're adding: 
"All menu inputs are handled correctly, independent of the menu layout." 
Things like the swap that we do within the PC menu and the drawing of images in the save_menu require additional input. Are you already accounting for this? Have you looked into the storage swapping and save_menu?

bitcraft:
Thanks for your patience with this matter!  Its nice to work with a group that doesn't get upset really easily....been there done that.  In regards to the PC storage menu, I've reviewed all the changes, and I do like how the player interacts with both menus.  Its intuitive and simple.  That being said, I don't have a ready solution to handling it off the top of my head.  I do have a couple ideas, but I'm not totally happy with them yet:
* make a state of each side.  the left/right keys will 'fall through' each state, the correct state grabs it and handles the change to the opposite menu.
* capture left/right keys and then just move the cursor around as needed
* pull out some of the control handling code and make a generic "player input area" that you can add list widgets, selectable items, things like that.

I'm actually leaning toward the last option the most, as it means developers can construct generic interfaces where the controls are handled automatically by making some parent container widget, then adding widgets (like a clickable button, a list, etc) to it.  It also means that I probably wont get to that right away.

As far as a timeline goes, I'm thinking two weeks.  I realize that I'm probably being super imposing here, and I hope that I don't block anything.  There is the least appealing option of just cleaning up what I have, merging it, and leaving the menus broken, but at least then I wont be blocking anyone.

tuxemon/resources/maps/bedroom_test.tmx
```diff
@@ -81,7 +81,7 @@
     <property name="act3" value="dialog_chain Go forth and inact my lampy will."/>
     <property name="act4" value="dialog_chain ${{end}}"/>
     <property name="cond1" value="is player_facing_tile"/>
-    <property name="cond2" value="is button_pressed K_RETURN"/>
+    <property name="cond2" value="is button_released K_RETURN"/>
```
bitcraft: Dialogs should use "pressed" and not "released".  Using KEYUP/released events was a workaround events leaking between states.  Since this particular issue is fixed, these events should use KEYDOWN/pressed, as it provides instant pop up dialogs.

Issue #22, opened 2015-08-19
Add free software mascot-inspired monsters


pizzamaker:
For ideas, check out: https://commons.wikimedia.org/wiki/Category:Free_software_mascots

ShadowApex:
To be more specific about this issue, we'd like to include monsters *inspired* by free software mascots instead of being directly based on them.

Sanglorian:
We have a few in this category now: Foxfire, Lambert and Mystikapi. So I think we can close this issue even though we don't have a Tux monster yet.

PR #94, opened 2016-01-27
Add Tux to game.

hyperdriveguy:
Partially addresses issue #22.


ShadowApex:
Thanks for this PR! While that issue was created to include free software mascots, we'd like to eventually have creatures who are *inspired* by free software mascots instead of being directly based on them (I'll update that issue). I'll still accept this request, but we'll probably want to tweak its design and name in the future.

I also noticed that the monster's name is still "Template" in its JSON file. If you change the name I'll be able to accept this PR.

Thanks again!

hyperdriveguy:
Fixed it! Also thank you for the information, very helpful.

PR #96, opened 2016-02-23
german translations

tamashihoshi:
Here are the very first german translations for the en_US.json file


tamashihoshi:
Nevermind! I'll make a bigger update..

PR #102, opened 2016-03-18
Translation

HugopGhost:
French translation added and German translation improved.


tamashihoshi:
Thank you alot for the translation files! Also prepare yourself for a long comment to read ^^'

- - - - -

About the german translation update:
Why did you choose "Zeitschrift" for "journal"?
"Zeitschrift" is more like a magazine while "Journal" is more like a log... I don't think "Zeitschrift" is a good translation...

since i can express myself a little better in german, i'll try to explain the translation a little more precisely:
Eine **Zeitschrift** ist eher ein Printmedium, ein Magazin oder eine Broschüre, in welchem herausgeber artikel veröffentlichen
**Journal** ist ein alter begriff für zeitschrift, allerings versteht man darunter auch eine art **Tagebuch**. Und genau um so ein Journal geht es in diesem Fall.. wie eine art questlog oder so, wo notizen und informationen für den spieler stehen.
Vergleich:
https://de.wikipedia.org/wiki/Zeitschrift
https://de.wikipedia.org/wiki/Journal_%28Buch%29

Short english translation of the text above:
Zeitschrift is a printmedium like a magazine or brochure in which publishers publish text (wow so much translation)
Journal is an old world for zeitschrift, but its main meaning is a journal like a diary. in this case, the journal is like a diary, like a questlog or something, in which notes and informations can be found for the player.

- - - - -

About comments:
you can not write comments in json ... i actually looked that up as well. there are a few ways to get around though
http://stackoverflow.com/questions/244777/can-i-use-comments-inside-a-json-file
IMO we shouldnt use comments... even if we push some fake-comments in, its just weird and makes the file ugly

- - - - -

about "KP" instead of "HP":
Was it really "KP" (kraftpunkte?) or "LP" (lebenspunkte?), im not too sure about that.. if you play pokemon, feel free to change HP into KP/LP

- - - - -

and now something **important**:

You actually made a pull request to the main branch of tuxemon, the translation files are on the delevopment branch though. The translations will be added to the main branch when the translator object is finished.. until then, please add all internationalization (i hate that word.) files to the development branch of tuxemon

so yeah.. thanks again very much, please do another pull request with the changes i mentioned (no comments, change HP into something better, let journal stay journal) into tuxemon/development

tamashihoshi:
correction:
ANY additions go to development branch, the main branch gets updated once in a while when we reach a new version

HugopGhost:
As you mentioned, journal is an old word. The first time I knew about it was while watching a documentation movie about Anne Frank, I didn't knew back then it is a valid German word and this is why I changed it. Maybe Notizbuch sounds better?

I didn't knew that JSON doesn't accept comments, but I am not sure if we should change the term "HP" or not because it would need to be changed everywhere, this is why I made the comments. Both KP and LP are valid, but I see KP more often.

Alright, going to add the commits in the right section.

HugopGhost:
https://github.com/Tuxemon/Tuxemon/pull/103

PR #105, opened 2016-03-20
fixed a typo

tamashihoshi:


tamashihoshi:
"all checks have failed" <- why? ._.

ShadowApex:
The checks are from TravisCI, which isn't really set up at the moment. I'll need to change the way TravisCI runs to not throw any errors.

PR #125, opened 2016-05-07
Moved add_item function.

r1g1l:
I Moved the add_item function from **_core.components.events.actions.player_**
**_to core.components.player_** so Npc's can be given items as well.


r1g1l:
My bad, I forgot to remove the 'game' parameter from the doc string, It's not necessary anymore.

r1g1l:
Found a bug Removed so i could fix will replace soon

PR #127, opened 2016-05-08
Moved add_item function

r1g1l:


tuxemon/core/components/player.py
```diff
@@ -698,6 +698,39 @@ def get_adjacent_tiles(self, curr_loc, game):
...
+                "Potion"
+            ]
+        }
+
+        """
+        item_to_add = item.Item(action.parameters[0])
+
+        # If the item already exists in the player's inventory, add to its quantity, otherwise
+        # just add the item.
+        if item_to_add.name in [i.name for i in self.inventory.keys()]:
```
ShadowApex: Hi,

This PR looks good, but I had a question about the `add_item` logic that you used here.

Was there any reason you used a list comprehension here? To check for a key in a dictionary in python you can simply use `if item_to_add.name in self.inventory`. Using a list comprehension adds a little unnecessary complexity.

tuxemon/core/components/event/actions/player.py
```diff
@@ -254,41 +254,6 @@ def add_monster(self, game, action):
         game.player1.add_monster(current_monster)
 
 
-    def add_item(self, game, action):
```
ShadowApex: We'd actually still need an `add_item` method here in `core.components.event.actions.player` so that you'd have the ability to add items to the player's inventory from an event defined in a map file. Everything in `core.components.event` are functions that you can use in the [Tiled map editor](http://mapeditor.org) for in-game events. The actual logic for adding items to a player's inventory should definitely be moved into the `core.components.player.Player` class though. Once the logic has been moved, this method could just call the player's `add_item` method like this:

```python
player = game.player1
player.add_item(item_to_add)
```

tuxemon/core/components/player.py
r1g1l: No that was a leftover from an earlier idea, sorry i forgot to take it out.

tuxemon/core/components/player.py
```diff
@@ -698,6 +698,39 @@ def get_adjacent_tiles(self, curr_loc, game):
...
+            self.inventory[[i for i in self.inventory.keys() if i.name == item_to_add.name][0]]['quantity'] += 1
```
r1g1l: I removed the pull request because of this line though. It will crash if the condition is never met. i.e. If there is no item by that name in the inventory.
This shouldn't be a problem since that should be impossible,
but i felt i should find a better fix withouth the possible crash.

PR #131, opened 2016-05-10
Moved add_item function.

r1g1l:
Moved add_item function logic login to _core.components.player.Player_.

### DIFF
--- a/tuxemon/core/components/event/actions/player.py
+++ b/tuxemon/core/components/event/actions/player.py
@@ -321,7 +321,7 @@ class Player(object):
         :rtype: None
         :returns: None
 
-        Action Parameter must be: create_item
+        Valid Parameters: None
 
         """
         # Get a copy of the world state.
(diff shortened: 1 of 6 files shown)


tuxemon/core/components/event/actions/player.py
```diff
@@ -253,42 +253,19 @@ def add_monster(self, game, action):
...
         game.player1.add_monster(current_monster)
 
-
-    def add_item(self, game, action):
-        """Adds an item to the current player's inventory. The action parameter must contain an
-        item name to look up in the item database.
-
-        :param game: The main game object that contains all the game's variables.
-        :param action: The action (tuple) retrieved from the database that contains the action's
+    def add_item(self,action):
```
ShadowApex: This method still needs the `game` argument. All event methods take `game, action` from the Event Engine defined in `core.components.event`.

tuxemon/core/components/player.py
```diff
@@ -698,6 +698,39 @@ def get_adjacent_tiles(self, curr_loc, game):
             adj_tiles.append((curr_loc[0]+1,curr_loc[1]))
         return adj_tiles
 
+    def add_item(self, action):
```
ShadowApex: Instead of passing `action` as an argument, we should just need to pass the "item to add" string. That way we can call `add_item` on a player/npc object from anywhere without having to build an action object to pass to this method.

ShadowApex:
As a general comment, you don't need to keep closing and re-opening pull requests. If you simply commit to your fork of the project, the changes will show up here.

ShadowApex:
I was reviewing this request to merge but found a few problems with it, mostly due to the fixes being applied to the `master` branch of your fork instead of the `development` branch. If you apply your changes to the development branch instead and re-open this request, I think everything will look good.

PR #144, opened 2016-06-21
Portuguese translation

HugopGhost:
Added Portuguese translation, should be the same for Portuguese from Brazil and Portuguese from Portugal.


ShadowApex:
Thanks for this PR!

PR #147, opened 2016-06-23
Damage Tracking

buddja:
Added simple damage tracking to the Combat state, as well as distributing exp after a monster faints.


bitcraft:
Looks good to me!

buddja:
No problem I've been lurking and playing around for a while and have a few commits i will push over as i test them.

Issue #145, opened 2016-06-22
level_range ignored for encounters


MirceaKitsune:
Levels are ignored for random battles. Every creature you encounter in the wild is level 1, and the level_range field specified in resources/db/encounter/#.json is ignored. Latest GIT development.

ShadowApex:
This was fixed in buddja's PR here: https://github.com/Tuxemon/Tuxemon/pull/148/files#diff-5675550d26d36370a371bb71934664beL258

The random encounter now uses the defined level range in the encounter data file.

Issue #132, opened 2016-05-10
MainCombatMenuState behavior


ghost:
How is the 'MainCombatMenuState' supposed to be used? 

I was trying to use multiple items in a row,(Potion,Capture Device,Potion) in that order if it matters. and i noticed that there were 2 instances of 'MainCombatMenuState' in game.active_states. Is this supposed to happen?

Here's the image i took when i found it
http://imgur.com/a/7aNnG

Their are 2 possible ways i can imagine it working.
1. The Combat menu state gets GC'd when you leave it the first time 
      --thus the need to create a new instance of it if when called again
2. The Combat menu state stays around the whole length of the battle.
      --certian functions combat modules need to account for this

I'm just trying to understand whether my problem is the state is not being GC'd correctly or if my problem is something unnecessarily creating and adding the state.

bitcraft:
I've noticed that states do not get GC'd, and its a problem that I've put off for a while.  The current code base shares many references to things and its just complicated to get through all of it.  Someday...we'll get that particular problem taken care of.

Not being GC'd will not cause multiple instances of a particular state, however.  I haven
t noticed the multiple copies of MainCombatMenuState.  If you link your changes, I can take a look.  The order that items are displayed in a menu is controlled somewhere else entirely.

The concept is that each menu item is a sprite, and a menu can have a group that you add items to.  The actual class to layout where the items are is linked below

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/sprite.py#L314-L453

I don't actually understand what you are trying to do.  Is there a problem with how items are laid out?

r1g1l:
The problem i'm trying to fix now is here: 
https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/states/items/__init__.py#L58-L60 

I changed the line from **_ return self.game.active_states[-1].name**   to
**_ return self.game.active_states[2].name_** ,because StartState never gets GC'd (atleast on my build it doesn't), so it is always the last item in self.game.active_states and therefor i couldn't use items the way it was.

But this function fails when there is an extra state because it returns one of the MainCombatMenuStates instead of CombatState like it should.

bitcraft:
Yes, as the comment states, it is a bit of a hack.  The controller/state system still has some rough edges.  Please don't rely on that particular function to be there.  Actually, I'd prefer that you don't change it until we can determine if the duplicated state is a result of your changes or not.

[...]

Issue #129, opened 2016-05-09
Default targets for items and techniques


bitcraft:
Items and techniques will need a new attribute called "targeting".  This attribute will be defined in the json for each.  This value will be read and used when the item/technique is being used.  This value will choose a default target.  For example "combat" targeting will default to an enemy monster, and "healing" targeting will default to an injured monster on your team.  The player will have the option of targeting an monster, this will just provide a default.

Here is a proposal for targeting types, feel free to add more:

- combat: enemy monster
- healing: injured monster on your team

tamashihoshi:
is the targetting for items used while running around on the worldmap or during combat? or both?

next to combat/offensive items we dont just have healing items but also buffs, right? (moar str for the battle after eating a lorem-tuxemon-berry or something)
i think the player should be able to use items like that on any monster; the default targetting should be the first monster (not the first injured or something)

also there are items which are used on the player.. like some stinking-ipsum-lotion to keep wild tuxemon away for .. like.. 100 steps or so. i think in pokemon it was some REPEL item if im not mistaken. default target here: player aka "noone visible"
these shouldnt be usuable during a battle i guess

next we have items, which are not usuable at all. like quest items. a keycard for some random high-tech door should have no target (not even the player); these can be triggered from map events...

maaaaaaaaaaaaaybe we could have items which target items. you could combine stuff to create new. nothing big of course, i wouldnt do a whole big crafting system. but combining stuff for story would be fun maybe? finding 3 pieces of an item by fighting 3 different bosses in no particular order, combine them to the key item and enter the big door. maybe small potion + small potion could be a mediocre potion.
what do you think of that?

so possible targeting types could be:
- combat: enemy monster
- healing: injured monster on your team
- buff: first monster on your team (or: any monster on your team)
- player: used on player
- none: unusable
- (item: used with another item)

bitcraft:
> is the targetting for items used while running around on the worldmap or during combat? or both? 

both.  it would be integrated into menus and combat.

> next to combat/offensive items we dont just have healing items but also buffs, right? (moar str for the battle after eating a lorem-tuxemon-berry or something) i think the player should be able to use items like that on any monster; the default targetting should be the first monster (not the first injured or something)

defaulting to the first monster on your team would be fine.  i suggested the first injured one as it is just a nice thing to do, but i'm not really pushing the idea.

[...]

PR #148, opened 2016-06-24
Item functionalities

buddja:
Fixed item functionality in both combat and world state, as well as issue #145.

### DIFF
--- a/tuxemon/core/states/combat/combat.py
+++ b/tuxemon/core/states/combat/combat.py
@@ -499,9 +499,8 @@ class CombatState(CombatAnimations):
 
         :returns:
         """
-        if technique is Technique:
-            technique.advance_round()
-            
+        technique.advance_round()
+
         result = technique.use(user, target)
 
         try:
--- a/tuxemon/core/states/combat/combat.py
+++ b/tuxemon/core/states/combat/combat.py
@@ -499,7 +499,9 @@ class CombatState(CombatAnimations):
 
         :returns:
         """
-        technique.advance_round()
+        if technique is Technique:
+            technique.advance_round()
+            
         result = technique.use(user, target)
 
         try:
--- a/tuxemon/core/components/event/actions/combat.py
+++ b/tuxemon/core/components/event/actions/combat.py
@@ -255,7 +255,7 @@ class Combat(object):
 
             # Set the monster's level
             current_monster.level = 1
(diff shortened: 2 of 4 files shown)


bitcraft:
Thanks for taking on this issue.  Everything looks good except the changes in combat.py.  The test you have for Techniques is going to fail all the time (the "is" operator), so it needs to be removed.  I would revert the changes to compat.py and add an empty method advance_round to the Item class.  Items and techniques should share the same interface, since it simplifies how they are used in the code (sorry if that is not clear or is omitted in the docstrings).  Eventually, multi-round use items can be implemented this way, too.  (I have actually done all these things locally, but you beat be to making a PR for it :)

I like that you added the dialogs after using items in the World State.  If you make those changes, I don't see a problem merging it.

In item.py in the Item class:

```python
def advance_round():
    pass
```

The changes in item.py line 59 can be simplified as follows:

```python
    def determine_state_called_from(self):
        # hackish way to just get this working...  may come back to this later, idk
        return self.game.active_states[1].name
```

buddja:
Made the suggested changes, I believe everything should be good.
If there are any more suggestion's i'll gladly make them until then i'll continue moving down the issue tracker.

tuxemon/core/states/combat/combat.py
```diff
@@ -499,7 +499,9 @@ def perform_action(self, user, technique, target=None):
 
         :returns:
         """
-        technique.advance_round()
+        if technique is Technique:
+            technique.advance_round()
+            
         result = technique.use(user, target)
```
bitcraft: Please remove this test.  :)

bitcraft:
Go ahead and revert combat.py and I'll merge it.  Thanks!

bitcraft:
I'm going to apologize here for a sec.  The complex way you were searching for the CombatState in the item screen was actually better than the simple one.  I tested it a few times and prefer your way.  Open a PR if you'd like to have it merged in.  Thanks again!

buddja:
I will put it in my next round of pull requests, probably sometime next week.

bitcraft:
Sounds great!  Just a FYI, I'm doing a bit of item work, so that the
capture device is useable.  The core issue has been establishing a way to
select targets for techniques and items.  #129 #132  I'm generally on
freenode #tuxemon if you'd like to try to coordinate things with chat.

PR #149, opened 2016-06-24
Missing "

HugopGhost:
Missing "


bitcraft:
Thank you!

ShadowApex:
Good catch. I did a rough check with Google Translate, but missed verifying the json syntax. Thanks again!

Issue #146, opened 2016-06-22
Player can walk one extra tile after random encounter is triggered


MirceaKitsune:
A minor issue but worth noting: When a battle triggered by a random encounter starts, you can still walk one extra tile before movement is blocked, during the animation preceding the battle screen.

Example: Let's say the player is at tile 10,15 and walking upward (toward 10,16). Once the player reaches tile 10,16, a random battle is triggered and the music changes while the screen flashes. However the player can walk one extra tile to say 10,17. Normally, the player should freeze on the spot at tile 10,16 before the actual battle screen appears.

buddja:
Just for the record I decided not to fix this right now. I removed it from my PR, see #150 for more info.

Airon90:
The problem is still present: I sometimes find a Tuxemon out of the grass :D
@buddja Could you  try again to fix this problem?

MirceaKitsune:
Problem still exists in the latest development branch. Wanted to bump this since it's one of the very obvious issues still remaining. Especially since if a tile near the edge of a grass patch starts a battle, and the player is walking from within that patch toward the outside, the battle will start with the player out of the grass altogether. Maybe there is an easy way to just freeze movement the moment a random battle has been triggered?

MirceaKitsune:
This still seems to happen in the latest development branch, we might want to reopen.

Truth be told, I'm not sure whether it might be a matter of perception at this point: The battle song and screen flash appear to be triggered the moment the player starts walking toward the triggering tile. The behavior most of us expect however is for the battle to be triggered the moment the player reached the tile and stopped walking, and of course they can no longer move again until the battle is over.

@ShadowApex Any thoughts on this? Is this an actual bug, if not is it an effect ordering issue, and if so should it be changed to work more as in my suggestion?

bitcraft:
I'm not sure what the exact issue is, but I've noticed that the timing can be off and the player will sometimes move an extra tile.

Qiangong2:
@bitcraft this has been fixed.

MirceaKitsune:
I believe I saw and can confirm the fix as well.

Issue #128, opened 2016-05-09
Combat actions not sorted


bitcraft:
This issue is about the order that events are executed during the combat phase.  Currently, they are executed by sorting all the actions during the round according to the speed of the monster that is executing them.  Below I've listed how I understand it should be....please feel free to comment and correct it.
- Running from combat will happen before any moves
- Items will happen next
- Healing will occur before attacks
- Everything else is sorted by speed of the monster

tamashihoshi:
I opened a thread in the forum to ask others:
https://forum.tuxemon.org/thread.php?pid=979#p979

IMO:
running -> speed of monster -> items -> healing

but maybe something variable is also possible, and the above order is the usual order..

bitcraft:
https://github.com/Tuxemon/Tuxemon/wiki/Combat-State-Overview

ShadowApex:
> So, I'm wondering about "Use items". Should trainers have a speed value? So that if both (or more) trainers decide to use an item, the faster trainer is first?

I think in the case of two trainers both using items, we can randomly select the order. In general I don't foresee use of any items that would give a player an advantage if they use it first. Items can only be used on your own monsters (with the exception of capture devices). This certainly could change in the future though, so we should take that into account to make it easier on ourselves if we decide to change it.

ShadowApex:
I think I've changed my mind on this actually. Trainer speed sounds good to add. That, or we take the monster's speed into account for using items.

tamashihoshi:
> That, or we take the monster's speed into account for using items.
Would it be weird if we say: if a tuxemon is slow, you can use items faster on it?
(slow tuxemon doesnt move that much, it might be easier for the trainer to give the potion etc to the tuxemon)

bitcraft:
Trainer speed seems like the best solution to me.

xirsoi:
I disagree. Up to this point trainers have been the simplest of actors, adding attributes to them doesn't really make sense. Especially since adding ways to affect those attributes doesn't make sense.

I'd argue that item use should have a fixed speed (or maybe even a per-item speed attr, but that's more complicated than necessary) and be treated just like an attack action. Very fast tuxemon may be able to get an ability off before the item can be used. This keeps things simple, and simple is good where complexity adds nothing to the game play.

bitcraft:
[...]

PR #150, opened 2016-06-24
It's been slow at work today

buddja:
#146, #128, and possibly #132 but I'm not sure what that one is asking.

### DIFF
--- a/tuxemon/core/components/event/actions/combat.py
+++ b/tuxemon/core/components/event/actions/combat.py
@@ -237,8 +237,6 @@ class Combat(object):
         # If a random encounter was successfully rolled, look up the monster and start the
         # battle.
         if encounter:
-            #stop player motion
-            player1.moving = False
 
             logger.info("Start battle!")
 
--- a/tuxemon/core/components/event/actions/combat.py
+++ b/tuxemon/core/components/event/actions/combat.py
@@ -213,9 +213,6 @@ class Combat(object):
         if not self.check_battle_legal(player1):
             return False
 
-        #stop player motion
-        player1.moving = False
-        
         # Get the parameters to determine what encounter group we'll look up in the database.
         encounter_id = int(action.parameters[0])
 
@@ -240,6 +237,9 @@ class Combat(object):
         # If a random encounter was successfully rolled, look up the monster and start the
         # battle.
         if encounter:
+            #stop player motion
+            player1.moving = False
+
             logger.info("Start battle!")
 
             # Stop movement and keypress on the server.
--- a/tuxemon/core/components/event/actions/combat.py
+++ b/tuxemon/core/components/event/actions/combat.py
@@ -213,6 +213,9 @@ class Combat(object):
         if not self.check_battle_legal(player1):
             return False
 
+        #stop player motion
+        player1.moving = False
+        
         # Get the parameters to determine what encounter group we'll look up in the database.
         encounter_id = int(action.parameters[0])
 
@@ -296,4 +299,3 @@ class Combat(object):
                 logger.warning("Cannot start battle, player's monsters are all DEAD")
                 return False
             else: return True
-
--- a/tuxemon/core/states/combat/combat.py
+++ b/tuxemon/core/states/combat/combat.py
@@ -348,9 +348,9 @@ class CombatState(CombatAnimations):
             monster = menuitem.game_object
             import bpython;bpython.embed(locals_=locals())
             if monster.current_hp == 0:
-                tools.open_dialog(self.game, ["Cannot choose because is fainted"])
+                tools.open_dialog(self.game, [trans("combat_fainted", parameters={"name":monster.name})])
             elif monster in self.active_monsters:
-                tools.open_dialog(self.game, ["Monster is already in play"])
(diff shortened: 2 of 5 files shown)


bitcraft:
Hahaha, slow day for me as well.  What is this line about bpython?

https://github.com/Tuxemon/Tuxemon/pull/150/files#diff-cc5c9d7880b63a288d60b3dcd89e3802R349

tuxemon/core/components/event/actions/combat.py
```diff
@@ -213,6 +213,9 @@ def random_encounter(self, game, action):
         if not self.check_battle_legal(player1):
             return False
 
+        #stop player motion
+        player1.moving = False
```
ShadowApex: The problem with simply setting the player moving to `False` is that it will stop the player mid-step between tiles. After a few encounters the player will start to become offset from the tiles like this:

buddja:
Oh its how i debug...it embeds the [bpython] (https://github.com/bpython/bpython) interpreter. It should have been removed already.

tuxemon/core/states/combat/combat_menus.py
```diff
@@ -67,6 +67,11 @@ def on_menu_selection(self, menuitem):
     def open_swap_menu(self):
         def swap_it(menuitem):
             monster = menuitem.game_object
+            if monster in self.game.get_state_name('CombatState').active_monsters:
+                tools.open_dialog(self.game, ["Monster is already in play"])
```
ShadowApex: Instead of hardcoding English dialogs, we should ideally create an entry in [resources/db/locale/en_US.json](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/resources/db/locale/en_US.json) and use the `translate` function like used here:

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/states/persistance/save_menu.py#L90

bitcraft:
Thanks for the reminder about bare, untranslateable strings, @ShadowApex.

tuxemon/core/components/event/actions/combat.py
buddja: So could we set player.moving to False then round the current player.tile_pos back to were it should be?

ShadowApex: That could work for now, but it might cause a slightly jarring snapping of the character into place on the tile when we do the rounding. If we do the rounding just before the world state comes back after combat, it might remove that jarring effect. 

The snapping back effect would happen because, when you're holding down an arrow key, your character will keep moving past the tile to make the movement appear fluid.

bitcraft: I think the core issue should be addressed....that it seems the order between player movement, keypresses, and event handling could be rearranged to prevent anomalies like this.  I'm sure there are other quirks related to it, but they are not obvious like this.

tuxemon/core/states/combat/combat.py
```diff
@@ -694,6 +710,9 @@ def end_combat(self):
         fadeout_action.parameters = [1000]
         event_engine.actions["fadeout_music"]["method"](self.game, fadeout_action)
 
+        #Round player movement in case battle started in the middle of a step
+        self.game.player1.tile_pos = (round(self.game.player1.tile_pos[0]),round(self.game.player1.tile_pos[1]))
```
ShadowApex: I was doing some testing and still found the player being offset from the tile grid when getting into random battles because of this.

It looks like you actually need to round the player's pixel position (instead of tile position) to the nearest multiple of the base tile size. The reason you have to do this is because the [player's tile position is actually derived from the player's pixel position](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/states/world/worldstate.py#L297) (thus it gets overwritten every frame).

We might need to do something like this instead here:
```python
self.game.player1.position[0] = int(prepare.TILE_SIZE[0] * round(float(self.game.player1.position[0])/prepare.TILE_SIZE[0]))
self.game.player1.position[1] = int(prepare.TILE_SIZE[1] * round(float(self.game.player1.position[1])/prepare.TILE_SIZE[1]))
```

I have this same calculation in `core.components.map` called ["round_to_divisible"](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/map.py#L570) that is used there a few times. We might be able to move that function into a more generic utility module if it's used in more places.

buddja: Alright if its ok with you i'll remove it from the PR. I think if that's the case we should probably wait to decide were it goes once we see what all needs to call it right?

ShadowApex: Yes, that's fine. I mainly just want to prevent fixing one bug while introducing another.

ShadowApex: After this part is changed, I should be able to accept the PR. :)

tuxemon/core/components/event/actions/combat.py
```diff
@@ -237,6 +237,9 @@ def random_encounter(self, game, action):
         # If a random encounter was successfully rolled, look up the monster and start the
         # battle.
         if encounter:
+            #stop player motion
+            player1.moving = False
```
ShadowApex: This line should be removed too if we're going to address #146 later. Otherwise the player will still become offset from the tile grid.

PR #151, opened 2016-06-28
Update fr_FR.json

druz320:


ShadowApex:
Thanks for the updated translation! I double checked these in Google Translate for a rough review of it.

HugopGhost:
The translation is great, he fixed some of my mistakes and added original Pokemon dialogue, I just translated the English one into French.

PR #152, opened 2016-06-28
Combat Targeting and Item Changes

bitcraft:
New Features:
* New combat targeting feature
* Techinques and items can be targeted during combat

Library Changes:
* New syntax for json db techniques and items
* New Menu API: can add items w/o the generator interface
* Menu items can be disabled: will be shown, but not selectable
* Menus can be fully transparent

Bugfix/Misc:
* Items now respect the combat order
* During combat, items can be used on monsters that are out of combat
* Monsters can be swapped in after one is fainted
* Fixed CTD when choosing empty monster slot
* Fixed issues when swapping monsters after one faints
* Fixed experience bug when monsters are fainted
* New test, fix of older one
* General cleanup of some menu code
* Docstrings and small comments, linting

Targeting works as expected but it is currently enabled just for techniques.  The targeting syntax is currently ignored, so the initial target is random (b/c based on dict).


bitcraft:
This was not merged.

PR #162, opened 2016-07-12
Feature/#158 change player attribute

xirsoi:
I think these three methods should cover all four of the originally proposed methods. I created one method for modifying by percentages because setting to 85% is the same as reducing by 15%.

This simplifies the events API a bit.

I didn't include any type checking in the set_player_attribute method intentionally, since it seemed perfectly valid to alter non-integers (such as the name) in script. If that's wrong, I'm perfectly willing to change it, and anything else.


ShadowApex:
These are great changes. Thanks for the PR!

Here's some comments based on my review of these changes:

* Right now it looks like you're only able to change attributes for NPCs on the map and not the player itself. We should either include the ability to change the player's attributes in this action or move this to `npc.py` and change the name to `set_npc_attribute` to ultimately have both the `set_npc_attribute` and `set_player_attribute` actions. Right now the current design pattern is leaning more towards the latter.
* The current action names: `set_`, `change_`, and `modify_` seem pretty ambiguous. I feel like I would need to reference the documentation every time I use these actions because they are named so similarly. I think we should try and come up with a better naming convention for these actions to make them more specific. Maybe something like `set_`, `add_to_`, and `multiply_by_`? I'm open to other names.

Other than that, the pull request looks great. Thanks again!

bitcraft:
I had proposed "set" to mean "set the attribute to exactly this value", and "modify" to mean "add this value to the existing value, while allowing a negative number (so it is subtraction as well)".  Its how it is done in Fallout 3/4, FWIW.

The "_percent" methods are useful because you could define a healing item/technique as "recover %50 hp", and you could simply do "modify_npc_attribute_percent(npc, "hp", .5)" and get the expected result.

I'm not set on those, but thought I should clarify.  Nice PR, its great to have some code coming in.

tuxemon/core/components/event/actions/player.py
```diff
@@ -359,3 +359,157 @@ def player_resume(self, game, action):
...
+                player = npc
+
+        if not player:
+            return
+
+        attribute = action.parameters[1]
+        value = action.parameters[2]
+        
+        # check for valid inputs
+        # trigger an ArgumentException if the property doesn't already exist
```
bitcraft: s/ArgumentException/AttributeError
s/property/attribute

xirsoi:
Thanks for your comments! I have to run to work, i'll have more replies during my lunch break.

tuxemon/core/components/event/actions/player.py
xirsoi: Haha, that's what happens when you spend all day working in C#.

xirsoi:
@ShadowApex NPC.remove_npc() includes the same loop and also checks 'not npc.isplayer', so I assumed that Player1 is in that list. Is there a better way to have a generic method to affect players and npcs? I'm naturally averse to useless repetition (DRY), so I'd like to avoid having two nearly identical methods. At the very least, lets make a single method that takes a player object, attribute, and value that the npc and player actions can call. This should reduce any maintenance headaches.

I really don't think the names are that ambiguous, but then again I wrote them. I'll do some thinking and try to come up with clear, concise names for the functions. 

@bitcraft If you want to give them +50% of their health, do modify_player_attribute "Gregory" "hp" "1.5"

I'll do some more thinking about these, thanks again for your comments!

bitcraft:
I think I'm a bit caught up on the name "modify" because it doesn't seem to be related to multiplication.  I agree with @ShadowApex...maybe we should rethink the names of these functions.  Other than that, the code looks great.

ShadowApex:
> @ShadowApex...maybe we should rethink the names of these functions.

We talked a little about this on IRC. Here's what I'm thinking/proposing we'd name these actions:    
* `set_player_attribute`: Sets the player's attribute to a set value.
* `modify_player_attribute`: Adds to the player's current attribute value. If "%" is found in the parameter string, the current attribute value is multiplied by the parameter instead.

I think this reduces the number of actions we'd need and makes it a little easier to tell the difference between the two actions, rather than having to try and figure out the difference between `change_` and `modify_`. Thoughts?

> NPC.remove_npc() includes the same loop and also checks 'not npc.isplayer', so I assumed that Player1 is in that list. Is there a better way to have a generic method to affect players and npcs? I'm naturally averse to useless repetition (DRY), so I'd like to avoid having two nearly identical methods.

The `npc.isplayer` property is actually used for multiplayer (multiplayer players are just NPC objects with the `isplayer` property set to `True`.

I totally agree that we should minimize duplication. I'd recommend that we create a shared function that could modify both NPCs and the player, then have two separate actions depending on which kind player you want to modify. Right now I think it's useful for map makers to have a clear separation of actions that affect the player vs NPCs. 

It might be useful if we created a `get_all_players()` method in the world state that would return a list (or dictionary) of all NPCs and `player1`. It might also be useful for us to change the `world.npcs` list to a dictionary so we can simply get the NPC object reference by name if we need to. In either case, we need to make sure that `player1`'s name doesn't collide with an NPC with the same name.

xirsoi:
I'll make those changes soon, but it might not happen until Friday evening (CST), the next two days are busy for me.

I'm going to create two new issues for the things in the last paragraph, I believe one of them requires further discussion.

xirsoi:
@ShadowApex where would be the best place to put this shared third function?

ShadowApex:
> @ShadowApex where would be the best place to put this shared third function?

I'm thinking we could add it as `core.components.event.actions.common`, since it will only be used by action plugins.

xirsoi:
Sweet, I will make some time to get this updated over the weekend. I've been itching to work on Tuxemon all week, but Life Finds A Way (to get in my way :( )

bitcraft:
Everything except event system should be using slugs now.  No plans to migrate the events to slugs ATM.

tuxemon/core/components/event/actions/player.py
```diff
@@ -32,6 +32,7 @@
 from core import tools
 from core.components import item
 from core.components import monster
+import Common
```
ShadowApex: This should be changed to:
```python
from core.components.event.actions.common import Common
```

Otherwise the game crashes on launch:
```python
    main()
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/main.py", line 57, in main
    control = PygameControl(prepare.ORIGINAL_CAPTION)
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/control.py", line 75, in __init__
    self.event_engine = event.EventEngine()
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/components/event/__init__.py", line 73, in __init__
    self.actions = plugin.get_available_methods(action_plugins)
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/components/plugin.py", line 148, in get_available_methods
    for plugin in plugin_manager.getAllPlugins():
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/components/plugin.py", line 93, in getAllPlugins
    m = importlib.import_module(module)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/tmp/xirsoi/Tuxemon/tuxemon/core/components/event/actions/player.py", line 35, in <module>
    import Common
ImportError: No module named Common
```

xirsoi: Thanks, I'll deal with this after my other PR gets merged.

xirsoi:
Finally got this updated with the latest branch. Everything seems to be working.

xirsoi:
@ShadowApex and/or @bitcraft , could you give this another look?

ShadowApex:
@xirsoi I'll try and review this PR this week. Been very busy recently.

xirsoi:
I understand, I've been busy too. Just wanted to make sure you saw that it was updated.

ShadowApex:
Finally had some time to review this. Everything looks great! Thanks for the PR!

Issue #160, opened 2016-07-08
End of combat crashes when using non-en_US locale.


ShadowApex:
Very odd error if you use a non-English locale at the end of combat when enemy monster faints. This does not happen when locale is set to `en_US`:

System Details:
Platform: `x86_64`
OS: `Ubuntu 16.04`
Python Version: `2.7.11`
Pygame Version: `1.9.1`

```python
Traceback (most recent call last):
  File "./tuxemon.py", line 60, in <module>
    main()
  File "/tmp/Tuxemon/tuxemon/core/main.py", line 109, in main
    control.main()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 512, in main
    self.main_loop()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 573, in main_loop
    self.update(time_delta)
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 203, in update
    state.update(dt)
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 156, in update
    self.update_phase()
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 328, in update_phase
    self.handle_action_queue()
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 338, in handle_action_queue
    self.check_party_hp()
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 614, in check_party_hp
    self.faint_monster(monster)
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 581, in faint_monster
    awarded_exp = monster.total_experience/monster.level/len(self._damage_map[monster])
KeyError: <core.components.monster.Monster object at 0x7fb24236ddd0>
```

PR #165, opened 2016-07-13
slug update

bitcraft:
I'm calling this the slug update.  In an effort to improve the ease of translating the game strings and to prevent bugs around game references to objects (see #160), the slug system was implemented.

Just to show I'm not making this up, slugs are used often in the web world to make short, easier-than-number identifiers for things.  A few google searches will prove I'm not crazy.

Slugs in Tuxemon will be used to create references in the code and maps to game objects using short underscore-spaced English words.  Slugs will never be displayed in the game.  Think of them as an alternative to remembering numeric IDs for things.

Here is the gist of the changes in the PR to make slugs a reality:

- Remove code references to object by their "name", and instead use their "slug"
- Change map references from English names of objects to the slug: "Capture Device" becomes "item_capture_device"
- Give Items and Techniques a slug field and make it somewhat simple ("item_potion")
- Make locale references more explicit: "item_potion" becomes "item_potion_name"
- Remove English descriptions of things into the locale db (moved to en_US.json)
- Items and techniques' names and descriptions are in the locale.  Do not make references to names as they will change depending on the locale.  instead use the slug.

One point to note is that slugs and locale references share the same format, but are not used the same way.  Slugs are used to create references to objects, while the locale references are used to display locale specific strings.

I've also included some very handy tools in the tests/interactive/locale folder.  Using the two scripts, maintainers can compare locales to the "Master" English locale and locate missing translations quickly.  Another script will also read the DB's and look for items that are missing translations in the Master locale.  I was able to find a few missing strings that were not caught in manual playtesting, so this will really help keep our translations clean.

Moving forward:

- In the future, if we get a nice build system, we can integrate the locale tests and generate reports of missing translations.
- No base English strings should be used in the DB objects.  The exception is the slugs, and locale references of course.
- Some wiki-ing will have to be done to make the slugs and what not clear to our great translation team.  We owe them a so much for their work so far!

So, thank you to the translators and I hope we can make even more of the game localized to your language and accessible to all!

Just for fun:
http://imgur.com/gallery/R9NCh


ShadowApex:
Everything looks great. I'm super happy with this change, and it's going to make working with other locales significantly easier. The script for looking for missing translation strings is especially awesome. I'll try and look into setting up an auto-generated page on tuxemon.org that runs this script against the dev branch.

PR #171, opened 2016-07-19
Monster slugs, some tools to verify

bitcraft:
Big-ish update:

- Migrate monsters to the slug system
- Change Technique names to slug that are found in monster def
- Update existing events to use monster slugs and not ids or names (which are localized)
- Add locale tags for monster names, categories, and descriptions
- Add some English monster categories and descriptions
- Add monster verification scripts
- Cleanup non-conforming monster defs
- Update the random encounter to include all useful monsters

The changes were pretty far reaching, but mostly accomplished by copy/paste across the repo (with sed).  Also, two new tools to verify the monster locales tags exist and sprites and to generate a encounter file from all the monsters in the db.  Some minor clean up as well.  Some monster defs didn't conform to the correct folders for their sprites (using 'resources' as root, instead of 'gfx').

Monster numeric ID's are basically obsolete now.  In the future, we should also retire ids for everything else (use slugs).


chiefexb:
It's awesome I spend one hour to make similar tools yeasterday but you made it better

bitcraft:
@chiefexb Great!  Hope it helps you with the translations.  You can copy/paste the program output into your translation so you don't need to type all the syntax.

PR #176, opened 2016-07-20
added spanish language

chagara:


bitcraft:
Using google translate, "Poison Sting" to translated back into English as "envenomation", which is a really cool word that I've never heard.

https://en.wikipedia.org/wiki/Envenomation

Seems to be a typo: "Cura a un montstruo por 150 HP.",  (monstruo)
What is this: Cojio?  (¡$ {{name}} Cojio daño Venenoso!")

Looks pretty good!

chagara:
Yeah google translate its like 40% correct. Barely ever gets anything
correct. Yeah ill fix the type monstruo
Cojio its like getting but, depends where you are from it can mean
different things.

bitcraft:
Please see #181, as it may impact your translation work.  Thanks!

Issue #163, opened 2016-07-12
Save/Load: changes to game data between saves doesn't update save file


bitcraft:
Long title, but a potential annoyance.  This issue relates to how save files are stored and I'll outline two known bugs that it causes:
1. Locale settings don't change all strings if you save and load after changing the locale
2. Changing monster parameters will not update old saves, forcing a new game to be started

Here are the steps for the locale bug:
- New game
- Save
- Change locale
- Load

Some things, such as technique names and items will stay at the original locale.  This seems to happen since items loaded from the db are saved directly into a save file.  When loading, the game simply reuses the saved items, without instancing them.  The problem is that the translated strings are saved along with db items.

Possible fixes:
- Rewrite save/load to not store python objects, and store and load DB items by id
- Remove the translated string from db items, and define new way to get names and descriptions from items

I think that this is somewhat of a minor bug, as most people will not go through this, but it does expose some limitations of the save and load code.  For example, a similar problem would happen when changing monster parameters.  Suppose you:
- New game
- Get Fruitera
- Save
- Edit Frutera's max HP (or anything else)
- Load

The save game wouldn't have the updated changes from Frutera.  Obviously not a problem when released, but for development, if forces developers to always create new games when testing changes, since save games will save all old parameters of DB objects.  If we go for a "rolling release", or expect to issue balance updates, locale changes, etc, then this bug will force workarounds to deal with the save files.

To clarify, "DB objects" refers to objects that have their parameters set with json files.  Currently, this is monsters, items, techniques, locale translations, encounters, and npcs.

Personally, I'd prefer to see Fix #1 used, as it would make the save files smaller (by storing ID's and not full objects) and make the game more stable when configuration or json objects change.  It also would just change the save/load code.  Fix #2 would possibly be nice as well, but I'm not sure what it would entail...maybe a more comprehensive DB of translations?

In any case I won't set a milestone yet as it is a extreme corner case, but I hope it will get fixed at some point.

xirsoi:
Why would you expect unsaved changes to show up after loading?

bitcraft:
That's not it.  The issue is that save games are saving DB objects and loading them.  If DB objects are changed between saves, the save files will be using old data.  Suppose a decision is made to make Potions heal 100 instead of 150.  A save file saved with Potions before the update will still recover the old 100. 

The more annoying part is that this applies to locale translations as well so your save file cannot be effectively change locales once saved for the first time.

[...]

PR #177, opened 2016-07-21
Changed save format to json. It will still load old save files.

dulsi:
Although files created before the slug field was added to the object will fail to save afterward.

This fixes issue #163 . I have not split monster instances and the species. This will allow techniques to be updated for monsters if the data changes but things like hit points and defenses will not.

### DIFF
--- a/tuxemon/core/components/save.py
+++ b/tuxemon/core/components/save.py
@@ -130,95 +130,55 @@ def load(slot):
         return
 
     saveData = dict()
-    try:
-        with open(save_path, 'r') as save_file:
-            json_data = json.load(save_file)
-            tempinv = dict()
+    with open(save_path, 'r') as save_file:
+        json_data = json.load(save_file)
+        tempinv = dict()
 
-            for slug,quant in json_data['inventory'].items():
-                tempinv1 = dict()
-                tempinv1['item'] = Item(slug)
-                tempinv1['quantity'] = quant
-                tempinv[tempinv1['item'].name] = tempinv1
-            saveData['inventory'] = tempinv
-
-            tempmon = list()
-            for mon in json_data['monsters']:
-                tempmon1 = Monster()
-                tempmon1.load_from_db(mon['slug'])
-                tempmon1.load_from_savefile(mon);
-                tempmon.append(tempmon1)
-            saveData['monsters'] = tempmon
-
-            # TODO: unify loading and game instancing
-            # Loop through the storage item keys and re-add the surface.
-            tempstorage = dict()
-            for keys, values in json_data['storage'].items():
-                if keys == 'items':
-                    tempinv = dict()
-
-                    for slug,quant in values.items():
-                        tempinv1 = dict()
-                        tempinv1['item'] = Item(slug)
-                        tempinv1['quantity'] = quant
-                        tempinv[tempinv1['item'].name] = tempinv1
-                    tempstorage[keys] = tempinv
-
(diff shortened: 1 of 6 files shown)


tuxemon/core/components/save.py
```diff
@@ -123,42 +129,96 @@ def load(slot):
...
+            tempstorage = dict()
+            for keys, values in json_data['storage'].items():
+                if keys == 'items':
+                    tempinv = dict()
+
+                    for slug,quant in values.items():
+                        tempinv1 = dict()
+                        tempinv1['item'] = Item(slug)
+                        tempinv1['quantity'] = quant
+                        tempinv[tempinv1['item'].name] = tempinv1
```
bitcraft: The name attribute is dependent on the locale set.  Is there a way to avoid its use here?

bitcraft:
Thanks for tackling this issue.  JSON is ok with me, though you are building dictionaries so pickle/shelve could be used as well if others think the pickle binary format is preferred.  If we keep the JSON format, then I don't think we need code to load the old files....i'd rather it not be around to reduce code maintenance.

I'd also prefer to move the "load/save_from_savefile" from the fusion and monster files to the save/load modules.  If the file format changes in the future, it may not be obvious to look in the monster and fusion modules.

Looks good so far!

dulsi:
Pickle/shelve is not a good idea. [Here](https://lwn.net/Articles/595352/) is an article detailing why.

I can certainly move the load/save functions from fusion/monster files. Is this just a personal preference or common practice under python?  The counter argument to your format change is that if new fields are added to monster and fusion, it may not be obvious that they need to look at save/load code. I'll make the change.

bitcraft:
To be clear, I'm not advocating pickle.  I brought it up because I wasn't sure if we all agreed it should be JSON...which I'm fine with.   Anyway, thanks for understanding about moving the save/load code; I don't think it is a 'python' thing, mostly just a nice way to separate concerns.  Since the save/load code is dependent on the monster class, but the monster class is not dependent on the save/load, then the glue code should be organized with the save/load code.  Hope that helps.

dulsi:
All the changes you asked for are in. Let me know if you see anything else you feel needs to be fixed up.

bitcraft:
Works great!  I've opened #178, which is somewhat related...how do we handle save files that the game cannot read?  (I realize you had a work around for loading old-style saves, but I don't think we need that function...)  Maybe we can get a dialog or something to handle it.

I've tested the changes and it seems to work fine; saving and loading a file worked as expected.  Thanks again!

ShadowApex:
> I brought it up because I wasn't sure if we all agreed it should be JSON...which I'm fine with.

I'm totally happy with saves being in JSON.

I also tested this out and confirmed it works great. Thanks for the PR!

The only reason I originally made saves using `pickle` was to have a small level of obfuscation for save files to discourage cheating by modifying your save file with a simple text editor. 

To help prevent this, maybe we could use a combination of json + base64 + zlib to keep save files small and not as easy to modify. This obviously won't stop people from cheating, but it may discourage casual modifying of save games. Something like this:

```python
import zlib
import base64
...
save_data = base64.b64encode(json.dump(json_data))
save_file.write(zlib.compress(save_data))
save_file.close()
```

Sanglorian:
Isn't it kind of cool if people edit their own save files? I can imagine that being a kid's first interaction with programming and code and so on. Having the details of their game readable to the player sounds like a feature to me, not a bug!

PR #180, opened 2016-07-22
Update it_IT.json

Airon90:
Updated Italian translations. I corrected some error and added some missing keys. However I need more info in order to translate:
- Lambert's description and category
- Nudiflot's category

As Italian is a gender-based language, I had to remove the key about the object in "item_cannot_use_here" as it would have needed a complex management of gender: an article between "il" and "lo" for masculine words, "la" for feminine words and "l'" in caso of a word which starts with wovel would have needed and "usato" could have been "usata" in case of feminine words


bitcraft:
Interesting problem there, @Airon90.  Tuxemon doesn't have any sense of gender for our words and items, but the change will not be simple.  If you would like, we could just merge this now and try to leave it in a "neutral" tone (if you can!) and just let the game use broken Italian for a while.  I imagine our Spanish and French translations have this problem, too but they didn't report it.

TL;DR I think the best course of action is to merge the "broken Italian" for now.  In the future, we can add gender rules for all languages and then update the locales.

I've opened #181 to address this gender problem.

Sanglorian:
Hi @Airon90 

What extra info do you need for Nudiflot and Lambert?

Nudiflot is called the Flopped tuxemon because when it sits on the beach it just sags/droops/dangles down like a fat lump.

Lambert is called the Gumnut tuxemon because its shell is a seed/nut from a gum tree/eucalyptus.

Airon90:
@bitcraft By now yes, keep the "broken" one and when #181 is fixed somehow I'll change the translation

@Sanglorian Thank you, I fixed the last translations as I wasn't able to do that without any explanation :)

Airon90:
Thank you, @ShadowApex!

PR #189, opened 2016-07-24
Implemented has_monster conditional and remove_monster action.

dulsi:
has_monster conditional and remove_monster action added to the event system. They are not currently used in the maps.

### DIFF
--- a/tuxemon/core/components/event/actions/player.py
+++ b/tuxemon/core/components/event/actions/player.py
@@ -378,4 +378,4 @@ class Player(object):
 
         monster = game.player1.find_monster(monster_slug)
         if monster:
-                game.player1.remove_monster(monster)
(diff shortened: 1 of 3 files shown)


tuxemon/core/components/event/actions/player.py
```diff
@@ -358,3 +358,24 @@ def player_resume(self, game, action):
...
+        :rtype: None
+        :returns: None
+
+        Valid Parameters: monster_slug
+        """
+        monster_slug = action.parameters[0]
+
+        monster = game.player1.find_monster(monster_slug)
+        if monster:
+                game.player1.remove_monster(monster)
```
ShadowApex: This line is indented too much, but other than that everything looks great! Thanks for the PR!

dulsi:
Indenting has been fixed.

Issue #197, opened 2016-07-28
Need dialog_choice to ask the user to select an option


dulsi:
Sometimes it would be useful to have an NPC ask the PC a question. Here is the example we discussed on the forum:

Event 1:
cond1: not has_monster txmn_pigabyte
act1: dialog I lost my pigabyte.

Event 2:
cond1: is has_monster txmn_pigabyte
act1: dialog_chain You found a Pigabyte. Can I have it?
act2: dialog_choice Yes:No:Maybe,myvariable
act1: dialog_chain ${{end}}

Event 3:
cond1 is variable_set myvariable:Yes
act1: remove_monster txmn_pigabyte
act2: dialog "Thank you so much!"
act3: set_variable myvariable:

Event 4:
cond1 is variable_set myvariable:No
act1: dialog "*whimper* I'll keep looking I guess"
act2: set_variable myvariable:

My prototype implementation only has translated_dialog_choice at the moment. This example uses dialog_chain because my implementation requires that the dialog_choice be at the end of a dialog_chain. Right now the choice box appears in the center of the text dialog which is not ideal.

PR #199, opened 2016-07-28
Adds dialog_choice and translated_dialog_choice to the event system

dulsi:
Implements issue 197 so that NPCs can ask questions.

### DIFF
--- a/tuxemon/core/components/event/actions/core.py
+++ b/tuxemon/core/components/event/actions/core.py
@@ -235,8 +235,9 @@ class Core(object):
 
         if text == "${{end}}":
             # Open a dialog window in the current scene.
-            open_dialog(game, self._dialog_chain_queue)
+            open_dialog(game, self._dialog_chain_queue, self._menu)
             self._dialog_chain_queue = list()
+            self._menu = None
         else:
             self._dialog_chain_queue.append(text)
 
@@ -481,6 +482,35 @@ class Core(object):
             if e['id'] == int(action.parameters[0]):
                 event_engine.execute_action(e['acts'], game)
 
+    def dialog_choice(self, game, action):
+        """Asks the player to make a choice.
+
+        :param game: The main game object that contains all the game's variables.
+        :param action: The action (tuple) retrieved from the database that contains the action's
+            parameters
+
+        :type game: core.control.Control
+        :type action: Tuple
+
+        :rtype: None
+        :returns: None
+
+        Valid Parameters: choice1:choice2,var_key
+        """
+        def set_variable(game, player, var_key, var_value):
+            player.game_variables[var_key] = var_value
+            game.pop_state()
+            game.pop_state()
+
+        # Get the player object from the game.
+        player = game.player1
+
+        var_list = action.parameters[0].split(":")
+        var_menu = list()
+        for val in var_list:
+            var_menu.append((val, val, partial(set_variable, game=game, player=player, var_key=action.parameters[1], var_value=val)))
+        self._menu = var_menu
+
     def translated_dialog_choice(self, game, action):
         """Asks the player to make a choice.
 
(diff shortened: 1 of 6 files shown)


ShadowApex:
Hi dulsi, thanks for the PR! Everything looks good, but I wanted fully test the `dialog_choice` and `translated_dialog_choice` in a map, which has taken me a little bit to get around to. Once I've finished testing I should be able to merge this in.

ShadowApex:
I did some testing and found some issues with the current implementation. Right now I have only tested `dialog_choice` and not `translated_dialog_choice`. I'd like to know if you can also replicate these issues. Here's what I found:

## Events
Here's what I currently have configured in the map. If I didn't create these events properly, let me know:
**Event 1**


**Event 2**


## Bugs
- Right now it appears that I have to trigger the dialog chain twice in order for the choices to appear:


- After picking the appropriate choice, it looks like the dialog choice continues to pop up:


- Not really a bug necessarily, but I think we should also allow `dialog_choice` after normal `dialog` and `translated_dialog` actions, and not just `dialog_chain` actions.

Other than those things, everything else seems to work pretty well!

dulsi:
dialog_chain ${{end}} needs to be after the dialog_choice. See issue #197 for my example. I agree ideally you want to be able to do it after dialog. Frankly I don't like dialog_chain and would prefer to have no need to specify dialog_chain ${{end}} but haven't looked into how.

ShadowApex:
> dialog_chain ${{end}} needs to be after the dialog_choice. See issue #197 for my example.

Thanks for linking to your example. After moving the `dialog_choice` before calling `${{end}}` everything worked perfectly!

I also agree that not needing to specify `${{end}}` would be ideal. The main reason it was originally implemented that way was because each action is inherently not aware of any of the other actions. For example, `act2 dialog_chain My name is Lampy\\, lord of all lamps` doesn't know about or have access to "act1", "act3", "act4", etc., and thus doesn't know if it's the last text in the dialog chain (once the last text in the dialog chain appears, it needs to call [open_dialog](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/event/actions/core.py#L234)). We might be able to solve this problem and other event problems like it by [passing the full "action_list" to every action that's executed by the Event Engine](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/event/__init__.py#L169).

dulsi:
I have two thoughts on the problem:

1) When an event is started a "context" object is created. Event system is passed this "context" object. _dialog_chain_queue and _menu are moved to this object. If after the event finishes, it finds a _dialog_chain_queue and/or -menu, it opens the dialog.

2) Dialog is opened immediately but all add future dialog/choice events add to the dialog object. How exactly that is done, I'm not sure. (I suppose you could say this is kinda the same idea. Just that the dialog object becomes the "context" object mentioned above.)

bitcraft:
Currently, the dialog_chain event will queue up any messages that it gets.  What if we dropped the normal dialog action and renamed dialog_chain to dialog?  Then in the event manager, we explicitly check for any messages in the dialog queue and open them at the end of the frame.  We'll shoehorn the dialog choices in the queue as well.

This would also drop the need for {{end}}.

Thoughts?

dulsi:
That is essentially what I'm suggesting with the "context" object. The reason I mentioned it as passing in a new object instead of just checking the _dialog_chain_queue/_menu globals at the end of running an event is that eventually the plan is to support multiplayer. If I have a single _dialog_chain_queue, that won't work.

Right now the event manager doesn't know anything about the event system. Moving the open_dialog there makes it more knowledgeable about the internals. Maybe the context could be a collection of "finish" functions. That way the event manager still doesn't know anything.

bitcraft:
I see what you are saying @dulsi.  Our suggestions will do essentially the same thing, but the implementation differs.  Right now, I don't see any need outside of the dialogs for a "context" system like you are proposing, but I do see the merits.

If you want to pursue this, it would help everyone if you could provide some pseudo code and a general plan of how it works and what the benefits will be.  Opening an issue would be a good start.

bitcraft:
After a bit of thought, it seems to me that we could pass the list of events without much effort, and we can fix a few issues with "one weird trick" (computer programers HATE him).  The context system can be bolted on later, because I don't see them creating any conflicts with each other.

ShadowApex:
@dulsi I'm going to accept this PR, but I created a separate issue for discussion about adding a context system to the event engine. It can be referred to here:  #208

PR #204, opened 2016-08-04
Fast map drawing

ChillyCider:
Benchmarks were taken at 320x240 resolution, on map1.tmx, using the cProfile module for Python. On my Lemote Yeeloong (CPU: Loongson 2f), the speed increase after these commits is ~40ms each frame. See commit messages for specifics.

I'm interested to hear if it increases framerate on other devices, too. :-)

### DIFF
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -301,43 +301,37 @@ class WorldState(state.State):
         self.highlayer_tiles = []
         self.medlayer_tiles = []
 
-        starting_tile_x = - \
-            (self.global_x / self.tile_size[0])
-             # How many tiles over we have to draw the first tile
-        starting_tile_y = - \
-            (self.global_y / self.tile_size[1])
-             # How many tiles down we have to draw the first tile
-        self.tile_buffer = 2  # This is how many tiles we should draw past the visible region
+        # What region of tiles should be visible?
+        left   = -int(self.global_x / self.tile_size[0])
+        top    = -int(self.global_y / self.tile_size[1])
+        right  = left + self.visible_tiles[0]
+        bottom = top + self.visible_tiles[1]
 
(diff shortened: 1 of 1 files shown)


ShadowApex:
These changes look great! Thanks for the PR.

I'll try and test this on the GCW-Zero too and see how it's improved the framerate. What FPS were you getting on your Longsoon architecture before and after the change on map1.tmx?

xirsoi:
40ms per frame is an incredible boost, great work @ChillyCider !

ChillyCider:
map1.tmx was really choppy before. When I moved my character, I would see just three frames of movement and then the movement was finished. I think it was like 8 to 12 fps, but I don't have an actual measurement.

I am now getting like 15 to 25 fps. :smile:

bitcraft:
* This PR breaks python3 compatibility with the use of xrange.  If @ShadowApex agrees, we could consider using the six module in Tuxemon (already required by pytmx) and this PR could be changed to use the range function provided by the six module.  Alternatives would be simply using range or futurize (outside the scope of this PR).

* For iterating multidimensional arrays, I like to use itertools.product...it reduces the indentation level and in some cases, is a bit faster than nested for loops.  You could consider using that in your PR (not needed, but since you use the itertools package, I thought it would fit in here!).  Here's an example: https://github.com/bitcraft/PyTMX/blob/master/pytmx/pytmx.py#L866-L867


This should be opened up against the development branch as well.  Overall, thanks for the great PR.  I like the liberal use of iterators.

ShadowApex:
> If @ShadowApex agrees, we could consider using the six module in Tuxemon (already required by pytmx) and this PR could be changed to use the range function provided by the six module.

I think using the `six` module would be good. As you mentioned, `six` is already required by PyTMX.

ChillyCider:
Thanks for the feedback guys! I have updated my development branch and put my changes there. I'll submit a new pull request using that as the source.

PR #206, opened 2016-08-04
Fast map drawing (redone for 'development' branch)

ChillyCider:
These commits speed up map drawing on slow systems. The speedup is significant (40ms every frame) on a Lemote Yeeloong laptop in map1.tmx.

It uses range & map from the **six** module for compatibility between Python 2 and 3. I have only tested with Python 2, however.

I experimented with itertools.product but IMO readability suffered.

### DIFF
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -33,6 +33,7 @@ from __future__ import division
 # Import various python libraries
 import logging
 import math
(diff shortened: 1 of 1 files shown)


ShadowApex:
Looks great! Thanks for re-opening this PR!

I tested this with python2 and python3 and both work well. I'll be trying to do some benchmarking on GCW-Zero in the near future and post my results.

PR #209, opened 2016-08-09
Fix pathfinding.

### LINKED ISSUE #196: pathfind doesn't work anymore.
I found this problem not too long ago when I was messing with some campaign maps. I was trying to make someone talk and then walk to lead the player. The person didn't pathfind. I created a new event that was to make the person pathfind to a tile if the player pressed enter on a certain tile. This didn't work either. 

Test map: https://www.dropbox.com/s/ao5tp40muz1i2nn/map1.tmx?dl=0 (tell me if it won't open for any reason.

Console Output:
screenshotdfkj

dulsi:
Fixes #196

### DIFF
--- a/tuxemon/core/components/player.py
+++ b/tuxemon/core/components/player.py
@@ -663,7 +663,7 @@ class Player(object):
     def pathfind_r(self, dest, queue, visited, depth, game):
         # recursive breadth first search algorithm
 
-        if queue:
+        if not queue:
             # does reaching this case mean we exhausted the search? I think so
             # which means there is no possible path
             return False
--- a/tuxemon/core/components/player.py
+++ b/tuxemon/core/components/player.py
@@ -650,7 +650,7 @@ class Player(object):
                 # last minute check to remove the top _plan step if
                 # it's the same as our location
                 if path[len(path)-1] == self.tile_pos:
-                    plan.pop()
+                    path.pop()
 
                 # store the path
                 self.path = path


bitcraft:
Have you tested this?  It seemed to me that the issue was related to all the npc references being changes from name to slug.  Maybe there are two issues involved?

dulsi:
Yes I tested it. Development doesn't have name to slug change for npcs. When slug is added the code to find the npc needs to be updated but that is unrelated to this bug.

xirsoi:
Don't worry, i'll make sure this doesn't get broken with the npc name->slug update.

bitcraft:
Sounds great.  If you have a small script that you used to test it, it would be useful to add it the the "tests" folder and add it to the PR.  Even if it isn't a unit test, it might be useful.  There isn't much in the way of unit tests at the moment, but I hope we can get there someday.

dulsi:
I tested it by using josepharaoh99's map. How can I run the unit tests?

bitcraft:
@dulsi, i've got a test in the works for this.  what i envision is a pool of maps that are designed to be run in a testing environment.  #212 has more details on the concept.

bitcraft:
@dulsi, I tested this locally, but got an error on the line below.  If you change it, then we can merge this in.

https://github.com/dulsi/Tuxemon/blob/4678d4b9f87d0bc165023e5ea61c0a6098b29f9c/tuxemon/core/components/player.py#L653

```python
                if path[len(path)-1] == self.tile_pos:
                    plan.pop()
```

To this:
```python
                if path[len(path)-1] == self.tile_pos:
                    path.pop()
```

After your change, then the one described, I was able to see a NPC move on the map correctly.

dulsi:
Fix checked in.

bitcraft:
Thanks for taking care of this.

Issue #156, opened 2016-07-06
No way to change monster name (or any other)


bitcraft:
Monsters and Trainers have the ability to have different names, currently.  So, it is technically possible to have a Rockitten named "Floofie", and Big Fin named "Tuna".  Right now, there isn't a way to actually change their names though.  So we need a state/menu to input text.

Right now, I see two needs for text input: Naming things and inputing IP/hostnames for multiplayer.  Here is a proposal to create just the Name Input state.
- New Menu/State
- Includes portrait of monster/trainer to be named
- Keyboard input
- Mouse/touch input
- Controller/Joystick input

This state should behave like a tablet computer/phone that we are all familiar with as well as a console where you press the arrows to move a cursor and select letters and numbers.

Here's a rough outline of how to make it:
- Subclass the Menu State
- Overload Menu.initialize_items: create menu items for each letter and symbol
- Overload Menu.on_menu_selection: build input string from selections

Thats the 10,000ft. overview.  Any takers?

buddja:
I have been working on a way to get a string from the user. I found extext.py in _core/components_ but it seems keyboard based from what i've seen (I havent played much with it though) witch wouldn't be what we wanted, so i'm trying to create a onscreen keyboard that the user can navigate with the arrow keys... 

Am i on the right path or did i miss something?

bitcraft:
Glad you've volunteered  :D. I've gone through eztext.py and the update() function would be the most useful part to keep.  Most of the rest will have to be written again to support the new menus and also be touch/joystick/keyboard aware.

Here's a few pointers to get you started:

- overload Menu.handle_event() to grab keypresses
- reuse the update code to enable keyboard input
- for just testing and debugging, store the input string as whatever you like
- for joystick and touch, create many buttons in Menu.initialize_items()
- the buttons should have their callback set to a function to add their key to the stored string
- To display the text, create a TextArea in the menu, and use Menu.alert() to show your text

If you follow those points, you will be 90% complete, with the following problems:

- The menu will be able to input text but each keypress or button will make the text box input clear all the text and draw it again.
- The backspace button will not work
- English letters only

If you get to that point, even with that problem, I will help complete it.  There will be a few issues with showing the input and also supporting other languages besides English.  If you get it mst of the way, I can help finish it.

buddja:
Do you think it should be implemented as a new state something like KeyboardState, or just as another game component that states can call?

 I'm unsure witch one would be better.

[...]

PR #211, opened 2016-08-10
input state & menu touch awareness

bitcraft:
Add the code for the input state.  Addresses issue #156. 

So far, the state is not being used, but it could be used for naming players and monsters, IP addresses & hostnames, etc.  Its a little rough, but can be cleaned up by adding a background and eventually adding different alphabets for locale changes.

So far there is no mechanism to collect the input values, but we can work that out later.

The prompt can be changed when pushing the state:

```python

state = game.push_state("InputState", prompt="Name?")
```

One possible way to collect input would be to use the "input_string" attribute, which contains the test entered:
```python

input_state = game.push_state("InputState", prompt="Name?")

...some other work...

print(input_state.input_string)
```

Allowing the user to touch/tap/click the letters in the menu meant adding touch support to menus.... So now all menus have the ability to respond to touches or clicks.  Right now, it is only enabled for the input state, but it is trivial to enable it for others.... Just change the touch_aware attribute to True.

https://github.com/bitcraft/Tuxemon/blob/development/tuxemon/core/components/menu/menu.py#L56

### DIFF
--- a/tuxemon/core/components/menu/__init__.py
+++ b/tuxemon/core/components/menu/__init__.py
@@ -3,7 +3,7 @@
 #
 # Tuxemon
 # Copyright (C) 2014, William Edwards <shadowapex@gmail.com>,
-#                     Benjamin Bean <superman2k5@gmail.com>
+# Benjamin Bean <superman2k5@gmail.com>
 #
 # This file is part of Tuxemon.
 #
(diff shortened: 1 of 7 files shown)


bitcraft:
Not ready to merge.  :disappointed:

bitcraft:
Ok, ready to merge.

PR #218, opened 2016-08-19
fully translated to ru_RU

nikitakurylev:
I have fully translated en_EN to ru_RU
I don't know why, but it crashes when I enter battle with ru_RU in config.
It logs this:
`UnicodeEncodeError: ascii codec can't encode characters in position 0-9: ordinal not in range(128)`

But on other side everything else works


bitcraft:
I'll open a bug report for the error you are having.  What version of python are you using?  Python 2.7?

nikitakurylev:
Python 2.7.8

PR #229, opened 2016-08-22
Rename sp_US.json to es_MX.json

IBPX:
Right now, the code being used for Spanish is `sp`, while the standard is `es`. Also, the country code I believe should be `MX` for Mexico, not `US` for the United States, because (correct me if I'm wrong here) I believe the dialects between the States and Mexico are almost identical, and are commonly referred to as Mexican Spanish.


ShadowApex:
`es_US` exists, actually, but you're right that the standard prefix is `es` for Spanish:

```bash
$ cat /usr/share/i18n/SUPPORTED | grep 'es_.. '
es_AR ISO-8859-1
es_BO ISO-8859-1
es_CL ISO-8859-1
es_CO ISO-8859-1
es_CR ISO-8859-1
es_CU UTF-8
es_DO ISO-8859-1
es_EC ISO-8859-1
es_ES ISO-8859-1
es_GT ISO-8859-1
es_HN ISO-8859-1
es_MX ISO-8859-1
es_NI ISO-8859-1
es_PA ISO-8859-1
es_PE ISO-8859-1
es_PR ISO-8859-1
es_PY ISO-8859-1
es_SV ISO-8859-1
es_US ISO-8859-1
es_UY ISO-8859-1
es_VE ISO-8859-1
```

Issue #213, opened 2016-08-12
Multiplayer menu option causes duplicate menu items.


Phrohdoh:
![screen shot 2016-08-12 at 12 53 45 pm](https://cloud.githubusercontent.com/assets/4861023/17632078/2c19cfe8-608c-11e6-876e-1e9aacc42c38.png)

Interact with the computer in the starting room and select `Multiplayer` to repro.

bitcraft:
Nice catch.  We'll look into it!

PR #231, opened 2016-08-23
Fix multiplayer menu, a Task bug

bitcraft:
Addresses #213 

* Multiplayer menu opens correctly
* Added some options in the MP menu
* Added en_US translations for MP menu
* Fixed a couple bugs with timing in Tasks
* Updated the Task class

A couple issues were at work with this bug.  A nice side effect is level transitions and dialogs open more smoothly on slow systems after fixing the Task bug.

There could be some problems with MP still.  I've just renabled the menus.  Still need to do IP address input and such.


ShadowApex:
I spent some time this weekend looking at this. Looks good! :) Thanks for this fix.

The way multiplayer works may change a bit when I have a more feature-complete server implementation, so IP address input isn't a huge problem.

PR #232, opened 2016-08-23
Fixed Esperanto translations

Airon90:
- estas venenigita → veneniĝis
- estas damaĝita de → damaĝiĝis pro
- kion faris? → kion faru?
- vi kaptis → estis kaptita
- uzeblas → ne uzeblas
- estas jam en ludado → jam batalas


Spagy:
Is it possible to make variables accusative just by adding an -n?
E.g. "${{user}} uzis ${{name}}n!"

I can't see why this isn't possible but it seems as though some of the strings have been constructed so as to not use the accusative.

Also sorry for taking such a long time to have a look.

Spagy:
I'll make comments on the strings assuming that using -n is possible.

Spagy:
I an seemingly only comment on lines that are close to edited lines. So I'll comment on other lines like this:
Line 4: Why use passive?
Suggestion:
"combat_used_x": "${{name}} estis uzata de ${{user}}!" → "${{user}} uzis ${{name}}n!"

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
     "combat_used_x_on_y": "${{name}} estis uzata al ${{target}} de ${{user}}!",
```
Spagy: The use of passive "al" make this seem unnatural to me.
Suggestion: "${{user}} uzis ${{name}}n kontraŭ ${{target}}!"

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
     "combat_call_tuxemon": "Iru, ${{name}}!",
     "combat_opponent_call_tuxemon": "${{user}} sendas ${{name}}!",
```
Spagy: This is a different tense from the English ('sent' compared to 'sendas'). Assuming this string is used as, or just after, a user has sent out a monster, this looks like it might be a more natural translation than using 'sendis'.
Just wanted to bring this up in case it was accidental.

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
-    "combat_state_poison_damage": "${{name}} estis venenigita!",
-    "combat_status_damage": "${{name}} estis damaĝita de ${{status}}!",
+    "combat_state_poison_damage": "${{name}} veneniĝis!",
+    "combat_status_damage": "${{name}} damaĝiĝis pro ${{status}}!",
     "combat_victory": "Vi gajnis!",
```
Spagy: 'Gajni' usually means 'acquire' or 'win' (in the sense of winning a prize). E.g. "Mi gajnis la grandan gajnon".
Although it can be used in the sense of "I won a game" (Mi gajnis ludon), I think it would be nicer to use 'venki' here because conveys the meaning of overcoming the opponent and it has nicer symmetry with the following string, which uses 'venki'.
Suggestion: "Vi venkis!"

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
     "combat_defeat": "Vi estis venkita!",
```
Spagy: The current string is fine, this is just nitpicking. If you disagree with me on this absolutely leave it as is, I won't feel too hurt. :)
I don't really like the passive form here because it seems to mirror the English, whereas, for me, "Vi malvenkis" feel natural and Esperantoey.
Suggestion: "Vi malvenkis"

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
-    "combat_capturing success": "Vi kaptis ${{name}}!",
+    "combat_capturing success": "${{name}} estias kaptita!",
```
Spagy: It looks like your trying to avoid the accusative here. That makes sense if we can't localise ${{name}}. Having something like "Vi kaptis Spookthingn" would be a mess.

But if we can localise ${{name}} "Vi kaptis ${{name}}n!" would seem better to be.

In any case maybe we could pick one tense rather than two for 'esti' ;)
(estias → estas)

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
     "combat_capturing_fail": "${{name}} liberiĝis!",
     "combat_replacement": "Elektu anstataŭanton!",
```
Spagy: Another nitpick and personal preference. Feel free to ignore. This could be 'anstataŭon', instead of 'anstataŭanton'. Assuming this is referring to a monster, removing '-anto' might be good as it is usually used only for people (with the -o ending that is). This would also shorten the word and make it more readable IMHO.

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
     "combat_replacement_is_fainted": "La monstro svenis!",
```
Spagy: Considering that a player probably has more than one monster, to me, it would seem more clear to use 'Tiu' instead of 'La'.

tuxemon/resources/db/locale/eo.json
```diff
@@ -5,18 +5,18 @@
...
     "combat_wild_appeared": "Sovaĝa ${{name}} aperis!",
-    "combat_monster_choice": "Kion ${{name}} faris?",
+    "combat_monster_choice": "Kion ${{name}} faru?",
     "combat_fainted": "${{name}} svenas!",
```
Spagy: Same comment about tense, the English string uses past, this string is present. This could be a good translation depending on context. Just making sure this isn't accidental.

IBPX:
It is acceptable to leave off the accusative if working with proper names, according to the PMEG; as long as the subject comes before the object, it's obvious enough. Leaving off the accusative in this way is what most programs I've used do. If you _really_ want to have the accusative, I believe `${{name}}-on` would be the only proper way to handle it, but don't avoid the natural form of `Subject verbed the object!".

Thanks for your work on the translation.

tuxemon/resources/db/locale/eo.json
IBPX: `${{name}} estis kaptita!` does not mean that you captured `${{name}}`. While that grammar makes sense in English, it is very unnatural in Esperanto; that sentence would seem to allude, "`${{name}}` _was_ captured, but not anymore," not that you just captured it.

Spagy:
I agree, it makes sense to leave off the accusative (or use -on) for user entered strings. If there are variables that will be replaced by something predetermined then it seems to me that using the accusative would be nice.

It's hard to see how the strings will be used because of the the game's current state.
But "${{user}} sendas ${{name}}!" seems good if ${{name}} is specified by the user.

However if ${{name}} is the type of monster, and this is defined in game, then "${{user}} sendas ${{name}}n!" seems better, to me so long as monster names are Esperantisised.

IBPX:
@Spagy I think it's better to leave off the accusative altogether. If you really, _really_ want to have it, ${{name}}-on would be better, because it functions for both Esperanto and non-Esperanto names. But really, in this case, you do not need to have the accusative, and would only be clunky to work with.

Spagy:
@IBPX I would be happy with only having the accusative where we know that the preceding variable is in Esperanto. Otherwise having no ending works for me. I don't want to sprinkle the accusative everywhere because, although it's nice for grammatically it's not so nice visually :)

IBPX:
@Spagy Well, we can't always count on the name being translated, firstly because I'm not sure if we are even _going to_ translate the names or just leave them in English, and names in third-party maps would not work well.

tuxemon/resources/db/locale/eo.json
Airon90: You're right about using "kontraŭ" instead of "al" but not about ```{{name}}n``` as explained below. I'll make the correction

tuxemon/resources/db/locale/eo.json
Airon90: The present form was already present in the transation I worked upon ([here](https://github.com/Clonewayx/Tuxemon/blob/3a85ae69b7507d413ce97eaffaa14b6aa498d0e5/tuxemon/resources/db/locale/eo_EU.json)). I didn't check it, my fault. I fix it

tuxemon/resources/db/locale/eo.json
Airon90: :+1: You're (again) right! :D

tuxemon/resources/db/locale/eo.json
Airon90: Very nice! It is so Esperantoey :) Changed!

tuxemon/resources/db/locale/eo.json
Airon90: Until the name of Tuxemon are not localized we should use the passive forme to avoid accusative. I dislike it so much but we are obliged to do so in order to avoid words like "Spookthingn". Moreover, I hope that will never be Tuxemon with a space in the name: the accusative form would be "Spook thingn". OMG
By now I'd keep the passive form. We could discuss it later. ;)

tuxemon/resources/db/locale/eo.json
Airon90: I kept the [older](https://github.com/Clonewayx/Tuxemon/blob/3a85ae69b7507d413ce97eaffaa14b6aa498d0e5/tuxemon/resources/db/locale/eo_EU.json) translation. I remove ```-anto``` as it was a bit weird to me. Now I know why ;)
Moreover [the dictionary Italian-Esperanto](http://www.swahili.it/glossword/index.php?a=term&d=2&q=sostituzione) doesn't use "anstataŭanto"

tuxemon/resources/db/locale/eo.json
Airon90: I'm not sure about "jam", as it is translated as "already". The translation would be "The monster is already fainted". It seems that it battled a little bit then it fainted as not so powerful. 
I'm not sure about "Tiu" instead of "La", too: "That monster is fainted". "That" who? 
Instead I suggest "Via" instead of "La": "Via monstro svenis", "Your monster is fainted". 
If you want to change the entire key you should see also the [original one](Tuxemon/Tuxemon/blob/development/tuxemon/resources/db/locale/en_US.json#L15) ;)

tuxemon/resources/db/locale/eo.json
Airon90: This string is not present, is imperative. I chose to translate it in imperative form as IMHO gives the nuance about the order.
The [original one](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/resources/db/locale/en_US.json#L17) is in future form, not in the past on. 
Obviously it doesn't have sense to translate in past form: "What NameOfTuxemon did?" but it already didn't do anything. LOL
I think that the imperative is correct but if you have a proposal, write it down there and let's discuss ;)

Airon90:
Some comments:
1. All names of Tuxemon have not been translated: see ([e](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L90)[x](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L96)[a](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#105)[m](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L108)[p](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L111)[l](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L118)[e](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L129)[s](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L138) (there are more)
2. I used everywhere the passive form as we don't know the content of the subject variable: see the sentence ```${{{user}}} uzis ${{{name}}}```. What if ```${{{name}}}``` is ```[Impera pocio](https://github.com/Airon90/Tuxemon/blob/patch-1/tuxemon/resources/db/locale/eo.json#L42)```, which is in Esperanto? If we add ```-n``` after ```${{{name}}}``` the result is ```${{{user}}} uzis Impera pocion```, which is grammatically incorrect. ```Impera pocio estis uzata de ${{{user}}}``` is a correct sentence ;)

tuxemon/resources/db/locale/eo.json
Spagy: I completely agree with your translation for line 17: "Kion ${{name}} faru?". I was talking about line 18: "${{name}} svenas!"

Airon90: Sorry, I didn't get what were you referring to ^^'
Oh, yes, you're right! :)

IBPX:
`Impera pocio estis uzata de ${{user}}` is a correct sentence, but it does not mean the same thing. As I said, it is acceptable to leave off the accusative, so it would be best to have the sentence be `${{user}} uzis impera pocio.`

Spagy:
To me, the sentences seem only to have a slight difference in nuance rather than a noticeable difference in meaning. Is the difference that you are talking about? If so, I agree that using the active feels slightly more natural but this doesn't seem like a major issue.

You're right, it is acceptable to leave off the accusative **on non-Esperantisised names** according to [PMEG](http://bertilow.com/pmeg/gramatiko/propraj_nomoj/ne-esperantigitaj.html#i-u5g) (I think is what you were referencing). However 'impera pocio' is not a non-Esperantisised name. So that specific part of PMEG doesn't apply. Is there another part that says dropping the accusative is okay?

IBPX:
`Impera pocio estis uzata de ${{user}}` means that `Impera pocio` _was being used_ by `${{user}}`.

If you look at any other translated program into Esperanto, they leave off the accusative when the object is a variable as well. That part of the PMEG doesn't apply to Esperanto names, but there is no system in place (and likely it won't be implemented) to check if an item's name has been translated, and then apply an accusative in the right way, so it is best to use the _natural_ form of _Subject - Verb - Object_, which even if the accusative is left off, is generally accepted. Since this form is the only way to say exactly that, it should be used.

I think the active with a small but acceptable grammatical error is better than a different but correct sentence which doesn't mean what you want to mean.

Spagy:
>`Impera pocio estis uzata de ${{user}}` means that `Impera pocio` *was being used* by `${{user}}`.

Agreed. While this might not carry exactly the same nuance as the English string I think it carries the same information and is understandable. It doesn't seem like the perfect solution to me, though, because, well… it's in the passive, which just feels odd here.

>If you look at any other translated program into Esperanto, they leave off the accusative when the object is a variable as well.

I've seen this happen in programs where the variable is in another language, usually English. I'm not sure I've seen this when the variable is in Esperanto though. Would you be able to point me to an example?

>there is no system in place (and likely it won't be implemented) to check if an item's name has been translated

Agreed. I don't think this is the problem though. As long as we translate names, then we shouldn't need a system to tell us whether a string has been translated because we know we've translated it.

> and then apply an accusative in the right way

Yeaaah, that's more of a problem. I don't know much about the current localisation system in Tuxemon but it doesn't seem to support that.

We could make sure that any string used as a variable is one word by using vortfaradaĉo and then add an -n after the variable in the parent string. E.g "${{user}} uzis ${{name}}n". But that would almost certainly produce some clumsy seeming translations though. (E.g. Imperipocio)

>so it is best to use the *natural* form of *Subject - Verb - Object*

This might be natural for some Esperanto speakers. It just feels a bit assumptive to say SVO is **the** natural form, even if it is the most common form.

>which even if the accusative is left off, is generally accepted

Maybe within the group of speakers you know, most might accept this. Quite a few Esperantists I know would be fairly strongly against dropping the accusative, because "grammar". (Quite a few also wouldn't care)

>Since this form is the only way to say exactly that, it should be used.

This seems like an odd sentiment to me. Translation is very rarely considered an exact science (or an exact art?) Some good translations aren't that similar to the originals, some are.

>I think the active with a small but acceptable grammatical error is better than a different but correct sentence which doesn't mean what you want to mean.

I think the passive with a small but acceptable semantic difference is better than a similar but grammatically incorrect sentence. :P

Here's my ranking of translations

1. "Vi uzis imperian pocion!"
2. "Imperia pocio estis uzata de vi!"
3. Vi uzis imperia pocio!"

I assume your ranking, @IBPX, would be

1. "Vi uzis imperian pocion!"
2. "Vi uzis imperia pocio!"
3. "Imperia pocio estis uzata de vi!"

Assuming the first option isn't possible, there's a bit of a problem because we disagree on which translation is better. I'm sure this is something we'd be able to resolve somehow but let's not try if there's no need.

Because the first option isn't necessarily impossible. With a better localisation system that has support for grammatical cases this would be completely possible.

A possible candidate would be [L20n](https://github.com/l20n/python-l20n).

It would seem a bit arrogant of us, as Esperantists to tell everyone they need to update to the new translation system so that we can use cases but…

Let's just take a look at the current languages that Tuxemon has been translated for

Language|Cases|Genders
---|---|---
Czech|Y|Y
German|Y|Y
English|N|N
Esperanto|Y|N
Spanish|N|Y
French|N|Y
Italian|Y|Y
Norwegian|N|Y
Portuguese|N|Y
Russian|Y|Y

(Information gathered by a quick search, may not all be correct)

Since L20n helps translating into languages with cases and with genders, we might find that quite a few translators would like the change. Though I certainly wouldn't go anywhere close to guaranteeing that.

I'm not a long-time contributor to Tuxemon so I don't feel like I've got a lot of weight to throw around and I definitely don't want to come and tread on peoples toes. But I hope this seems like a good way of making translations higher quality and more natural not just in Esperanto but across the board.

TL;DR L20n is pretty cool

IBPX:
> I've seen this happen in programs where the variable is in another language, usually English. I'm not sure I've seen this when the variable is in Esperanto though. Would you be able to point me to an example?

I can't think of any program translated into Esperanto that changes it's behavior based on whether something is an Esperanto word or not. Words translated into Esperanto may not necessarily end in a way that a computer can easily apply an accusative; There can be multiple words, some of which may be in other languages, nouns can end in ' (e.g. Esperant' instead of Esperanto), etc, and I think it's best not to impose further rules on people making custom content.

> This might be natural for some Esperanto speakers. It just feels a bit assumptive to say SVO is the natural form, even if it is the most common form.

I worded that poorly; I wasn't sugguesting (or at least wasn't meaning to sugguest) that SVO is **the** natural form, just that if an accusative is left off, it is generally still understood.

> > which even if the accusative is left off, is generally accepted

> Maybe within the group of speakers you know, most might accept this. Quite a few Esperantists I know would be fairly strongly against dropping the accusative, because "grammar". (Quite a few also wouldn't care)

I mean that in the sense of a program, where it is hard to use the accusative correctly in this sense, it is generally permisable. I too would highly object to outright dropping the accusative from Esperanto.

> I think the passive with a small but acceptable semantic difference is better than a similar but grammatically incorrect sentence. :P

I think that it is better to have a minor grammatical error (which is seen in many applications), than to use a sentence that for non-English speakers, doesn't make as much sense.

> A possible candidate would be L20n.

I had not before heard of L20n. That could be cool, though I'm not sure if everyone else would agree.

Airon90:
I (but I'm sure I'm not the only one) am completely against using non-correct translations ("Vi uzis imperia pocio"), so if you want to have the active form of the sentences ("Vi uzis imperian pocion") instead of the passive one ("Imperia pocio estis uzata de vi") and you know that it cn be done by using L20n, please open a new bug request about L20n.
It seems that L20n supports gender with Italian words so it solves another problem in Italian l10n of Tuxemon.
By now IMHO it is better to keep the sentence grammatically correct even if not completely equal to the original one (sincerely I don't know why you say the passive sentence is not equal to the active one).

Spagy:
I don't really have experience with using L20n in python, or really using python for that matter, but I'd be happy to have a look and see how feasible this is. If it seems like a reasonable thing to do I'll follow the suggestion of submitting an issue and maybe a pull request.

Airon90:
Keep in mind also that Tuxemon is in alpha stage. So many things will be changed. ;)

Spagy:
Maybe the translation system will be one of those things

Airon90:
I think so :)

Airon90:
ANother solution is to use 2 different keys, instead of only "combat_used_x": if a move is used a translation could be ```${{user}} atakis per ${{name}}``` while if a item is used the passive form can be good. What do you think?

Meanwhile, I change some translated key as I tested the translation

Spagy:
That seems like it would be a good (hopefully temporary) solution

bitcraft:
There is a bit of friction here because Tuxemon doesn't have very robust handling of grammar.  I will just recommend that translations efforts continue, but just don't expect that the results will make it into Tuxemon right away.  What I mean is, the community can maintain the correct translations for things, and eventually we will integrate them.  I don't want to stand in the way and say "We will not support this, so stop.", but realistically, we don't have enough developers to address the grammar issues right now.

I wouldn't mind if the community started a "scratch file" or "workspace" for recording the true translations.  If the community would like to make PR for a "candidate translations" file with correct grammar, I can see a place for that, and we can support it.

As far as getting the dialog in the game, I'm sorry to say that realistically the community can either: 

- accept awkward or bad translations using the current system, or
- not add translations directly, but continue to work without updating the game (make translations, but the work won't make it into the game until later).

I've looked at Transiflex and it seems like a great fit for Tuxemon, but before we dive into it, we need to decide what flavour of localization to support (ie gettext, L20n, etc), see how it related to the code we already have and move on.

I don't see the core developers addressing this until after the next release, .4, and even then, we have a lot to do to support great multiplayer experience and campaigns.

I want to say thanks for all your work so far, and I'm glad we can have this discussion!  I'm learning a lot about localization by reading the comments here and researching the libraries/services that have been recommended so far.

bitcraft:
@Airon90 @spagy  @IBPX @ShadowApex 

This has been open for a long time.  While Tuxemon won't support advanced grammar rules for a while, do you feel comfortable merging this in?  Or would you like it to stay open for reference?

ShadowApex:
I'll merge this in for now. For more on translation support, we also have the [Tuxemon forums](https://forum.tuxemon.org) for further discussion on localizing details.

Issue #208, opened 2016-08-09
Event Engine Context System


ShadowApex:
Pull request #199 prompted a discussion about potentially adding a "Context" system to the Tuxemon event engine to better coordinate things like opening dialog chains and player choices. 

With the current event system, the `dialog_chain` action requires using a special keyword (`${{end}}`) to denote the end of a dialog chain. This was originally implemented in this way because each action is inherently not aware of any of the other actions. For example, `act2 dialog_chain My name is Lampy\\, lord of all lamps` doesn't know about or have access to "act1", "act3", "act4", etc., and thus doesn't know if it's the last text in the dialog chain (once the last text in the dialog chain appears, it needs to call [open_dialog](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/event/actions/core.py#L234)). 

To resolve these kinds of problems, we should implement one or both of the following:
1. Pass the entire action list to each action, providing the executing action access to all the other actions in the action group:
   
   **core/components/event/**init**.py**    
   
   ``` python
   ...
           for action in action_list:
               self.actions[action.type]["method"](game, action, action_list)
   ...
   ```
2.  Implement new plugable context objects similar to how actions and conditions are defined. New context objects can be placed under `core/components/event/contexts`, and all loaded contexts can be executed after all events every frame:
   
   **core/components/event/**init**.py**
   
   ``` python
   ...
               if should_run:
                   self.execute_action(e['acts'], game)
   
       for context in self.contexts:
           context.execute(game)
   ...
   ```

PR #247, opened 2016-08-27
Implemented a context system

dulsi:
This implements a context system described in issue #208 . Right now dialog and dialog_chain are both kept. Eventually dialog_chain should be removed. (Same with translated versions.)


ShadowApex:
Hey @dulsi, thanks for this pull request. I've been looking at it and wanted to open a general discussion about this context system to other members of the project before we decide to pull it in. 

One of my main concerns is whether or not having a context system would add a level of unnecessary complexity to the event system. I think that having a simple and easy to understand event system allows more accessibility for outside contributors to help add new event types, etc. I think a simpler solution to the dialog chain problem would be to simply pass the whole event (with all conditions/actions) to each action method. This would operate similarly to a context, but not require you to define context objects.

Again, I mostly wanted to just start a dialog about the context system. I may be completely wrong and there may be a huge benefit for us if we add contexts. Thoughts?

dulsi:
The problem I see with passing in all the whole event is that it makes expansion difficult and you would need to pass in the number of the action as well. Why do you need to know the number of the action consider:

act1 dialog Must get donuts.
act2 dialog Must get donuts.

If you don't know which action you are you don't know you are the last dialog. For expansion consider what happens if we add "enter_word" action. You would have to update dialog, translated_dialog, dialog_choice, and translated_dialog_choice. What if we decide we want "dialog_with_icon" or "select_from_color_wheel". Each time you modify the possible dialog boxes you need to update all the events if you are passing the event object instead of the context.

ShadowApex:
Thanks for updating this PR. I think that since there's no other objects to the inclusion of the context system, I'll go ahead and accept this request. Thanks again for working on this!

Issue #251, opened 2016-09-02
[Event] Actions to execute on map load


bitcraft:
It would be very useful for maps to have a set of actions that will execute each time a map is loaded, without any user interaction.

I'll leave this issue up for any discussion about its implementation and details.

PR #258, opened 2016-11-10
Add init and interact objects to the maps.

dulsi:
This implements issue #251 

The init object type is run when the map is loaded. This allows us to avoid constantly checking events that only run once.

The interact object runs when the player is facing the tile and pressing return. This makes adding interactive squares easier for map designers. It also decreases the number of events we need to check if return is not being pressed.

Right now the code cheats a little for the interact check. I created an EventObject (I'm open to better name for the object). It has x and y coordinates. That is passed into the check_condition function since we don't have a Condition object. Ideally I'd like to see check_condition take the EventObject and Condition object. That way x and y information is only stored once and is removed from the Condition object. Same could be done with width and height but that is not in the patch at the moment.

When should the wiki be updated? Should I add information about the init and interact objects now? Wait until it is committed to the Tuxemon/Tuxemon repository? Wait until it is merged into master?

### DIFF
--- a/tuxemon/core/components/event/__init__.py
+++ b/tuxemon/core/components/event/__init__.py
@@ -97,6 +97,27 @@ class EventEngine(object):
         """
 
         if self.state == "running":
+            for e in game.inits:
+                should_run = True
+
+                # If any conditions fail, the event should not be run
+                for cond in e['conds']:
+                    # Conditions have so-called "operators".  If a condition's operator == "is" then
+                    # the condition should be processed as usual.
+                    # However, if the condition != "is", the result should be inverted.
+                    # The following line implements this.
+                    # I am not satisfied with the clarity of this line, so if anyone can express this better,
+                    # please change it.
+                    if not self.state == "running":
+                        return
+                    check_condition = self.conditions[cond.type]['method']
+                    should_run = (check_condition(game, cond) == (cond.operator == 'is'))
+                    if not should_run:
+                        break
+
+                if should_run:
+                    self.execute_action(e['acts'], game)
+            game.inits = list()
             for e in game.events:
                 should_run = True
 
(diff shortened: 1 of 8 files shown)


ShadowApex:
Thanks for this PR! I'll try and finish reviewing these changes this weekend so we can merge this. Once these changes have been accepted into the development branch, we can update the wiki for both interacting with objects and map init.

ShadowApex:
I noticed that the "init" events are normal Tiled objects with custom properties, similar to normal "event" objects. I know that Tiled also supports [custom properties](http://doc.mapeditor.org/reference/tmx-map-format/#properties) within the map itself (in Tiled you can go to Map-> Map Properties). I was originally thinking that "init" events could be put there, but thinking about it now, it might be harder to put init events that require other conditions. Do you think future "init" events might require conditions other than just to start on map load?

Since we are also defining more than one custom object for events now, we might be able to create predefined properties using the [Object Types Editor](http://doc.mapeditor.org/manual/custom-properties/) to make it easier to show map creators what kind of events are available directly from within Tiled. The editor looks like this:
Object Types Editor Screenshot

With predefined objects, we could have one for `event`, `interact`, etc. with default properties.

dulsi:
I think you may want conditions other than just first load. It could still be done like it was before in those cases but wouldn't be as elegant. I've only learned the features of Tiled through the Tuxemon map editing information so init and interact were added with my limited knowledge of what could be done. I'll take a look at the Object Types Editor when I have a chance. (It's National Novel Writing Month so don't expect much until December.)

ShadowApex:
I agree, I think it's probably better the way you implemented it. It also follows the way we currently implement normal events and interactions.

I went over this and everything looks good. I'll open another issue to address the Object Types Editor, which can be looked at separately.

Thanks again for the PR!

PR #262, opened 2017-01-16
Implement healing action

MirceaKitsune:
Today I decided it's time to attempt my first pull request for Tuxemon. I noticed that although a lot of features got done, the healing center is still unable to heal your Tuxemon! Since this is one of the basics that's still missing, and a fairly easy thing to fix, I decided to start off by giving it a shot.

This implements a set_monster_health action. It takes two optional parameters: The first is which monster slot to heal, if left empty then all are healed. The second is the amount of health to give them, if left empty the health is maxed out. The healing center map was also updated, and talking to the counter will actually heal your Tuxemon now.

Note: Although I try to code as cleanly and efficiently as possible, I'm not familiar with Tuxemon's code and Python in general. My changes might need to be checked and modified by a more experienced developer. I have tested them to be work properly, but extra testing wouldn't hurt.

### DIFF
--- a/tuxemon/core/components/event/actions/player.py
+++ b/tuxemon/core/components/event/actions/player.py
@@ -283,6 +283,9 @@ class Player(object):
         monster_health = action.parameters[1]
 
         if monster_slot:
+            if len(game.player1.monsters) < int(monster_slot):
+                return
+
             monster = game.player1.monsters[int(monster_slot)]
             if monster_health:
                 monster.current_hp = int(monster.hp * min(1, max(0, float(monster_health))))
@@ -331,6 +334,9 @@ class Player(object):
         monster_status = action.parameters[1]
 
         if monster_slot:
+            if len(game.player1.monsters) < int(monster_slot):
+                return
+
             monster = game.player1.monsters[int(monster_slot)]
             if monster_status:
                 monster.status.append(monster_status)
@@ -379,6 +385,9 @@ class Player(object):
         monster_level = action.parameters[1]
 
         if monster_slot:
+            if len(game.player1.monsters) < int(monster_slot):
+                return
+
             monster = game.player1.monsters[int(monster_slot)]
             if monster_level:
                 monster.level = max(1, monster.level + int(monster_level))
--- a/tuxemon/core/components/event/actions/player.py
+++ b/tuxemon/core/components/event/actions/player.py
@@ -293,6 +293,96 @@ class Player(object):
                     monster.current_hp = monster.hp
 
 
+    def set_monster_status(self, game, action, contexts):
+        """Changes the status of a monster in the current player's party. The action parameters
+        may contain a monster slot and the new status to be appended. If no slot is specified,
+        all monsters are modified. If no status is specified, the status is cleared.
+
+        :param game: The main game object that contains all the game's variables.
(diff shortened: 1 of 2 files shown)


MirceaKitsune:
Something I forgot to mention: Healing doesn't currently clear the poison registry! I wasn't sure how to do that, but hopefully it can be fixed before merging.

MirceaKitsune:
Fixed the problem mentioned above: I also implemented a set_monster_status, and while at it also a set_monster_level which allows in-world events to level up or reset monsters.

As always, the functions were implemented the way I saw fit at the moment; If there is a better way to use the parameters, modify my code as needed! Also I don't know how to squash commits on Github, so if you want one commit for everything please do so before merging.

ShadowApex:
Thanks for this pull request! Don't worry about having multiple commits per PR. As long as each PR fixes or implements a single feature, I don't mind having a few extra commits.

Overall the way you created these actions is written perfectly fine! I had one or two comments about this particular change before I accept the PR though:
```python
+        monster_slot = action.parameters[0]
+        monster_health = action.parameters[1]
+
+        if monster_slot:
+            monster = game.player1.monsters[int(monster_slot)]
+            if monster_health:
+                monster.current_hp = int(monster.hp * min(1, max(0, float(monster_health))))
+            else:
+                monster.current_hp = monster.hp
+        else:
+            for monster in game.player1.monsters:
+                if monster_health:
+                    monster.current_hp = int(monster.hp * min(1, max(0, float(monster_health))))
+                else:
+                    monster.current_hp = monster.hp
```

1. We should have a check on the length of `game.player1.monsters` before trying to get a particular monster index. Right now if a map event is created with `act1 set_monster_health 3,100`, but the player only has one monster in their party, the game will crash.
2. Just as a stylistic comment, I would remove the extra `else` statement and just make the first condition return. Something like this:

```python
+        if monster_slot:
+            monster = game.player1.monsters[int(monster_slot)]
+            if monster_health:
+                monster.current_hp = int(monster.hp * min(1, max(0, float(monster_health))))
+                return
+            monster.current_hp = monster.hp
+            return
+ 
+        for monster in game.player1.monsters:
+            if monster_health:
+                monster.current_hp = int(monster.hp * min(1, max(0, float(monster_health))))
+                continue
+            monster.current_hp = monster.hp
```

I'm not super strict on style though, so if you think it's clearer the way you currently have it, I can still accept it in its current state. Right now there hasn't been a clearly defined style other than trying to conform to the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/).

MirceaKitsune:
@ShadowApex Added that extra check, indeed it should be there. And I compared your suggestion to the current format, but to me it looks a lot cleaner and easier to understand with if statements instead of returns... therefore I decided to leave that part as is for now.

ShadowApex:
@MirceaKitsune Thanks for the change! With the added change, there's still an edge case where the game could crash though. If the player has more than one monster but not enough specified in the map action, it will still crash.

Basically we should check if the monster slot given in the action is less than the player's party size:

```python
+        if len(game.player1.monsters) < int(monster_slot):
+            # we should also log an error, ideally
+            return
```

MirceaKitsune:
Fixed.

PR #263, opened 2017-01-17
Append quantity to item names

MirceaKitsune:
I noticed that, when viewing items in the bag, there's no way of knowing how many you have. This patch adds the quantity to the right of each item, in a classic "name x #" format.

screenshot_20170117_013040

An obvious and important mention: The way I'm doing this isn't really the nicest and cleanest, though after looking through the code I concluded it's the only one possible. Basically I need to change the item's real name before the menu entry is generated, then revert it back afterward. The developers know best whether a better method is possible, in which case this should be redone by someone more knowing in the menu code.

### DIFF
--- a/tuxemon/core/states/items/__init__.py
+++ b/tuxemon/core/states/items/__init__.py
@@ -129,8 +129,17 @@ class ItemMenuState(Menu):
         """
         for name, properties in self.game.player1.inventory.items():
             obj = properties['item']
+            quantity = properties['quantity']
+
+            # temporarily add the quantity to the name
+            old_name = obj.name
+            obj.name += " x " + str(quantity)
+
             image = self.shadow_text(obj.name, bg=(128, 128, 128))
             yield MenuItem(image, obj.name, obj.description, obj)
+
+            # restore the original name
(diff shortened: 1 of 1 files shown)


xirsoi:
You should omit the quantity in the case there is only one.

MirceaKitsune:
@xirsoi I wanted to do that initially, but ultimately changed my mind. I looked through how most Pokemon games do this for inspiration, and noticed that some also showed " x 1" for single items. So I thought it's cleaner and more consistent to always have the number there.

If changing that is better, just add `if quantity > 1:` above `obj.name += " x " + str(quantity)`

xirsoi:
It was discussed as an idea on the forum and I liked it personally. I think it keeps the UI cleaner. I'll not argue to strongly if no one agrees with me 😉

MirceaKitsune:
There is one more thing that could be improved, though I don't know how so it's up to the developers: Making the " x #" appear at the right side of the table (right-to-left). This way it's always in the same spot, and doesn't move based on the length of the name... which is both prettier and easier to see. Hopefully this pull offers a good base on which to do that easily.

PR #268, opened 2017-01-17
Monster properties in dialogs

MirceaKitsune:
A little patch adding new text replacement variables for dialogs, capable of listing several monster properties. They operate based on party slot, so for instance ${{monster_0_name}} is the name of the first monster in your party. Properties include: Name, description, type, hp, maximum hp, level. The patch also includes a simple test case in the player's house.

The main reason for this addition is allowing NPC's to make various comments about the player's Tuxemon. For instance, someone you encounter in the world may say "Wow... you have your own Fruitera, and it's even a level 10!". As an idea for later, conditionals for these values could be implemented as well, though I don't know how to do that right now.

screenshot_20170117_150923

### DIFF
--- a/tuxemon/core/components/event/actions/core.py
+++ b/tuxemon/core/components/event/actions/core.py
@@ -59,6 +59,15 @@ class Core(object):
         text = text.replace("${{name}}", game.player1.name)
         text = text.replace(r"\n", "\n")
 
+        for i in range(len(game.player1.monsters)):
+            monster = game.player1.monsters[i]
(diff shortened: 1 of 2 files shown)


ShadowApex:
I reviewed these changes and everything looks good. Thanks for submitting this PR!

In regards to conditionals, we'd probably need to add a new condition in `core.components.event.conditions.player` to check for certain properties of the monsters in the player's party. Right now we only have the `has_monster` condition to check the type of monster.

Issue #264, opened 2017-01-17
Collision property to force continuous movement


MirceaKitsune:
I initially attempted to code this myself, but the task seems to be beyond my understanding of the Tuxemon code at this date. So for now I decided to post this as an issue, and explain what I wanted to do and why I believe it's needed.

Suggested functionality: Collision entities on maps can already take "from" and "to" parameters, for both "enter" and "exit" directions. For example: If a collision has "enter_from up" and "exit_to down", you can only step on that tile from above and exit it below. The addition I'm suggesting is a "force_to" parameter, which constrains the player to continue walking and doesn't allow them to stop at that tile. If no "exit_to" parameter is set, the player keeps walking forward... if one is set, the player is forced to walk toward the direction specified.

There are mainly two reasons for requesting this: The first are ledges, which are already used on the existing map... you can't climb them (walk upward) but can descend from them (walk downward). The problem is that when you do step down a ledge, you aren't forced to walk all the way across it, and can simply stop in the middle which looks pretty bad (a screenshot of that below). The second reason is that this will allow conveyor belts similar to classic Pokemon games... where if the player steps on a belt tile, they must travel in the direction of the belt until reaching a dead end.

![screenshot_20170117_040316](https://cloud.githubusercontent.com/assets/825418/22005218/13fc852c-dc6a-11e6-8e1d-ebda0e3e9ad3.png)

If anyone attempts this task: The most likely place to implement it seems to be _continue_move in core/components/player.py, where self.tile_destination and self.direction might be forced to a further tile once reaching the initial destination tile.

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/player.py#L862

Airon90:
Duplicate of #221

MirceaKitsune:
@Airon90 Thanks for mentioning that. Kind of the same thing, though somewhat different still: #221 seems to request actual jumping, which would be better for ledges once it's doable. My suggestion is about chained movement instead... which would also solve the ledge problem in the meantime, but will be needed separately for other things too (such as conveyor belts).

MirceaKitsune:
I've actually managed to create a pull request implementing this feature! See #269

PR #269, opened 2017-01-17
Collision property to force continuous movement

MirceaKitsune:
Pull request for #264

This implements a 'continue' parameter for collisions, which forces the player to keep walking in the specified direction after reaching the tile. For example: If a tile has "continue down" and the player steps on it from the left, the player will walk right then immediately down.

This implements support for features like ice and conveyor belts. Until jumping is implemented, I've also enabled it for the cliffs on the default map, so the player is forced to walk all the way over a ledge.

My implementation is simple and clean, so overall I'm happy with it. I tested the feature and it seems to be working, but someone else should definitely do extra testing before this is merged.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -495,4 +495,4 @@ Special Thanks
 
 ### Patreon Supporters
 * [Chris Sakkas](https://www.patreon.com/user?u=40864)
-
+* [Kelvin Shadewing](https://www.patreon.com/kelvin)
--- a/tuxemon/core/components/map.py
+++ b/tuxemon/core/components/map.py
@@ -476,6 +476,8 @@ class Map(object):
                                 tile_conditions['enter'] = enters
                             if "exit" in key:
                                 tile_conditions['exit'] = exits
+                            if "continue" in key:
+                                tile_conditions['continue'] = collision_region.properties[key]
                         collision_map[collision_tile] = tile_conditions
 
         # Similar to collisions, except we need to identify the tiles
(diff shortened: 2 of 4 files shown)


xirsoi:
I think this is exactly how "jumping" over ledges was intended to be implemented. I think we would simply want to extend this either to cover the specific case of jumping down ledges (a set of jump_<direction> script commands that would do the force move and also trigger a jump animation) or to the general case of triggering an animation in addition to the forced move.

MirceaKitsune:
@xirsoi I guess we could later add a parameter to this for a jump animation. Note that the protagonist still plays the walk animation when being moved by a tile, so perhaps a flag for disabling that could be added later as well. For now, I'd think the base functionality is what matters.

MirceaKitsune:
It seems there are actually a few issues with my code. I'm not sure how to solve them however... even this patch took a bit for me to figure out. I'd appreciate it if someone could take a look please.

1. If the continue parameter points to a colliding direction, the player can walk through the wall. I was not expecting this since I thought self.move_destination only tells the code you want to go in that direction, so it could verify collisions and manage the movement on its own.

2. On rare occasions, the player's sprite will disappear or blink, when being moved by a continuous tile. This seems to be probabilistic, and only happened a few times for me.

3. Can someone verify whether continuous tiles push NPC's as well? I don't know how to spawn them, and IIRC moving NPC entities have issues and don't work anyway. I only made sure the player is pushed, but ideally this should work for all moving entities.

PR #270, opened 2017-01-18
Offset walk animation speed with movement

MirceaKitsune:
This makes the player's animation speed properly scale with the speed of their movement. By default the effect can be seen while sprinting: Previously the player appeared as if they were gliding, whereas now they look like they're running correctly.


ShadowApex:
I wanted to see how this change looked with a run animation, instead of a walk animation before accepting this PR. After testing, I think it's a good change. Thanks for this PR!

Issue #240, opened 2016-08-26
When fighting against a wild Tuxemon don't show foe's available Tuxemons


Airon90:
See the picture:

![schermata da 2016-08-26 13-30-31](https://cloud.githubusercontent.com/assets/139841/18005409/157c3b90-6b9a-11e6-9895-f0a78684e1fe.png)

Under Tumblecoon's life there are 6 circles that must not be shown if fighting against a wild Tuxemon

dulsi:
What should the circles look like if there are tuxemons? Is the active monster a specific color? Are fainted a different color?

Airon90:
That's another question, which is not part of this bug report. Open another bug report to choose the colour of circles. By now, as there isn't any trainer in the game, it is not important.

bitcraft:
I think the circles are supposed to represent "slots" where a capture
device could go.  Currently, the slots are drawn, but they are not being
filled in with the graphic which represents an available monster.  The
images in this folder, I think, are supposed to fill the empty circles.

https://github.com/Tuxemon/Tuxemon/tree/development/tuxemon/resources/gfx/ui/icons/party

MirceaKitsune:
This appears to be fixed in the development branch, should this issue be closed? Here is a screenshot: The circles (slot icons) no longer show for the opponent, if they are a wild Tuxemon.

![screenshot_20170513_010528](https://cloud.githubusercontent.com/assets/825418/26018658/db719a84-3778-11e7-9405-f87720aeed53.png)

PR #281, opened 2017-04-07
Do not show available tuxemons for wild tuxemon.

dulsi:
This fixes issue #240.


ShadowApex:
Looks good! I like how you use the opponent's max party size to determine how many slots to draw. We'll later need to set the party size depending on if it's a wild encounter or "trainer" battle.

Thanks for the PR!

Issue #207, opened 2016-08-05
Splash screen bugs out


bitcraft:
Not sure what the cause is.  Seems to be related to hitting buttons quickly or timing related.  Occasionally, the splash screen will glitch and crash the game.  Occasionally the screen will not fade.

PR #283, opened 2017-04-10
draw matches end properly 

bitcraft:
Draw matches, where both parties are defeated in the same round are now properly handled.  There are details like, "where does experience go", etc, but this PR just addresses the bug where it will not end at all.

Also, linting, and a new locale string for draw matches.


bitcraft:
Now with more changes!  Accidentally also pushed a bugfix for #207.  :/

ShadowApex:
Looks great from my testing. Thanks for this PR @bitcraft !

PR #284, opened 2017-04-15
Update the professor's house

Qiangong2:


ShadowApex:
Hi @Qiangong2 ,

Thanks for this pull request! The professor's house looks much better than the empty room we had before.

One thing I noticed was that you used new modified versions of the tilesets that don't have any transparency. Was there any particular reason why?

Qiangong2:
The backgrounds turned out black in the game, that's why I colored them, A black backsplash and a black floor don't look good.

ShadowApex:
Instead of making a new tileset with a floor color, you can actually use a floor tile and just fill the entire first layer with that floor tile. Then if you make a second layer, and put the table, furniture, etc. on the second layer, you won't have that black floor problem.

I made a video explaining a little more about tile layers here: https://www.youtube.com/watch?v=-qbeFy1qcl8

If you change the map to use floor tiles instead of new modified tilesets, I should be able to accept this pull request.

Thanks!

PR #285, opened 2017-04-17
Update the Professor's house

Qiangong2:


Qiangong2:
layering is correct now, I added backsplash too so it does not look off

ShadowApex:
Looks great @Qiangong2 . Thanks for these changes!

Issue #266, opened 2017-01-17
Swapping in battle doesn't work


MirceaKitsune:
Now that capturing other monsters is finally implemented, I also tested in-battle swapping. Unfortunately it seems to be completely broken and has some strange side effects.

Whenever I select Tuxemon in the middle of a battle and choose to swap to another monster, the player basically attacks their own Tuxemon with a supposed swap attack! I get the message "Player used swap!" while the attack sound plays, and the first time I even see the teeth animation over my own Tuxemon plus them flashing as if they took damage! The swapping itself never occurs, and you're still stuck with the Tuxemon you started the battle with.

There is however an odd exception: If you have more than one Tuxemon, and the one currently in battle faints, you can swap to the other one without any issues. The mentioned problem only appears to occur if you choose to swap of your own accord, instead of attacking or running or using an item.

xirsoi:
I'd like to take a crack at this, if no one minds.
I've found the methods that allow swapping, and I think I can figure out why the menu part doesn't work.

MirceaKitsune:
Sure as far as I go: I'm only familiar enough with the code to play with simple things, and usually get more motivated by working on features over bugfixes. Thanks!

ShadowApex:
Right now when you select the "TUXEMON" menu item, it will call the `open_swap_menu` method in `combat_menus.py`. This method will basically push the "MonsterMenuState" into the currently active state and set up a function to be executed (called `swap_it`) when a monster is selected. The `swap_it` function will create a new "swap" technique, which right now doesn't swap the active monster.

Instead of creating a "swap" technique, you could alternatively look into making `open_swap_menu` call the `ask_player_for_monster` method inside the "CombatState":

```python
        combat_state = self.game.get_state_name("CombatState")
        for player in combat_state.human_players:
            combat_state.ask_player_for_monster(player)
```

The `ask_player_for_monster` method will not clear out the existing monster from play, though, so some additional logic might be needed to make it operate cleanly.

@bitcraft Any suggestions on how to handle switching monsters that haven't fainted?

xirsoi:
Figured it out.

`ask_player_for_monster' calls 'add_monster_into_play`

which does what it says on the tin. I think that should fix it! I'll make the change, do a couple quick tests, and have a PR up tonight.

xirsoi:
[...]

PR #291, opened 2017-04-21
Swap fix, combat sorting

bitcraft:
Fixes for #266.  Some progress on #128.  Includes misc data maintenance on the JSON DB, linting.

#266 was a quick fix, but I opted to rewrite a few key portions of the Technique and Items to make headway for #128.  Here is a summary of new features added to simplify "Meta" techniques...techniques that always run before all other actions and don't directly affect other monsters.  So far, this is Run and Swap.

#266 Fixes
=============

- Items and Techniques can define their own messages during battle (and outside)
- New messages are added the the JSON DB
- For example, Swap has no "It worked" message.
- It is possible to say "BIGFIN ate an apple!", by defining it in JSON (didn't implement this string)
- Swapping during battle works
- Items and Techniques have a "sort" attribute, which defines how they are sorted
- Combat state has new method that will sort queue.  Some work was already done, I extended it further.
- Implemented new function to rewrite actions in queue, so previous actions will hit new target

In addition to the major changes above, I also removed some duplicate data in some monster JSON files, and added AI to monsters that were missing it.

NOTE: commit message is misleading.  While I do have fixes for #207, splash screen fix, it was not commited here, but the commit message was left over.


ShadowApex:
Hey @bitcraft 

Looks great overall. I did get a crash when trying to use a potion with these changes:

```python
    main()
  File "/tmp/Tuxemon/tuxemon/core/main.py", line 91, in main
    control.main()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 542, in main
    self.main_loop()
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 603, in main_loop
    self.update(time_delta)
  File "/tmp/Tuxemon/tuxemon/core/control.py", line 206, in update
    state.update(dt)
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 153, in update
    self.update_phase()
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 388, in update_phase
    self.handle_action_queue()
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 400, in handle_action_queue
    self.perform_action(*action)
  File "/tmp/Tuxemon/tuxemon/core/states/combat/combat.py", line 656, in perform_action
    if result["capture"]:
KeyError: 'capture'
```

Maybe this should be changed to `if "capture" in result and result["capture"]`? Or just ensure that the healing effect has a "capture" key?

bitcraft:
Thanks for your detailed testing, as usual.  Seems to be working now.

Issue #290, opened 2017-04-20
Battle problems with npcs -cannot battle against npc's, only wild battle


Qiangong2:
start_battle does nothing with the npcs. No error or anything, just nothing happens. I've uploaded it in my code.

ShadowApex:
Hey @Qiangong2 . Thanks for opening this issue. I'll try and look into testing NPC battles with the latest development branch when I can.

Qiangong2:
Update: start_battle does work, but only as a wild battle. It will follow the JSON file, but it will not say "(blank) has challenged you to a battle"

bitcraft:
This is a good point to plan out how the start_battle command should work.  @ShadowApex may have some insight.  As is, there is no code to open a dialog.

@Qiangong2 what do you think start_battle should do?  We could modify it, or make a new command for what you are doing.

Qiangong2:
start_battle right now works for wild, random encounters. Maybe npc_battle? With an "animation id" in the corresponding npc JSON file. That way, it would be able to discern whether it is a "Tuxemon breeder" or "Hiker" or "Tuxemon Trainer" and the corresponding person would slide in from the side and then slide out again. Like how the player shows up during the battle at the beginning. The npc would do the same, but from the other side of the screen.

bitcraft:
Thanks for your comments.  I think that right now, start_battle is supposed to be for NPC's, but it just doesn't have the correct animation for trainers; this could be an issue with the combat state, or the action itself.  I've never tested battle with another trainer.

Next week, I will investigate other trainers in the battle.  For now, you can use start_battle.  Eventually, it will support a trainers and his monsters on the other side.  During testing, you can just pretend it works correctly.  :D

Qiangong2:
Okay, I will :D

Qiangong2:
@ShadowApex @bitcraft Well, its been six months...

Any update on this? It's not necessarily pressing at the moment, just an annoyance.

bitcraft:
Its not a priority for me at the moment, but I will get to it sometime.

Sanglorian:
Just adding to this - as well as opening dialogue and the trainer sprite/animation, it would be good to have dialogue at the end of the battle too - with different results for whether the protagonist won or lost the battle.

xirsoi:
What's the issue here? It's possible to battle a trainer already, though the experience needs improvement.

Qiangong2:
#613 and #604 need to be addressed first before we can close

[...]

Issue #289, opened 2017-04-20
remove npc sprite issue


Qiangong2:
A similar issue to before, but this time you cannot remove an npc. I added a test in my code for the npc, and you can see that it does not get removed. Now they won't stop showing :D

@ShadowApex

Qiangong2:
There is no error code

ShadowApex:
Hey @Qiangong2 , I'll take a look at this when I can. Posting your events' actions/conditions in this issue would also be helpful for anyone else who might want to take a look at this problem.

Thanks!

Qiangong2:
![screenshot - 04262017 - 11 45 04 am](https://cloud.githubusercontent.com/assets/16091790/25451087/ddc97c9e-2a75-11e7-89c3-531add453eb7.png)

Qiangong2:
There's the actions/conditions

bitcraft:
I've tested this against the development branch, and remove_npc seems to works as expected.  Maybe the conditions are not correct?  @ShadowApex thoughts?

Qiangong2:
@bitcraft Have you tried it against my dev branch?

bitcraft:
No, I haven't tested it.  What map are you having problems with?

Qiangong2:
The tux mart map, I posted the conditions I used here. If you download my source, you can see when you talk to the professor in the mart

bitcraft:
I tested your branch and remove_npc still works.  Most commands use "slugs" to make references to game objects.  In this case, your action is incorrect.  Use the slug, and it will work.  Change Act6 to "remove_npc npc_professor".  I've tested this locally.  Nice work so far.

PR #293, opened 2017-04-28
Add route configs

Qiangong2:
And signage. #289 and #290 need fixing


ShadowApex:
Hi @Qiangong2 

I took a look at these changes and they look like a great start! It's good to see the Tuxemon maps start to have more events and dialog that will help start to shape it into a fully playable game.

One addition that that we'd need before accepting this pull request is credit to the artist that made the `femaletrainer` and `professor` sprites in `CREDITS.md`. We want to ensure that everyone gets proper attribution for their artwork. Once that's done, we should be able to accept this PR.

Thanks!

PR #294, opened 2017-05-04
Add route configs

Qiangong2:
#289 and #290 need fixing.

This is updating the previous commit with credits update

Also adding Sphalian Town


Qiangong2:
@bitcraft 

Your debug code automatically gives the player two tuxemon and items, If someone were to just try out the development snapshot, it would give them an unfair advantage. I have commented it out, you can always uncomment it if needed.

bitcraft:
@Qiangong2 yes, the debug code would be unfair.  thanks for correcting it, i must have forgotten to uncomment it.

tuxemon/resources/maps/bedroom_test.tmx
```diff
@@ -61,17 +61,6 @@
     <property name="cond3" value="is button_pressed K_RETURN"/>
    </properties>
   </object>
-  <object id="11" name="Get Fruitera" type="interact" x="144" y="112" width="16" height="16">
```
bitcraft: do we want to remove fruitera from the plant?

Qiangong2: It wouldn't make sense to have it as usually you get your first Pokemon (in this case, Tuxemon) from the professor. Besides, if you leave that in, there is no reason to talk to the professor in the first place and follow the story.

ShadowApex:
@Qiangong2 Thanks for working on this!

There's still a few problems with these maps that I'll note, but that's ok! I'm going to accept this pull request because it's a great step forward in starting the actual campaign. Thanks again for your work on this! Creating maps and events also helps us figure out what parts of the engine still need work.

Also, since you've been active in creating maps, I'd highly recommend you check out the [Tuxemon Forums](https://forum.tuxemon.org). There's quite a few other contributors that also have great maps they've created. It would be great if we could work together on these. Check out this thread in particular: https://forum.tuxemon.org/thread.php?pid=1881#p1881

With that, here's what I found in my testing. Going forward, we should try and fix these:
* Some of the collision areas aren't set up yet, which lets the player walk over things they shouldn't.

* Some maps don't exist yet, which crash the game (like trying to go inside the cave):
```python
IOError: [Errno 2] No such file or directory: '/tmp/Tuxemon/tuxemon/resources/maps/route1_cave.tmx'
```
* Some of the tiles aren't on the correct layer, which can make the player look like they're standing on things they shouldn't be. 

* NPC sprites shouldn't be used as tilesets in the map. Instead, use the `create_npc` action to create NPCs. This will draw them correctly and let you use stuff like pathfinding, facing, and other NPC actions.
* When creating maps, try and make them slightly bigger than they actually are, and copy a portion of existing maps they connect to so they seem to connect seamlessly. This will give the impression that you're walking around in a cohesive world. See how `map1.tmx` connects with the route on the left. When we don't do this, we see hard black areas that we should try and avoid for outdoor areas:


Thanks!

ShadowApex:
Also, another quick note, for dialog, I would recommend using the [`translated_dialog`](https://github.com/Tuxemon/Tuxemon/wiki/event-reference#translated_dialog) action instead of just `dialog`. With `translated_dialog`, we'll also be able to display the dialog in other languages, so people from all over the world can play Tuxemon in their language. 

To use it, you just need to edit `resources/db/locale/en_US.json` with a text editor and add a new entry there with your dialog. For example, you could add an entry that looks like this:    
```json
    "professor_dialog1": "I've heard that you want to learn about Tuxemon.",
```

Then, in the map, you would have an action that looks like this:    
`translated_dialog professor_dialog1`

Using dialog that way will allow the game to check to see if there's translations for it in other languages.

Qiangong2:
@ShadowApex okay, I'll work on it. I've just been going off what's been in the source, I guess that's why the maps are somewhat broken. I'll change it to the translated_dialog as well.

Are the maps available in tmx form? If they aren't that's fine, I can still work with it

ShadowApex:
@Qiangong2 Sanglorian recently uploaded them to his fork here: https://github.com/Sanglorian/Tuxemon/tree/master/tuxemon/resources/maps/Draft%20maps

They are in tmx format.

PR #295, opened 2017-05-04
fix dialog_open condition test

bitcraft:
An easy one for you!  Previously, this condition would cause a crash.  Fixed.


ShadowApex:
Looks good! Thanks bitcraft!

PR #297, opened 2017-05-07
Draft maps

Sanglorian:
As suggested by @ShadowApex, I've put all my draft maps into the git. Also updated the credits.


ShadowApex:
Thanks @Sanglorian !

ShadowApex:
The only problem I had with this was that the tileset paths didn't appear to be correct. For future maps, if you make sure that you save the map in `resources/maps` and tilesets in `resources/gfx/tilesets`, then the tileset paths should come out ok in the map file.

I just pushed a fix that updates the paths here: https://github.com/Tuxemon/Tuxemon/commit/75ddb276b627a96ee42b208e2d57a3faf5949636

PR #306, opened 2017-05-17
map animations restored

### LINKED ISSUE #300: Animated tile effects not showing
The leaf effect no longer shows when walking through tall grass. Normally you should see a sprite animation over the tile you walked to / from, showing leafs flying outward. Sprite and command:

`tuxemon/resources/animations/tileset/grass.###.png`

`act1 play_map_animation grass,0.1,noloop,player`

Perhaps such animations got broken with the introduction of pyscroll?

bitcraft:
- fix #300 
- linting, docstrings.
- moved some animation loading to core.tools

Map tile animations that are created with actions are working again.  

Note, when testing, there are some patches of tall grass that will not animate due to the conditions in it.  In the patches nearest the player home, you must have a monster in your party, due to cond3.  Consider it a map bug.

To test without monsters, find the tall grass in the upper left corner.

Map bug:
capture

### DIFF
--- a/tuxemon/core/components/event/actions/map.py
+++ b/tuxemon/core/components/event/actions/map.py
@@ -26,20 +26,15 @@
 from __future__ import absolute_import
 
 import logging
-import os
-import re
-
-import pygame
 
 from core import prepare
-from core.components import pyganim
+from core import tools
 
 # Create a logger for optional handling of debug messages.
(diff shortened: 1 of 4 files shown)


ShadowApex:
Looks great. Thanks @bitcraft !

PR #308, opened 2017-05-20
Fix installation steps in Ubuntu

zapashcanon:
Hi,

`requirements.txt` was not available before cloning the repo.
LibShake's Makefile was expecting a backend.

I also noticed there's no more mention of Debian in the `README` from the dev branch, is there a reason it's been removed ? (Compared to `master` branch.) Steps are the same, so maybe it could be mentionned like: `Debian or derivatives: Ubuntu, Mint, etc.` ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,10 @@ Installation
 
 ```sh
 sudo apt-get install python python-pygame python-pip python-imaging python-six git
-sudo pip install -U -r requirements.txt
 git clone https://github.com/Tuxemon/Tuxemon.git
-cd Tuxemon/tuxemon
+cd Tuxemon
+sudo pip install -U -r requirements.txt
+cd tuxemon
 ./tuxemon.py
 ```
 
--- a/README.md
+++ b/README.md
@@ -27,18 +27,17 @@ Installation
 **Ubuntu**
 
 ```sh
-sudo apt-get install python python-pygame python-pip python-imaging python-six git
+sudo apt install python python-pygame python-pip python-imaging python-six git
 git clone https://github.com/Tuxemon/Tuxemon.git
 cd Tuxemon
 sudo pip install -U -r requirements.txt
-cd tuxemon
-./tuxemon.py
+tuxemon/tuxemon.py
 ```
 
 *Optional rumble support*
 
 ```sh
-sudo apt-get install build-essential
+sudo apt install build-essential
 git clone https://github.com/zear/libShake.git
 cd libShake/
 make BACKEND=LINUX; sudo make install BACKEND=LINUX
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ cd tuxemon
 sudo apt-get install build-essential
 git clone https://github.com/zear/libShake.git
 cd libShake/
-make; sudo make install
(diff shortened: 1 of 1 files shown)


bitcraft:
I think the Ubuntu/Debian instructions were just merged out of convenience, maybe?  I didn't know Ubuntu's packager is now just "apt".

BTW, if you wouldn't mind duplicating and modifying the instructions for Debian that would be awesome!  I recently closed an issue with a person who couldn't run Tuxemon on Stretch/Testing because the deps were so out of date.... using pip to install the requirements fixed it, so it would be nice to have those instructions explicitly listed for Debian as well.

zapashcanon:
>  I didn't know Ubuntu's packager is now just "apt".

It's from Debian before being from Ubuntu. Explanation [here](https://debian-handbook.info/browse/stable/sect.apt-get.html): 

> APT is a vast project, whose original plans included a graphical interface. It is based on a library which contains the core application, and apt-get is the first front end — command-line based — which was developed within the project. **apt is a second command-line based front end provided by APT which overcomes some design mistakes of apt-get**

I don't think there's a need for duplicating and modifying. I'm using Debian Unstable and it works like that. What I would recommend is just moving the section title from `Ubuntu` to `Debian or derivatives: Ubuntu, Mint, etc.`. No need for two distinct sections.

But I didn't checked everything's fine on Testing, I'll try.

bitcraft:
I think the heading would be fine "Ubuntu and Debian-based Linux".  I've found that some people don't really know that Ubuntu is derived from Debian, so you can't assume they will know that Debian commands/packages generally work for Ubuntu (and other) as well. 

Are you sure the following line works:
```
tuxemon/tuxemon.py
```

You need the ./ for the shell (bash) to execute the file.

zapashcanon:
Ok, I'll fix the heading.

> ./tuxemon/tuxemon.py

No. It's needed only if you launch a script in the current directory. E.g:

```
$ ls
script.sh  subfolder
$ script.sh 
zsh: command not found: script.sh
$ ./script.sh 
hi
$ subfolder/script.sh
hi
```

bitcraft:
> No. It's needed only if you launch a script in the current directory.

Didn't know that.  Thanks!

> python-pygame is 1.9.1 in Testing and 1.9.3 in Unstable. 1.9.1 seems pretty old, do you know if Tuxemon relies on 1.9.3 or is 1.9.1 ok too ? If it's ok with 1.9.1, then, everything is fine.

1.9.3 is several years old.  1.9.1 is even older than that!  I don't believe that Tuxemon relies on any new functionality, or that there are any bugs because of it.  So, for now, I'm fine with using debian-supplied pygame.  The pygame maintainers have recently become more active with it and builds are available on pip, but I don't see this project requiring, since most of the changes seem to be bug fixes and a few functions with obscure modules like surfarray, pixelcopy, etc.

Thanks for checking this all out.

Have you tried the OS X install recently?  Last I knew there was a libpng bug that required it to be built from an older version.  .....no need to change the install instructions, i'm just curious if it has been fixed or not.

Everything looks good to me, thank you.

zapashcanon:
> Have you tried the OS X install recently?

No, I just saw the same error than the one in the script above so I fixed it. But I'll try.

PR #309, opened 2017-05-20
Fix png files

zapashcanon:
Hi,

I fixed the `libpng warning: iCCP: known incorrect sRGB profile` warning.


bitcraft:
What was your fix for the files?

zapashcanon:
I don't remember exactly, but it was something like what's described [here](http://stackoverflow.com/a/22747902).

bitcraft:
Ok.  Looks like you just stripped the icc profile.  I think that you could consider that the 'nuclear option', and IMO probably not the best one in the long term.  I'm not familiar with the graphics programs used for our sprites, but using incorrect color profiles, (or none at all!) can create inconsistencies across platforms and displays while editing.  The best solution would be to import the files into and export from a graphics program with good working color profile handling, but i'm not against just removing them, since it seems to be the easiest method, and recommended by some.

http://juanttil.kapsi.fi/profiletest/

The link above demonstrates what an incorrect icc profile can do.  As it relates to Tuxemon, I doubt it will make a significant impact, considering that stripping all the ICC profiles effectively 'merges' them together, and i'm not even sure if pygame uses that information.

So, my only reservation is that stripping the PNG files might render their color information different when editing....but I don't really care either way...maybe our artists are not using very sophisticated software IDK.

I'll leave this open for a while, and link to the forums, in case any graphics people have an opinion about it.

https://forum.tuxemon.org/thread.php?pid=1991

bitcraft:
Looks good to me.  Should be merged in

Issue #278, opened 2017-03-28
Platform tweaks and database


bitcraft:
It would be useful for Tuxemon to include a database of supported platforms.  This information could be used to activate platform work-arounds for the display, sound, and controls.

Here is a proposal of the information to be included.  The storage format could be JSON, like everything else:

- Platform name
- Native resolution
- Preferred keyboard and joystick button mapping
- Include touch controls or not
- Sound latency (high on alsa/pulse, low on windows)
- Preferred framerate (lower on battery operated platforms)
- Clock accuracy (high on computers, low on battery)

Tuxemon could start in a compatibility-type mode to allow platform selection, or the platform could be stored in the saved settings.

Here is a partial list of platforms that would benefit:

- PC
- Android (include the sub types, like model of device)
- Raspberry Pi
- Pocket Chip & Chip

There are probably other data points that should be included.  Feel free to add them here.

Sanglorian:
This info could also be collected on the wiki while we wait for this issue to be implemented.

PR #310, opened 2017-05-20
Increase audio buffer size

### LINKED ISSUE #260: Mint 17: ALSA underrun occured
I downloaded the dev version (by cloning the repo), installed the dependencies in a virtual environment, and started the game.

Right away the sound doesn't work. I'm getting this message repeated many times:
`ALSA lib pcm.c:7843:(snd_pcm_recover) underrun occurred`

At some point the music starts playing faintly, then gets restored to normal, but I don't know how to reproduce this, it seems random.

zapashcanon:
Hi,

Currently, with `512` there'll be a lot of warnings like: `ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred` on a lot of linux machine.

While I agree `4096` is probably too much (the pygame default), `512` is not enough. I think `1024` is better - it's the default for ALSA.

So, I switched to `1024` as it won't have important impact on latency while fixing the issue with ALSA.

A better solution would be to let the user decide or set the value according to the platform, but I would involes more complicated changes. But until it's done (with #278), it seems better to fix it to the minimum value compatible with all platforms.

Fix #260.

### DIFF
--- a/tuxemon/core/platform/__init__.py
+++ b/tuxemon/core/platform/__init__.py
@@ -31,7 +31,7 @@ def init():
     # reduce sound latency.  the pygame defaults were ok for 2001,
     # but these values are more acceptable for faster computers
     if _pygame:
-        mixer.pre_init(frequency=44100, size=-16, channels=2, buffer=512)
+        mixer.pre_init(frequency=44100, size=-16, channels=2, buffer=1024)
 
 def get_config_path():
     if android:


bitcraft:
I agree that the audio latency needs a good configuration default for different systems.  Using the platform database outlined in #278 seems like the best way forward.  On my windows PC, for example, audio seems stable at 512.  Linux audio.....sigh.  So much pain.  I don't see a problem with merging this, if it helps out Linux folks.

Does this actually remove that message though?  In my experience, I see it in a broad range of games and apps on linux.  If there is sound dropping out, then it should be increased.  If the sound is fine, but there is a message, I'm more hesitant to raise the value past 1024, just to remove a error message (which can be supressed!).

FWIW, that buffer size is only one part of the problem, it also has to do with PulseAudio and poor defaults for the sound.

zapashcanon:
> Does this actually remove that message though?

Yes. That's why I opened a PR. ;)

Issue #141, opened 2016-06-19
Mod management


MirceaKitsune:
Proposing a mod management system, that would allow custom creatures / campaigns / other to be distributed and related individually: https://forum.tuxemon.org/thread.php?pid=1058

My personal suggestion is this: Right now, all resources are stored in the Tuxemon/resources directory. Why not allow multiple resource directories, which can have unique names and depend on one another? Names and dependencies could be specified in a json file located in the base of the directory, such as a depends.json. Minetest is a good example of this structure used successfully!

For example: Let's say the current Tuxemon/resources is moved to Tuxemon/mods/default and named "default". Someone wants to make a mod which adds a new set of Tuxemon... so they name the directory Tuxemon/mods/mytux and include a depends.json that tells to the engine this mod is named "mytux" and has no dependencies. Now someone wants to create a campaign using Tuxemon from this mod but tilesets from the default mod... so they put it in the directory Tuxemon/mods/mycamp, and its depends.json lists the mod as "mycamp" while specifying that it requires the presence of "default" and "mytux".

From what I'm seeing, the existing game can already be modified rather easily. But the way everything is structured into a strict set of directories makes that look more like hacking instead of doing so naturally. You also can't update various components individually, whereas this way every individual mod could be even maintained from its unique Github repository (cloned in Tuxemon/mods). Lastly, players could select which mods they want to enable or disable from a menu later on, whereas with the current structure they'd have to move files in and out of directories each time.

MirceaKitsune:
ShadowApex clarified that data packs will come in ~/.tuxemon/data. So to update my idea: Every individual mod would ideally be located under ~/.tuxemon/data/mymod. As for the default game, I imagine it could be moved from ./resources to ./data/default to follow the same architecture and be treated like every other mod.

xirsoi:
Might I suggest taking some inspiration from the modding in Paradox Interactive's Clausewitz engine?

bitcraft:
@xirsoi If you have any links or information to link in the discussion, that would help.

xirsoi:
Of course, I'm sorry!

An overview of how the modding works in the engine, as it pertains to Crusader Kings II specifically, is [here](http://www.ckiiwiki.com/Modding)

bitcraft:
That is a pretty common way of handing things where there is a hierarchy of assets.  I don't think that Tuxemon will be much different.

MirceaKitsune:
[...]

PR #318, opened 2017-05-27
Use xdg directory for config, data and cache

### LINKED ISSUE #314: XDG Base Directory Specification
Hi,

Currently, on a linux system, after running Tuxemon, the folder: `~/.tuxemon/` is created. It contains `tuxemon.cfg`, `data/` and `saves/`.

We should check if some environment variables are defined and follow [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

E.g. if `$XDG_CONFIG_HOME` is set and if it's value is `~/.config/`, we should create a folder `~/.config/tuxemon/` and put `tuxemon.cfg` into it.

I'm willing to do a PR for this. But I've got some questions. What's the `data/` folder used for ? It's empty for me.

zapashcanon:
Hi,

Fix #314.

There are a few things left to do:

* `from xdg import (XDG_CACHE_HOME, XDG_CONFIG_HOME, XDG_DATA_HOME)`  I can't put this a few lines before with the other import for a reason I don't understand (I don't use Python).
* `xdg` provides default values when the env. var is not defined, I don't know if they're valid on Androïd as I don't use and can't test on Androïd, so I kept the `/sdcard/org.tuxemon/` as a base directory, let me know if you think it should be handled differently. 
* something was strange in the locale discovery, I had to fix it but I'm not sure of what I did.
* now we have `USER_DATA_PATH = DATA_PATH`, it looks weird, I can remove this and use `DATA_PATH` everywhere if you want.
* `SAVE_PATH = DATA_PATH + "saves/slot"` looks weird too as `slot` is actually part of future files name like `saves/slot1.png` and `saves/slot2.save`. But it has nothing to do with this PR. I just noticed it.
* I also added the cache directory. If you think it will never be useful, I can remove it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ pillow
 pygame
 six
 neteria
+xdg
--- a/tuxemon/core/components/locale.py
+++ b/tuxemon/core/components/locale.py
@@ -59,8 +59,8 @@ class Translator(object):
         directories = [prepare.BASEDIR + LOCALE_PATH]
         for item in os.listdir(prepare.USER_DATA_PATH):
             item = prepare.USER_DATA_PATH + "/" + item
-            if os.path.isdir(item):
-                locale_directory = item + LOCALE_PATH
+            locale_directory = item + "/" + LOCALE_PATH
+            if os.path.isdir(locale_directory):
                 directories.append(locale_directory)
 
         return directories
--- a/tuxemon/core/platform/__init__.py
+++ b/tuxemon/core/platform/__init__.py
@@ -7,6 +7,8 @@ from os.path import expanduser
 
 _pygame = False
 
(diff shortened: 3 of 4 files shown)


bitcraft:
I think the cache directory could be useful if we decide to ship compressed data at some point, like a big zip file of resources.

MirceaKitsune:
I wonder if this could be fixed together with #141 It suggests allowing secondary data directories, in which people can install mods without having to put them alongside the original files.

xirsoi:
This PR has been hanging around for quite some time now...

Qiangong2:
@bitcraft @shadowapex

Any reason this hasn't been merged or closed yet?

Issue #317, opened 2017-05-26
Plugin format


bitcraft:
Doing some plugin work right now, part of a reorganization to support #302.  Question for developers: what do you think about simplifying the plugin format?  I'd like to remove the requirement for the ".plugin" files, and move all the plugin metadata to module-level variables.  My rationale is that it seems like busy work and a slight inconvenience to maintain 2 files for each plugin.  I'd also like to break down the plugins and make them 1:1 class:file.

As I've worked with them, I would get stuck wondering where a particular bit of code for something was and not sure if it is in the "common plugin", the "core plugin", the "player plugin, etc".  Moving all the code to a individual files would make maintenance easier, imo.

Here is a sample plugin with the proposed changes:

```python
# -*- coding: utf-8 -*-
#
# Tuxemon
# Copyright (C) 2014, William Edwards <shadowapex@gmail.com>
#
# This file is part of Tuxemon.
#
# Tuxemon is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# Tuxemon is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with Tuxemon.  If not, see <http://www.gnu.org/licenses/>.
#
# Contributor(s):
#
# William Edwards <shadowapex@gmail.com>
# Leif Theden <leif.theden@gmail.com>
#
from core.components.event.eventaction import EventAction

__plugin_meta__ = {
   "authors": ["William Edwards", "Leif Theden"],
   "version": 1.0,
   "website": "http://www.tuxemon.org",
   "description": "Pause for seconds plugin."
}

class WaitForSecsAction(EventAction):
    name = "wait_for_secs"

    def execute(self, game, action, contexts):
        """Pauses the event engine for n number of seconds.

        :param game: The main game object that contains all the game's variables.
        :param action: Retrieved from the database; contains the action's parameters
        :param contexts: Other actions that have been executed during the frame

        :type game: core.control.Control
        :type action: namedtuple
        :type contexts: dict

        :rtype: bool

        Valid Parameters: duration

        * duration (float): time in seconds for the event engine to wait for

        **Examples:**

        >>> action.__dict__
        {
[...]

Issue #302, opened 2017-05-12
combat_started error


Qiangong2:
I am trying to make it so that when you talk to an npc before the battle, it only shows once, then the battle starts.

when the condition "not combat_started" is added, you get this error:

**AttributeError: 'WorldState' object has no attribute 'battle_transition_in_progress'**

bitcraft:
Could you post the actions and condtions that you are using?

Qiangong2:
![capture](https://cloud.githubusercontent.com/assets/16091790/26023220/437fbcb2-376c-11e7-9557-4ef96d1576df.PNG)

Qiangong2:
there it is @bitcraft

bitcraft:
Qinagong2, I've tested a fix for this problem and it doesn't crash and seems to work.  Please close the issue if your problem is fixed ^_^

Qiangong2:
@bitcraft 

It still does not work, even when combat has started. The dialog box will still show up, right after the battle sequence is initiated

bitcraft:
I cannot verify this, but it seems that the dialog is opened, then the combat state is started, so technically, it is working as it should.  The issue could be that the combat state starts after the dialog, and therefore gets control.  After the battle, the game return to the dialog.  

It seems that you want to wait for the dialog to close, then start combat.  I don't know if there is a way to wait for a closed dialog.  Maybe @shadowapex knows.

Qiangong2:
What I am trying to do is have 2 dialogs. The first appears when you have not battled the npc yet or battled and lost, starting a battle, and the second only shows after you have battled the npc and won.

bitcraft:
Ok.  I think that this situation would be solved best with a new condition: won_last_battle.  It will be true if the last battle was won.

You are doing a lot of great work thank you!  You have been using more parts of the maps than there is code for...  Your opinion is important, if there is an action or condition that will help you, or if you want to change something, please tell us.

What do you think about the condition "won_last_battle"?

Qiangong2:
No problem, I've had fun working on this :)

If it could be used for a specific npc, that would work. We wouldn't want it to just activate after you battle someone else or a random encounter.

bitcraft:
Yeah.  There would have to be some kind of restriction for it to work well.

ShadowApex:
> It seems that you want to wait for the dialog to close, then start combat. I don't know if there is a way to wait for a closed dialog. Maybe @ShadowApex knows.

[...]

PR #323, opened 2017-06-23
various fixes for #302.

bitcraft:
Various fixes to support issue #302, #317, and new actions that can span multiple frames.

I've done basic testing of the game and it seems to work well without any regressions.  One goal was to remove the need for the ${{end}} token in dialog chains, but given the fact this PR has taken several weeks and has already made significant changes, I've decided to not wait any longer and release before that could be fixed.

After being merged, Tuxemon will be able to support more complex MapEvents by allowing actions to pause execution of the actions in that list.  A pause will not stop the entire engine, just that list.

Most actions will complete in less than one frame, so the "long running action" feature is optional.  For actions that do not need it, the behavior will be the same as it was before.

It is now possible to have a MapEvent defined in Tiled like the following:
```
act1 dialog Hey\, want to battle me?  I won't take no for an answer!
act2 dialog_choice Yes:No,battle_jerk
act3 start_battle npc_jerk
act4 dialog Hey\, you cheated!  No fair!
```
In this case, the dialog actions will wait until the dialog box is closed and the start_battle action will pause the list, so the final dialog will appear after the battle has finished.

The new EventEngine class API does not require complex namedtuple or parameters to work.  Here is an example of a simplified way to call an event:

```python
game.event_engine.execute("add_monster", ("txmn_tux", 25))
```

Included in the PR is some basic code cleanup, linting, docstrings, logger statements, headers, etc.

Major Changes:
* New super class for Actions and Condition
* New EventEngine class API
* Events and conditions are now 1:1 file/class for easy of maintenance
* Actions no longer need to be completed in one frame, they can span over several
* Dialog-related actions will now pause the the action list until the dialog is closed
* The battle-related actions will pause the action list until dialog is closed
* Action and Condition plugins are now a class, rather then a method
* Actions have a "valid_parameters" attribute which is a basic syntax checker
* Greatly simplified API for actions, no need to mange namedtuples
* No need for "action contexts", but the functionality is still around

Minor Changes:
* Linting
* Remove python shebang (#!/usr/bin/python) from files which are not scripts
* Remove most print statements, replace with logging statements
* New docstrings and logging messages
* added copyright header to several files


tuxemon/core/components/event/actions/start_battle.py
```diff
@@ -57,6 +57,7 @@ class StartBattleAction(EventAction):
     def start(self):
         # Don't start a battle if we don't even have monsters in our party yet.
         if not check_battle_legal(self.game.player1):
+            logger.debug("battle is not legal, wont' start")
```
airon90: "battle is not legal, wont' start" → "Battle is not legal, won't start"

Issue #326, opened 2017-07-01
Missing module components.locale


MirceaKitsune:
Latest commit of the development branch is no longer working, and crashes immediately after starting up Tuxemon. It complains about the module components.locale missing from core.components.locale import translator. Console backlog:

`BOO
Enabling logging for module: neteria.server
Enabling logging for module: neteria.core
Traceback (most recent call last):
  File "./tuxemon.py", line 61, in <module>
    main()
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/main.py", line 53, in main
    control = PygameControl(prepare.ORIGINAL_CAPTION)
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/control.py", line 101, in __init__
    self.event_engine = core.components.event.eventengine.EventEngine(self)
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/event/eventengine.py", line 117, in __init__
    self.load_plugins(path, "actions")
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/event/eventengine.py", line 132, in load_plugins
    classes = self.load_classes_from_plugins(path, category)
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/event/eventengine.py", line 151, in load_classes_from_plugins
    for cls in plugin.get_available_classes(plugins):
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/plugin.py", line 170, in get_available_classes
    for plugin in plugin_manager.getAllPlugins():
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/plugin.py", line 94, in getAllPlugins
    m = importlib.import_module(module)
  File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/event/actions/__init__.py", line 24, in <module>
    from core.components.locale import translator
ImportError: No module named components.locale`

MirceaKitsune:
Seems to be fixed in #326 Tested and confirming it works now so closing.

PR #327, opened 2017-07-01
#326: py27 regression.  fixed by futurize.  linting.

bitcraft:


MirceaKitsune:
Tested in development branch and confirming it fixes the problem.

PR #329, opened 2017-08-09
Development

Sanglorian:
I have removed non-free assets, added music for the soundtrack, added some encounters and tweaked some maps.


Qiangong2:
@Sanglorian One thing you should change is the gaming hall map. Its in csv format.

Another thing is that all maps should be lowercase without spaces (you could use underscores). It would be easier for linking without errors that way.

Qiangong2:
Overall great job though

ShadowApex:
Hi Sanglorian, thanks for this pull request!

This has a large number of changes, so it'll take me a little bit of time to review before merging these in. In future pull requests, try creating a new branch, and add only one or two maps with their assets and make a pull request from that. With fewer changes, it's much faster to review and accept.

Sanglorian:
Thanks @Qiangong2 for pointing that out. I'll fix that in the gaming hall; I've noticed it in a few other maps as well. Is there an easy way to tell which maps are in csv format, or should I just open them one by one in Tiled and see?

Thanks @ShadowApex. I'll do that in future (there's one more pull request coming that I've already done, so that'll also have a large number of changes in the one pull request).

Qiangong2:
To find if it I in CSV, open each map in a code editor like notepad++ or VSCode and if the maps have multiple lines of numbers separated by commas, then you know it is in CSV. Base64 (which every map should be in) is just one long string of letters and numbers.

CREDITS.md
```diff
@@ -34,18 +34,6 @@ Graphics
 
 ### Character Sprites
 
-* ["Player Sprite"](https://git.tuxemon.org:3000/tuxemon/tuxemon/blob/development/resources/sprites/player_front.png)
```
ShadowApex: Is there a reason why you are removing this attribution from the credits? Mike Bramson was the artist who created the player sprite that you reverted back to. In general, I would avoid removing people from the credits. If it's unused, we can put their attribution in another section of the credits to attribute people who helped make intermediate artwork.

tuxemon/resources/db/encounter/.~lock.JSON-o-matic for Encounters.ods#
```diff
@@ -0,0 +1 @@
+,BILLACER/Bill,BillAcer,08.08.2017 22:25,file:///C:/Users/Bill/AppData/Roaming/LibreOffice/4;
```
ShadowApex: This file, `tuxemon/resources/db/encounter/.~lock.JSON-o-matic for Encounters.ods# `, should not be included in the commit.

Sanglorian:
Thanks ShadowApex for the thorough look. 

> As stated in my previous comment, we shouldn't ever really remove attributions from the credits, unless under certain circumstances. Some of the attributions you removed still have artwork in the game.

I've restored Mike's credit for the protagonist overland sprite, sorry about that. I've restored the other credits but moved them to a section for assets not currently in use. 

> I'm pretty sure the maps you've removed in this commit were originally yours that you created, but as a note, we should refrain from deleting maps, even if they're unused. It's ok for them to stick around for a while, in case we might decide to use them in other parts of the game.

I don't think any maps have been removed - just renamed. If any have been removed, that's unintentional. 

> There are some layering issues on some maps, as seen here ...

These should be mostly fixed in the second omnibus pull request. 

> Some music files don't exist

I've added peasant_kingdom. If there are others missing, I think they'll be present in the second omnibus pull request. 

> The encounter files you created are missing the slug field. Try looking at default_encounter.json for an example of how it should be formatted. Right now you will get this error when trying to enter a battle

I've updated the files (and the wiki entry which had "id" instead of "slug"

> This file, tuxemon/resources/db/encounter/.~lock.JSON-o-matic for Encounters.ods#, should not be included in the commit. 

An artifact introduced by Windows; I think it's gone now.

PR #336, opened 2017-08-22
Update eo.json

airon90:
I know that it is not a priority and the game could change but I updated the translation. 

I changed the name of Karianna and Maple with [Lidja](https://en.wikipedia.org/wiki/Lidia_Zamenhof) and [Klara](https://eo.wikipedia.org/wiki/Klara_Zamenhof) in order to "esperantoize" the game but they can change later while the characters are ready: I assumed that both are good persons. I already translated the name of Tuxemon and the word "Tuxemon" itself. I don't know how to translate "Xero" (I have to know if the institute is good or bad, ...) so by now I won't change it.

I'm open to any kind of change


ShadowApex:
Looks good to me! Thanks @airon90 !

airon90:
Thank you too but I need to know if Karianna, Maple nd Xero Institute are good or bad entities :)

PR #337, opened 2017-08-22
Update en_US.json

airon90:
Fixes: capital letters and missing "s".
If the plural form of "Tuxemon" exists, the "xero_grunt2" dialog should be changed to "[...] My Tuxemons love them!"


xirsoi:
I believe Tuxemon is like Deer, in that it's plural and singular forms are the same.

Qiangong2:
Tuxemon is singular and plural, like Pokemon

ShadowApex:
I agree that Tuxemon is both a singular and a plural. The rest of these changes look good. Thanks @airon90 !

PR #344, opened 2017-09-05
Fix grammar and spelling errors in cs_CZ.json

chylex:
There are several more issues I did not fix in the PR, because they are much more complex than just changing a couple characters:

There's an issue where Czech and several other languages inflect words depending on the gender of the talking character or the subject, right now all the text assumes the player, combat opponents, and monsters are male - this does not have an easy solution, but if the engine cannot deal with this by providing an alternative for each gender, it's possible to add both options into the text, for example:

* `VYHRÁL JSI!` --> `VYHRÁL/A JSI!`
* `${{name}} byl otráven!` --> `${{name}} byl/a otráven/a!`
* `${{name}} by chtěl Duel!` --> `${{name}} by chtěl/a Duel!`
* and a lot more, most of the combat texts and several other ones scattered around the file would require context about the object's or subject's gender
* also I don't think the language has any gender-neutral constructs that can be used without dehumanizing the person (note that items also often have male/female gender, and the third gender is used pretty much exclusively when referring to the other items, and sometimes animals and babies), so more input on this would be useful

It is also impossible to directly translate texts like `"combat_used_x": "${{user}} použil ${{name}}!"` due to inflection rules. These would require inflection options for most nouns in the game, which would also need to be implemented in the engine and make things a whole lot more complicated. Unfortunately, in this case there isn't any elegant solution, the only one I can think of is referencing a subject that would normally require different inflection indirectly, for example (the indirection is marked with italics):

* `"combat_opponent_call_tuxemon": "${{user}} posílá ven ${{name}}!"`
  * would become `${{user}} posílá ven Tuxemona ${{name}}!`, which literally translates to "{{user}} is sending out *the Tuxemon* {{name}}!"
* `"received_x": "Obdržel ${{name}}!"`
  * would become `Obdržel/a jsi předmět {{name}}!`, which literally translates to "You have received *the item* {{name}}!" and also combines the mentioned approach for bypassing gender specificity
  * (I'm not sure whether this line is talking about *you* or *him/her/them*, but hopefully you get the idea)
* ...and so on

I mainly just wanted to fix some glaring spelling and grammar mistakes, but I figured it'd be worth pointing out that the engine is not prepared to handle localization for languages with these kinds of subtleties English-only speakers aren't aware of. I also just edited the file using GitHub's file editor, so hopefully the formatting doesn't get messed up.


bitcraft:
Its been discussed in the past the limitations of the localization used in Tuxemon.  At the time when we started the feature, we decided to just enable "mostly correct, but possibly wrong/minor" translations....until we/somebody can implement a better solution.

It will take a non-trivial change to enable the correct context to choose the correct strings...things like inflections, noun gender, all require a lot of changes we're currently not prepared to handle (as you know!).

**For languages with incompatible grammar with English, we just ask to make the best choices you can.**

Also, I wrote a lot about this in the past (don't feel like bringing it up, atm): In general, feel free to add translations....but please don't try an extreme amount, because I don't want the translators to keep writing/rewriting everything as the game changes.

Thank you for your efforts.  As I am not a native speaker, I can merge this since it doesn't seem to change much, and I'll trust that your changes are good for native speakers.

Some day, we will implement a better system, and the translations can be updated.

chylex:
Mozilla has a nice collection of resources on localization, like a [localization framework](http://l20n.org) with a [few examples](http://l20n.github.io/tinker/), or specific stuff like a [list of plural forms](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals). Granted, it's more focused on JS and websites, but there is a [python impl](https://github.com/l20n/python-l20n) you can look into if you want.

PR #351, opened 2017-10-18
Place Chloragon family json files in the correct directory

treetrnk:
I also added two encounters for testing out the new tuxemon.

I plan to try and update CREDITS.md today when I get a chance. I'll throw it in this PR if it's still open.


treetrnk:
Credits have been added.

Issue #347, opened 2017-10-17
Monster JSON files should not contain repetetive data


treetrnk:
Currently the monster JSON files list out things like the path to sprites and other repetitive information. Would it make more sense to set these things somewhere else for all monsters rather than specifying it in every individual db file? This could help avoid typos, and it also would make creating new monster.json files as easy as copying a template and renaming it.

## Examples of repetitive data:
```json
		"slug": "txmn_eaglace",
		"name_trans": "txmn_eaglace_name",
		"description_trans": "txmn_eaglace_descr",
		"category_trans": "txmn_eaglace_category",
 ...
		"sprites": {
				"battle1": "gfx/sprites/battle/eaglace-front.png",
				"battle2": "gfx/sprites/battle/eaglace-back.png",
				"menu1": "gfx/sprites/battle/eaglace-menu01.png",
				"menu2": "gfx/sprites/battle/eaglace-menu02.png"
		},
```

I'm thinking we could set things like `battle1 = "gfx/sprites/battle/[tuxemon]-front.png"` in some sort of config and fill the bracketed part each time we look for a monster's sprite.

Does this sound like a good idea?

Qiangong2:
Since all of the images have to be different, it isn't really too much harder to just type everything out. Also, since it is hard coded into the JSON, we don't have to worry about something else accidentallly breaking it.

ShadowApex:
@treetrnk Yeah, we could definitely simplify much of the monster JSON files if we stick to a defined naming convention. If we follow the naming convention, we can assume the path to the sprites, name, description, etc.

The only thing we'd lose from doing that is the flexibility of using sprite files that don't match the name of the monster (e.g. a monster using `placeholder-front.png` if it doesn't have a front sprite.)

bitcraft:
I agree with @ShadowApex.  It would be great to have one naming convention, then allow the resources to be found deterministically, instead of explicitly as it is now.  It has become very tedious to maintain the JSON files...I'd like to see them simplified, and ideally contain only information that cannot be determined automatically, such as initial stats, leveling characteristics, techniques, etc. 

I'd recommend something along the lines of using the unique "slug" identifier.  There was a time when there was more variance in the data, but after a few cleanups, it has gotten pretty consistent as you see it now.

As for the placeholder, I don't see a problem with the monster loader just using placeholder images automatically if the file doesn't exist in the file system.

treetrnk:
[...]

PR #357, opened 2017-10-24
Add a placeholder for missing sprites

treetrnk:
This commit adds missing.png and tells the game to use it if a sprite cannot be found instead of throwing an error and crashing the game. In case you want to test, both merlicun and firomenis have missing back sprites. Without this PR, the game crashes when you have caught one of them and try to open the tuxemon menu or when you use either of them in battle.

At some point, I vote we replace it with our own missingo, but this will work for now. 

Missingo

I also added in a couple of minor fixes like correctly making the merlicun family fire type, changing the levels of the rare_txmn encounter and removing a white line from the left side of velocitile's back sprite.


bitcraft:
Do you think it would be more appropriate to move the missing file handling to the monster class?   I like the idea of a universal "missing image" for loading, but I don't think the question mark sprite is appropriate.  A solid color that shows the dimensions of the sprite would be better for the general case, I think.  (I like the question mark is for monsters).

move this:
https://github.com/Tuxemon/Tuxemon/pull/357/files#diff-997a72d7854e725c6c93f5cecb1022e3R145

to here, maybe?
https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/monster.py#L405

treetrnk:
Yeah, I can do that. That's actually where I had it originally (before commits) because I figured it made more sense to do it once instead of three times. That makes sense though, since it's not just monster sprites using that function. 

I should be able to move it over tonight.

treetrnk:
This has been moved. I created a new method just so I didn't have to put the same if statement in there three times and in case we wanted to use it for anything else in the future. If you'd rather it be in three seperate if statements though, let me know and I'll move it all into the load_sprites method.

treetrnk:
In testing this more this morning, I found out that the new placement of this doesn't actually handle front sprites when they are pulled up for a battle. I went into a random encounter against Chillimp, but his front sprite name had a typo in it, so the game crashed. Based on the traceback, it looks like `core.states.combat.combat_animations` calls `core.components.tools.load_sprite` directly, and bypasses where this checks in monster.py. 

Any ideas of a universal way to implement this for battle sprites?

bitcraft:
> If you'd rather it be in three seperate if statements though, let me know and I'll move it all into the load_sprites method.

I'd prefer a generic function, like you implemented it.  Thats fine.

> Any ideas of a universal way to implement this for battle sprites?

In general, when I come across object loading data for pygame use (like the monster class loading pygame surfaces), I've been trying to rewrite the game state to load the graphics.  The idea being that only the game state (WorldState, CombatState, etc) should load and use images.  Otherwise, the images will remain loaded in memory for the duration of the game.  Also, moving the surface loading out of the class enables them to be using in a headless environment for automated testing or as a true headless server.

So, what I am saying is, I would make a similar function in the combat state that checks for a battle sprite and loads a default one on failure.  If you are feeling adventurous, while you are at it, maybe even play around with #347 and find the correct filename without touching the JSON.

ShadowApex:
Thanks for this PR @treetrnk ! I'm going to merge this in :)

treetrnk:
Cool. I still plan on implementing something for all battle sprites like what @bitcraft suggested above, if that sounds good to you. Hoping to get to it this week at some point.

PR #358, opened 2017-10-26
NPC Movement Mega Patch

bitcraft:
NPC Movement Loving
==================

Not many changes here for the content creators, but lots of work "under the hood", to support really nice movement of NPCs and Players.  All of the various ways of moving entities about the map have mostly been rolled into a single API that is not affected by bad framerates or small screens.  This PR addresses the following issues: #320 #114 #26 #146 #194 #166 and possibly others.

For Designers:
==============

* new npc_move command #320 
* new npc_speed command (currently, just run or walk)
* new wait command; pauses the execution of an event chain
* pathfinding npcs will try new path if they become blocked
* npc moving with npc_move will wait if they become blocked, then resume
* npcs can detect other npcs much better (some overlap is still possible)
* add starting position option in map data (configure in tiled)
* map starting position is no longer in the config (use the map editor)
* player sprite can be changed in the config
* renamed "player" sprites to "red"

For developers:
===============

* new entity class for game objects on the map (WIP)
* basic 3d physics for game entities
* all game entities use tile/second movement
* refactor npcs and players into entities
* unified pathfinding and defined path movement (much easier api)
* move pathfinding into world state (needs to be in own class eventually)
* new method of sharing data between EventConditions (based on player_moved condition)
* various other changes in effort to remove pygame hard dependencies
* remove pygame deps for use in a headless/server environment
* linting/pep8, per the norm
* players are now normal npc/entites, but controlled by game events (mostly!)
* players are returned with npcs when queried #166 
* removing special handling of player sprites makes multiplyer easier
* new debug drawing to show player sprite center point (camera debug)

Bug Fixes:
==========
* fix map animations
* fix maps being loaded twice
* fix bug where some game events were not being passed
* fix bug where "player_moved" condition was always true
* player_moved condition no longer is true when leaving a tile, only when entered
* fix lagging framerates causing glitchy movement
* screen size no longer affects npc movement speed
* fix issue where sprites on map appear to move while map scrolls
* player sprite is now centered correctly
* much, much better player input locking when changing maps
* better walking animation timing (less glitchy when stopping)
* fixed a few broken JSON entries
* prevent some EventActions from running after changing maps
* prevent movement after starting random battle (still changes facing, wont move)

And now for something completely different....

Overall, many many small changes.  I've been working on this PR now for a
couple months off and on... An hour here and an hour there sort of thing.
I'm happy to do it, but I'm going to get a bit personal here and just say
that I have found it much harder to find time to contribute to the
[...]

PR #363, opened 2017-10-28
Customizable resources directory

MirceaKitsune:
An attempt to partly address #141 for the time being. This makes the "resources" directory customizable from tuxemon.cfg, allowing the presence of multiple resource sets the user can easily switch between. While this doesn't permit the combination of multiple data directories, it allows self-contained mods that provide all necessary base assets to be selected independently.

I have already tested the changes: Temporarily renaming the directory "resources" to "myresources" then adding the option "data = myresources" to tuxemon.cfg under the section "[game]" works as expected without any noticeable flaws.


MirceaKitsune:
Fixed conflicts and adapted my code to the new map loading system. This is now compatible with the post #358 development branch. New testing indicates everything should be working fine.

MirceaKitsune:
@ShadowApex @bitcraft Any thoughts on this so far? Shouldn't take long to review it briefly, as the change is itself pretty simple.

I figure the question is whether everyone agrees with this setting being an okay solution, be it temporary or permanent (I made it with the later option in mind). I see no harm in having such a setting personally, even after a mod loader may exist later down the road; It adds a welcome bit of flexibility, and can make it easier to do testing and play around with various modified data sets.

ShadowApex:
Hi @MirceaKitsune 

I prefer that this is handled by a resource loading module that will first look in the user's directory for assets and then fall back to the built-in assets, but overall I don't mind this change. I'll go ahead and merge this at least as a stop-gap solution until we have a more robust system for mods and custom campaigns.

Thanks!

Issue #356, opened 2017-10-21
Grass doesn't draw correctly


bitcraft:
Grass tiles are not sorted correctly before drawing, so grass tiles will cover the top of tall sprites, like the player.

MirceaKitsune:
Grass does seem to draw properly for me; The lower half of the tile covers the player, whereas the upper half is drawn behind the player sprite as it should.

The only issue I'm seeing is with this order looking wrong when the player is walking through grass vertically rather than horizontally; It will appear as if the lower half of the grass is cutting through the player, due to the direction of the movement.

![screenshot_20171026_203900](https://user-images.githubusercontent.com/825418/32068568-5459f092-ba8f-11e7-9eb1-5e1c66c91f07.png)

bitcraft:
partially fixed with latest pyscroll (2.17)  some more tile sorting is needed to get more correct rendering of the top sprite of the grass.

bitcraft:
Fixed with latest pyscroll and tall sprites support.

PR #366, opened 2017-11-01
#356 Grass drawing & start_battle fix

bitcraft:
Pyscroll has a new feature that will correctly sort sprites when they are taller than the tile size.  This means that sprites like NPCs will correctly render over grass tiles.  This change enables the pyscroll feature, preliminarily called "tall sprites".

```start_battle``` is fixed by correcting the Npc loading, missing music.  a minor change to ```random_battle``` & ```start_battle``` music loading as well to support #141.


ShadowApex:
Looks good to me! I should be able to merge this in after resolving the conflicts with @MirceaKitsune 's change that allows changing the resources directory.

bitcraft:
I think that just merging my changes for start_battle and random_encounter would work.  Then make a change to ```tools.transform_resource_filename``` to support the new datadir.

PR #369, opened 2017-11-03
Story additions and text fixes

Qiangong2:
Add more story, and add more text mainly

Also, fixed text, changing from dialog_chain to translated_dialog


ShadowApex:
Thanks @Qiangong2 !

Looks good to me!

PR #374, opened 2017-11-05
Setting for clamp_camera

MirceaKitsune:
Fix for #373 This lets players choose whether to enable camera clamping, limiting the view within the map boundaries. Some players dislike the black borders visible beyond the edge of a map, whereas others dislike the player not always being in the center of the screen... with this you can choose your favorite compromise among the two solutions for handling map edges.

Issue #373, opened 2017-11-05
Limit camera view within map boundaries


MirceaKitsune:
One of biggest visual problems still pertaining to the engine, is the lack of a cropping function to limit the camera view within the map boundaries as the player approaches an edge. This allows the player to see the nothingness beyond the world's limits, represented by a black strip that looks really ugly.

I created a visual example: The top image is an unedited screenshot. The bottom image was edited to represent how my suggestion would look in comparison, with the player standing on the same tile.

![screenshot](https://user-images.githubusercontent.com/825418/32415286-04810e3a-c240-11e7-8c15-1c7e813ec5ee.png)

This should be easy to fix with a simple bound check under the camera code:

```
camera_tile_x = min(camera_tile_x, map_tiles_x - (window_resolution_x / tile_size) / 2)
camera_tile_x = max(camera_tile_x, 0 + (window_resolution_x / tile_size) / 2)
camera_tile_y = min(camera_tile_y, map_tiles_y - (window_resolution_y / tile_size) / 2)
camera_tile_y = max(camera_tile_y, 0 + (window_resolution_y / tile_size) / 2)
```

MirceaKitsune:
I feel slightly embarrassed for not noticing this before creating the issue, which I did as I was going to attempt a pull request for it; Pyscroll already offers a solution to this exact problem, which is simply disabled by a boolean. It's located at core/components/map.py line 325:

`return pyscroll.BufferedRenderer(visual_data, prepare.SCREEN_SIZE, clamp_camera=False)`

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/map.py#L325

Is there any reason why we can't set clamp_camera to True? I gave this a very brief test, and it appears to work perfectly at a first glance.

MirceaKitsune:
I feel slightly embarrassed for not noticing this before creating the issue, which I did as I was going to attempt a pull request for it; Pyscroll already offers a solution to this exact problem, which is simply disabled by a boolean. It's located at core/components/map.py line 325:

return pyscroll.BufferedRenderer(visual_data, prepare.SCREEN_SIZE, clamp_camera=False)

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/components/map.py#L325

Is there any reason why we can't set clamp_camera to True? I gave this a very brief test, and it appears to work perfectly at a first glance.

@ShadowApex @bitcraft

bitcraft:
[...]

PR #376, opened 2017-11-06
Implement map properties, camera clamping

MirceaKitsune:
Improved replacement for #374 in providing a solution to #373

This PR implements a skeleton for loading map properties from Tiled. The property "edges" with the value "clamped" will indicate the usage of clamp_camera in the PyScroll renderer. This allows maps to limit the view within the world boundaries, in case avoiding black borders at the cost of the player not always being in the center of the screen looks better on a given map.

The new property is designed to also allow an alternate value for map stitching, for later when that feature will be implemented. Setting "edges" to "stitched" should connect opposite corners and edges of the map together, allowing the player to loop indefinitely in any direction.

I took the liberty of enabling the clamped camera on all outdoor maps, without using it indoors where the classic camera looks better. Other players shared my opinion that the bounded camera doesn't make sense indoors but may look good outside, thus it seemed worth defaulting to. This was done in a separate commit, so if the map changes are rejected it can be easily reverted.

The changes were tested and seem to work fine without negative side effects. Checks are present to ensure that the properties exist, in order to avoid crashes on maps where they are absent.

### DIFF
--- a/tuxemon/core/components/map.py
+++ b/tuxemon/core/components/map.py
@@ -187,6 +187,9 @@ class Map(object):
         self.size = None
         self.renderer = None
 
+        # Initiate the properties of the map at their default values
+        self.edges = ""
+
         # Get the tile size from a tileset in our map. This is used to calculate the number of tiles
         # in a collision region.
         self.tile_size = (0, 0)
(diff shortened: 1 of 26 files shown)


ShadowApex:
This looks good @MirceaKitsune . Thanks for this change!

I also tested this and it appears to be working well. Would you also be able to add this feature to wiki.tuxemon.org in one of the map making sections so it's visible as a feature to new map creators?

Thanks!

MirceaKitsune:
@ShadowApex Done: I documented the setting under https://wiki.tuxemon.org/index.php?title=Map_Editor_Events At the end there's now a section called "Map Properties".

PR #379, opened 2017-11-08
Fix Command in Readme

Wilma456:
The tuxemon.py has no Shebang, so it will not work to run the file as a program in the Shell. To start tuxemon, you have to run it with Python.


ShadowApex:
Thanks @Wilma456 

It used to have a `#!/usr/bin/python` line, but it looks like it was removed some time in the past. We should probably have people use the `python` command anyway for those who have python installed in a non-standard location.

PR #384, opened 2017-11-16
Read settings from tuxemon.cfg when absent in user config

### LINKED ISSUE #382: Tuxemon crashes when some settings aren't present in tuxemon.cfg
Tuxemon refuses to start when certain settings are missing from the user's tuxemon.cfg file. This is especially an issue for those who run the development version, as they have to regenerate their config or manually add some new options.

As an example, I removed net_controller_enabled from my "~/.tuxemon/tuxemon.cfg" file. Instead of ignoring this and using a default value, Tuxemon crashes with the following output:

```
mircea@linux-qz0r:~/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon> python ./tuxemon.py
    from core.main import main
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/main.py", line 33, in <module>
    from . import prepare
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/prepare.py", line 77, in <module>
    CONFIG = config.Config(CONFIG_FILE_PATH)
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/config.py", line 71, in __init__
    self.net_controller_enabled = self.config.get("game", "net_controller_enabled")
  File "/usr/lib64/python2.7/ConfigParser.py", line 618, in get
    raise NoOptionError(option, section)
ConfigParser.NoOptionError: No option 'net_controller_enabled' in section: 'game'
mircea@linux-qz0r:~/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon> 
```

The solution seems pretty obvious: All options in tuxemon/core/components/config.py should be defined using the try / except Python functions. This is done for newer settings, only older ones expect the config to be present at all times. An example of how this is done right:

```
try:
    self.locale = self.config.get("game", "locale")
except configparser.NoOptionError:
    self.locale = "en_US"
```

@ShadowApex @bitcraft Do you agree that doing this for all entries is the correct solution? If so I'll make a pull request and fix all of the settings listed there.

MirceaKitsune:
Fixes #382 This implements a customized function for reading settings, capable of using default options from the builtin tuxemon.cfg file when they're missing from the user's tuxemon.cfg. Previously Tuxemon would crash if the user config was missing most options.

Additionally this cleans up the settings in config.py, grouping them under categories and placing them in the same order as tuxemon.cfg. Default values are also no longer hardcoded.

### DIFF
--- a/tuxemon/core/components/config.py
+++ b/tuxemon/core/components/config.py
@@ -38,56 +38,59 @@ import pygame
 from core.components.animation import Animation
 Animation.default_transition = 'out_quint'
 
+# read by Config and HeadlessConfig
+FILE_DEFAULT = configparser.ConfigParser()
+FILE_CONFIG = configparser.ConfigParser()
+
+# customized version of config.get
+# tries the user config first, uses the default config if the setting is missing
+def config_get(section, setting):
+    try:
+        return FILE_CONFIG.get(section, setting)
+    except configparser.NoOptionError:
+        return FILE_DEFAULT.get(section, setting)
 
 class Config(object):
     """Handles loading of the configuration file for the primary game and map editor.
 
     """
-    def __init__(self, file="tuxemon.cfg"):
-        self.config = configparser.ConfigParser()
-        self.config.read(file)
+    def __init__(self, file_default = "tuxemon.cfg", file_config = "tuxemon.cfg"):
+        FILE_DEFAULT.read(file_default)
+        FILE_CONFIG.read(file_config)
 
-        self.resolution_x = self.config.get("display", "resolution_x")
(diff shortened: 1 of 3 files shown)


MirceaKitsune:
@ShadowApex @bitcraft This has been open for a week so thought I'd poke about it. As far as I'm aware it should be ready to merge safely; I tested the changes and they work just as expected.

ShadowApex:
@MirceaKitsune Thanks for this change! Looks good!

PR #387, opened 2017-12-02
Add head bob to player front/back sprites

drummyfish:
There was head bob animation for left/right movement for the player but not up/down, which made the player look very static and inconsistent with the other direction. This change also resembles the original Pokémon games. Try it out and see for yourself :)


bitcraft:
@drummyfish Please reopen against the development branch. We don't accept PRs against master.  Also, some animations or screenshots will help community members decide if they want to make this change or not.

PR #399, opened 2018-03-19
Changes to work in Windows 10 with Python 3.6

vnmabus:
It is an error to use os.sep, as '\\' has been replaced by '/'. Also, there were repeated images with different case. If these modifications are applied, the program works in Windows 10 with Python 3.6.


bitcraft:
I'm not sure what problems you've had in the past using `os.sep`.  There are a some places that need better path handling, but hard-coding the separator isn't the best solution.  If you can, replace these with `os.path.join` instead, since it will always do the right thing for each platform.

vnmabus:
The problem is that os.sep is "\\" and that has been replaced by "/" in the previous lines.

bitcraft:
hmm.  well, i guess in this case, os.sep is not the right choice when used in a regular expression.  It seems like an easy mistake to make.  what are the changes to the images?

vnmabus:
Windows filesystem is not case sensitive by default, and these images were duplicated with different case. I have deleted the duplicated copies, and it works in Windows. However, it should be tested in a case sensitive filesystem, because it is possible that the maps are trying to access the images using both names.

bitcraft:
i've run into that problem a couple times too, where files are duplicated because of a capital letter somewhere.  super annoying.  i think we should just lower case everything and make a policy that new assets are lower case only.

i think your changes are ok, after considering the rationale behind using os.sep.  the better solution would be to rewrite the plugin loader in such a way that the regex is not used, but i think its fine for now.

PR #405, opened 2018-04-09
ru_RU locale fixes

stas730:
Warning: replace ""attempting_capture": "Попытка поимки…"," with ""attempting_capture": "Попытка поимки..."," if no Unicode support.


stas730:
Note: "Tux" can’t be translated but "Такс" is right transliteration.

bitcraft:
> if no Unicode support.

Is this an issue?  When is there not unicode support?

Issue #411, opened 2018-05-10
Save game while poisoned


andrew-lei:
Game will not save when poisoned. This is because tuxemon.core.components.technique.Technique is not JSON serialisable and tuxemon.core.components.save.save's json_data["monsters"][index]["status"] is a list of Techniques.

This seems to be fixed if save_monster and load_monster have checks for key == "status" and repeating the code for "moves", or instead of

(in save_monster)
```python
if key == "moves":
    save_data["moves"] = [i.slug for i in mon.moves]
```

```python
if key in ("moves", "status"):
    save_data[key] = [i.slug for i in getattr(mon,key)]
```

and

(in load_monster)
```python
if key == "moves":
    mon.moves = [Technique(i) for i in value]
```

```python
if key in ("moves", "status"):
    setattr(mon, key, [Technique(i) for i in value])
```

bitcraft:
Yeah it’s a known issue.  What needs to happen is a overhaul of the save system, which is why it’s been put off.  Basically, all objects which can be saved need to implement a method which returns their internal state as a python object.  Once that is done, saving and loading will be much less problematic.

bitcraft:
Related #265

kerizane:
https://github.com/Tuxemon/Tuxemon/pull/417

Issue #265, opened 2017-01-17
Saving breaks and save files become corrupted


MirceaKitsune:
After starting a new game and playing for a short amount of time, Tuxemon will report problems with saving. Attempting to load your last save also gives you a "BROKEN SAVE" label.

I'm not fully sure what triggers this yet, although I get the impression it occurs after battling and attempting to catch other monsters. When running Tuxemon from a bash console, the only relevant message I noticed is the following:

`<core.components.technique.Technique object at 0x7f64662c63d0> is not JSON serializable`

![screenshot_20170117_042334](https://cloud.githubusercontent.com/assets/825418/22005543/d90cd40a-dc6c-11e6-91c8-9e9948c7eec9.png)
![screenshot_20170117_042345](https://cloud.githubusercontent.com/assets/825418/22005544/d90da420-dc6c-11e6-84bb-216131575920.png)

bitcraft:
Thanks for reporting this.  I'll look into it.

bitcraft:
There have been several issues with the save/load code.  I'd like to overhaul it, unless any one objects.  Specifically, this is how I would approach it:

* Isolate the functions to serialize Tuxemon objects like Monsters, NPCs, and Items
* Remove redundant implementations that handle serialization
* Group the encode/decode functions
* Create class to read game state and convert it to basic python types (dict, list, int, etc)
* Feed that data to some encoder, like JSON

This approach would give some flexibility in terms of how save games are stored.  The actual implementation would be a bit like pickle, but more secure, since the ways objects are created is handled explicitly.

Also, I'd like to serialize the game screenshots and add them to the save file.

Thoughts?

ShadowApex:
Sounds great to me. In terms of serializing the screenshots, are you thinking of just base64'ing the image data and including it in the JSON save file?

What were you thinking in terms of isolating the serializing functions?

bitcraft:
> are you thinking of just base64'ing the image data and including it in the JSON save file?
That is exactly what I will do 

> What were you thinking in terms of isolating the serializing functions?
Short answer is to find all the places where game state is manipulated from save files.  It's all over the save menu, components.save, and some is in the actions...maybe some in the control class.  I'd like to create one mechanism for saving and loading state, then have just one game event trigger saving/loading.

Long answer...

[...]

PR #417, opened 2018-06-20
Save system improvements

kerizane:
Implements the redesign detailed in:
https://github.com/Tuxemon/Tuxemon/issues/265

Fixes:
https://github.com/Tuxemon/Tuxemon/issues/411

Code is backwards compatible with current save files. 

Screenshots taken with the old system will be lost, didn't seem any point in preserving that.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+cbor
 pytmx>=3.20.17
 pyscroll>=2.17.9
 pillow
--- a/tuxemon/core/components/fusion.py
+++ b/tuxemon/core/components/fusion.py
@@ -190,6 +190,15 @@ class Body(object):
         self.body_image = Image.open(self.body_image_path)
         self.face_image = Image.open(self.face_image_path)
 
+    def get_state(self):
+        if self.name:
+            return self.__dict__
+
+    def set_state(self, save_data):
+        # TODO: There's no point optimising this until Body is actually used.
+        if save_data:
+            for attr, value in save_data.iteritems():
+                setattr(self, attr, value)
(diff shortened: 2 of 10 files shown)


bitcraft:
Wow nice work man!  Its cool that you were able to solve a couple issue with this one PR.  I only have a cpuple concerns about it

* cbor is great (i use it, too!) but i'd prefer to keep the json format since its useful for debugging
* when you edit a method/function signature, please keep the docstrings up to date.

Sorry about the cbor thing, because I can see you put a lot of effort into preserving the json format, but I don't want that to be merged as-is.  I could accept it if there was a config file option to change the save/load format, then remove the logic & fallbacks when loading...for example: https://github.com/Tuxemon/Tuxemon/pull/417/files#diff-fada4d66045cd1c7a22769a1840722f0R113

Keeping the docstrings is important because smart editors like pycharm, or static analysis tools (pyflake/pylint, etc) can run a check the source code before runtime to eliminate bugs and regressions
when signature changes.  Please correct the function signatures.

Myself and ShadowApex are often in the discord. so in the future, it would be awesome to check with us about big changes like the save game format.  JSON files make it possible debug and edit save file to help development, and it would make my debugging more difficult without it.

Other than that, I'd love to merge this and fix the really annoying "cannot save when poisoned" bug!  Thank you!

kerizane:
Np, I'll get it done.

kerizane:
I'll add the config option to choose which method to save, but leave it able to load either type (no reason to break things when we don't need to).

kerizane:
Done.

Issue #409, opened 2018-05-01
"Locale Error" message when talking to salesman


MirceaKitsune:
Talking to the salesman in the Tuxemart will show a "Locale Error" message rather than the correct text.

![screenshot_20180502_024822](https://user-images.githubusercontent.com/825418/39498864-5f15f9e2-4db3-11e8-91de-29da3a71fdce.png)

Qiangong2:
I think that is because the dialog is not in the en_US json file :P 

it is probably due to me moving stuff to Translated_dialog and forgot to move that one. I will work on it in my latest commit

PR #418, opened 2018-06-21
Dialogue fix

kerizane:
Very quick fix for: https://github.com/Tuxemon/Tuxemon/issues/409

I wrote a dirty script (not worth sharing yet) that checked the rest of the dialogue for en_US, all good.


bitcraft:
Looks good to me.  I added some locale analysis tools a couple years ago.  I actually have not used the recently, but I suspect they still work:

https://github.com/Tuxemon/Tuxemon/tree/development/tests/interactive/locale

The `verify_master_locale` script only works with JSON entries.  Adding support the the TMX maps local checks to the script would be awesome,

kerizane:
Sure, in progress.

Issue #296, opened 2017-05-05
Tuxemon permanently reordered after being swapped in battle


MirceaKitsune:
When swapping Tuxemon during a battle, the party becomes permanently reordered. To test:

- Make sure you have at least two Tuxemon in your party.
- Note down their order in the Tuxemon menu.
- Enter a battle.
- Swap out the Tuxemon you started with, and swap in another Tuxemon from your party.
- Finish the battle.
- Look in your Tuxemon menu again, and note that the order has now changed.

bitcraft:
Yes, I see that happening.  So, to be clear, swapping in battle should not
change the order outside of battle?

MirceaKitsune:
@bitcraft Oh no, not at all: Swapping in-battle is only a temporary battle action, whereas permanent party reordering is always done from the main menu. In fact, I think the order of Tuxemon in this list should remain unmodified during the battle as well, as swapping shouldn't affect it at any time.

bitcraft:
I understand swaps not affecting order after battle, that is a straight
forward change.


> In fact, I think the order of Tuxemon in this list should remain
unmodified during the battle as well, as swapping shouldn't affect it at
any time.

I don't understand how this would work.  What should happen instead?
Monsters have to have their order changed for a swap to occur.

MirceaKitsune:
@bitcraft I was thinking the order of Tuxemon in the party shouldn't need to change even during battle and when swapping. I believe this is how all other monster fighting games work, by keeping the list constant unless you edit it from the main menu exclusively.

If there's a technical limitation that requires the list to be temporarily reordered during battle for swapping to work, I am okay with that, although I'm not sure if it's ideal. But the list should be reverted after the battle ends at least, which is the primary problem in this case.

bitcraft:
Ok, forgive me for being dense.  I think I may have a solution now, but it isn't going to be a quick fix.  It relies on an unimplemented part of the battlefield, where there are positions that monsters can be placed.   

In the source, it is called "layout", for now.  Right now, their position is determined by the monster order...but that has always been a cheap hack to get battle order to work correctly.  The correct way forward would be to complete the battlefield layout code, then allow monsters to be placed in the layout, independent of their order in the trainers party.

I can't make this a priority, but I will keep the issue open.  The following links shows the code for battlefield layout as it exists now.

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/states/combat/combat_animations.py#L55-L80

[...]

PR #438, opened 2018-07-24
Swapping tuxemon does not change party order

kerizane:
Fix https://github.com/Tuxemon/Tuxemon/issues/296


bitcraft:
Looks good, thank you!

Issue #428, opened 2018-07-11
Add npc shop interface


bitcraft:
Part of #421 
=======

Add interface for shopping/trade with NPCs or shops.  Ideally, one interface can be used for both.  This will require a mockup layout.  Talk to @bitcraft for help with the menu/widget system, as it is undocumented.

bitcraft:
verified

Issue #426, opened 2018-07-11
Allow npc inventory to be set in JSON DB


bitcraft:
Part of #421 
=======

Implement and test setting the starting inventory for a npc using the JSON database.

ultidonki:
I've looked at the buy/sell interface, and this can be closed, by pull request #439
You can set items in /mods/tuxemon/db/inventory, then assign it to a shop using the set_inventory <npc> action. 
So this works!

PR #439, opened 2018-07-25
Inventory template and Open Shop

kerizane:
Implemented #428 and #426.

Buying and selling works, however there's no such thing as money (what a world).
UI can use some improvement.

If you set the quantity of the item as "null" in the jsondb, then the npc will never run out of the item.


bitcraft:
Nice work, thanks!  Looking forward to playing with this.

PR #440, opened 2018-07-25
Set command line arg to allow a save file to be launched immediately

kerizane:
tuxemon.py -l 1/2/3 will launch the game and automatically boot the first/second/third save slot.


bitcraft:
Good idea!

PR #441, opened 2018-07-28
Optimise pathfinding

kerizane:
Made a pretty good improvement to the performance of path-finding.

- Stopped the game from crashing when you tried to path-find too far (about halfway) across the cotton-town map.
- After fixing that, it would take 6s to plot the path from one corner to another.
- Now it takes 0.01s!

See pathfind-wip for a more granular break-down of the time saves, if you like.

No change to the algorithm itself, the main thing was to use a set, instead of a list, for storing visited nodes. Also fixed an issue where the pathfinding would wander out of bounds, causing it to analyse extra nodes.

### DIFF
--- a/tests/test_map.py
+++ b/tests/test_map.py
@@ -1,0 +1,47 @@
+from os.path import join
+from unittest import TestCase
+import mock
+import pygame as pg
+
+from tuxemon.core import prepare
+from tuxemon.core.states.world import WorldState
+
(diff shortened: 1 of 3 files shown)


kerizane:
Yeah, it works.

bitcraft:
Very interesting it is so quick now.  Thank you!

Issue #446, opened 2018-07-30
Text drawing is slow on slow systems


bitcraft:
For embedded or mobile systems, dialog text (and all text) is slow to render.  The cause is how tenxt is rendered, where the entire line of text will be rendered each time a new letter is added to the dialog.

The fix will be to only render and blit the changed letters, not the entire line.

kerizane:
Is this still an issue with 447 and 450?

bitcraft:
@Qiangong2 ^^^

PR #447, opened 2018-08-01
Only blit the newest character for text boxes, not the entire line.

kerizane:
#446


bitcraft:
This PR eliminates the `fill` here: https://github.com/Tuxemon/Tuxemon/pull/447/files#diff-6037188ddcb61beb06998eab88e8b0f2L24

Without it, there could be artifacts in some cases if a background is required or the font is not rendered with a background color.  I don't think I've actually run into the problem before, so its probably not a big deal if the `fill` is removed.

PR #448, opened 2018-08-04
Fix issues caused by the fps refactor

kerizane:
The for loop caused Backgroundstate to be triggered, killing the active state.


bitcraft:
Looks great, thanks!

PR #450, opened 2018-08-11
Remove explicit garbage collection

kerizane:
Explicit garbage collection harms the frame-rate.

On the world map fps is 47, within a battle it drops to 30. Text drawing is slow too.

Profiling showed that garbage collection took up 50% of the run-time. 

Removing explicit gc brings the fps back up to 60 at all times.

### DIFF
--- a/tuxemon/core/control.py
+++ b/tuxemon/core/control.py
@@ -28,7 +28,6 @@ from __future__ import absolute_import, division
 
 import logging
 import time
-from collections import deque
 
 import pygame as pg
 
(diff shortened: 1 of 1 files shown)


bitcraft:
i'm fine with this for now.  eventually, i would like to enable it again because it exposes problems in the code base.  it should not be taking 50% of run time, so it makes me think there are issues can can be solved with wekref to reduce cyclical references that cause slow gc collection.

bitcraft:
I've made a comment about the fps counter.  it should be tracking the number of draws per second, not game ticks, right now anyway.  the whole refactor to decouple game ticks and drawing is not quite complete.

kerizane:
I can't see another comment about the fps. Bit confused, currently each game tick contains a draw doesn't it?

bitcraft:
I changed the event loop recently to only draw whenever it is needed.  When not drawing, it is either waiting a short amount of time, or updating the game engine.  its not optimized right now, and will need some tweaking in the future.  The FSP comment is visible if you look at "Files Changed"

kerizane:
I'm tracking the draws, not the clock ticks. It does get written every clock tick though.

bitcraft:
Ok, I see that now.  The implementation isn't very clear about what is happening.

bitcraft:
Ok, I see that now.  The implementation isn't very clear about what is happening.

Issue #453, opened 2018-08-29
can't save ( gentoo gnu/linux)


marius851000:
When I want to save I obtain a message saying that there was a problem saving ( with a newly created account ). In the log, there is these message : 
`Enabling logging of all modules.
2018-08-29 16:28:00,417 - tuxemon.core.components.rumble - ERROR - No rumble backend
s available.
2018-08-29 16:28:00,544 - tuxemon.core.components.locale - ERROR - Key 'npc_red_name
' does not exist in 'en_US' locale file.
2018-08-29 16:28:22,131 - tuxemon.core.states.persistance.save_menu - ERROR - Unable
 to save game!!
2018-08-29 16:28:22,132 - tuxemon.core.states.persistance.save_menu - ERROR - Object
 of type 'bytes' is not JSON serializable
`

This is with a just created game ( just named the character )

marius851000:
I've found that if I comment the three line which add the screenshot of the game in the save funtion, it work.

kerizane:
It works with python2, not with python3.

Needs fixing though.

bitcraft:
@marius851000 looks like a potential fix was pushed.  Do you mind testing it again?

marius851000:
Ok, save work and load work ( and display image of the save also works ).

PR #456, opened 2018-09-19
Fix save

kerizane:
#453 
Fix the screenshot saving.
Pretty-print the save file for easier debugging.


bitcraft:
Looks good, ty!

PR #460, opened 2018-10-03
Update en_US.json

airon90:
Fix:
* Tuxemon are in alphabetic order
* Tuxemon names are in capital letters, capital letters for Omnichannel)


bitcraft:
Nice attention to detail!

Issue #458, opened 2018-10-01
Improve extracting information from the Tuxepedia


AndyMender:
Currently, there is a script called `sync-wiki-monsters.py`, which as far as I could tell scrapes the Tuxepedia for Tuxemon data. I would like to improve the script and perhaps refactor it into a Python module so that the module can later be included in the main game files and the Tuxemon information used in-game (in the _Tuxedex_?), for instance from a JSON file or from a SQLite database.

I haven't worked on this project before, but I do have extensive programming experience so I would like to give a helping hand :).

**EDIT:** This is potentially related to https://github.com/Tuxemon/Tuxemon/issues/347 and might address some issues there.

**EDIT2:** After some considerations I decided it would be most convenient to use SQLite as the backend for the local copy of the Tuxepedia (highly recommended for other parts of the project as well! :)) and provide a convenient API to pull information from it into the game.

bitcraft:
I'm not sure that we are ready for a new database API.  I do like SQLite, but tuxemon targets Android, so you would have to be sure that whatever you are proposing would work there.  If you find improvements for converting to JSON, those are certainly welcome.  Anything more than that would require some planning to get right, but willing to consider it.  Before any database changes are implemented, I'd like to see #347 closed, as well as cleanup with all the related "database" classes.

AndyMender:
I'll try to have a look at #347, too. I'm honestly not hellbent on using SQLite, though I do believe it would work on Android just as well. Regardless of the presumed backend, a unified API might be more important here.

Thanks for the feedback :).

bitcraft:
There are a couple things to consider, but overall, I do support a SQL based backend.
* Many of our contributors are not programmers, so JSON was a compromise between allowing programmers and non-programmers to edit the game.  Using SQLite would need an interface to edit the game that anyone can use.
* It should be generic enough to not compromise our ability to save games.  Currently, the JSON entities are serialized to disk, so whatever API/backened is used needs to support serializing the objects to disk and loading them back.

AndyMender:
> - Many of our contributors are not programmers, so JSON was a compromise between allowing programmers and non-programmers to edit the game. Using SQLite would need an interface to edit the game that anyone can use.

In a personal project of mine I used JSON to define item and monster data, precisely because those were easier to edit and more transparent. In a group project I contributed to we used YAML with a very well defined structure for that purpose. An SQLite backend would of course require a clean and readable API not only for the rest of the game, but also for users more adept at coding.

[...]

Issue #437, opened 2018-07-24
Support alternative screen ratios


bitcraft:
Currently, everything is laid out assuming a 16:9 screen format.  It would be nice to support 4:3 screen, like some handheld units, or retro CRTs.  Mostly, the combat state is affected, but I am sure there are other places.  There should be a check when setting the screen resolution, which aspect ratio it is, and all states should have layouts for each.

bitcraft:
Very basic need to support:
* 16:9
* 4:3

PR #461, opened 2018-10-04
458 Tuxepedia and tuxemon JSON interfaces

AndyMender:
- Addresses #458 (opened by me)
- Partially addresses #347 

This PR introduces:
- A `tuxepedia` module located at "tuxemon/core/tuxepedia"
- A class for scraping the [Tuxepedia](https://wiki.tuxemon.org) wiki page called `TuxepediaWebExtractor` located in `tuxepedia.extractor`
This class is an interface based on [scripts/sync-wiki-monster.py](https://github.com/Tuxemon/Tuxemon/blob/development/scripts/sync-wiki-monsters.py), however made in a more modular fashion and importing even the tuxemon cry sound files.
- An API to sync with the Tuxepedia using `TuxepediaWebExtractor` as the backend and extract JSON data for individual tuxemon
- The API contains auxiliary methods to manage SQLite connections if that is needed in the future
- An `__init__` file defining standard local file paths as well as remote URL paths for easy reference by the API and `extractor`
- Sprites are stored at "tuxemon/resources/gfx/sprites/tuxemon"
- Sounds are stored at "tuxemon/resources/sounds/tuxemon"
- Tuxemon JSON records are stored at "tuxemon/resources/db/tuxemon"

Everything is modular so it doesn't interfere with existing code :).

Example JSON record for Aerodeer. Unfortunately, `sync-wiki-monsters.py` seems to be broken and fails when a sprite HTML `src` attribute is missing.
```json
{"Aerodeer":  
     {
         "tuxepedia_url": "https://wiki.tuxemon.org/index.php?title=Aerodeer", 
         "types": ["Untyped"], 
         "sprites": {
             "battle1": null,
             "battle2": null,
             "menu1": null,
             "menu2": null
         },
         "blurp": "",
         "call": null
     }
}

```


bitcraft:
So if I understand correctly, this is usable from the scripts/sync-wiki-monster.py script?  I think it would be nice to not have competing implementations of the sync, so could you verify that it is 1:1 replacement (no missing features?)  If so, please remove the existing code and set the script to call this one instead.

bitcraft:
Also, I think you meant to link #347 in your comment, not #437.

AndyMender:
> So if I understand correctly, this is usable from the scripts/sync-wiki-monster.py script? I think it would be nice to not have competing implementations of the sync, so could you verify that it is 1:1 replacement (no missing features?) If so, please remove the existing code and set the script to call this one instead.

The only thing I removed was the `is_valid_monster` feature since the sync is only meant to download the sprites/sounds and prep the JSONs, not to validate whether a monster record is complete. This could be done at a later stage when loading a JSON for a specific tuxemon into the game. I left the original script in place for backward compatibility. I will check both the script and my implementation for any discrepancies and fix them if there are any :).

The other issue is that the existing tuxemon JSON files contain a lot more information than just the type, sprites, sounds, name, etc. `TuxemonStore.sync_with_remote` considers this and when it encounters an existing tuxemon JSON, it will only update the fields/paths which are new and only if `overwrite` is set to `True` (default behavior).


> Also, I think you meant to link #347 in your comment, not #437.

Yes, sorry, I corrected the original post promptly :).

bitcraft:
I really like that the script can update existing monsters.  With all the changes, I think it is a good to merge.

PR #465, opened 2018-10-07
Delete fr_FR.txt

airon90:
This file is useless to the game itself and it must be moved either to the forum or to the wiki


airon90:
@marius851000

bitcraft:
Actually, I don't really mind having translation notes in the repo.  I can bring this up in the discord and get other opinions, but for now, I don't really mind and will leave this open.  We have comments and notes in the source code, having them in the translations should be useful for translators....without digging into github issues.

Issue #474, opened 2018-10-08
When suddenly leaving a new scenario the playing character disappears


airon90:
When you go in a scenario and suddenly you turn back and go in the previous one, the playing charcter disappear and all you can do is to restore the game. This doesn't happen if you wait a bit and then go back.

Qiangong2:
I have had this issue too. It isn't too frustrating on faster systems, but on slower systems like the GameShell, it can take as long as 2-3 seconds depending on which areas you move between

bitcraft:
Do you mean the game will glitch when changing maps?

airon90:
Yes, if you change the map and then change it again rapidly, the game is stuck.

bitcraft:
Ok that is def a bug that should be fixed.  FWIW, I have brought up in the
past that we don't really need to split the game map into so many
sections.  There are not any technical limitations that I am aware of to
creating big game maps
 with the Tuxemon game engine.

Qiangong2:
At least on the GameShell, the load time between big maps and smaller maps is greater than between small maps and other small maps. The More NPCs that are on a map also seem to play a factor

bitcraft:
Do you have any benchmarks on load time?  I don't have a low power device handy to check myself (my pocketchip is bricked RIP).  It would be worthwhile maybe to revisit the load times.

airon90:
OT: Is there a list of (debug) buttons used on Tuxemon?

kerizane:
@bitcraft This is fixed.
@airon90 That's the only one that I know of. If you have any suggestions feel free to add them to https://github.com/Tuxemon/Tuxemon/issues/346.

PR #480, opened 2018-10-20
Fix "map freeze" when we exit a map immediately

kerizane:
https://github.com/Tuxemon/Tuxemon/issues/474

This was an interesting bug! When we are transitioning to a new map we unlock the controls before the transition is complete; this helps the movement feel a bit more fluid. 

This means we're able to pass through the triggering tile whilst the transition is still taking place. The event is triggered multiple times but never does anything, as we don't allow two transitions to take place at the same time. We pass all the way through, and end up out of bounds, on a tile that can no longer trigger the transition.

From now on we should wait until the previous transition is completed, and then start the next one. I can't think of any issues this could cause.

### DIFF
--- a/tuxemon/core/components/event/actions/screen_transition.py
+++ b/tuxemon/core/components/event/actions/screen_transition.py
@@ -35,8 +35,12 @@ class ScreenTransitionAction(EventAction):
     ]
 
     def start(self):
+        pass
+
+    def update(self):
         world = self.game.get_state_name("WorldState")
 
         if world is not None:
             if not world.in_transition:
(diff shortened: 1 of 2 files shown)


kerizane:
Oh yeah, I didn't need to explicitly raise StopIteration, that's raised by default when I exhaust the generator.

        self._transition_updater = self.game.event_engine.execute_action_non_blocking(
            "screen_transition",
            [self.parameters.transition_time]
        )

class EventEvengine(object):
    ....
    def execute_action_non_blocking(self, action_name, parameters=None):
         ....
        return action.execute_non_blocking()

Issue #462, opened 2018-10-06
Missing locale items - improve locale/translation system?


AndyMender:
Upon starting the game with default settings I get the following error:
```
2018-10-06 21:19:52,989 - tuxemon.core.components.locale - ERROR - Key 'npc_red_name' does not exist in 'en_US' locale file.
```

Related locale issue: #409

I noticed the locale files are actually JSON files and contain a lot more than just regular translations (for instance, dialog options). It might be worthwhile to revisit the locale system and implement something more robust. A former project I participated in, Unknown Horizons, used Weblate and .po files to contain the translated text. Translation was carried out through simple text wrappers. I could try implementing something similar in Tuxemon.

Thoughts? :)

**EDIT:** I did some research and established a basic in-code translation framework using PO, POT and MO files compatible and based of `gettext`. Since some of the in-game text requires inserting variables into templates, I believe using `Jinja2` would be a good fit here :). All that's left is inserting the text wrappers around in-game text and generating text templates.

bitcraft:
The current translation system was put together in a day or two, just to get the feature working.  We have considered Jinja2 in the past.  There are a few threads in the issues (hopefully not closed), that detail the limitations.  I'll list off a couple from memory:
* cannot handle gendered nouns
* capitalization rules
* pluralization for nouns, etc.

I'm not super familiar with gettext, but as I recall it is basically used everywhere and would be a good fit.  If you'd like to work with the translation system, feel free to.

AndyMender:
Thank you for the feedback :). I'm actually already on it!

AndyMender:
As a side note, I added a `constants` module to the `tuxemon` top module to handle directory/file paths, leveraging `os.path`. Regular string concatenation is extremely risky. All standardized and reusable paths are now accessible via the `tuxemon.constants.paths` module.

AndyMender:
So apparently it's not so easy to create POT translation templates using some of Python's default tools, because the text is often hidden in JSON database files and is not readily accessible in the code. That's why it takes so long... 

The other problem is that the current translation system extracts information from the locale JSON and from the JSONs for specific objects (items, monsters, NPCs, etc.). With `gettext` based translations this is no longer necessary. The locale PO and MO files contain a full list of all translated strings (avoids duplication if a string is called multiple times in the code) for a particular locale and the JSON files defining objects can then store the actual details, not code keys like in the below JSON:
```json
{
  "slug": "item_apple",
  "name_trans": "item_apple_name",
[...]

PR #481, opened 2018-10-21
Strip repeated info from jsons

kerizane:
#347

Can still override them if necessary.

### DIFF
--- a/tuxemon/core/components/db.py
+++ b/tuxemon/core/components/db.py
@@ -39,6 +39,15 @@ from tuxemon.core import prepare
 # Create a logger for optional handling of debug messages.
 logger = logging.getLogger(__name__)
 
+SLUG_PREFIX_MAP = {
+    'encounter': None,
+    'inventory': 'inv',
+    'item': 'item',
+    'monster': 'txmn',
+    'npc': 'npc',
+    'technique': 'technique',
+}
+
 
 def process_targets(json_targets):
     """ Return values in order of preference for targeting things.
@@ -92,7 +101,6 @@ class JSONDatabase(object):
         else:
             self.load_json(directory)
 
-
     def load_json(self, directory):
         """Loads all JSON items under a specified path.
 
@@ -113,6 +121,8 @@ class JSONDatabase(object):
             with open(self.path + directory + "/" + json_item) as fp:
                 try:
                     item = json.load(fp)
+                    if not item.get('slug'):
+                        item['slug'] = get_slug_name(json_item, directory)
                 except ValueError:
                     logger.error("invalid JSON {}".format(json_item))
                     raise
@@ -122,7 +132,6 @@ class JSONDatabase(object):
             else:
                 logger.error(item, json)
                 raise Exception("Error: Item with this slug was already loaded.")
-
 
     def lookup(self, slug, table="monster"):
         """Looks up a monster, technique, item, or npc based on slug.
@@ -137,8 +146,10 @@ class JSONDatabase(object):
         :returns: A dictionary from the resulting lookup.
 
         """
-        return self.database[table][slug]
-
+        return set_defaults(
+            self.database[table][slug],
+            table
+        )
 
     def lookup_sprite(self, slug, table="sprite"):
         """Looks up a monster's sprite image paths based on monster slug.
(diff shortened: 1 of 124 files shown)


kerizane:
Do we actually need the prefixes in the slugs (txmn_, item_, technique_)? I find them a bit annoying.

AndyMender:
@kerizane we don't and the new translation system I am working on will not have them. An example monster JSON with the new translation system:
```json
{
  "slug": "bigfin",
  "category": "island",
  "description": "bigfin_description",

  "attack_base":            60,
  "defense_base":           7,
  "hp_base":                50,
  "special_attack_base":    6,
  "special_defense_base":   10,
  "speed_base":             7,

  "attack_mod":             1,
  "defense_mod":            1,
  "hp_mod":                 1,
  "special_attack_mod":     1,
  "special_defense_mod":    1,
  "speed_mod":              1,

  "exp_give_mod":           3,
  "exp_req_mod":            27,

  "ai": "RandomAI",
  "moveset": [
    {
      "level_learned": 2,
      "technique": "technique_bite"
    },
    {
      "level_learned": 2,
      "technique": "technique_water_shot"
    }
  ],
  "sprites": {
    "battle1": "gfx/sprites/battle/bigfin-front.png",
    "battle2": "gfx/sprites/battle/bigfin-back.png",
    "menu1": "gfx/sprites/battle/bigfin-menu01.png",
    "menu2": "gfx/sprites/battle/bigfin-menu02.png"
  },
  "types": [
    "water",
    "wood"
  ],
  "weight": 50
}
```
The monster `category` field is common across multiple Tuxemon, hence it doesn't need to be verbose in the monster JSON. Same goes for the `description`. Currently, item JSON files look as follow:
```json
{
  "slug": "apple",
  "description": "apple_description",
  "use_item": "combat_used_x",
  "use_success": "generic_success",
  "effects": [
    "heal"
  ],
  "power": 20,
  "sort": "food",
  "sprite": "gfx/items/apple.png",
  "target": {
    "enemy monster": 1,
    "enemy team": 0,
    "enemy trainer": 0,
    "own monster": 2,
    "own team": 0,
    "own trainer": 0,
    "item": 0
  },
  "type": "Consumable",
  "usable_in": [
    "MainCombatMenuState",
    "WorldState"
  ]
}
```
The `slug` is also a base for the name translation, hence no need for a `name` field anymore. Apologies if my work for #462 overlaps with this issue, but I had to simplify some `slug`s to avoid extra work in the new translation system.

kerizane:
Cool, well I don't want to interfere with your work. I'll drop the trans, description part.

The sprites stuff can stay, right?

There might still be value in me smoothing the way for your change. 

How about if I change the slugs to have no prefix, and deal with any knock-on from that? I want to put in a formal save-system backwards compatibility method. ATM changing the slugs will cause some data loss.

kerizane:
Cool, well I don't want to interfere with your work, I'll drop the trans, description bit, but keep the sprites.

I think there may be value in me smoothing the way for your change. How about I change the slugs to have no prefix, and deal with the knock-on for that? The save system will need a small change to avoid data loss, and I'd like to put a proper system for handling backwards incompatible updates.

I'm also pretty good at vim, if there's any mass JSON changes you're despairing at, let me know.

kerizane:
Cool, well I don't want to interfere with your work. 

If you don't mind, I'll remove the prefixes from the slugs, and deal with any knock-on impact of that. 
This will cause data-loss upon loading the save files, so I'd like to implement a backwards compatibility feature.

I'll leave the sprites change in, but won't bother with the name, category, descr changes.

kerizane:
Cool, well I don't want to interfere with your work. 

If you don't mind, I'll remove the prefixes from the slugs, and deal with any knock-on impact of that. 
This will cause data-loss upon loading the save files, so I'd like to implement a backwards compatibility feature.

I'll leave the sprites change in, but won't bother with the name, category, descr changes.

AndyMender:
The biggest difficulty is refactoring JSON fields and values into the PO `msgid` and `msgstr` system. For now I'm working with the en_US locale. I'll let you know if I run into any issues, though :).

bitcraft:
@kerizane @AndyMender Thank you for working out the JSON conflicts.  Just a heads up, there could be some blocking bugs after these are merged in, and I also have a beg PR in the works for input handling.  In light of these three big PRs, Id like to ask that development be slowed down or freeze for a week after they are merged in, so that I/we can go through the code base and just clean everything up.

Any scripts that you produce for the migration, please add to the "scripts' folder in case they become required to clean up the JSON after merging.  @AndyMender you may also consider updating your wiki sync script to do the following:

* Compare what is in the local DB to the wiki
* Log/print information found in the local DB, but not wiki

With the above feature, you can give @Sanglorian a list of missing info (if any), to update the wiki.  Then we can clear the monster db, and create a new fresh one from the script using the new format with changes from @kerizane's combat PR.  That would potentially save a lot of manual effort transcribing the old JSON format to the new one, after these big PRs.  This is optional for you, but it might save you a lot of effort.

bitcraft:
@kerizane 
> Do we actually need the prefixes in the slugs (txmn_, item_, technique_)? I find them a bit annoying.

Lets move the discussion here: #482

The prefixes are to prevent collisions when searching for slugs.  Suppose we have a txmn named "Potion", and also a item named "Potion".  Stupid example, I realize, but there could be some situation like that eventually.  It makes it less ambiguous when making reference to other things from the JSON.

I agree though, it worked well for a while, but it is tedious and should be updated.

PR #483, opened 2018-10-22
Figure out sprite location if not set in json

kerizane:
Split this off from my aborted PR, as it shouldn't conflict with Andy's work.

### DIFF
--- a/tuxemon/core/components/db.py
+++ b/tuxemon/core/components/db.py
@@ -137,8 +137,10 @@ class JSONDatabase(object):
         :returns: A dictionary from the resulting lookup.
 
         """
-        return self.database[table][slug]
-
+        return set_defaults(
+            self.database[table][slug],
+            table
+        )
 
     def lookup_sprite(self, slug, table="sprite"):
         """Looks up a monster's sprite image paths based on monster slug.
@@ -162,3 +164,27 @@ class JSONDatabase(object):
 
         return results
 
+
+def set_defaults(results, table):
+    if table == "monster":
+        slug = results['slug']
+        name = slug.partition('_')[2]
+
+        sprites = results.setdefault(
+            "sprites",
+            {}
+        )
+
+        for key, view in (
+                ('battle1', 'front'),
+                ('battle2', 'back'),
+                ('menu1', 'menu01'),
+                ('menu2', 'menu02'),
+        ):
+            if not results.get(key):
+                sprites[key] = "gfx/sprites/battle/{}-{}".format(
+                    name,
+                    view
+                )
+
+    return results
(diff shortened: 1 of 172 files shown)


tuxemon/core/components/db.py
```diff
@@ -162,3 +164,27 @@ def lookup_sprite(self, slug, table="sprite"):
...
+            {}
+        )
+
+        for key, view in (
+                ('battle1', 'front'),
+                ('battle2', 'back'),
+                ('menu1', 'menu01'),
+                ('menu2', 'menu02'),
+        ):
+            if not results.get(key):
```
bitcraft: I think we should go all-in on automatically finding the art assets and drop support for specifying images in the json.  If a monster doesn't have the right format for the images, then the fix is to rename the images, not edit JSON.  The less code the better :D

So, look for the image and raise FileNotFound error with the monster and filename if the image cannot be found.

bitcraft:
https://github.com/Tuxemon/Tuxemon/pull/483/files#r227096262

tuxemon/core/components/db.py
AndyMender: As far as I remember, the wiki scraping script I wrote creates generic file names ('menu01`, `menu02`, etc.), retaining the correct image file extension (mostly .png or .jpg). It did not address the in-JSON mapping, though.

kerizane: I could make it deal with it having an unknown file extension.

Do we want to enforce a specific file extension, probably png?

bitcraft: file extension is a good point.  i don't see any drawbacks to enforcing png at this time.  jpg is a lossy format and not really suitable for game development.  PIL/Pillow/imagemagick/pygame can all convert any jpg assets from the wiki into png, but we should log the jpg ones and get the artists to provide proper pngs.

Sanglorian: The majority of the images will be GIF - I don't think we have any JPG sprites. Which as I understand is not lossy, provided the palette is small enough - is that right?

AndyMender: So far I've seen .gif, .png and .jpg. I have no opinion on which format is better, though.

bitcraft: * GIF only supports 256 colors and does not compress well.  It really needs to be phased out.
* JPG is lossy and really only suitable for backgrounds, and only for release.
* PNG compresses well, supports 32-bit color, alpha, lossless, well supported.

bitcraft: So just clarifying, go ahead a collect all file types for now, but in the future, I would like to only support png images for all assets.

kerizane:
Done.

bitcraft:
Looks good, thank you!

PR #485, opened 2018-11-01
Update en_US.json

airon90:
Just a proposal. Use "Tuxemon" instead of a generic "monster". Moreover, which one is correct, "That Tuxemon *has* fainted!" or "That Tuxemon *is* fainted!"?

### DIFF
--- a/tuxemon/resources/db/locale/en_US.json
+++ b/tuxemon/resources/db/locale/en_US.json
@@ -23,7 +23,7 @@
     "combat_capturing_fail": "${{name}} broke free!",
     "combat_recharging": "${{name}} needs a break before using ${{move}} again...",
     "combat_replacement": "Choose a replacement!",
-    "combat_replacement_is_fainted": "That Tuxemon has fainted!",
+    "combat_replacement_is_fainted": "That Tuxemon already knocked out!",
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thanks @airon90  -- I've gone back and forth on "monster" or "Tuxemon" - am interested to hear what others think too.

"That Tuxemon _has_ fainted!" or "That Tuxemon _is knocked out!_"

AndyMender:
> "That Tuxemon has fainted!" or "That Tuxemon is knocked out!"

I believe there is a very strong contextual and semantic difference between those two. `That Tuxemon has fainted!` may refer to a current event (for instance, a Tuxemon in battle has just run out of HP), while `That Tuxemon is knocked out!` could be a notification when the player is trying to swap in a Tuxemon which fainted before. To make it even clearer, I think the notification should read `That Tuxemon/monster is already knocked out!`

bitcraft:
I agree, Tuxemon is preferable over monster, and "already knocked out!" is more clear.  If we all agree, then please make those changes and I can merge it in.

airon90:
It can be merged.

bitcraft:
Looks good, ty!

PR #486, opened 2018-11-01
Update it_IT.json

airon90:
Not sure about those translations. I have to check them in the game, but by now they are ok. I couldn't translate one of the techniques


Sanglorian:
Thanks @airon90 ! Which was the technique you couldn't translate?

bitcraft:
Partial translations are fine, thank you for the update!

airon90:
@Sanglorian I translated all but "Tonguespears". I need to know better what it does in order to translate it. I also need to know better other techniques in order to make other translations better but by now I think it iss the best

Sanglorian:
Ah yes, Tonguespear involves stabbing the opponent with a sharp tongue, and then sucking the blood out. Like how a mosquito sticks its needle or proboscis into your skin to suck your blood.

bitcraft:
Maybe use the word `proboscide` in it? https://translate.google.com/#en/it/proboscis

airon90:
> Ah yes, Tonguespear involves stabbing the opponent with a sharp tongue, and then sucking the blood out. Like how a mosquito sticks its needle or proboscis into your skin to suck your blood.

Thank you! I will think about a translation, which is not used in Pokémon too in order to avoid any kind of copyright infringement. Where can I find these info for other techniques too? I think the [wiki](http://wiki.tuxemon.org) should have all of them.

> Maybe use the word `proboscide` in it? https://translate.google.com/#en/it/proboscis

Thank you but `Proboscide` is used to translate `Proboscis`, which is another technique.

Sanglorian:
> Thank you! I will think about a translation, which is not used in Pokémon too in order to avoid any kind of copyright infringement. Where can I find these info for other techniques too? I think the [wiki](http://wiki.tuxemon.org) should have all of them.

No worries!

The wiki does have [all the techniques now](https://wiki.tuxemon.org/index.php?title=Category:Technique), but it sounds like you're also thinking of descriptions of each technique - is that right? We haven't written any of those so far, but it's a good idea to do at some point.

airon90:
> you're also thinking of descriptions of each technique - is that right?

Yes. I imagine the wiki as a place where all possible non-coding information about something in Tuxemon is present, like [Bulbapedia](http://bulbapedia.bulbagarden.net/) for Pokémon (see [example](https://bulbapedia.bulbagarden.net/wiki/Thunderbolt_(move))).

> it's a good idea to do at some point

If you tell me where to find information I can help you. I also asked for an account on the wiki but I have problems.

Sanglorian:
It's actually just that the non-coding information doesn't exist anywhere! But thanks for the offer to help :)

Sorry about the account delay on the wiki - @ShadowApex , are you able to help @airon90 ?

PR #494, opened 2018-11-26
Don't provide Fruitera as a starter

kerizane:
Fruitera has temporarily been taken out of the game as it's incomplete.

I'm removing Fruitera as a starter for now, to stop the game from crashing if you choose it.


bitcraft:
Removing the Fruitera choice sounds fine.  Map editing is a pain though, do you mind just removing the option to select it on line 69, but leaving the conditions and actions in?  It shouldn't affect the game, and will be easier to add Fruitera back as an option when it is fixed.

Issue #495, opened 2018-12-21
Starter Tuxemon have 0 HP


MirceaKitsune:
When the professor gives you your starters (tested with Rockitten) they come at 0 HP. Oddly enough, talking to the Tuxcenter lady to heal them doesn't fix it either.

bitcraft:
Is this with latest development version?

MirceaKitsune:
Latest Git on development branch, yes.

AndyMender:
I can confirm that this is the case also in my branch under refactoring (rebased from `development`, of course).

MirceaKitsune:
Correct: I could enter combat too, but instantly died the moment I was hit.

AndyMender:
> Ok. This needs two fixes:
> 
>     Potions, spells, and events should be able to raise HP if HP is 0. (verify this is correct)
>     When Tuxemon are created, or when event creates them, HP should default to 100%

Not sure about spells (you mean techniques?) and events, but potions can raise HP if it's at 0 already. That part works. I even tested healing a _freshly minted_ Tuxemon at a Tuxecenter and this too works as intended. The primary issue is defaulting Tuxemon HP to 100% when received from the Professor (but not when captured, I think!).

bitcraft:
Yeah s/spells/techniques.  HP and status are not directly related.  Meaning that HP could be 0, while the status is not KO'd.  So, you've verified #1.
> The primary issue is defaulting Tuxemon HP to 100% when received from the Professor

Right.  The txmn given to the player are created by events (and the Tuxemon class) so those need to be checked and fixed so that the event creates TXMN with 100% HP.

bitcraft:
@MirceaKitsune Please update and test then close issue if the problem is fixed.

MirceaKitsune:
Confirming this is solved in latest Git version. Healing at the Tuxcenter also works again.

PR #499, opened 2019-01-02
add_monster creates monster with full health

kerizane:
#495 
Probably a regression from my combat update.

FYI, monsters were able to enter battle with 0hp as we have an explicit "fainted" status, which is added when lethal damage is done in battle.

It might be worth simplifying this in the future, but I don't think it's urgent.


bitcraft:
looks good, thanks!

PR #503, opened 2019-01-23
Fix trainer battles

kerizane:
#290 

- Battle starts with "x wants to battle",
- Trainers can use tuxemon,
- Stops the player running from trainer battles.


bitcraft:
Looks good!  I'll merge this and we can get some map makers and testers on it.

PR #504, opened 2019-01-23
Remove additional effects from dollfins moves

kerizane:
When I was testing my combat update I made some changes to dollfins moves. Looks like I forget to reset them.


bitcraft:
Thanks for cleaning it up!

PR #511, opened 2019-02-18
Verbal conjugation is WRONG!

AndrejHatzi:
I changed all verbs ending in a 's' like 'fugistes' because that does not exist in portuguese!


bitcraft:
Thanks for looking into the translation.  Just a heads up, another contributor recently changed our system from json to gettext/PO files.  These json files are no longer being used, so these changes won't show up in the game.  I will can merge this anyway, but for the translation to work in the game, it will need to be updated to the PO format.  Unfortunately, I guess the other contributor didn't migrate this with the rest.

AndrejHatzi:
Where are exactly those files?

bitcraft:
Sure.  Copy the spanish locale folder (https://github.com/Tuxemon/Tuxemon/tree/development/tuxemon/resources/l18n) and then modify it for portuguese.

bitcraft:
@AndrejHatzi I'll merge this so the source json translation is correct when somebody tackles the gettext translation.  Keep in mind changes will not show up in game until gettext files are fixed.  Thanks again for your contribution.

PR #513, opened 2019-03-08
Don't allow escaping on start menu

kerizane:


kerizane:
Pressing escape on the "New Game", "Load Game" window removes the window, forcing us to reboot.

PR #518, opened 2019-03-14
Stop menu misbehaviour after trying to run

kerizane:
Fixes an issue where the menus would go slightly crazy after trying to run from a trainer battle.


bitcraft:
Thanks for the fix!

PR #520, opened 2019-04-06
Don't crash if we try to load an empty slot

### LINKED ISSUE #519: Game Crashes While Trying to Load an Empty File
Upon attempting to load a game from an "Empty Slot" in the Load menu, the game crashes with the following error: 
`File "/home/Projects/Tuxemon/tuxemon/core/states/persistance/load_menu.py", line 20, in on_menu_selection
    save_data = save.load(self.selected_index + 1)
  File "/home/Projects/Tuxemon/tuxemon/core/components/save.py", line 124, in load
    save_data["error"] = "Save file corrupted"
TypeError: 'NoneType' object does not support item assignment`

kerizane:
Fixes #519

### DIFF
--- a/tuxemon/core/components/save.py
+++ b/tuxemon/core/components/save.py
@@ -120,14 +120,17 @@ def load(slot):
     save_data = open_save_file(save_path)
     if save_data:
         return upgrade_save(save_data)
+    elif save_data is None:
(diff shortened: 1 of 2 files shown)


bitcraft:
Thanks for the great fix, as always.

Issue #523, opened 2019-04-09
Error when entering tall grass southwest of Cotton Town


agervase:
![image](https://user-images.githubusercontent.com/24658277/55816280-e93f3d80-5abf-11e9-88d6-0df5994891c4.png)

When trying to enter the grass in this area, you receive the following error. Is this just because the area hasn't been fully developed yet? Or is this a bug that needs to be fixed?

```
Traceback (most recent call last):
File "/home/projects/Tuxemon/tuxemon.py", line 50, in <module>
tuxemon.core.main.main(load_slot=args.slot)
File "/home/projects/Tuxemon/tuxemon/core/main.py", line 101, in main
control.main()
File "/home/projects/Tuxemon/tuxemon/core/control.py", line 570, in main
update(clock_tick)
File "/home/projects/Tuxemon/tuxemon/core/control.py", line 619, in update
self.event_engine.update(time_delta)
File "/home/projects/Tuxemon/tuxemon/core/components/event/eventengine.py", line 340, in update
self.update_running_events(dt)
File "/home/projects/Tuxemon/tuxemon/core/components/event/eventengine.py", line 411, in update_running_events
action.start()
File "/home/projects/Tuxemon/tuxemon/core/components/event/actions/random_encounter.py", line 70, in start
encounters = monsters.database['encounter'][encounter_slug]['monsters']
KeyError: None
```

kerizane:
Fixed

PR #526, opened 2019-04-14
Fix route1 random enounters, and don't crash on KeyError

kerizane:


kerizane:
#523

bitcraft:
Thanks for looking into this.  In the past I believe there were more exception handlers like you PR has.  I have removed many like this.  I prefer to allow the game to crash in cases where the maps or config is broken, because it tells you right away that there is an issue.  Most players will never see the log output, so it could be frustrating to spend time debugging issues if the errors are being silently ignored.

One of my general goals is to have less code that needs to be maintained, and less tolerance for bad config.  For me, the game shouldn't handle a bad config....that should be fixed by the map creators.  If we keep adding workarounds for bad config, the game will quickly have lots of exception handling code, and could create more bugs around broken config.

There are some exceptions, and I am open for allowing this, but in general, I would prefer to see the game config fixed, and avoid adding new code.

Alternatives to ignoring errors like this would be better log output or maybe even a high-level exception handler in the game, which will open a dialog with the error.

Let me know if you want to merge it.  I'm ok with it for now.

kerizane:
That's fine, I removed the error handling.

bitcraft:
Looks good, thanks!

Issue #522, opened 2019-04-09
Fainted Tuxemon sent into battle first


agervase:
I have not been able to reliably reproduce this error, but sometimes when entering combat, a fainted Tuxemon will be sent out first (even though the party has non-fainted Tuxemon), and then the Tuxemon's first action in battle will just be to faint.

![image](https://user-images.githubusercontent.com/24658277/55813466-cbbba500-5aba-11e9-94d4-e6bfa92d57fa.png)

kerizane:
Thanks for reporting.

Are you running the latest code? There's a bug that was fixed 6th of January which could have caused this. Mons were mistakenly being created with zero health, but weren't marked fainted. 

Alternatively, if the mon was caught using the old code, and you hadn't used or healed them since updating, then the issue would still persist.

kerizane:
If the above isn't true, then any other info would help.

How many times have you seen the error? 
What enemy and attack did your tuxemon faint against, was there anything unusual about how it was fainted?

kerizane:
Should be fixed by #532

bitcraft:
@kerizane #532 is merged, please check if this issue can be closed or not.

kerizane:
Fixed, please close.

PR #527, opened 2019-04-14
Check current_hp when we check if the tux is fainted

kerizane:
#522 

Not sure what the root cause is (maybe related to #495 ?) , but this should help a little.


bitcraft:
This works for now.  Thanks!

Issue #528, opened 2019-04-19
Encounters are nearly guaranteed on route2


Qiangong2:
On route2, when you walk through the grass, you are lucky not to encounter tuxemon due to the chance that is coded in. If it was less (like 0.25) you'd actually be able to walk through.

This might be just a change I could do, but if an action could be created to set the chance amount (between 0 and 1) of an encounter, we could do a lot with it

kerizane:
Would this action control encounters in general, or each individual encounter?

What's the use case for this?

Qiangong2:
individual encounters would be ideal. That way for event blocks only one wide, they don't have the same chance as ones that are 8 or 9 blocks wide.

For the use, try walking through route 2 without an encounter. It's nearly impossible

kerizane:
Currently the encounter jsons are like this:

{"slug": "route2","monsters": [
{ "monster": "cardiling","encounter_rate":10,"level_range": [3,6]},
{ "monster": "aardorn","encounter_rate":10,"level_range": [3,6]},
{ "monster": "eyenemy","encounter_rate":10,"level_range": [3,6]},
{ "monster": "axylightl","encounter_rate":5,"level_range": [4,7]},
{ "monster": "cataspike","encounter_rate":10,"level_range": [3,6]},
{ "monster": "cataspike","encounter_rate":1,"level_range": [5,8]}
] }

And we call it in the map like so:
  <object id="97" name="random battle" type="event" x="160" y="192" width="16" height="48">
   <properties>
    <property name="act1" value="random_encounter route2"/>
    <property name="cond1" value="is player_at"/>
    <property name="cond2" value="is player_moved"/>
    <property name="cond3" value="is party_size greater_than,0"/>
   </properties>
</object>

We could add a modifier to the random_encounter action, so it scales the probabilities.
e.g.    <property name="act1" value="random_encounter route2, 0.5"/> # Overall chance of an encounter is 50%

Or would you prefer:
<property name="act1" value="random_encounter cataspike, 10, 3, 6/>
<property name="act2" value="random_encounter aardorn, 10, 3, 6/>
etc...

The second one seems verbose and difficult to maintain.

Qiangong2:
That would be great if it was enabled

kerizane:
Done, please test.

bitcraft:
@Qiangong2 can you test this?

Qiangong2:
@bitcraft sure. Might be a day or two though.

[...]

PR #530, opened 2019-05-06
532: Fix random encounter formula and add config

kerizane:
#528 
random_encounter can have total_prob passed to scale the probabilities.
Scaling is from 0 - 100, not 0 - 1.0. Decimals are allowed.

### DIFF
--- a/tuxemon/core/components/event/actions/random_encounter.py
+++ b/tuxemon/core/components/event/actions/random_encounter.py
@@ -40,12 +40,15 @@ class RandomEncounterAction(EventAction):
     "monster.db" database. The chance that this will start a battle depends on the
     "encounter_rate" specified in the database. The "encounter_rate" number is the chance
     walking in to this tile will trigger a battle out of 100.
+    "total_prob" is an optional override which scales the probabilities so that the sum is
+    equal to "total_prob".
 
-    Valid Parameters: encounter_slug
+    Valid Parameters: encounter_slug, total_prob
     """
     name = "random_encounter"
     valid_parameters = [
         (str, "encounter_slug"),
+        ((float, None), "total_prob"),
     ]
 
     def start(self):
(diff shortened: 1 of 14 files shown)


tuxemon/core/components/event/actions/random_encounter.py
```diff
@@ -130,3 +99,40 @@ def start(self):
     def update(self):
         if self.game.get_state_name("CombatState") is None:
             self.stop()
+
+
+def _choose_encounter(encounters, total_prob):
+    total = 0
+    roll = random.random() * 100
```
kerizane: The previous implementation was buggy:

1. Random number was 1 - 1000 so the probability was 10% of what it should have been.
2. Even with the correct range, the probabilities were lower than they should have been.

tuxemon/resources/db/encounter/mansion.json
```diff
@@ -3,10 +3,10 @@
 
 
 
-{ "monster": "cairfrey","encounter_rate":10,"level_range": [13,16]},
-{ "monster": "polyrock","encounter_rate":10,"level_range": [13,16]},
-{ "monster": "djinnbo","encounter_rate":5,"level_range": [14,17]},
-{ "monster": "pigabyte","encounter_rate":10,"level_range": [13,16]}
+{ "monster": "cairfrey","encounter_rate":6,"level_range": [13,16]},
```
kerizane: As the calculation doesn't underestimate the probabilities any more, the numbers need lowering a bit.

tuxemon/resources/maps/map1.tmx
```diff
@@ -229,7 +229,6 @@
     <property name="act2" value="play_map_animation grass,0.1,noloop,player"/>
     <property name="cond1" value="is player_moved"/>
     <property name="cond2" value="is player_at"/>
-    <property name="cond3" value="is party_size greater_than,0"/>
```
kerizane: Don't need these.

PR #532, opened 2019-06-05
Make tuxemon faint properly

kerizane:
Fixes bug caused by: #527

The game wasn't running removing the monsters from the action queue. If the opponent attacked first then the fainted mon would still try to attack and the game would crash.


kerizane:
#537

Issue #535, opened 2019-07-17
Translate on install


jgibbons94:
I am working on a guix package for tuxemon (because I like guix more than I like the other distros).  I got it to install (though I have to skip tests), but it crashes on launch with the following output:

> pygame 1.9.4
> Hello from the pygame community. https://www.pygame.org/contribute.html
> Traceback (most recent call last):
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/bin/.tuxemon-real", line 11, in <module>
>     load_entry_point('tuxemon==0.4.1', 'gui_scripts', 'tuxemon')()
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/__main__.py", line 63, in main
>     from tuxemon.core.main import main
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/core/main.py", line 36, in <module>
>     from .components.player import Player
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/core/components/player.py", line 35, in <module>
>     from tuxemon.core.components.npc import NPC
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/core/components/npc.py", line 38, in <module>
>     from tuxemon.core.components.item import decode_inventory, encode_inventory
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/core/components/item.py", line 39, in <module>
>     from tuxemon.core.components.locale import T
>   File "/gnu/store/3sn0551440vllfmqb1vys3rhrdbh88hn-tuxemon-0.4.0-alpha/lib/python2.7/site-packages/tuxemon/core/components/locale.py", line 273, in <module>
>     T = TranslatorPo()
[...]

PR #536, opened 2019-07-17
build translations on install

jgibbons94:
Fix issue #535 
-Build all translations in setup.py
-Check that compiled translations do not exist in tuxemon/core/components/locale.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,10 @@ from setuptools import setup
 import fnmatch
 import os
 
+#build the translations
+from tuxemon.core.components.locale import T
+
+
 # Find all the python modules
 modules = []
 matches = []
--- a/tuxemon/core/components/locale.py
+++ b/tuxemon/core/components/locale.py
@@ -76,7 +76,7 @@ class TranslatorPo(object):
             outfile = os.path.join(os.path.dirname(infile), "base.mo")
 
             # build only complete translations
-            if os.path.exists(infile):
+            if os.path.exists(infile) and not os.path.exists(outfile):
                 with open(infile, "r") as po_file, open(outfile, "wb") as mo_file:
                     catalog = read_po(po_file)
                     write_mo(mo_file, catalog)


bitcraft:
@jgibbons94 thanks for interest in the tuxemon project.  as for the translations, its not a great idea to include the package you are installing as an import for setup.py.  i've always had problems with this and i'd like to avoid it.  could you find an alternative solution which doesn't require importing tuxemon into setup.py?

jgibbons94:
Thanks for reviewing my pull request!

On Sun, 2019-12-29 at 21:25 -0800, Leif Theden wrote:
> @jgibbons94 thanks for interest in the tuxemon project. as for the
> translations, its not a great idea to include the package you are
> installing as an import for setup.py. i've always had problems with
> this and i'd like to avoid it. could you find an alternative solution
> which doesn't require importing tuxemon into setup.py?

Do you mean the line "from tuxemon.core.components.locale import T"?
What sort of problems have you had with that?

The two important things if the guix package manager it to include
Tuxemon are:

1. Tuxemon should set up the translations on build because (IIRC)
tuxemon crashes without the translations built, and

2. Tuxemon should not try to build the translations on run if the
translatiosn exist because guix builds its packages from source and
installs them to a read-only directory identifiable by hash and the
program will crash if it tries to write to a read-only location.

The line in question is sufficient to build the translations when they
need to be built by Guix and is probably the simplest way to accomplish
this.

I suppose I could put the line into a "build-translations.py" script
and run that when guix builds. Would that be acceptable despite such a
script being only one line?

-Jesse

bitcraft:
Its totally possible that I am just wrong here, but in my experience packaging python programs where setup.py includes the package itself means the venv/env that builds it must also import it and all of its requirements.  Its much easier to use a vanilla venv to build.  Specifically, I've had python packages not build because runtime dependencies were not available (deps are required for running, not building).  Its easier to maintain a simple build venv if you don't import the world into it.

It makes sense that building the translations should happen before installing.  I'm sure we would need this fixed for other platforms.  I'm sure our windows builds are broken by this too...

Before I get too far into it (as I am not super familiar with this issue), could you confirm the following:
* You're build doesn't work b/c your distribution files must be read-only
* Tuxemon generates translation files on startup if they are not present
* These translation files end in extension "mo", and are in resources/l18n

I have an idea that should work for you, as well as other platforms without requiring another script.

jgibbons94:
On Thu, 2020-01-02 at 16:50 -0800, Leif Theden wrote:
> Its totally possible that I am just wrong here, but in my experience
> packaging python programs where setup.py includes the package itself
> means the venv/env that builds it must also import it and all of its
> requirements. Its much easier to use a vanilla venv to build.
> Specifically, I've had python packages not build because runtime
> dependencies were not available (deps are required for running, not
> building). Its easier to maintain a simple build venv if you don't
> import the world into it.
That does make sense. I don't have much of a Python background, so I
didn't consider this. Thank you for your explanation.

> 
> It makes sense that building the translations should happen before
> installing. I'm sure we would need this fixed for other platforms.
> I'm sure our windows builds are broken by this too...
Last I checked, the Android builds crash on startup. I think it might
be related to this issue.
> 
> Before I get too far into it (as I am not super familiar with this
> issue), could you confirm the following:

Since Guix doesn't support a rolling release, I'm using the most recent
commit c8358cabbcf01ed4797d8de0192f7b4bc0fa9a0f to check the official
repository as-is. Before I forget, I should mention that it would be
best to bump the stable release version in the official repository
after this other pending issues with available solutions are resolved
because the guix maintainers prefer it when the package references a
tag.

> 
> You're build doesn't work b/c your distribution files must be read-
> only
I'm guessing that by "distribution files" you mean all the files
installed when the tuxemon package is installed. If that is the case,
then, long story short, yes. 

> Tuxemon generates translation files on startup if they are not
> present
Before my changes, Tuxemon generates translation files on startup every
time no matter what. My change to tuxemon/core/components/locale.py
checks that outfile does not exist.

The official repository as-is does attempt to build the translation
files to the install directory on-launch. I'm attaching output.txt if
you're interested in the error.

> These translation files end in extension "mo", and are in
> resources/l18n
The translation files are compiled from .po files in
tuxemon/resources/l18n into .mo files in resources/l18n. When I build
with my changes, the .mo files are found with the .po files in

<install-location>/lib/python2.7/site-packages/tuxemon/resources/l18n/. 

> I have an idea that should work for you, as well as other platforms
> without requiring another script.

Cool. I look forward to resolving this in the best way possible.

-Jesse

pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/bin/.tuxemon-real", line 11, in <module>
    load_entry_point('tuxemon==0.4.1', 'gui_scripts', 'tuxemon')()
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/__main__.py", line 65, in main
    from tuxemon.core.main import main
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/main.py", line 38, in <module>
    from .components.player import Player
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/player.py", line 36, in <module>
    from tuxemon.core.components.npc import NPC
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/npc.py", line 42, in <module>
    from tuxemon.core.components import db, monster, pyganim
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/monster.py", line 42, in <module>
    from tuxemon.core.components.locale import T
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/locale.py", line 278, in <module>
    T = TranslatorPo()
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/locale.py", line 60, in __init__
    self.build_translations()
  File "/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/core/components/locale.py", line 86, in build_translations
    with io.open(outfile, "wb") as mo_file:
IOError: [Errno 30] Read-only file system: '/gnu/store/33sgxjn5gyhy49d03dxra4qalpxagrkn-tuxemon-0.4.0-alpha-0/lib/python2.7/site-packages/tuxemon/resources/l18n/es_MX/LC_MESSAGES/base.mo'

bitcraft:
ok cool.  after considering the problem, i think that generating the translation files when launched is not the right approach (what we are doing now).  it is likely slow on some hardware and should only be done once.  the resources really should be in a separate folder, outside of the package.  code/data generation and then sticking those files back into the package seems problematic long term.  here's what i have in mind:
* refactor locale.py so that all of the building code is separate
* refactor setup.py so that the build step will import tuxemon (scoped in the function) then build

Nice to have:
* refactor the translations steps to be separate from tuxemon, so the only dep is gettext
* add config option which will build translations if not found.  default is true.

Eventually (just listing here for future notes):
* Move resources into a folder outside of the package (tuxemon folder)
* Create a folder for the MO files
* Create keep all source resources in one folder, all compiled resources (like MO) go in another

Proposal for setup.py
```python
# https://stackoverflow.com/questions/1321270/how-to-extend-distutils-with-a-simple-post-install-script/1321345#1321345
import distutils.core
from distutils.command.install import install
...
class my_install(install):
    def run(self):
        install.run(self)
        # Custom stuff here
        # distutils.command.install actually has some nice helper methods
        # and interfaces. I strongly suggest reading the docstrings.
...
distutils.core.setup(..., cmdclass=dict(install=my_install), ...)
```
I think this snippet is a bit out of date...im not sure that distutils is recommend to use anymore.  

Anyway, I think that your PR has most of this covered.  Let me know if you want to tackle the setup.py script.

bitcraft:
ive tested our debian packaging tools and this issue also breaks packaging in debian.  so this is now officially a high priority to get fixed because i'd really like to clean up our packaging scripts since we've just implemented a bucnh of new features.  

@jgibbons94 have you looked at the changes required in setup.py?

bitcraft:
I've merged this as is.  Ill open a follow up ticket to address the concerns i've raised in this issue.  thank you again for your contribution.

jgibbons94:
On Sat, 2020-01-18 at 07:59 -0800, Leif Theden wrote:
> I've merged this as is.  Ill open a follow up ticket to address the
> concerns i've raised in this issue.  thank you again for your
> contribution.
I apologize for not following up. I can help change setup.py to be
independent of the tuxemon package when time permits.
-Jesse

bitcraft:
no worries

vnmabus:
Since this was merged, it is not possible to install tuxemon if you do not have its dependencies previously installed.

bitcraft:
Yeah, I'm aware of that.  At the moment I would like to get the solved, but considering its very easy to run the game in the folder directly, it hasn't gotten much love.  If anyone wants to provide a fix, I think we would need to manage a cache of the compiled translation files, or somehow keep them in memory so the files are not required.

vnmabus:
In fact you could remove the import, as the imported module has now another name, and nowadays the translations are not built in the `__init__` method, so it would be a NOP. Thus, currently this is not working...

bitcraft:
There is an existing issue about this problem here https://github.com/Tuxemon/Tuxemon/issues/535

vnmabus:
Yes, but supposedly this PR solves that issue... Although I now see that it is still open.

Issue #550, opened 2020-01-01
Crash on startup


MirceaKitsune:
Latest Git currently crashes on startup with the following error:

```
mircea@linux-qz0r:~/Games/Small/Python/Tuxemon/Tuxemon_GIT> python ./tuxemon.py
pygame 1.9.6
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "./tuxemon.py", line 42, in <module>
    import tuxemon.core.main
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/main.py", line 38, in <module>
    from .components.player import Player
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/player.py", line 36, in <module>
    from tuxemon.core.components.npc import NPC
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/npc.py", line 42, in <module>
    from tuxemon.core.components import db, monster, pyganim
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/monster.py", line 42, in <module>
    from tuxemon.core.components.locale import T
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/locale.py", line 277, in <module>
    T = TranslatorPo()
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/locale.py", line 60, in __init__
    self.build_translations()
  File "/home/mircea/Games/Small/Python/Tuxemon/Tuxemon_GIT/tuxemon/core/components/locale.py", line 84, in build_translations
    with open(infile, "r", encoding="UTF8") as po_file, open(outfile, "wb") as mo_file:
TypeError: 'encoding' is an invalid keyword argument for this function
```

kerizane:
#542 breaks python2, fixing.

Until this gets fixed you can run it with python3

PR #551, opened 2020-01-02
Fix issue 550: Crash on startup

kerizane:
TypeError: 'encoding' is an invalid keyword argument for this function


kerizane:
Fixes #550

PR #552, opened 2020-01-02
Fix broken maps

kerizane:
Map was crashing in city park.
Misa was skipping dialog.
A sign was skipping dialog.

Mostly related to the locale change from ages ago.


bitcraft:
cool, ty again

PR #555, opened 2020-01-02
Route1 tidy up

kerizane:
Tidy up sprites:
- there were trees without tops
- weird pixels by the xero institute
- trees piled on top of each other by the xero institute
- flowers that would hover over the player

Resize map (slight cropping as I didn't think the empty space on the left and bottom of the map looked good)
Remove grunts after you meet Misa (don't want them to be blocking you in).


kerizane:
Probably wasn't worth resizing it, but oh well.

Tested the guard placement, Misa conversation, teleportation to the route above, and teleportation to the route to the right.

bitcraft:
You're stepping up your work on the maps.  Thanks for the fixes!

PR #557, opened 2020-01-02
Display a message when a monster has leveled up

MirceaKitsune:
It's difficult to tell when a monster has leveled up after battle. For this reason I attempted to implement a message that will inform you when this happens. At the moment this pull request works but contains a few issues, I will need someone else to help with the following problems:

- The message is displayed immediately after the enemy monster faints. Because of this messages containing other information, such as what move was just used or that the enemy monster fainted, are being hidden. Also if multiple tuxemon are leveling up at the same time, the messages will likely cover one another and only the last one will be displayed.
- Ideally this message should be shown by the level_up function in core/components/monster.py not faint_monster in core/states/combat/combat.py: That way it would also work if something else makes monsters level up in the future. I wasn't able to place it there as monster.py doesn't have access to the original self object, only the monster (defined as self), the alert method requires the player.
- I don't know how to do the translations for other languages, so currently I used the english version in all language files. We should fix the rest before committing this.

Screenshot_20200102_234215

### DIFF
--- a/tuxemon/core/components/monster.py
+++ b/tuxemon/core/components/monster.py
@@ -349,8 +349,8 @@ class Monster(object):
 
         :type amount: Integer
 
-        :rtype: None
-        :returns: None
+        :rtype: Boolean
+        :returns: True if the monster has leveled up.
 
         **Example:**
 
@@ -360,6 +360,9 @@ class Monster(object):
         if self.total_experience >= (self.experience_required_modifier * (self.level + 1) ** 3):
             #Level up worthy monsters
             self.level_up()
+            return True
+
+        return False
 
     def apply_status(self, status):
         """ Apply a status to the monster
--- a/tuxemon/core/states/combat/combat.py
+++ b/tuxemon/core/states/combat/combat.py
@@ -726,7 +726,8 @@ class CombatState(CombatAnimations):
             # Award Experience
             awarded_exp = monster.total_experience / monster.level / len(self._damage_map[monster])
             for winners in self._damage_map[monster]:
-                winners.give_experience(awarded_exp)
+                if winners.give_experience(awarded_exp):
+					self.alert(T.format('level_up', {"name": winners.name, "level": winners.level}))
 
             # Remove monster from damage map
             del self._damage_map[monster]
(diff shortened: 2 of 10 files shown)


Qiangong2:
I don't know if it applies in this case, but when multiple dialogue boxes open at once in the main world, it's usually due to condition overlaps. I get around it be explicitly listing conditions required. Maybe you need to do something similar here?

Also, you could probably link to the translation files we already have. All we need is to declare it in each language

bitcraft:
This will be a neat feature when its done.  As-is though I don't think that the implementation is very clear.  In general, I'm not a fan for return values used this way.  The side-effects of functions (increasing the level if needed) shouldn't be returned.  For example, suppose later we want to know their new level, or new techniques earned from giving experience....should we change the boolean to a dictionary?  What if we just want to test if experience will trigger a level up?...  What are the maintenance cost of that long term?

For a simple alternative, you could store the monster level before adding experience, then test it after...if the value is greater, display a dialog.

I am hesitant to merge this as-is, because of the return value.

As for the translations, I would prefer to keep English out of them.  Its obvious now that its english, but as the files get larger, this pattern of adding English might mean the wrong translations "slip through".  If they are not found in the PO files, Tuxemon should be showing either the translation slug or the English.  If it is not, then that is a separate issue.

Please consider removing the return value and the English from the other translations.

bitcraft:
I'm not a Pokemon expert... but I would think that Pokemon should level up at the end of a battle?  I think with our current implementation of "experience is earned when monster faints" violates a traditional rule where you cannot keep experience if you run or lose.  Does that make sense?  Maybe while you are making changes, you could fix this (if needed!) so that experience is earned at the end of a battle if you win it.

MirceaKitsune:
Thanks for the feedback. For the time being I'm going to close this, if I manage to improve it I may reopen or make a new pull request: It's the best solution I could manage, but even I'm not that happy with how it turned out, and you confirmed it needs fixing.

On top of that I'm separately working on implementing the exp bar... expect that one tomorrow! With that addition this won't be so urgent to have, as there's going to be a way to keep track.

PR #569, opened 2020-01-03
Implement experience bar

### LINKED ISSUE #558: Monsters always have an experience value of 0
I was testing monsters leveling up earlier today. I noticed that after every battle, my tuxemon always increased by one level which seemed peculiar. As part of another pull request I'm working on I later added some debug prints; Through those messages I discovered that at the moment, the `give_experience` function is always being called with a value of 0!

It appears that `monster.total_experience` and `monster.experience_required_modifier` are always 0 no matter where I check in the code, including after a battle where exp was given. This seems like a bug but I don't know what's causing it, could be code or monster configs.

Note that I'm currently working on implementing the exp bar for the player HUD and am halfway done with its patch. It would help if this could be addressed in the next hours in case it's a simple issue, so I can properly test the calculation tomorrow before I submit a pull request for the bar.

MirceaKitsune:
This implements the experience bar for the player HUD. Tested and appears to be working exactly as intended: The exp bar of your monster will increase upon the enemy monster fainting, reflecting the amount of experience you just received... when the bar fills the monster levels up.

Additionally this fixes #558 which caused all monsters to give 0 exp due to a multiplier defaulting to 0.

Screenshot_20200103_174937


bitcraft:
Cool looks pretty good.  I would have preferred that you modify the existing hp bar class and reuse that existing code (by substituting colds and the border), but its not a huge deal.

Please:
* add a comment to the HP and EXP bar classes noting that they share a lot of code
* refactor the math about next levels into a common function, then change those classes/functions so that it is shared.  for formulas, there should be only a single place where they are defined.

Have you noticed any graphical glitches when a monster levels up?

MirceaKitsune:
Done. I managed to make a common Bar class for HpBar and ExpBar where I moved everything that doesn't need to be specific to each one, only left a few things that broke functionality or caused crashes as duplicates. There's also now a common function for monsters to get the exp requirement for the next level, with an optional offset so 1 can be used to check requirements for the next level.

As far as visual glitches go, nothing I can see during the vigorous testing I did today. There's only one small annoyance which I don't currently know how to solve: If your monster levels up, the player HUD won't update to reflect this, neither the exp bar nor the level displayed (the later was already present). This isn't a big issue as the match ends soon after so it's just a visual annoyance... except maybe for trainer battles with multiple monsters. This is a separate existing issue and can be fixed later though.

bitcraft:
cool ty for those changes.  please open an issue to document the remaining issues.  i'll merge this for now.

MirceaKitsune:
Thank you; I opened #577 regarding the update issue when leveling, which is really the only problem I found during testing. I used debug prints which confirm the bar should be updated too after leveling up... something might be blocking HUD updates under some circumstances, no idea on this one.

PR #578, opened 2020-01-04
Add event action to change the faint location

### LINKED ISSUE #392: Losing battle doesn't send player to last Tuxemon center
In the Pokemon games, when you lose a battle, you are sent back to the latest Pokecenter that you visited and your pokemon are restored completely. If you haven't visited any, it sends you back to your house.

This should be implemented as every pokemon game has it and it is convenient for the user.

MirceaKitsune:
Attempts to fix #392 by implementing an event to change the location where the player will be teleported upon fainting in battle. The hardcoded map remains default but can now be overridden.


MirceaKitsune:
@bitcraft Is my implementation correct or do you believe any changes are needed? I didn't verify the new action to change location as it requires editing a map, but the old functionality is unchanged so the new faint variable should be working properly.

bitcraft:
Do you mind investigating the effort the make the spawn location a game variable?  I think it would simplify the implementation (no need for a new action, just set a game variable to the amp and coords).  Im not a fan of adding variables onto the player object because those values eventually will need special handing when saving, loading, multiplayer serialization, etc.  I'd prefer to store values like this in a standard place...that game variables.

MirceaKitsune:
I wasn't sure if a game variable was the preferred idea, it does seem like the best option. Latest commit fixes this with one minor annoyance: You can't separate parameters by "," as this causes a crash, likely because it's interpreted as multiple properties... I had to use " " instead. So to set a new faint location you now simply use:

`act1: set_variable teleport_faint:map1.tmx 33 45`

I tested and it works: Once this variable is set, fainting takes you to that exact map and position.

bitcraft:
thanks for making those changes. looks great.

PR #587, opened 2020-01-05
Allow gaining multiple levels at once

### LINKED ISSUE #579: Should level up multiple times if there's enough exp
If you have enough exp then you should gain multiple levels, not just one.
This could happen if a low-level tmon helps defeat a high-level one.

MirceaKitsune:
Should fix #579 Previously monsters could only gain one level per call, now they should get the full amount their exp increase allows for.

### DIFF
--- a/tuxemon/core/components/monster.py
+++ b/tuxemon/core/components/monster.py
@@ -369,7 +369,7 @@ class Monster(object):
         self.speed = shape["speed"] * multiplier
 
     def level_up(self, levels):
-        """Increases a Monster's level by one and increases stats
+        """Increases a Monster's level and stats
         accordingly
 
         :rtype: None
--- a/tuxemon/core/components/monster.py
+++ b/tuxemon/core/components/monster.py
@@ -369,8 +369,7 @@ class Monster(object):
         self.speed = shape["speed"] * multiplier
 
     def level_up(self, levels):
-        """Increases a Monster's level and stats
-        accordingly
+        """Increases a Monster's level and stats accordingly
 
         :rtype: None
         :returns: None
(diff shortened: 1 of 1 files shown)


kerizane:
To test: edit the save file so someone has lots of experience.

MirceaKitsune:
Thanks: Did that and can confirm it works exactly as intended.

Sanglorian:
I think ideally monsters will level up one level at a time, but keep levelling until they have fully leveled up. That way events that trigger at a particular level won't be missed.

MirceaKitsune:
Are there such events? The issue is that the full leveling needs to be done during battle; Glitches can occur when the exp doesn't match the level range... current functionality included, this is pretty much a bugfix. Even if we could safely do only do one level up per battle, that would be a deliberate limitation and feel really broken and annoying.

bitcraft:
I also think I would prefer to see the `level_up` method called once for each level gained.  I don't recall playing a game which combined levels.  As @Sanglorian mentioned as well, if we have level triggers implemented in the future, this would reduce any potential issues for events which depend on a specific level.  This can be done easily by 
```python
for level gained_levels:
   self.level_up(1)
```

MirceaKitsune:
Ah, just call the function multiple times instead of once with a level count? Sure, I'll change that: Just seemed more logical that way, this would actually be even cleaner though.

MirceaKitsune:
Strange: If I revert the level_up function and call it each time in my loop, it only seems to give some levels at random. For instance: I have a tuxemon at level 5, hacked the save to give it 1000 exp... but instead of getting to the maximum level after the first battle, it first gets to level 7 or 9 at first then another +2 / +4 extra levels per battle. This is what I'm seeing it with:

```
levels = 0
while self.total_experience >= self.experience_required(levels + 1):
    self.level_up()
    levels += 1
```

MirceaKitsune:
Done and retested, works the same way but by calling the level_up function per level as suggested.

bitcraft:
Cool, looks good, ty!

PR #592, opened 2020-01-06
add test map and simple rewrites

Qiangong2:
1. Added test map for experimentations
2. start rewrite of city park
3. re-add music on route2


Sanglorian:
When you modify the maps, can you please keep a copy of the original with a new name? E.g. citypark_spyder.tmx?

That's because we still have the Spyder in the Cathedral campaign to implement with the original maps.

PR #594, opened 2020-01-06
Custom battle environment and music

MirceaKitsune:
Battles now have the ability to configure their theme (island graphics) and music. This includes both random encounters, trainer battles, and the experimental multiplayer battle system.

I ultimately decided against adding them as a game variable: It's not a value that needs to be stored constantly as if the game needs to access it anytime, it's just a property being called on the spot. I'm also uncertain about adding so many hardcoded game variables in general... this is the cleanest and easiest way to implement such options I believe.

Example of giving a custom theme and music to a random encounter:

`act1: random_encounter new_txmn,10,water,8bit airship.ogg`

Last but not least, this branch also adds battle islands for the sand and snow theme, which are the other two themes available in the standard tileset. Now we can have desert and snow routes without the battles looking like they're taking place in grass! Here are screenshots of the 3 available options:

theme_grass
theme_sand
theme_snow


MirceaKitsune:
Cave and water islands now also added! This should cover every environment usable right now.

theme_cave
theme_water

Qiangong2:
Looks good! What about actual backgrounds though? Do those require something more to be done?

bitcraft:
my comments from discord:

i think its a good start.  im not happy with overloading all these values into the various commands because it increases complexity of them and because they now share the "environment" option, they need to be updated together.  i know you are excited to get this feature in, but im pretty confident that this isn't the final way it will be implemented.  this approach is too britle and goes against the design of simple commands which we've discussed before.
ill test this out and might merge it to get the feature in, but i don't want to keep adding options and arguments to commands.  please keep that in mind that we should be moving towards "more simple commands, less complexity" and not "a few powerful complex commands"
@MirceaKitsune 3 options per command is probably the "sweet spot", since we are inputing commands into the Tiled editor.  much more than that gets difficult to deal with and this is bumping up the amount to 5 and thats not even covering all the customizations we need for battles
consider all the options we want for battles, then consider what it would look like to add a value for each one on a single line
this is why i advocated the "variable" approach.  i don't agree that variable are limited and should be avoided.

MirceaKitsune:
If you believe this is the best way, I'll use two game variables instead. It just felt overkill as this isn't something that needs to be stored constantly, nor something that needs to be set separately. At the same time, having too many parameters can get complicated as well, so neither way is perfect in the end.

Sanglorian:
Nice! 

> Looks good! What about actual backgrounds though? Do those require something more to be done?

If people are interested, there are some full screen backgrounds on the wiki that could be used for testing:

https://wiki.tuxemon.org/index.php?title=The_World#Battle_backgrounds

MirceaKitsune:
Fixed and retested, works fine and takes a little less code to do it this way. The new formula now is going to be:

```
act1: set_variable battle_environment:water
act2: set_variable battle_music:8bit airship.ogg
act3: random_encounter new_txmn
```

That's one reason why I preferred the parameters: The one random_encounter call stays shorter, but now requires two additional set_variable commands just to customize every battle. Either way is fine om yjr rmf, just depends on which it's easier to think is cleaner.

MirceaKitsune:
@Sanglorian Those backgrounds look beautiful! I'm legitimately tempted to try integrating them... should be a cinch to do with this implementation. But what if the difference is too big for some players?

bitcraft:
@ShadowApex @kerizane @Qiangong2 @xirsoi could i get your input here on this PR?  should we add more arguments onto the battle command as suggested here, or should we break it up into several commands, such as "set sprites", "set music", etc.  there is more discussion in the discord if you need context

MirceaKitsune:
Following discussion with @ShadowApex on Discord, environments have been implemented as a new database. A single game variable called "environment" is used to define which is used.

As he described there are several advantages to this approach. One is that you don't have to override an existing theme if you wish to define your own, another is that it can be expanded with properties to override even more content in the future based on the theme of the map (eg: menu skins).

I was going to make the battle background customizable as well... unfortunately there's no easy way to do this: background_filename is defined as a core class property and pretty much unchangeable. I'm sure in the future we'll find a way to fix that as well, for now this PR meets its intended purpose.

bitcraft:
Thank you @MirceaKitsune for working with us and finding a solution we are all happy with.  and thanks for reviewing @kerizane.

xirsoi:
I know I'm late to the party, but I think what was decided is best. It allows a mapper to create events that set a default battle theme/environment and only customize it for special occasions, rather than specify the background in Every. Single. Battle. Event.

bitcraft:
i think the point that i raised was missed where i suggested the ability to customize all aspects of the battles using commands, but also have another command which simple runs a bunch of those batter-setters based on json parameters.  so it would be "flexible customization" and also "simple defaults for common types".

the pr as is just enforces a group of defaults for different types of battles, but you cannot "build your own" battle by maybe customizing the background for a specific fight, like a boss.  so, i do think it misses the mark a bit, but those things can be added later.

MirceaKitsune:
Currently the battle background can't be customized due to how it's defined: I couldn't find a way to change it before the battle code reads the default. In the future that may be possible to fix as well.

The environment can otherwise be customized per-battle, by changing the variable before battle starts.

I have plans to add more settings and features to environments. The current one I have in mind is for implementing a day-night cycle, where I plan to customize the colors in the environment settings... I'm still thinking out the details on that one though.

bitcraft:
sounds good.  please open a new issue for those features when appropriate.

PR #595, opened 2020-01-07
Npc speed

bitcraft:
npc_walk: set npc to the default walk speed (3.75 default)
npc_run: set npc to the default run speed (7.35 default)
npc_speed: set arbitrary speed in tiles/second

run/walk commands are available so that NPC speeds are consistent and easy to remember.  the speed command is available to use when more control is needed


bitcraft:
thanks for the review @kerizane

PR #596, opened 2020-01-07
Improved arguments

### LINKED ISSUE #591: Specify starting map and location from the CLI
This is a feature to all the player to start the game on a new map without using the config file or editing source code.  It will assist developers and mappers when debugging and testing maps.

Here is a suggested way to do it: 
```
python tuxemon.py --starting-map test1.tmx --starting-position 10,12
```
There is already an established way to get CLI options using `ArgumentParser` in `tuxemon.py`.  We already configure a starting position for new games from the config, this will be another way to do it.  Eventually, the config file option will be removed when we implement campaigns.

Please feel free to suggest alternative syntax.

MirceaKitsune:
Fixes #591 by implementing a --starting-map argument that tuxemon.py can be called with, alongside a --data directory argument. A new prepare.ARGS variable can be accessed throughout the code just like prepare.CONFIG, making it easy to implement any other arguments in the future.

### DIFF
--- a/tuxemon.py
+++ b/tuxemon.py
@@ -44,9 +44,11 @@ import tuxemon.core.main
 
 if __name__ == '__main__':
     parser = ArgumentParser()
-    parser.add_argument('-l', dest='slot', metavar='1,2,3', type=int, nargs='?',
+    parser.add_argument('-l', '--load', dest='slot', metavar='1,2,3', type=int, nargs='?',
                         default=None, help='The index of the save file to load')
-    parser.add_argument('-d', dest='data', metavar='resources', type=str, nargs='?',
+    parser.add_argument('-d', '--data', dest='data', metavar='resources', type=str, nargs='?',
                         default=None, help='The data directory to use')
+    parser.add_argument('-s', '--starting-map', dest='starting_map', metavar='bedroom.tmx', type=str, nargs='?',
+                        default=None, help='The starting map')
     args = parser.parse_args()
     tuxemon.core.main.main(args)
--- a/tuxemon.py
+++ b/tuxemon.py
@@ -46,6 +46,7 @@ if __name__ == '__main__':
     parser = ArgumentParser()
     parser.add_argument('-l', dest='slot', metavar='1,2,3', type=int, nargs='?',
                         default=None, help='The index of the save file to load')
+    parser.add_argument('-d', dest='data', metavar='resources', type=str, nargs='?',
+                        default=None, help='The data directory to use')
     args = parser.parse_args()
-    tuxemon.core.main.main(load_slot=args.slot)
-
+    tuxemon.core.main.main(args)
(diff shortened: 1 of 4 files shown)


tuxemon.py
```diff
@@ -44,8 +44,11 @@
 
 if __name__ == '__main__':
     parser = ArgumentParser()
-    parser.add_argument('-l', dest='slot', metavar='1,2,3', type=int, nargs='?',
+    parser.add_argument('-d', '--data', dest='data', metavar='resources', type=str, nargs='?',
+                        default=None, help='The data directory to use')
+    parser.add_argument('-l', '--load', dest='slot', metavar='1,2,3', type=int, nargs='?',
                         default=None, help='The index of the save file to load')
+    parser.add_argument('-s', '--starting-map', dest='starting_map', metavar='bedroom.tmx', type=str, nargs='?',
```
bitcraft: `metavar` is not a default, it is an alternate name for the option, used in help or listing the options.  Please change this from `bedroom.tmx`.

https://docs.python.org/3/library/argparse.html#metavar

tuxemon/core/states/start/__init__.py
```diff
@@ -85,9 +85,11 @@ def set_player_name(text):
 
         def new_game():
             # load the starting map
+            starting_map = prepare.CONFIG.starting_map
+            if prepare.ARGS.starting_map:
```
bitcraft: remove `prepare.ARGS`.  there should be a single object to get game options from

tuxemon/core/prepare.py
```diff
@@ -180,3 +183,12 @@ def init():
     # Initialize PyGame and our screen surface.
     if PLATFORM == 'pygame':
         pygame_init()
+
+# Set the arguments
+def args(arguments):
+    global ARGS
+    global DATADIR
+
+    ARGS = arguments
```
bitcraft: remove `prepare.ARGS`

tuxemon/core/main.py
```diff
@@ -39,8 +39,7 @@
 
 logger = logging.getLogger(__name__)
 
-
-def main(load_slot=None):
+def main(arguments=[]):
```
bitcraft: please revert this change.  we should not be using arguments from the CLI internally.

bitcraft:
Sorry, Im not really happy with the idea of passing argument from the CLI into the game directly.  The CLI arguments are just options, and they should be changing the config.  The game, state, components, actions, etc should all have a "single source of truth" for the game configuration, and at this time that is the `prepare.CONFIG` object.  Please change this so that the CLI arguments change the `CONFIG` object.  The ability to set the starting map is already a feature there, so not much should change.

The origin line reading `def main(load_slot=None):` could be refactored, but please use the CONFIG object, not arguments.  also keep in mind that `main()` might be called from other places that expect an option, so please verify that you can save/load from the game menus and the CLI

MirceaKitsune:
Fixed. Upon better understanding the code, it turns out I can override everything directly from tuxemon.py without needing to make changes to any other files! I've done so and retested, it's working perfectly with this new simplistic approach.

bitcraft:
ok this is much better, thank you.

PR #597, opened 2020-01-07
NPC wandering

MirceaKitsune:
Support for making an NPC randomly wander around the map. Uses pathfinding with randomly selected targets. To make your NPC move around simply use the following function:

`npc_wander npc_slug,2,4`

The first paramater is the NPC slug, the second is the frequency (how often to walk), the third is the distance the NPC may travel in one go (bigger values good for larger spaces).

A default use case is already included into the game: If you go into the maple house next to yours, you will see the professor's wife walking around the room.

### DIFF
--- a/tuxemon/core/components/event/actions/npc_wander.py
+++ b/tuxemon/core/components/event/actions/npc_wander.py
@@ -24,6 +24,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import math
 import random
 
 from tuxemon.core.components.event import get_npc
@@ -31,14 +32,15 @@ from tuxemon.core.components.event.eventaction import EventAction
 from tuxemon.core.components.map import dirs2
 
 
-def single_dir(origin, direction, tiles, collisions):
-    origin = tuple(origin) # make copy to prevent modifying the npc
-    for i in range(tiles):
-        origin_ofs = dirs2[direction]
-        if tuple(origin + origin_ofs) in collisions:
-            break
-        origin += origin_ofs
-    return origin
+# Picks a random position from the cubic area around the origin
+def pick_target(origin, tiles, collisions):
+    origins = []
+    for x in range(origin[0] - tiles, origin[0] + tiles):
+        for y in range(origin[1] - tiles, origin[1] + tiles):
+            org = (x, y)
+            if (org) not in collisions:
+                origins.append(org)
+    return random.choice(origins)
 
 
 class NpcWanderAction(EventAction):
@@ -63,35 +65,22 @@ class NpcWanderAction(EventAction):
             if npc.moving or npc.path:
                 return
 
-            # Choose a direction
-            direction_random = random.random()
-            direction = "up"
-            if direction_random > 0.75:
-                direction = "down"
-            elif direction_random > 0.5:
-                direction = "left"
-            elif direction_random > 0.25:
-                direction = "right"
-
-            # Choose a distance
-            tiles_max = 4
+            # Choose a random distance
+            tiles_max = 8
             if self.parameters.tiles:
                 tiles_max = self.parameters.tiles
-            tiles = int(round(tiles_max * random.random(), 0))
+            tiles = int(max(1, math.ceil(tiles_max * random.random())))
 
-            # Move is there's at least one tile, change direction otherwise
-            if tiles > 0:
-                parameters = [direction + " " + str(tiles)]
-                npc.pathfind((single_dir(npc.tile_pos, direction, tiles, collisions)))
-            else:
-                npc.facing = direction
+            # Pick a new location to move to
+            origin = (int(npc.tile_pos[0]), int(npc.tile_pos[1]))
+            target = pick_target(origin, tiles, collisions)
+            npc.pathfind(target)
 
         def next():
             # Check that the NPC still exists
             if npc is None:
                 return
 
-            # Issue the move command
             move()
 
             # Run this function again between time / 2 and time
@@ -99,8 +88,6 @@ class NpcWanderAction(EventAction):
             if self.parameters.time:
                 time_max = self.parameters.time
             time = min(10, max(0.1, (time_max / 2) + ((time_max / 2) * random.random())))
-
-            # Schedule the next move
             world.task(next, time)
 
         # Initialize the schedule function
--- a/tuxemon/core/components/event/actions/npc_wander.py
+++ b/tuxemon/core/components/event/actions/npc_wander.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,94 @@
...
+
+            # Choose a random distance
+            tiles_max = 4
+            if self.parameters.tiles:
+                tiles_max = self.parameters.tiles
+            tiles = int(max(1, math.ceil(tiles_max * random.random())))
+
+            # Pick a new location to move to
+            origin = (int(npc.tile_pos[0]), int(npc.tile_pos[1]))
+            target = pick_target(origin, tiles, collisions)
```
bitcraft: `WorldState.get_exits()` will give you tiles that can be moved into

tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,94 @@
...
+            npc.pathfind(target)
```
bitcraft: If possible, please use `NPC.move_one_tile()` here.  I don't think we should be using pathfind here.

tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,94 @@
...
+
+        def next():
+            # Check that the NPC still exists
+            if npc is None:
+                return
+
```
bitcraft: If the npc is not moving and does not have a path, then it should quit here to prevent it from moving during dialogs, etc

tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,94 @@
...
+    valid_parameters = [
+        (str, "npc_slug"),
+        (float, "frequency"),
+        (int, "tiles")
+    ]
+
+    def start(self):
+        npc = get_npc(self.game, self.parameters.npc_slug)
+        world = self.game.get_state_name("WorldState")
+        collisions = world.get_collision_map()
```
bitcraft: Try to use `WorldState.get_exits()` to avoid reinventing tile exit checking.

bitcraft:
Please see my review.  Overall looks good.  Try to use the existing logic for tile exits and don't call pathfind.

tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,94 @@
...
+            if self.parameters.tiles:
+                tiles_max = self.parameters.tiles
+            tiles = int(max(1, math.ceil(tiles_max * random.random())))
+
+            # Pick a new location to move to
+            origin = (int(npc.tile_pos[0]), int(npc.tile_pos[1]))
+            target = pick_target(origin, tiles, collisions)
+            npc.pathfind(target)
+
+        def next():
```
bitcraft: next is a builtin name for python.  please rename this.

bitcraft:
I see what you are trying to do using pathfind.  Can I ask if we could instead do the following?  each time the event is called, pick a random exit (open tile) and move into it?  we are already experiencing issues with pathfinding and i would like to avoid using it.  this "single tile movement" approach will also avoid accidentally picking a random tile which the npc cannot pathfind into

MirceaKitsune:
It's a shame pathfinding can't be safely used here. Perhaps I will create a new mode with that later on, where you can give an NPC a list of locations on the map it will periodically travel between objectives. For now yeah, it feels easier to just move the NPC one tile each occurrence.

MirceaKitsune:
I noticed that get_exits doesn't really work, so I'm probably going to use the old way for now. I tried all 3 of these options, and all seem to cause the NPC to still attempt walking into a wall:

```
world = self.game.get_state_name("WorldState")
collisions = world.get_collision_map()
world.get_exits(origin)
world.get_exits(origin, collisions)
world.get_exits(origin, collisions, set())
```

tuxemon/core/components/event/actions/npc_wander.py
MirceaKitsune: Doesn't work, I tried every combination I could think of. Using `if origin not in collisions:` works perfectly however, and is just as elegant if not even more.

tuxemon/core/components/event/actions/npc_wander.py
MirceaKitsune: If the cycle ended whenever an NPC doesn't move or have a path, it would stop wandering the moment it takes a break and pauses, npc_wander would need to be called constantly to re-enable it. I'm thinking of using a special flag to mark when wandering should be suspended so they stop moving during dialogues.

tuxemon/core/components/event/actions/npc_wander.py
MirceaKitsune: Yes... objective based movement should be a different project, I'll keep that in mind for later on. Changed it to use move_one_tile which greatly simplified the code too.

tuxemon/core/components/event/actions/npc_wander.py
MirceaKitsune: `WorldState.get_exits()` doesn't work for me, I tried it in every combination I could think of and it fails to detect the collision. `if origin in collisions` works like a charm though.

MirceaKitsune:
I've fixed most of the suggestions: The ones I couldn't solve I commented on as to why. The latest commits also fix the last two obvious issues left:

1. NPC wandering is now paused while a dialog is active, to avoid the character you're talking to running around as you're speaking with them. Normally this should only be done for the NPC you're talking to, instead of freezing everyone in place... still this is a good solution for now.
2. The NPC checked for collisions with map walls, but not with the player and other NPC's. I found a simple and relatively elegant way to fix this, if there's a better one let me know.

tuxemon/core/components/event/actions/npc_wander.py
bitcraft: that approach is fine for now i think.  if you look in the NPC class there is some variable that is/was used to indicate that the NPC was supposed to wander.  i don't really want to use instance attributes like this, but its probably the best solution at the moment.

tuxemon/core/components/event/actions/npc_wander.py
bitcraft: its a matter of not duplicating code.  consider that the maps can define "one way" exits.  we don't want all the special case code in the world map to be duplicated here.  i agree that `if origin not in collisions` is pretty, but its not the best solution and you should be using the existing mechanisms we have.  as i understand it, you are now doing the "one tile at a time" movement, so please do not re-implement functions we already have in use for the world map.

MirceaKitsune: I don't mind either way. The only issue is that get_exits isn't working for some reason. Please see my recent comment in the main discussion about that, I tried various combinations and they all fail to detect collisions. Maybe another bug?

tuxemon/core/components/event/actions/npc_wander.py
```diff
@@ -0,0 +1,84 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import math
+import random
+
+from tuxemon.core.components.event import get_npc
+from tuxemon.core.components.event.eventaction import EventAction
+from tuxemon.core.components.map import dirs2
```
bitcraft: no longer needed

PR #601, opened 2020-01-08
Mod loader 

### LINKED ISSUE #576: Remove resources folder from the code folder
Game code and story assets/resources should be split.  The resources folder should be inspected and all default campaign items should be moved outside the tuxemon folder.  Any assets that are living in the code folder should be moved to resources as well.

The root folder should look something like this:
```
concept
docs
scripts
tests
tuxemon
resources
```

Once this split is made, we can open another issue to tackle choosing a folder to load data from at startup; ie selecting a campaign.  The resources folder could be renamed or modified at that time to fit the campaign idea.

MirceaKitsune:
Fixes #576 and implements a game loader in the process. Resources will first be loaded from root/games/tuxemon, with the last part (tuxemon) customizable via the game setting (previously called data)... if not found there it will fallback to root/resources.


tuxemon/core/components/config.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, config_path=None):
         self.music_volume = cfg.getfloat("sound", "music_volume")
 
         # [game]
-        self.data = cfg.get("game", "data")
+        self.game = cfg.get("game", "game")
```
bitcraft: please do not change this.  the word "game" has a specific meaning in the code and this will be confusing.

tuxemon/core/components/config.py
```diff
@@ -162,7 +162,7 @@ def get_defaults():
             ("music_volume", 1.0),
         ))),
         ("game", OrderedDict((
-            ("data", "resources"),
+            ("game", "tuxemon"),
```
bitcraft: ditto.  do not change data to game.

tuxemon/core/components/event/actions/play_map_animation.py
```diff
@@ -62,7 +60,7 @@ def start(self):
         # "position" can be either a (x, y) tile coordinate or "player"
         animation_name = self.parameters.animation_name
         duration = self.parameters.duration
-        directory = os.path.join(paths.BASEDIR, prepare.DATADIR, "animations/tileset")
+        directory = prepare.fetch("animations", "tileset")
```
bitcraft: this looks clean.  i don't think i like that it is `prepare.fetch`, so maybe we can find a better name, but i do think this format would work great.

tuxemon/core/components/config.py
MirceaKitsune: Will revert in a following commit.

tuxemon/core/components/config.py
MirceaKitsune: Will revert in a following commit.

tuxemon/core/prepare.py
```diff
@@ -180,3 +177,16 @@ def init():
     # Initialize PyGame and our screen surface.
     if PLATFORM == 'pygame':
         pygame_init()
+
+# Fetches a resource file
+def fetch(*args):
+    path = '/'.join(i for i in args if i is not None)
```
bitcraft: use `os.path.join`

MirceaKitsune: I think I tried and it didn't work, will attempt this again and see what fails if anything does.

tuxemon.py
```diff
@@ -43,16 +43,16 @@
 
 if __name__ == '__main__':
     parser = ArgumentParser()
-    parser.add_argument('-d', '--data', dest='data', metavar='resources', type=str, nargs='?',
-                        default=None, help='The data directory to use')
+    parser.add_argument('-g', '--game', dest='game', metavar='tuxemon', type=str, nargs='?',
```
bitcraft: @ShadowApex @xirsoi @kerizane 
This is a proposal to add another folder for looking up game resources.  Historically we have been using the term "campaign".  tbh, I'm not really a fan for using the word "game" here.  could i get your opinions on what we should be calling this feature?

tuxemon.py
```diff
@@ -43,16 +43,16 @@
...
+                        default=None, help='The game to be loaded')
     parser.add_argument('-l', '--load', dest='slot', metavar='1,2,3', type=int, nargs='?',
                         default=None, help='The index of the save file to load')
     parser.add_argument('-s', '--starting-map', dest='starting_map', metavar='map.tmx', type=str, nargs='?',
                         default=None, help='The starting map')
     args = parser.parse_args()
 
-    if args.data:
-        prepare.CONFIG.data = prepare.DATADIR = args.data
+    if args.game:
```
bitcraft: please revert this to data as well

tuxemon.py
MirceaKitsune: For now I'm reverting it back to data which was the old name. I believe it can be altered in the future.

tuxemon/core/prepare.py
MirceaKitsune: The change fails with the following error:

`AttributeError: 'generator' object has no attribute 'startswith'`

tuxemon/core/components/config.py
bitcraft: also, please keep the word tuxemon.

tuxemon/core/prepare.py
bitcraft: please post the code that raised that traceback

MirceaKitsune: `path = os.path.join(i for i in args if i is not None)`

bitcraft: there should never be a `None` passed in.  there should always be an argument, so why not
```path = os.path.join(*args)```?

MirceaKitsune: Never mind, a much easier solution was available all along:

`path = os.path.join(*args)`

tuxemon/core/components/event/actions/play_map_animation.py
MirceaKitsune: Can be renamed later, too many big changes at once like you said.

MirceaKitsune:
Tested everything I could possibly think of with the latest commit: New game, loading an old save, creating and loading a new save, reading signs, talking to NPC's, swapping tuxemon in / out and before / after battles, using potions and capture devices during battle, getting exp and leveling up, even winning and letting myself faint: No crashes, nor errors, no missing graphics or audio. As far as I'm concerned this is ready to merge now.

bitcraft:
@MirceaKitsune I've tested this locally and it seems fine.  I'm going to merge this now.  Please be ready to make fixes in case there are breakages related to this PR.

PR #609, opened 2020-01-18
Extra monster and battle info

MirceaKitsune:
Adds 3 features that allow mappers to make better use of monster and battle data. I'm planning to use such checks in my game / mod and wanted to suggest this in preparation. What it changes:

- The combat state now records the relevant properties of the last monster you've battled using game variables: Name, level, slug, category, shape. They're primarily intended for random battles, to be used both in conditional checks as well as NPC dialogues (eg: A man standing near the patch of grass watching you fight can say "wow Player, I can't believe you defeated a ${{battle_last_monster_level}} especially a ${{battle_last_monster_slug}}!")
- Monster category and shape can now be translated as well for dialogues.
- A conditional to check similar properties for the player's monster on a specific slot was added, allowing certain events to execute if a given Tuxemon matches an expectation. For example: If you want an event to run only if the monster in the first slot is tiny_bat you can now use the syntax: `cond1: monster_property 0 category tiny_bat`

### DIFF
--- a/tuxemon/core/event/conditions/monster_property.py
+++ b/tuxemon/core/event/conditions/monster_property.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from tuxemon.core.event.eventcondition import EventCondition
+
+
+class MonsterPropertyCondition(EventCondition):
+    """ Checks to see if a monster has a given property
+    """
+    name = "monster_property"
+
+    def test(self, game, condition):
+        """Checks to see if a monster has a given property
+
+        :param game: The main game object that contains all the game's variables.
+        :param condition: A dictionary of condition details. See :py:func:`core.map.Map.loadevents`
+            for the format of the dictionary.
+
+        :type game: core.control.Control
+        :type condition: Dictionary
+
+        :rtype: Boolean
+        :returns: True or False
+
+        """
+        slot = condition.parameters[0]
+        prop = condition.parameters[1]
+        val = condition.parameters[2]
+        monster = game.player1.monsters[slot]
+
+        if prop == "name":
+            return monster.name == val
+        elif prop == "level":
+            return str(monster.level) == val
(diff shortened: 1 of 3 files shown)


tuxemon/core/event/conditions/monster_property.py
```diff
@@ -0,0 +1,70 @@
...
+            return False
+
+        monster = game.player1.monsters[slot]
+        if prop == "name":
+            return monster.name == val
+        elif prop == "level":
+            return str(monster.level) == val
+        elif prop == "level_reached":
+            return monster.level >= int(val)
+        elif prop == "type":
```
bitcraft: should we use type here?  seems confusing if a monster can be "water type", etc.  perhaps we need another name, such as "prototype", "lineage", something like that.

tuxemon/core/states/combat/combat.py
```diff
@@ -304,8 +313,10 @@ def transition_phase(self, phase):
             # TODO: proper match check, etc
             # This assumes that player[0] is the human playing in single player
             if self.remaining_players[0] == self.players[0]:
+                self.players[0].game_variables['battle_last_won'] = 'true'
```
bitcraft: what happens if you run from the battle?  this value should be cleared at the beginning of a battle.

MirceaKitsune: I mainly went with that name to mirror the existing name in the locale string translation.

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/locale.py#L304

bitcraft: fair enough.  i think you meant to reply to my other comment?

tuxemon/core/event/conditions/monster_property.py
MirceaKitsune: I mainly went with that name to mirror the existing name in the locale string translation, as it would likely be best to keep all of them identical whenever possible.

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/locale.py#L304

tuxemon/core/states/combat/combat.py
MirceaKitsune: I didn't think of that. The values are replaced and recorded when any battle starts, so it's essentially the name of the last monster you met.

What if instead I changed the battle_last_won variable to have 3 values: Won, lost, ran? That way it will still record the last monster you encountered, but at the same time what happened when you met them... an NPC can then poke fun of you for seeing you run away from said monster.

MirceaKitsune: Yes, sorry about that: I moved my comment above on its line.

MirceaKitsune:
Renamed battle_last_won to battle_last_status, now used to record exactly how your last random encounter ended. The values can be "won", "lost", "draw", "ran".

tuxemon/core/states/combat/combat.py
bitcraft: "last_battle_result" seems fine to me.  possible values are lost, ran, won.  are there draw matches?

tuxemon/core/event/conditions/monster_property.py
bitcraft: sounds good

tuxemon/core/states/combat/combat.py
bitcraft: very minor, please rename "last_battle_status" to "last_battle_result"

MirceaKitsune: There is a draw match phase. It might be used or could be in the future, so it's not a bad idea to add a scenario for that as well.

MirceaKitsune: Renamed it to battle_last_result. I'd rather keep the battle_ prefix, as such variables might be used for other things in the future and this keeps them cleanly organized.

bitcraft:
cool looks good.  any more changes?

MirceaKitsune:
It's fine as it is for me, your inputs were the last ones.

PR #611, opened 2020-01-19
city park story rewrite

Qiangong2:
City Park is almost completely finished (main story, not side quests). I am waiting on a couple sprites to be made, otherwise the scene will look out of place.


bitcraft:
Merged

PR #612, opened 2020-01-20
Party level conditional

MirceaKitsune:
Add a conditional to check if the party's lowest, highest, or average level has reached a given value. Useful to create story points where the player must have at least one monster of level X before advancing further (mandatory grinding), as well as special areas (eg: tournaments) where you can only enter if you're high level enough. Intended only for carried monsters, storage doesn't count.

```
cond1: party_level highest 10 // True if at least one monster is level 10 or higher
cond2: party_level lowest 5 // True if at least one monster is level 5 or higher
cond3: party_level average 3 // True if the average levels of all monsters are 3 or higher
```


tuxemon/core/event/conditions/party_level.py
```diff
@@ -0,0 +1,94 @@
...
+            return level_average >= level
+
+        # Check to see if the party's lowest level monster has reached this level.
+        elif check == "lowest":
+            return level_lowest >= level
+
+        # Check to see if the party's highest level monster has reached this level.
+        elif check == "highest":
+            return level_highest >= level
+
```
bitcraft: suggestion: store the lowest, highest, and average levels as variables so that a dialog can use them.  For instance "You cannot join this tournament because your highest level monster is only X!"

bitcraft: For that to work properly though, you would have to run this action, then a dialog after it.

MirceaKitsune: That's actually a very good idea! Only problem with it is that, it would require a new action, and I'm not sure if adding one just for this makes sense. It would be somewhat absurd to set variables upon checking a conditional. So not sure how to go about this one.

tuxemon/core/event/conditions/party_level.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        player = game.player1
+        for monster in player.monsters:
+            
+        """
+        if len(game.player1.monsters) == 0:
+            return False
+
+        level_lowest = 100
```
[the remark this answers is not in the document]
MirceaKitsune: Can I add a MAX_LEVEL constant in monsters.py? That would be easier to work with I think. I'll default it to 9999 and make sure the level_up function caps to it as well.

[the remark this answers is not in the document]
bitcraft: yeah, that sounds great

tuxemon/core/event/conditions/party_level.py
bitcraft: i see what you are saying.  actually, since i am thinking about it, a condition runs at least once per frame, so the variables will always be updated.  as long as the map has this condition, then these variables will be up to date.  i was mistaken by saying "run this action first"....since its a condition, the game is checking it often.

bitcraft: its maybe a potential bug if a map designer doesn't add this condition if they want these values in a dialog, but i think its a very easy fix and they likely will have this condition on the map if they have a dialog about it, so i don't see a huge problem.

MirceaKitsune: Actually, in this case, I can simply remove the conditional and do this as an action entirely. You can simply check the variables directly then, no need for extra functions.

In fact... I could do this internally altogether! Whenever a monster is added or levels up, the game variables with the party levels are automatically updated, no need for any more map actions. Should I go with this approach?

bitcraft: thats not a bad idea.  its somewhat problematic because monsters would have to have a reference to their player object.  monsters shouldn't have party logic in them.  npcs shouldn't be directly modifying game variables because variables are for the game player.  suppose a trainer npc levels up a monster, then it would overwrite the game variables.  i do think your idea is best overall, but i think it would require more changes to do it correctly, and the game isn't really at a point to support it.  maybe adding a conditional 'if self is game.player1` might work.  feel free to experiment.

MirceaKitsune: The latest commit contains what seemed like the best compromise solution: An NPC function records party status variables now. This function is filtered by self.is_player so running it for another NPC will make no changes.

xirsoi:
What if you wanted to have something restricted to "max level must be *below* X" ?

MirceaKitsune:
@xirsoi Simply reverse the check. IIRC you just use `not` instead of `is` as the first parameter.

PR #615, opened 2020-01-21
Allow dialogs to contain avatars

MirceaKitsune:
Dialog event actions may now specify an optional additional parameter, which can be used to draw any avatar you'd like that dialog to have. This avatar will be displayed in the upper-left corner of the box, next to the name of the person speaking.

Avatars may currently be specified in 3 ways: A monster slot, a monster name, or the full path to the sprite file. Later on we can add the ability to give NPC's avatars and specify those by name, but for now the automatic translation only works for monsters. Examples:

```
act1: dialog 'Whatever I'm saying',monster 0
act1: dialog 'Whatever I'm saying',monster fruitera
act1: dialog 'Whatever I'm saying',gfx/sprites/battle/fruitera-menu01.png
```

Screenshot_20200121_185255

### DIFF
--- a/mods/tuxemon/maps/sphalian_town_house.tmx
+++ b/mods/tuxemon/maps/sphalian_town_house.tmx
@@ -54,7 +54,7 @@
   </object>
   <object id="24" name="Watch TV" type="event" x="32" y="96" width="16" height="16">
    <properties>
-    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}.,monster 0"/>
+    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}.,0"/>
     <property name="cond1" value="is party_size greater_than,0"/>
     <property name="cond2" value="is player_at 2,6"/>
     <property name="cond3" value="is player_facing up"/>
--- a/mods/tuxemon/maps/sphalian_town_house.tmx
+++ b/mods/tuxemon/maps/sphalian_town_house.tmx
@@ -54,7 +54,7 @@
   </object>
   <object id="24" name="Watch TV" type="event" x="32" y="96" width="16" height="16">
    <properties>
-    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}."/>
+    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}.,monster 0"/>
     <property name="cond1" value="is party_size greater_than,0"/>
     <property name="cond2" value="is player_at 2,6"/>
     <property name="cond3" value="is player_facing up"/>
--- a/tuxemon/core/monster.py
+++ b/tuxemon/core/monster.py
@@ -39,6 +39,8 @@ from tuxemon.core import tools
 from tuxemon.core import ai, db, fusion
 from tuxemon.core.locale import T
 from tuxemon.core.technique import Technique
+from tuxemon.core.sprite import Sprite
+from tuxemon.core.pyganim import PygAnimation
 
 logger = logging.getLogger(__name__)
 
@@ -451,11 +453,11 @@ class Monster(object):
         :returns: The surface of the monster sprite
         """
         if sprite == "front":
-            surface = tools.load_sprite(self.front_battle_sprite, **kwargs)
+            surface = tools.load_and_scale(self.front_battle_sprite)
         elif sprite == "back":
-            surface = tools.load_sprite(self.back_battle_sprite, **kwargs)
+            surface = tools.load_and_scale(self.back_battle_sprite)
         elif sprite == "menu":
-            surface = tools.load_sprite(self.menu_sprite, **kwargs)
+            surface = tools.load_and_scale(self.menu_sprite)
         else:
             raise ValueError("Cannot find sprite for: {}".format(sprite))
 
@@ -463,10 +465,16 @@ class Monster(object):
         for flair in self.flairs.values():
             flair_path = self.get_sprite_path("gfx/sprites/battle/{}-{}-{}".format(self.slug, sprite, flair.name))
             if flair_path != MISSING_IMAGE:
-                flair_sprite = tools.load_sprite(flair_path, **kwargs)
-                surface.image.blit(flair_sprite.image, (0, 0))
+                flair_sprite = tools.load_and_scale(flair_path)
+                surface.blit(flair_sprite, (0, 0))
 
-        return surface
+        images = [(surface, .1)]
+        tech = PygAnimation(images, False)
+        sprite = Sprite()
+        sprite.image = tech
+        sprite.rect = surface.get_rect(**kwargs)
+
+        return sprite
 
     def set_flairs(self):
         """Sets the flairs of this monster if they were not already configured
--- a/tuxemon/core/monster.py
+++ b/tuxemon/core/monster.py
@@ -39,8 +39,6 @@ from tuxemon.core import tools
 from tuxemon.core import ai, db, fusion
 from tuxemon.core.locale import T
 from tuxemon.core.technique import Technique
-from tuxemon.core.sprite import Sprite
-from tuxemon.core.pyganim import PygAnimation
 
 logger = logging.getLogger(__name__)
 
@@ -470,8 +468,8 @@ class Monster(object):
         for flair in self.flairs.values():
             flair_path = self.get_sprite_path("gfx/sprites/battle/{}-{}-{}".format(self.slug, sprite, flair.name))
             if flair_path != MISSING_IMAGE:
-                flair_sprite = tools.load_and_scale(flair_path)
-                surface.image.blit(flair_sprite, (0, 0))
+                flair_sprite = tools.load_sprite(flair_path, **kwargs)
+                surface.image.blit(flair_sprite.image, (0, 0))
 
         return surface
 
--- a/tuxemon/core/monster.py
+++ b/tuxemon/core/monster.py
@@ -468,11 +468,13 @@ class Monster(object):
                 flair_sprite = tools.load_and_scale(flair_path)
                 surface.blit(flair_sprite, (0, 0))
 
-        images = [(surface, .1)]
-        tech = PygAnimation(images, False)
+        images = [(surface, 0.1)]
+        anim = PygAnimation(images, True)
+        anim.play()
         sprite = Sprite()
-        sprite.image = tech
-        sprite.rect = surface.get_rect(**kwargs)
+        sprite.image = anim
+        sprite.rect = anim.get_rect(**kwargs)
+        sprite.image.blit(anim, (0, 0))
 
         return sprite
 
--- a/tuxemon/core/monster.py
+++ b/tuxemon/core/monster.py
@@ -251,7 +251,8 @@ class Monster(object):
         self.sprites = {}
         self.front_battle_sprite = ""
         self.back_battle_sprite = ""
-        self.menu_sprite = ""
+        self.menu_sprite_1 = ""
+        self.menu_sprite_2 = ""
 
         self.set_state(save_data)
         self.set_stats()
@@ -305,7 +306,8 @@ class Monster(object):
         # Look up the monster's sprite image paths
         self.front_battle_sprite = self.get_sprite_path(results['sprites']['battle1'])
         self.back_battle_sprite = self.get_sprite_path(results['sprites']['battle2'])
-        self.menu_sprite = self.get_sprite_path(results['sprites']['menu1'])
+        self.menu_sprite_1 = self.get_sprite_path(results['sprites']['menu1'])
+        self.menu_sprite_2 = self.get_sprite_path(results['sprites']['menu2'])
 
         # Load the monster AI
         # TODO: clean up AI 'core' loading and what not
@@ -453,11 +455,14 @@ class Monster(object):
         :returns: The surface of the monster sprite
         """
         if sprite == "front":
-            surface = tools.load_and_scale(self.front_battle_sprite)
+            surface = tools.load_sprite(self.front_battle_sprite, **kwargs)
         elif sprite == "back":
-            surface = tools.load_and_scale(self.back_battle_sprite)
+            surface = tools.load_sprite(self.back_battle_sprite, **kwargs)
         elif sprite == "menu":
-            surface = tools.load_and_scale(self.menu_sprite)
+            surface = tools.load_animated_sprite([
+                self.menu_sprite_1, 
+                self.menu_sprite_2],
+                0.25, **kwargs)
         else:
             raise ValueError("Cannot find sprite for: {}".format(sprite))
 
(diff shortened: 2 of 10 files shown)


MirceaKitsune:
There is currently only one issue, I'd like to have this fixed before the branch is merged: The sprite won't animate. Can anyone point me to an example of how I can make the menu01.png and menu02.png sprites switch between one another every 0.x seconds?

tuxemon/core/tools.py
```diff
@@ -385,7 +385,7 @@ def calc_dialog_rect(screen_rect):
     return rect
 
 
-def open_dialog(game, text, menu=None):
+def open_dialog(game, text, avatar=None, menu=None):
```
bitcraft: in the future, please don't change the order of function parameters.  just add new ones to the end.  this could break existing code.  might be ok now, but its good practice to leave parameter order alone.

tuxemon/core/states/dialog/__init__.py
```diff
@@ -60,3 +63,29 @@ def next_text(self):
...
+        params = self.avatar.split(" ")
+        if params[0] == "monster":
+            if params[1].isdigit():
+                player = self.game.player1
+                slot = int(params[1])
+                avatar_sprite = player.monsters[slot].get_sprite("menu")
+            else:
+                avatar_monster = monster.Monster()
+                avatar_monster.load_from_db(params[1])
+                avatar_monster.flairs = {} # Don't use random flair graphics
```
bitcraft: i see why you added this line to prevent flairs.  please open a follow up ticket to organize the flairs into groups according to the sprite type they overlay, such as front, back, menu, etc

bitcraft: ideally, if we have flairs, they should also be on the menu sprite (like hair ribbons, etc) and hiding them or not drawing them would be completely automatic.

tuxemon/core/states/dialog/__init__.py
```diff
@@ -60,3 +63,29 @@ def next_text(self):
...
+
+    def draw_avatar(self):
+        if not self.avatar:
+            return
+
+        # If the prefix is "monster" the second parameter refers to a monster
+        # If this parameter is a number, we're referring to a monster slot in the party
+        # If this parameter is a string, we're referring to a monster name
+        # If there's no prefix, the parameter represents the path to a sprite file
+        params = self.avatar.split(" ")
```
bitcraft: im not really a fan of this.  if we are storing items on the dialog class, such as `avatar`, it shouldn't be a raw string that we need to parse.  items stored on objects shouldn't require additional steps to get their meaning.  I think the most clean way to would be storing the "avatar" as a pygame surface, so it is just blit in the correct place.

this logic for decoding the meaning of the "avatar string" belongs in the code where the string is read; it should be in the dialog event action code.  please move it there and set the dialog avatar as a pygame surface.

bitcraft: Better yet would be setting the avatar as a pygame sprite, but i'm not going to expect that.  a surface here is fine and it can be cleaned up later and refactored as a sprite.  In general, sprites should be our basic "building block" of graphical elements.  Again, not needed now, but just an FYI

tuxemon/core/states/dialog/__init__.py
```diff
@@ -60,3 +63,29 @@ def next_text(self):
             return text
         except IndexError:
             return None
+
```
bitcraft: To animate the avitar, pass apyganim object as the avatar.  Then create a function hook here called "draw":
```python
def draw(self, surface):
   super(DialogState, self).draw(surface)
   # self.avatar.blit(...)
```

tuxemon/core/states/dialog/__init__.py
bitcraft: TLDR; please move this entire function to the dialog event.  only pass a pygame surface (or pyganim amimation) to the dialog state as the avatar.

tuxemon/core/states/dialog/__init__.py
```diff
@@ -60,3 +63,29 @@ def next_text(self):
...
+                player = self.game.player1
+                slot = int(params[1])
+                avatar_sprite = player.monsters[slot].get_sprite("menu")
+            else:
+                avatar_monster = monster.Monster()
+                avatar_monster.load_from_db(params[1])
+                avatar_monster.flairs = {} # Don't use random flair graphics
+                avatar_sprite = avatar_monster.get_sprite("menu")
+        else:
+            avatar_sprite = tools.load_sprite(self.avatar)
```
bitcraft: I think i would rather not load any avatar if the menu sprite is not available.  do you think it would look strange?  im not commited to this either way though

tuxemon/core/states/dialog/__init__.py
```diff
@@ -21,13 +22,15 @@ class DialogState(PopUpMenu):
...
         self.text_queue = kwargs.get("text", list())
+        self.avatar = kwargs.get("avatar", None)
         self.menu = kwargs.get("menu", None)
         self.text_area = TextArea(self.font, self.font_color)
         self.text_area.rect = self.calc_internal_rect()
         self.sprites.add(self.text_area)
 
     def on_open(self):
         self.next_text()
+        self.draw_avatar()
```
bitcraft: please remove this and see another comment about drawing.  drawing should only happen inside the "draw" method.

tuxemon/core/tools.py
MirceaKitsune: Will keep in mind. It only required few changes, so like you said I'll leave it like this for now as it looks cleaner. In the future I shall add them at the end.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: I see what you mean. Only issue is that there are several dialog events (at least 4): This would add yet more code duplication if done everywhere instead of a common function. Where should such a function be placed though?

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: It might indeed be a good idea to remove that. We'll need to support avatars for NPC's later down the road, but it should indeed be done using a similar syntax to monsters now.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: This is only an exception since we don't want random flairs when referencing a monster by name (that's not an actual monster in the world). I'll keep that in mind for later... though I'd rather not make flairs more complex, they are grouped in a simple but efficient way currently.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: Latest commit refractors this: The function to pick the avatar was moved to tools.py and only used once when initializing the dialog. The dialog state is now called with the sprite that was fetched, and is only responsible for adding it on startup the same way it adds the text area.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: Fixed in the latest commit: The parameter now takes only one value, which is either a number or a string. When a number it's a monster slot, when a string it refers to a monster by name. Direct file paths can no longer be specified.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: Fixed in the latest commit: The parameter now takes only one value, which is either a number or a string. When a number it's a monster slot, when a string it refers to a monster by name. Direct file paths can no longer be specified.

The function to pick the avatar was moved to tools.py and only used once when initializing the dialog. The dialog state is now called with the sprite that was fetched, and is only responsible for adding it on startup the same way it adds the text area.

tuxemon/core/tools.py
```diff
@@ -385,7 +386,28 @@ def calc_dialog_rect(screen_rect):
...
+        slot = int(avatar)
+        return player.monsters[slot].get_sprite("menu")
+    else:
+        avatar_monster = tuxemon.core.monster.Monster()
+        avatar_monster.load_from_db(avatar)
+        avatar_monster.flairs = {} # Don't use random flair graphics
+        return avatar_monster.get_sprite("menu")
+
+
+def open_dialog(game, text, avatar=None, menu=None):
```
bitcraft: please search for "open_dialog" and double check that you've found and fixed all the times it is used.  because you have changed the order of the parameters, there may be bugs now.  python isn't static typed, so you won't find the bug until the game is running.  please resolve this after youve looked again.

bitcraft: alternatively, you could go back to the original order and save yourself the trouble.

MirceaKitsune: Will take an extra close look after I figure out the animation issue. Already looked but will be extra sure.

MirceaKitsune: I did a grep through the whole code and looked carefully: All instances of open_dialog that specified the menu parameter were adjusted, the parameters should be in order everywhere. Even if somehow I managed to miss the obvious, I'm always around and could fix it at once.

tuxemon/core/states/dialog/__init__.py
MirceaKitsune: I resolved the animation part by creating a new load_animated_sprite function in tools.py: It can be reused for future animations, and also serves to spare others of having to rediscover what I spent a day finding out... mostly the same as load_sprite except it takes a list of images and a delay then returns an animated surface instead. That draw function was not even necessary, as issuing animation.play() when creating the sprite activates it just fine.

MirceaKitsune:
The latest commit includes all changes planned as well as fixing all previous observations: This is ready to be merged as far as I'm aware. I tested various signs and dialogues as well as battle... everything's working as intended with no warnings or errors on my end.

MirceaKitsune:
The latest commit includes all changes planned as well as fixing all previous observations: This is ready to be merged as far as I'm aware. I tested various signs and dialogues as well as battle... everything's working as intended with no warnings or errors on my end.

MirceaKitsune:
To help make it easier to review this, I figured I'd do a description of the order of functions and what everything does so it's easier to keep track:

https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-b156dfa5ef8d9752d7b6e1eb459aeed0R64
https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-2b46fed98d86c7bc67d84811d07fda40R81
https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-2edb01ad5be01b07071bece9517c14eaR66
https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-2edb01ad5be01b07071bece9517c14eaR66

Each dialog event action uses `tools.get_avatar` which returns the PyGame surface of an avatar.

https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-997a72d7854e725c6c93f5cecb1022e3R397

The `tools.get_avatar` function described above. Its purpose is to return an avatar surface based on the parameter given. Future NPC avatar support shall be added here, currently it handles monsters by using their `monster.get_sprite` function now responsible for fetching all monster sprites.

https://github.com/Tuxemon/Tuxemon/pull/615/files#diff-997a72d7854e725c6c93f5cecb1022e3R169

New function similar to `tools.get_sprite` except for animated sprites. Works the same way except it takes a list of images rather than one image, as well as a float specifying the animation rate. Intended to be reused throughout the code in every case where someone needs a simple sprite cycle. In this case it's now called by `monster.get_sprite` for fetching the animated menu image.

MirceaKitsune:
@kerizane As you helped with testing the other branches I made, could you also confirm this looks good to you please? It's been here for a week and I'm also preparing to use the feature in the dialogues of a game I'm now developing.

tuxemon/core/tools.py
```diff
@@ -385,7 +440,7 @@ def calc_dialog_rect(screen_rect):
     return rect
 
 
-def open_dialog(game, text, menu=None):
+def open_dialog(game, text, avatar=None, menu=None):
```
kerizane: Add a docstring for avatar and menu

tuxemon/core/tools.py
```diff
@@ -165,6 +166,31 @@ def load_sprite(filename, **rect_kwargs):
...
+    """ Load a set of images and return an animated pygame sprite
+
+    Image name will be transformed and converted
+    Rect attribute will be set
+
+    Any keyword arguments will be passed to the get_rect method
+    of the image for positioning the rect.
+
+    :param filename: Filename to load
+    :rtype: core.sprite.Sprite
```
kerizane: docstring for delay.
dunno about kwargs

tuxemon/core/tools.py
MirceaKitsune: I'm not familiar with docstring and couldn't find that in the code. Can you point me to where that is please?

mods/tuxemon/maps/sphalian_town_house.tmx
```diff
@@ -54,7 +54,7 @@
   </object>
   <object id="24" name="Watch TV" type="event" x="32" y="96" width="16" height="16">
    <properties>
-    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}."/>
+    <property name="act1" value="dialog You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}.,0"/>
```
bitcraft: Last time i tried this without any monsters the game crashed.  has that been fixed?

tuxemon/core/monster.py
```diff
@@ -455,7 +457,10 @@ def get_sprite(self, sprite, **kwargs):
         elif sprite == "back":
             surface = tools.load_sprite(self.back_battle_sprite, **kwargs)
         elif sprite == "menu":
-            surface = tools.load_sprite(self.menu_sprite, **kwargs)
+            surface = tools.load_animated_sprite([
+                self.menu_sprite_1, 
+                self.menu_sprite_2],
```
bitcraft: what happens if a monster JSON does not implement the 2nd sprite?

mods/tuxemon/maps/sphalian_town_house.tmx
MirceaKitsune: Just retested with the latest commit: No crash either when there's no parameter nor when an invalid one is used.

tuxemon/core/monster.py
MirceaKitsune: Not aware of any cases: I assumed that's considered a base requirement for any working monster. At worst the 2nd sprite can be copied and renamed.

I don't have a guard against that in place. If you believe one is necessary I can do so.

bitcraft: Please test this with a monster that doesn't implement the 2nd sprite.  If it works, then lets use that.  If it doesn't work, please make sure that animated avatars is optional because I don't want to make work for the artists for this feature.

MirceaKitsune: If the second menu sprite is missing, it will instead animate between the first sprite and the "?" sprite (missing monster image).

MirceaKitsune: Fixed. If a file isn't found it's simply not added to the animation any more: Both the first and second menu sprites can be missing, the other will just show as a static image instead.

tuxemon/core/tools.py
MirceaKitsune: Oh... they're the comments at the beginning of functions! Fixed.

tuxemon/core/tools.py
MirceaKitsune: Fixed for delay. The original get_sprite doesn't have a comment for the rect either and I followed its model.

bitcraft:
```python
:param int delay: ...
```

tuxemon/core/tools.py
```diff
@@ -368,6 +395,35 @@ def play(self):
...
 
 
+def get_avatar(game, avatar):
+    """Gets the avatar sprite of a monster or NPC.
+
+    If avatar is a number, we're referring to a monster slot in the player's party
+    If avatar is a string, we're referring to a monster by name
+    TODO: If the monster name isn't found, we're referring to an NPC on the map
+
+    :rtype: Pygame surface
```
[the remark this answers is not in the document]
bitcraft: mention here that this function is used to parse the string values for dialog event actions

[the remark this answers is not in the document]
MirceaKitsune: Fixed.

tuxemon/core/tools.py
bitcraft: Please reread the comment above.  it is missing "int"

bitcraft:
thank you again!

Issue #618, opened 2020-01-23
Allow maps to specify which tile layer is drawn above the player


MirceaKitsune:
Mappers are currently limited in how they can organize their map tiles by a hardcoded default of 3 layers drawn under the player. This means that if you have more than 3 tile layers, layer 4 and above will be automatically drawn on top of the player and other NPC's.

This limit should be customizable and possible to override by the map. The most logical option seems a new map property, specifying which layer will cover the player when set. If not set the default should of course remain the same as to not break existing maps.

xirsoi:
The most intuitive thing, imo, would be to have the property specify which layer characters (PC and NPCs) will be drawn, rather than the first layer that will be drawn above them. But that's a detail :P

xirsoi:
I'd like to take a crack at this.

xirsoi:
Okay, think I have it, just need to actually test it. Hopefully I'll have time tomorrow.

MirceaKitsune:
Thanks! I probably could have as well but am already juggling two branches and working on other things, so it does help.

xirsoi:
I've been wanting to get more involved anyway, thought this was an easy enough thing to work on.

xirsoi:
New map property sprite_layer added.

PR #619, opened 2020-01-24
Support for map property to specify character drawing layer #618

xirsoi:
Adds a new map level property called pc_layer. The world state passes this to the NPC.get_sprites() method, where it's inserted into the returned sprite tuples.

The value defaults to 2 (same as before) in the Map init.

I added this property to two of the maps (temporarily) for testing.


tuxemon/core/npc.py
```diff
@@ -244,11 +244,14 @@ def load_sprites(self):
         # at the same time, so that way they'll always be in sync with each other.
         self.moveConductor.add(self.sprite)
 
-    def get_sprites(self):
+    def get_sprites(self, layer):
```
bitcraft: i think that the layer attribute should be on the NPC.  its not common, but occasionally it is useful to change NPC layers to add depth to the world.  In that case, the layer attribute would be unique to each sprite/NPC/entitiy.  Please consider moving the layer attribute to the NPC class.

xirsoi: I'm not sure how that will work with maps specifying the layer to draw sprites, even if the npc overrides it I think it'll end up kinda ugly.

bitcraft: yeah.  its not worth thinking about too much now, but i think we should have a comment so that we don't keep adding to this kludge.  its a fine kludge, but when/if NPCs can move to different layers, i want this refactored and removed.

xirsoi:
Okay @bitcraft , ready for your final review.

bitcraft:
thank you!

PR #623, opened 2020-01-26
Ensure mods are set before loading locale and database paths

### LINKED ISSUE #620: Locale and db not being read from mod path
There appears to be an issue with the mod loader, which blocks the locale from being read from the correct mod directory. base.po and base.mo are always loaded exclusively from mods/tuxemon and never from mods/<my_mod>

This appears to be caused by the locale code running before mods are set: I used a debug print to confirm that by the time locale.py calls fetch in prepare.py, the mods list only contains "tuxemon" and no other entries... only later calls to prepare.fetch are able to see both. The locale code should be executed after CONFIG.mods has been set.

MirceaKitsune:
Fixes #620

Due to the fact that locale and database paths were read instantly at startup, they did not have time to catch the mods being added to the mod list. As a result some resources could only be read from mods/tuxemon and never mods/<my_game> making the mod loaded unusable.

This required a bit of refractoring to locale and databases, which now receive the signal to load their data from pygame_init after the base settings have been set. As a bonus, this PR cleans up the database code and also removes several cases of databases being pointlessly loaded more than once.

I tested this as thoroughly as I could and am not seeing any issues: This PR loads fine for both stock Tuxemon as well as my custom game directory. Further testing is encouraged still.

### DIFF
--- a/tuxemon/core/db.py
+++ b/tuxemon/core/db.py
@@ -61,18 +61,8 @@ class JSONDatabase(object):
 
     """
 
-    def __init__(self, dir=None):
-        self.database = {
-            "item": {},
-            "monster": {},
-            "npc": {},
-            "technique": {},
-            "encounter": {},
-            "inventory": {},
-            "environment": {},
-        }
-        if dir:
-            self.load(dir)
+    def __init__(self):
+        self.database = {}
 
     def load(self, directory="all"):
         """Loads all data from JSON files located under our data path.
@@ -84,6 +74,16 @@ class JSONDatabase(object):
         :returns: None
 
         """
+
+        self.database = {
+            "item": {},
+            "monster": {},
+            "npc": {},
+            "technique": {},
+            "encounter": {},
+            "inventory": {},
+            "environment": {},
+        }
 
         self.path = prepare.fetch("db")
         if directory == "all":
@@ -193,27 +193,5 @@ def set_defaults(results, table):
 
     return results
 
-# Database storage class
-class Databases(object):
-    def __init__(self):
-        self.item = JSONDatabase()
-        self.monster = JSONDatabase()
-        self.npc = JSONDatabase()
-        self.technique = JSONDatabase()
-        self.encounter = JSONDatabase()
-        self.inventory = JSONDatabase()
-        self.environment = JSONDatabase()
-        self.all = JSONDatabase()
-
-    def collect_databases(self):
-        self.item.load("item")
-        self.monster.load("monster")
-        self.npc.load("npc")
-        self.technique.load("technique")
-        self.encounter.load("encounter")
-        self.inventory.load("inventory")
-        self.environment.load("environment")
-        self.all.load("all")
-
-# Register default databases
-databases = Databases()
+# Global database container
+databases = JSONDatabase()
--- a/tuxemon/core/db.py
+++ b/tuxemon/core/db.py
@@ -62,7 +62,6 @@ class JSONDatabase(object):
     """
 
     def __init__(self, dir=None):
-        self.path = prepare.fetch("db")
         self.database = {
             "item": {},
             "monster": {},
@@ -86,6 +85,7 @@ class JSONDatabase(object):
 
         """
 
+        self.path = prepare.fetch("db")
         if directory == "all":
             self.load_json("item")
             self.load_json("monster")
@@ -192,3 +192,22 @@ def set_defaults(results, table):
                 )
 
     return results
+
+# Register default databases
+env_db = JSONDatabase()
+techniques_db = JSONDatabase()
+items_db = JSONDatabase()
+inventory_db = JSONDatabase()
+npc_db = JSONDatabase()
+monster_db = JSONDatabase()
+encounter_db = JSONDatabase()
+
+# Load default databases
+def collect_databases():
+    env_db.load("environment")
+    techniques_db.load("technique")
+    items_db.load("item")
+    inventory_db.load("inventory")
+    npc_db.load("npc")
+    monster_db.load("monster")
+    encounter_db.load("encounter")
--- a/tuxemon/core/db.py
+++ b/tuxemon/core/db.py
@@ -61,8 +61,19 @@ class JSONDatabase(object):
 
     """
 
-    def __init__(self):
-        self.database = {}
+    def __init__(self, dir=None):
+        self.path = None
+        self.database = {
+            "item": {},
+            "monster": {},
+            "npc": {},
+            "technique": {},
+            "encounter": {},
+            "inventory": {},
+            "environment": {},
+        }
+        if dir:
+            self.load(dir)
 
     def load(self, directory="all"):
         """Loads all data from JSON files located under our data path.
@@ -74,16 +85,6 @@ class JSONDatabase(object):
         :returns: None
 
         """
-
-        self.database = {
-            "item": {},
-            "monster": {},
-            "npc": {},
-            "technique": {},
-            "encounter": {},
-            "inventory": {},
-            "environment": {},
-        }
 
         self.path = prepare.fetch("db")
         if directory == "all":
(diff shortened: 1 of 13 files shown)


MirceaKitsune:
I was wondering if we could put this one on a fast track please. Unlike my other PR's which add new features or fix minor issues, this is a major bug fix. I'm working on my own game at the moment using the new mod system, and am presently only able to run it by using this branch.

I understand @bitcraft is busy and don't wish to rush him more than I have in the past. I was wondering if other developers are still available, and any one of you could review and merge this. @ShadowApex @pastaq @superman2k5 can you take a look at this and approve it please?

In terms of functionality I don't doubt it works: I've been using it for more than a day to constantly test changes to my mod all over the place. I never got any crashes, and missing file cases would almost always cause a crash at startup too. Only thing is that all subdirectories need to be present in your mod's db subdirectory (even if they're empty), also some hardcoded status techniques must be copied over... this is the case with the default mod too, mixing them is a feature for later.

kerizane:
Looks fine, will do a quick test this evening.

kerizane:
Error in cotton_town:

Python2
  File "/home/k/Tuxemon/tuxemon/core/event/actions/create_npc.py", line 76, in start
    sprite = databases.npc[slug].get('sprite_name')
TypeError: 'JSONDatabase' object has no attribute '__getitem__'

Python3
  File "/home/k/Tuxemon/tuxemon/core/event/actions/create_npc.py", line 76, in start
    sprite = databases.npc[slug].get('sprite_name')
TypeError: 'JSONDatabase' object is not subscriptable

tuxemon/core/event/actions/create_npc.py
```diff
@@ -72,7 +73,7 @@ def start(self):
                 slug
             )
         else:
-            sprite = db.JSONDatabase('npc').database['npc'][slug].get('sprite_name')
```
kerizane: sprite = databases.npc.database['npc'][slug].get('sprite_name')

kerizane:
`sprite = databases.npc.database['npc'][slug].get('sprite_name')`

This is pretty verbose, we should simplify it. 
Can be done in another ticket though

kerizane:
Other than the create_npc crash, I'm happy.

I don't have merge rights though.

tuxemon/core/event/actions/create_npc.py
MirceaKitsune: Really strange that I wasn't seeing it; I've been able to enter the maple house and talk to the professor's wife just fine. I tested your change and it works great so I committed it: Please double-check that this made the issue go away.

MirceaKitsune:
Yeah, we should keep that simplification in mind for later. The rest of the database structure is the same, I only created a class to set and store all databases in one place; This was needed so pygame init could set them after mods are configured rather than each script (monster.py, techniques.py, etc) doing it instantly themselves... it's also cleaner and avoids some duplicate db reads.

tuxemon/core/event/actions/create_npc.py
kerizane: Different code path, only affects some npcs

tuxemon/core/event/actions/create_npc.py
```diff
@@ -72,7 +73,7 @@ def start(self):
...
+            sprite = databases.npc.database['npc'][slug].get('sprite_name')
```
bitcraft: this new format is extremely obtuse.  please find a way to make it more simple:
```python
db.npc['npc'][slug].get('sprite_name')
```
would be much better.

bitcraft:
Overall, I'm not happy with this change.  I didn't expect a huge refactor of a core system without discussion first.  Please see my comment.  My primary objection is that the db system now requires remembering 3 names to access items.  Deeply nested structures like this should be avoided.  Please continue refactoring and make sure everything is tested.  You should be able to play until cotton town, have a monster, some items, and not crash.

MirceaKitsune:
Latest commit removes the unnecessary level of indentation, using a single global database instead of an object spawning a new db instance for each field. I really should have seen this earlier.

tuxemon/core/event/actions/create_npc.py
MirceaKitsune: Latest commit should fix this. It is now just `databases.database['npc']`, a single instance of JSONDatabase() is used as was most likely intended.

tuxemon/core/db.py
```diff
@@ -86,11 +75,22 @@ def load(self, directory="all"):
 
         """
 
+        self.database = {
```
bitcraft: why is this moved?

tuxemon/core/db.py
```diff
@@ -86,11 +75,22 @@ def load(self, directory="all"):
...
+            "item": {},
+            "monster": {},
+            "npc": {},
+            "technique": {},
+            "encounter": {},
+            "inventory": {},
+            "environment": {},
+        }
+
+        self.path = prepare.fetch("db")
```
bitcraft: instance attributes shouldnt be delcared outside of init.  python lets you do that, but its not a good practice.  please move "databases" back to init, and you can set path to None there.  leaving instance variables in init helps developers learn about the class and what it does.

tuxemon/core/db.py
```diff
@@ -86,11 +75,22 @@ def load(self, directory="all"):
...
         if directory == "all":
             self.load_json("item")
             self.load_json("monster")
-            self.load_json("technique")
             self.load_json("npc")
+            self.load_json("technique")
```
bitcraft: why is this moved?

tuxemon/core/db.py
```diff
@@ -192,3 +192,6 @@ def set_defaults(results, table):
                 )
 
     return results
+
+# Global database container
+databases = JSONDatabase()
```
bitcraft: `databases` is misleading.  this is a single object.  please change to `database`, or simply `db`

tuxemon/core/db.py
MirceaKitsune: I noticed it wasn't in the same order as the database properties declared in init. "technique" comes after "npc" there, but below it was above it... I figured this would be more clean just in case.

tuxemon/core/db.py
MirceaKitsune: I think I misunderstood one of our conversations on Discord. Placed it back in the last commit.

tuxemon/core/db.py
MirceaKitsune: Fair enough, I shouldn't have missed that. I put it back in init as None... the load() function must set it for the mod issue to be solved, so only its configuration is done there now.

tuxemon/core/db.py
MirceaKitsune: Renamed to db.

MirceaKitsune:
Last commit tested briefly, no crashes or errors or missing content; I tried changing maps and battling and / or reading signs, for both vanilla tuxemon and my mod... both seem to be working perfectly now.

tuxemon/core/monster.py
```diff
@@ -266,7 +262,7 @@ def load_from_db(self, slug):
         """
 
         # Look up the monster by name and set the attributes in this instance
-        results = monsters.lookup(slug)
+        results = db.lookup(slug)
```
kerizane: results = db.lookup(slug, table="monster")

tuxemon/core/monster.py
```diff
@@ -477,7 +473,7 @@ def set_flairs(self):
         if len(self.flairs) > 0 or self.slug == "":
             return
 
-        results = monsters.lookup(self.slug)
+        results = db.lookup(self.slug)
```
kerizane: results = db.lookup(slug, table="monster")

tuxemon/core/monster.py
MirceaKitsune: Ouch... thanks for catching that.

tuxemon/core/monster.py
MirceaKitsune: Fixed.

bitcraft:
Ok this looks like a good start.  there are still minor issues with it but they can be fixed in later PRs.  thanks again!

PR #624, opened 2020-01-26
Don't allow collision tiles with explicit exits to push the player into other solids

### LINKED ISSUE #617: Collision objects with exit_to allow the player to step into other collisions
I found a little bug with collision objects: When a tile uses the `exit_to` parameter, it allows the player to walk in that direction even when it's blocked by another collision. Normally tiles that use the enter_from exit_to and continue parameters should only allow an NPC to go in that direction if it's itself not blocked, no letting them walk into a solid.

Example: Let's say that at tile 12,16 you have a collision box. This collision object has the property `exit_to up down`. Under and over it at positions 12,15 and 12,17, you place two separate collisions with no properties (they collide in all directions). The player is normally not able to step on 12,15 and 12,17... however if they walk to 12,16, they can enter those two locations from there.

### LINKED ISSUE #629: Players can walk outside of map bounds
When a map lets the player walk to its edges without surrounding them with collision objects, the player is allowed to walk off the screen and into infinity. Further more, the player is unable to walk back from this position and is considered stuck. Players and NPC's should be forbidden to move to positions outside of the map bounds.

MirceaKitsune:
Fixes #617 and now also fixes #629

Previously collision tiles with the "exit_to" property would always let the player exit in that direction, even if another collision was present which doesn't allow entry from that position. This resulted in the player walking into walls and getting stuck or ending up in out of bounds areas, if stepping onto such a tile which neighbored another solid. Collisions with the exit_to parameter will now only allow the player to exit there if that direction isn't itself blocked.

Tested on my maps which make use of complex enter_from and exit_to statements and collision mixtures. Everything seems to be working exactly as intended after this change.

### DIFF
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -588,7 +588,7 @@ class WorldState(state.State):
         # does the tile explicitly define exits?
         try:
             adjacent_tiles = list()
-            for direction in tile["exit"]:
+            for direction in tile['exit']:
                 exit_tile = tuple(dirs2[direction] + position)
                 if exit_tile in skip_nodes:
                     continue
@@ -617,14 +617,11 @@ class WorldState(state.State):
         if skip_nodes is None:
             skip_nodes = set()
 
-        # if there are explicit way to exit this position use that
-        # information only and do not check surrounding tiles.
+        # if there are explicit way to exit this position use that information,
         # handles 'continue' and 'exits'
         tile_data = collision_map.get(position)
         if tile_data:
             exits = self.get_explicit_tile_exits(position, tile_data, skip_nodes)
-            if exits:
-                return exits
 
         # get exits by checking surrounding tiles
         adjacent_tiles = list()
@@ -634,6 +631,11 @@ class WorldState(state.State):
                 ("up", (position[0], position[1] - 1)),
                 ("left", (position[0] - 1, position[1])),
         ):
+            # if exits are defined make sure the neighbor is present there
+            if exits and not neighbor in exits:
+                continue
+
+            # check if the neighbor region is present in skipped nodes
             if neighbor in skip_nodes:
                 continue
 
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -641,7 +641,7 @@ class WorldState(state.State):
 
             # We only need to check the perimeter,
             # as there is no way to get further out of bounds
-            if position[0] in self.invalid_x or position[1] in self.invalid_y:
+            if neighbor[0] in self.invalid_x or neighbor[1] in self.invalid_y:
                 continue
 
             # check to see if this tile is separated by a wall
@@ -921,8 +921,8 @@ class WorldState(state.State):
         self.map_size = map_data["map_size"]
 
         # The first coordinates that are out of bounds.
-        self.invalid_x = (-1, self.map_size[0] + 1)
-        self.invalid_y = (-1, self.map_size[1] + 1)
+        self.invalid_x = (-1, self.map_size[0])
+        self.invalid_y = (-1, self.map_size[1])
 
         # TODO: remove this monkey [patching!] business for the main control/game
         self.game.events = map_data["events"]
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -581,14 +581,14 @@ class WorldState(state.State):
 
         # does the tile define continue movements?
         try:
-            return [tuple(dirs2[tile['continue']] + position)]
+            return [tuple(dirs2[tile["continue"]] + position)]
         except KeyError:
             pass
 
         # does the tile explicitly define exits?
         try:
             adjacent_tiles = list()
-            for direction in tile['exit']:
+            for direction in tile["exit"]:
                 exit_tile = tuple(dirs2[direction] + position)
                 if exit_tile in skip_nodes:
                     continue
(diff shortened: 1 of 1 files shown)


MirceaKitsune:
Additionally fixed #629 in the latest commit. The map bound check was broken, allowing the player to walk outside the world and into infinity. That issue was in the same code area and easy to resolve.

tuxemon/core/states/world/worldstate.py
```diff
@@ -588,7 +588,7 @@ def get_explicit_tile_exits(self, position, tile, skip_nodes):
         # does the tile explicitly define exits?
         try:
             adjacent_tiles = list()
-            for direction in tile["exit"]:
+            for direction in tile['exit']:
```
bitcraft: Do not change the quotes.  Double quotes is part of the PEP-8 style guidelines, which we should try to conform to.  Eventually, we will follow the "Black" standard, which will force all to double quotes.  Please revert this.

tuxemon/core/states/world/worldstate.py
```diff
@@ -634,12 +631,17 @@ def get_exits(self, position, collision_map=None, skip_nodes=None):
...
+                continue
+
+            # check if the neighbor region is present in skipped nodes
             if neighbor in skip_nodes:
                 continue
 
             # We only need to check the perimeter,
             # as there is no way to get further out of bounds
-            if position[0] in self.invalid_x or position[1] in self.invalid_y:
+            if neighbor[0] in self.invalid_x or neighbor[1] in self.invalid_y:
```
bitcraft: why is this changed?

tuxemon/core/states/world/worldstate.py
```diff
@@ -617,14 +617,11 @@ def get_exits(self, position, collision_map=None, skip_nodes=None):
...
             skip_nodes = set()
 
-        # if there are explicit way to exit this position use that
-        # information only and do not check surrounding tiles.
+        # if there are explicit way to exit this position use that information,
         # handles 'continue' and 'exits'
         tile_data = collision_map.get(position)
         if tile_data:
             exits = self.get_explicit_tile_exits(position, tile_data, skip_nodes)
-            if exits:
```
bitcraft: why is this changed?

bitcraft:
It's been a while since I've worked on this part.  I mostly don't like these changes because if a map defines a broken exit, then the game should not be silently fixing broken maps; the map should be fixed.  Its not the engines responsibility to fix broken maps.  This kind of attitude leads to deeply rooted hacks.

MirceaKitsune:
Will attend to the changes tomorrow. Just wanted to clarify the collision tile changes aren't meant to prevent issues caused by lazy mapping: They're to ensure that collisions with the exit_to property can safely sit next to other collision objects as intended (?)

Without it you cannot for instance place a thin fence vertically the right side of a house: You can give the entire length of fence `exit_to left up down` to let the player walk alongside it from the inside but not past it. But if you go too high up and bump into the house, you'll be able to walk into it.

I don't believe this was the desired functionality either when the feature was added, it feels and works in a broken way IMO. My new maps use such collision tiles around special corners... the old functionality let the player walk out of bounds all over the place.

kerizane:
```
        self.invalid_x = (-1, self.map_size[0])
        self.invalid_y = (-1, self.map_size[1])
```
FYI, this bit is there to stop the pathfinding planner from considering routes that are out of bounds. It helps the performance.

The reason the planner could wander out of bounds is that the planner is unaffected by the route_transitions and will happily walk through them and off the map.

MirceaKitsune:
@kerizane Good to know. Looking at the code, invalid_x and invalid_y are only used in the check I fixed. It's the same check that should prevent the player from walking out of bounds.

That check was already implemented but broken in two places: First it was checking the player's current position, not where they're walking toward. Second the +1 added an extra tile to the right and bottom... maps tiles are counted from 0, so for instance 36 is actually 35.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: I moved it below in the `for direction, neighbor` loop. This is what allowed the player to walk into other collisions from a neighboring collision defining an `exit_to` toward it: If exits were found here, they were immediately returned as a valid position. With my change, if exits are defined, we instead ensure that the target position is present in them... if not then it's invalid, if yes then it's valid but we now also proceed with the other checks to make sure the position can be walked to.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: The out of bounds check is an entirely different bug. This PR fixes two separate issues: Collision tiles with `exit_to` letting the player walk into neighboring solids, and map bounds not being checked properly. I was going to make a different PR for each one, but it's the exact same piece of code so I did them both here to avoid Git conflicts.

Regarding the `exit_to` check I explained in the review below how and why I changed that. As far as the map boundary fix goes, it was broken in two places: First it was checking the player's current position, not where they're attempting to walk toward... the modification quoted here fixes that part. Second issue was that the + 1 added to invalid_x and invalid_y further down added an extra tile to the right and bottom, since map tiles are counted from 0 so for instance 36 is actually 35... this allowed the player to walk one extra tile out of bounds in those directions.

I already tested those modifications individually to make sure they each do what I understood they do, by walking into each edge of two differently sized maps. My tests showed that each change was necessary to fix a particular problem.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: Fixed in the last commit. I noticed that single and double quotes were both used in different places, and since single quotes are most common I switched to that thinking it's more aesthetic. I asked about this on Discord and was clarified that double quotes are the way to go, so I reverted this and also fixed a few similar cases further down in the same code area.

bitcraft:
Please stand by for regression fixes.  As Kerizane mentioned, a change was made to code that handles pathfinding.  Thanks again!

PR #625, opened 2020-01-26
Fix transition_teleport being triggered twice when teleporting to the same map

### LINKED ISSUE #621: transition_teleport triggers two times when teleporting to the same map
There's a strange issue with the `transition_teleport` event which feels like it should be a bug. Whenever this event teleports you to a location on the same map, it will execute twice and teleport you two times. This doesn't happen when the map you're being teleported to is a different one.

This should be possible to replicate using the following steps: Create or edit any map, for instance map1.tmx. On it add an event, at lets say position `37,48`. Give it the following properties:

```
act1: transition_teleport map1.tmx,37,49,0.25
cond1: is player_at
```
Now open that map and walk to tile 37,48. You should be teleported one tile down to tile 37,49. This does happen but it occurs two times: First the screen fades to black and back as you move to the new location, but right after you've moved there the screen will fade to black once more and put you at that position a second time. This second occurrence must not happen, as the event is filtered by the player_at conditional which should prevent a following execution after you've teleported.

I take it the code is queuing the teleport event twice while player_at is still seen as true. I'm assuming the solution is to immediately block the execution of any further events during a teleport transition phase, and only re-enable triggering of new events once the teleport has completed.

MirceaKitsune:
Fixes #621

When transition_teleport takes you to the same map the event was run on, it will trigger two times instead of just once as it does otherwise. I fixed this by ensuring the event doesn't execute again if a transition is still registered as running.

I don't know if this is considered the cleanest solution. I tried everything I could think of from transition_teleport.py but nothing else worked. This solves the issue without breaking anything else.

### DIFF
--- a/tuxemon/core/event/actions/transition_teleport.py
+++ b/tuxemon/core/event/actions/transition_teleport.py
@@ -57,15 +57,19 @@ class TransitionTeleportAction(EventAction):
     ]
 
     def start(self):
-        # Start the screen transition
-        params = [self.parameters.transition_time]
-        self.transition = self.game.event_engine.get_action("screen_transition", params)
-        self.transition.start()
+        # Start the screen transition if one isn't already underway
+        world = self.game.get_state_name("WorldState")
+        if not world.in_transition:
+            params = [self.parameters.transition_time]
+            self.transition = self.game.event_engine.get_action("screen_transition", params)
+            self.transition.start()
 
     def update(self):
-        if not self.transition.done:
+        if not hasattr(self, 'transition'):
+            self.stop()
+        elif not self.transition.done:
             self.transition.update()
-        if self.transition.done:
+        else:
             self.transition.cleanup()
             # set the delayed teleport
             self.game.event_engine.execute_action("delayed_teleport", self.raw_parameters[:-1])


bitcraft:
I'd like of like to not commit any new code that checks for and interacts directly with the WorldState.  Is it possible to refactor using actions only?  My concern is that I am actively working to enable some multiple features, and interacting with the WorldState directly is problematic because it assumes a single player game.

MirceaKitsune:
I looked for several solutions when I gave this a try, but nothing else worked for me. I didn't feel this is the ideal solution either. It isn't an urgent fix, so maybe we can think of better ideas to solve it.

One would be: Why is player_at not preventing this double execution? Perhaps it's also at fault for seeing the player there for too long, and this could be fixed on its end.

tuxemon/core/event/actions/transition_teleport.py
```diff
@@ -57,15 +57,19 @@ class TransitionTeleportAction(EventAction):
...
+        world = self.game.get_state_name("WorldState")
+        if not world.in_transition:
+            params = [self.parameters.transition_time]
+            self.transition = self.game.event_engine.get_action("screen_transition", params)
+            self.transition.start()
 
     def update(self):
-        if not self.transition.done:
+        if not hasattr(self, 'transition'):
+            self.stop()
```
bitcraft: if we never get a transition, what will cause this action to stop?

bitcraft:
I don't think "player_at" condtions are triggering this, i think it is because the specific situation here causes 2 actions to trigger the screen transition.

tuxemon/core/event/actions/transition_teleport.py
MirceaKitsune: I had to the hasattr check after noticing that not setting a transition in `def start(self)` would cause this update statement to crash. I could just use a pass statement for it... `self.stop()` was just me wanting to be extra cautious and ensure everything was stopped accordingly.

MirceaKitsune: I just tested by replacing `self.stop()` with `pass`. Apparently that stop is needed: Without it, teleport events stop working after two teleports. No errors or warnings, but after being teleported twice on the same map touching the event stops doing anything.

MirceaKitsune:
One thing I totally should mention: It is not just the transition or even the teleportation that repeats, but the whole event is being triggered two times.

I added a second event to set the player's facing direction. Before the second occurrence finishes fading out, I have time to run a few tiles away. After the second teleport, the player doesn't just snap to the teleported position again, but the facing direction also changes a second time now.

bitcraft:
obsolete

PR #627, opened 2020-01-26
Prevent crashing when loading a map that contains no tilesets

MirceaKitsune:
I noticed that if you load a map which has no tiles painted on it, the engine will crash when attempting to get the tileset height / width (even when a tile layer does exist). While normally no one should ever be doing this, it does hinder a few testing cases. As a solution was simple I tested and submitted a fix.

### DIFF
--- a/tuxemon/core/map.py
+++ b/tuxemon/core/map.py
@@ -253,7 +253,7 @@ class Map(object):
         self.size = self.data.width, self.data.height
 
         # Get the tile size of the map
-        self.tile_size = self.size
+        self.tile_size = 16, 16
         if len(self.data.tilesets) > 0:
             self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
 
--- a/tuxemon/core/map.py
+++ b/tuxemon/core/map.py
@@ -253,7 +253,9 @@ class Map(object):
         self.size = self.data.width, self.data.height
 
         # Get the tile size of the map
-        self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
+        self.tile_size = self.size
+        if len(self.data.tilesets) > 0:
+            self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
 
         # Load all objects from the map file and sort them by their type.
         for obj in self.data.objects:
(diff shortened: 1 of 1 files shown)


tuxemon/core/map.py
```diff
@@ -253,7 +253,9 @@ def load(self, filename):
         self.size = self.data.width, self.data.height
 
         # Get the tile size of the map
-        self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
+        self.tile_size = 16, 16
```
kerizane:
```
        if self.data.tilesets:
            self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
        else:
            self.tile_size = 16, 16
```

MirceaKitsune: Fixed.

MirceaKitsune:
Hmmm. The comment says `# Get the tile size of the map`. I assumed this is for reading what the tilemap size is, for when we'll support +32px tiles later on.

Then again if there is no tilemap, the tile size might not matter? Except the engine still needs to know how far the player must walk. Would it introduce another crash if self.tile_size is never set?

tuxemon/core/map.py
```diff
@@ -253,7 +253,10 @@ def load(self, filename):
         self.size = self.data.width, self.data.height
 
         # Get the tile size of the map
-        self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
+        if len(self.data.tilesets) > 0:
```
bitcraft: ```if self.data.tilesets:```

tuxemon/core/map.py
```diff
@@ -253,7 +253,10 @@ def load(self, filename):
...
+            self.tile_size = self.data.tilesets[0].tilewidth, self.data.tilesets[0].tileheight
+        else:
+            self.tile_size = 16, 16
```
bitcraft: use the value from PREPARE.  i think it is called `TILE_SIZE`

tuxemon/core/map.py
MirceaKitsune: Fixed.

tuxemon/core/map.py
MirceaKitsune: Fixed.

PR #632, opened 2020-01-29
Feature/#200 combat calls

xirsoi:
This PR is pretty big, but I'll try to break it down.

The core thing that was done is adding a "sounds" dictionary to the monster model. You can now assign a "combat_call" and a "faint_call" to a monster. These will be played when a monster enters/leaves combat or faints, respectively.

If a monster does *not* have a call or faint assigned a generic one based on it's primary type will be used. If no file exists with the slug "sound_<type>_<faint/call>" (e.g. "sound_fire_faint" or "sound_fire_call") then an error will be logged and the mon will not have a call or faint play at the appropriate time.

I've created placeholder faint sounds by creating copies of the existing calls but played backwards.

The *major* change was making the sounds and music files be handled indirectly via the database. This means that from now on sound and music files need to have a corresponding entry in the new sounds and music db folders. You'll find examples in there already as I've converted all monster and interface sounds and all music (in use by maps) to using slugs.

The system will currently still work if you reference a filename directly, but this behavior is now officially deprecated.

A small change was made to the db code that allows sound and music records to be stored as a list, allowing multiple records in a single file. The correct format can be seen in the files I've provided.


tuxemon/core/states/combat/combat.py
```diff
@@ -643,7 +643,7 @@ def perform_action(self, user, technique, target=None):
             message = ''
 
         try:
-            tools.load_sound(technique.sfx).play()
+            tools.load_sound(db.lookup_file("sounds", technique.sfx)).play()
```
bitcraft: `tools.load_sound` is already supposed to take the pain out of path/file management.  consider refactoring this to accept a slug to make it more ergonomic.

xirsoi: 👍 i'll do some thinking and see what i can come up with

MirceaKitsune:
Initially I felt a bit thorn on creating a database for storing sounds and music. Upon better thinking, this is a good idea and cleanly implemented. I'll create them for my own game once this is in.

bitcraft:
looks good.  please stand by to fix any regressions.  overall im happy, thanks for making the change to load the sounds by the slug name.

PR #633, opened 2020-01-30
Fix avatar conflicting with value translation parameters in translated dialog events

MirceaKitsune:
translated_dialog and translated_dialog_chain used parameters following the text to replace certain variables. This conflicts with avatars, which added a second parameter oblivious to this. The issue is resolved by treating params containing var1=val1 as replacements and everything else as an avatar.


bitcraft:
thanks for the fix

PR #635, opened 2020-02-02
Numeric variable checks and operations

MirceaKitsune:
I was surprised to notice that the ability to check whether a numeric variable is grater or lesser than a given number was never implemented. This PR adds a new condition that will return true if a var is under or over a given number.

A companion event action is also added as a bonus, which allows basic mathematical operations to be preformed on numeric variables.

Simple usage example:

```
act1: variable_math myvar,+,1
cond1: variable_is myvar,<,10
```

### DIFF
--- a/tuxemon/core/event/actions/variable_math.py
+++ b/tuxemon/core/event/actions/variable_math.py
@@ -34,7 +34,7 @@ def number_or_variable(value):
     else:
         try:
             return float(player.game_variables[value])
-        except KeyError:
+        except (KeyError, ValueError, TypeError):
             logger.error("invalid number or game variable {}".format(value))
             raise ValueError
 
--- a/tuxemon/core/event/actions/variable_math.py
+++ b/tuxemon/core/event/actions/variable_math.py
@@ -25,18 +25,10 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 from tuxemon.core.event.eventaction import EventAction
+from tuxemon.core.tools import number_or_variable
 
-
-# Value fetching: If it's a string get a game variable, if it's a number use it as such
-def number_or_variable(value):
-    if value.isdigit():
-        return float(value)
-    else:
-        try:
-            return float(player.game_variables[value])
-        except (KeyError, ValueError, TypeError):
-            logger.error("invalid number or game variable {}".format(value))
-            raise ValueError
+import logging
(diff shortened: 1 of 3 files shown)


MirceaKitsune:
I'm thorn on whether to use 3 parameters, or one in which the items are separated by a space. It feels a little overhand to have to write `myvar,+,1` instead of `myvar + 1` even if that's implemented more cleanly code wise. @bitcraft Which way do you believe is best?

tuxemon/core/event/conditions/variable_is.py
```diff
@@ -0,0 +1,87 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from tuxemon.core.event.eventcondition import EventCondition
+
+
+# Value fetching: If it's a string get a game variable, if it's a number use it as such
+def number_or_variable(value):
```
bitcraft: code looks good.  please move to tools though.  docstrings :D

```
:param value: Union[str, float, int]
:rtype: float
:raises: ValueError
```

tuxemon/core/event/actions/variable_math.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Preform the operation on the variable
+        if operation == "+":
+            player.game_variables[var] = operand1 + operand2
+        elif operation == "-":
+            player.game_variables[var] = operand1 - operand2
+        elif operation == "*":
+            player.game_variables[var] = operand1 * operand2
+        elif operation == "/":
+            player.game_variables[var] = operand1 / operand2
```
bitcraft: Please add an `else` here and raise `ValueError` b/c the operation is not supported

tuxemon/core/event/conditions/variable_is.py
```diff
@@ -0,0 +1,87 @@
...
+        elif operation == "!=":
+            return operand1 != operand2
+        elif operation == ">":
+            return operand1 > operand2
+        elif operation == ">=":
+            return operand1 >= operand2
+        elif operation == "<":
+            return operand1 < operand2
+        elif operation == "<=":
+            return operand1 <= operand2
```
bitcraft: please add an `else` here and raise a `ValueError` b/c the operation is not supported.

bitcraft:
Great PR, thank you!

PR #637, opened 2020-02-03
#156 rename player and rename monster actions

xirsoi:
Two new scripts, rename_player and rename_monster.

The former simply brings up the input screen and prefills it with the player's current name. The later prompts the player to first select the Tuxemon from their party they want to rename.

Small issue; I'm having trouble popping the MonsterMenu from the stack at the end of the action, so the player has to press ESC to get out. Not ideal, but I'm not sure what I need to do to fix it, as the ItemMenu's method of just popping both the input and monster menu states when the work is done causes a crash.


MirceaKitsune:
Another great feature I'm very glad to see. Can't wait! Thanks again for the help with further improving the engine and finishing the remaining basic implementations.

bitcraft:
cool looks good to me.  i really like the shady guy renaming your monsters.

PR #640, opened 2020-02-03
Update windows installation instructions

yairm210:
Windows instructions were woefully out of date.
Newer versions of python come bundled with pip, so we can download everything from the regular requirements file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,16 @@ Installation
 
 **Windows**
 
-[Detailed here](https://www.tuxemon.org/windows-install.html)
+Install the latest version of python 2.7 from [here](https://www.python.org/downloads/)
+
+Pip is installed by default of newer versions of python, so now you just need to run:
+
+```cmd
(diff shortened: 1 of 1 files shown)


bitcraft:
We should be pointing people to download python 3.

README.md
```diff
@@ -26,7 +26,16 @@ Installation
...
 
-[Detailed here](https://www.tuxemon.org/windows-install.html)
+Install the latest version of python 3 from [here](https://www.python.org/downloads/)
+
+Run:
+
+```cmd
+git clone https://github.com/Tuxemon/Tuxemon.git
+cd Tuxemon
+python -m pip install -r requirements.txt
```
bitcraft: please change to `pip install -U -r requirements.txt`.  The `-U` flag is needed in the event that a dependency is installed but out of date.

bitcraft:
Thank you!

PR #641, opened 2020-02-03
Random variables and delays

MirceaKitsune:
This is kind of an amendment to 250062d566bdbd51e76ce681909c82bd6b9300ca as I forgot to add a few extra things that might come in handy. The PR makes two changes, centered around allowing random variables and delays to be set by the mapper:

- Allow the new number_or_variable function to return a random number in a range. This is done by specifying two numbers separated by a line, such as `5-10`. Each parameter is checked to verify that it's a valid number: If either is not the code continues to check for a single number then a game variable... there is no conflict if you're referencing a var named `foo-bar` or `foo-10` or `5-foo`.
- Make the wait and wait_for_secs event action use number_or_variable. This allows delays to be set to either a variable or a random interval.


tuxemon/core/event/actions/variable_math.py
```diff
@@ -49,12 +49,14 @@ def start(self):
...
         var = self.parameters.var
-        operand1 = number_or_variable(game, var)
+        operand1 = number_or_variable(self.game, var)
         operation = self.parameters.operation
-        operand2 = number_or_variable(game, self.parameters.value)
+        operand2 = number_or_variable(self.game, self.parameters.value)
 
         # Preform the operation on the variable
-        if operation == "+":
+        if operation == "=":
```
bitcraft: please use `==` so that the operations are similar to python and other langauges

bitcraft:
@ShadowApex @kerizane @Qiangong2 @xirsoi @yairm210
I'd like your input on this proposal to change the standard value parsing for numbers/variables to include this syntax change for random values.

tuxemon/core/event/actions/variable_math.py
MirceaKitsune: This is the event action though, not the conditional: It's setting the variable rather than checking it. This should mirror the common standard, as = is used to set and == is used to check.

MirceaKitsune:
Code wise this was simple to implement with minimal extra code and by using common functions. My only concern would be that there are a lot of overall syntaxes for mappers to remember... on the other hand this is necessary for certain features to be possible. I'm mainly thorn on whether using the line as a separator is the best way to describe a random range, as `3 15` or `10:40` might be cleaner than `8-16`... let me know if anyone prefers another separator.

I hope this can be accepted in some form however: Randoms open the way to many new mapping features. Currently you can only write perfectly deterministic stories. With this you can add in certain surprises or paths that will branch whenever you start a new game, which makes things a lot more intriguing and increases replayability as well.

tuxemon/core/tools.py
```diff
@@ -520,6 +521,7 @@ def load_image(rect=None, flags=None):
 def number_or_variable(game, value):
```
xirsoi: I know this was added by the previous commit, but I just wanted to point out that the Event system has a Common module to hold things like this function. Of course, if something outside the Event system needs to use this then tools would be the better location.

xirsoi:
Personally, I think using a colon as the delimiter would make more sense, but I think we should let the mappers weigh in first, as they'll be writing way more content.

tuxemon/core/tools.py
MirceaKitsune: Good to know. For now I think it makes sense to have this in tools.py, as it's a generic function that could have multiple uses throughout the code. At the same time I don't see those syntaxes being used outside of event actions, so I'll wait for more feedback on this.

MirceaKitsune:
@xirsoi Thanks for the feedback. So that's one vote for using a colon... two actually since I also feel it feels cleaner, plus the : symbol is already used in set_variable / variable_set so it's following an existing standard. I'll wait a bit more in case anyone else wishes to weigh in on this, then I'll change it.

bitcraft:
imo because the colon is already used, the meaning of it is more ambiguous using it two unrelated ways.

MirceaKitsune:
Makes sense. At the same time "-" could be interpreted as minus even if there's no space in between, so it might not be the best option either. Space is looking like the solution.

bitcraft:
no i don't think a space is the solution either.  space is a common delimiter between distinct things.

tuxemon/core/event/actions/variable_math.py
bitcraft: ah good point.  i had though this was comparison.

tuxemon/core/tools.py
bitcraft: @xirsoi i do agree.  tools.py is a great place for random bits of shared code, but since this particular function deals with parsing strings from events, i do agree that it could be better placed.  for now, i don't think its worth refactoring, but could be done later.

xirsoi:
Perhaps a ~ then?

bitcraft:
Yeah that could work.  I’m still not super excited about complicating the text parser.   I’d honestly prefer another action which generates a random number and stores that in a variable, then reference that variable in other actions or conditions.  I’d like to keep this open and wait for others to leave their opinions.

MirceaKitsune:
Could work but it's so much easier to use the existing math action: Too many event actions also add complexity, especially when it's for something so simple.

~ sounds very good. Also waiting to hear what others think before committing the change.

tuxemon/core/tools.py
```diff
@@ -529,7 +531,10 @@ def number_or_variable(game, value):
     :raises: ValueError
     """
     player = game.player1
-    if value.isdigit():
+    values = value.split("-")
```
xirsoi: I believe we decided to change this to a tilde, yes?

MirceaKitsune: It's fine either way by me. But yes, I'm also waiting for confirmation to know if I should leave it as is or change that to a tide.

bitcraft:
My vote is for the tilde ~

MirceaKitsune:
Changed in the latest commit. Anything else left to do for this one?

tuxemon/core/tools.py
MirceaKitsune: As this was the most voted choice, latest commit uses ~ to divide the random range.

Qiangong2:
I guess I'm confused, what does this take away that makes syntaxes harder?

bitcraft:
@Qiangong2 if you are referring to my comments, my concern is that its adds to the laready cramped syntax of events.  i don't want it to end up like perl.  i beleive that i am the minority here though, so ill test this and merge if it works as specified

MirceaKitsune:
@bitcraft As a general rule I believe this is inevitable as more features are added. Tuxemon uses a powerful and efficient system to define actions directly on the map, but I think that requires a smart and compact way to define how various things work. Only thing I believe is missing is a page on the wiki to easily describe how every action and syntax works, so a new map developer doesn't have to ask around to find them but can look in one place to know everything relevant.

bitcraft:
I don't agree that the final form of a syntax is perl as more features are added.  complex != complicated.  i could list a few different ways to do the same thing without resorting to syntax changes.  its not really worth a debate over, im just sharing my opinion.

the wiki is out of date, right.  we need to need it up to date with this grammar change.

MirceaKitsune:
I think it depends on each case and if weird grammar is unavoidable. Like in this case: We could just add a new action called `set_random_variable 0 1`, but then you have to create a temporary game variable if you want to pipe that into something else like a wait event. The separator technique allows doing it more directly, the only price being that you must use a `0<some_separator>1` syntax.

bitcraft:
sorry, im not really interested in a debate.  ive already said i would merge it, please just let the discussion end.

bitcraft:
```Traceback (most recent call last):
  File "tuxemon.py", line 62, in <module>
    main.main(load_slot=args.slot)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/main.py", line 117, in main
    control.main()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/control.py", line 257, in main
    update(clock_tick)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/control.py", line 305, in update
    self.event_engine.update(time_delta)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/event/eventengine.py", line 295, in update
    self.update_running_events(dt)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/event/eventengine.py", line 367, in update_running_events
    action.start()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/core/event/actions/wait.py", line 48, in start
    secs = number_or_variable(self.parameters.seconds)
TypeError: number_or_variable() missing 1 required positional argument: 'value'
```

bitcraft:
@MirceaKitsune Ive tested this and it crashes with the traceback above.  please go through and make sure that all existing calls to `number_or_variable` include the required positional arguments (was missing game here)

MirceaKitsune:
Sorry for taking so long to respond to this. Had a lot going on at once, tried dedicating today to getting back to unfinished code issues too.

@bitcraft I'm seeing an automated message above regarding conflicts that need to be solved. Any chance you could take care of those? I'm not sure if you can push to my branch though... maybe in this situation you can make the final change and push the whole thing? That crash might have been a single case of me forgetting add the game prefix, but it should of course be all tested again.

kerizane:
Fixed the merge conflicts for you. Please test it again.

MirceaKitsune:
I found and fixed some remaining issues. A quick test shows the random numbers are being generated properly with the latest notation, other functions in the default game work too (didn't test too far to not spoil myself on recent story additions). Example of what I used to set it:

`<property name="act10" value="variable_math myrand,=,1~10"/>`

kerizane:
What happened with this? @MirceaKitsune what were the issues you found?

MirceaKitsune:
I haven't been following the project closely for a while unfortunately. Wasn't this eventually implemented in some form? I'm okay with any syntax, just one of the basic features I felt had been missing altogether.

kerizane:
Yes, there is now a specific action for setting a random variable.

Ok, I think there is still some good stuff in this. If you remove the syntax for random variables that should be fine.

Do you want to do this, or should I?

MirceaKitsune:
I haven't reconfigured the project lately so I'd appreciate if you or others could take a look at it, optimize this as you feel is best.

PR #642, opened 2020-02-03
Choice menus have an optional escape_key_exists parameter

yairm210:
This is both a 'learning assignment' for me and a player UX thing.
I personally found it very annoying that at the shop, you need to either pick 'buy' or 'sell' and you can't just exit without picking either one
I think that there will probably be may choices like this where you will just want to exit the conversation, and so I added an optional parameter to enable that
All other existing usages of ChoiceState are unchanged

### DIFF
--- a/tuxemon/core/states/choice/__init__.py
+++ b/tuxemon/core/states/choice/__init__.py
@@ -16,11 +16,12 @@ class ChoiceState(PopUpMenu):
     * if there are no more messages, then the dialog will close
     """
     shrink_to_items = True
+    escape_key_exits = None
 
     def startup(self, **kwargs):
         super(ChoiceState, self).startup(**kwargs)
         self.menu = kwargs.get("menu", list())
-        self.escape_key_exits = kwargs.get("escape_key_exits",False)
+        self.escape_key_exits = kwargs.get("escape_key_exits", False)
 
     def initialize_items(self):
         for key, label, callback in self.menu:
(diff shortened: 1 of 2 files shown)


tuxemon/core/states/choice/__init__.py
```diff
@@ -16,11 +16,11 @@ class ChoiceState(PopUpMenu):
     * if there are no more messages, then the dialog will close
     """
     shrink_to_items = True
-    escape_key_exits = False
```
bitcraft: minor.  please leave this because it is used by linters to find out what attributes could be used by a class/instance.  if removed and added in any other method besides `__init__` linters will complain that it is not a value defined in the class.  You could set it to None here so that the default is not specified in two places.

bitcraft:
cool. i like the idea.  one minor comment and it is good to go.

tuxemon/core/states/choice/__init__.py
```diff
@@ -16,11 +16,11 @@ class ChoiceState(PopUpMenu):
...
 
     def startup(self, **kwargs):
         super(ChoiceState, self).startup(**kwargs)
         self.menu = kwargs.get("menu", list())
+        self.escape_key_exits = kwargs.get("escape_key_exits",False)
```
kerizane: Add a single space after the comma, please.

bitcraft:
Looks great!  Thanks for working with us on these small changes.

PR #643, opened 2020-02-03
Flavor text dialogs for the tuxemart

yairm210:
Another 'figuring this out' PR =)

### DIFF
--- a/mods/tuxemon/maps/tuxe_mart.tmx
+++ b/mods/tuxemon/maps/tuxe_mart.tmx
@@ -164,5 +164,45 @@
     <property name="cond2" value="not variable_set tuxchoice:end"/>
    </properties>
   </object>
+  <object id="62" name="cash_register_flavortext" type="event" x="64" y="128" width="16" height="16">
+   <properties>
+    <property name="act1" value="translated_dialog cash_register"/>
+    <property name="cond1" value="is player_at"/>
+    <property name="cond2" value="is player_facing left"/>
+    <property name="cond3" value="is button_pressed K_RETURN"/>
+   </properties>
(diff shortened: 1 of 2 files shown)


mods/tuxemon/maps/tuxe_mart.tmx
```diff
@@ -164,5 +164,45 @@
     <property name="cond2" value="not variable_set tuxchoice:end"/>
    </properties>
   </object>
+  <object id="62" name="cash_register_flavortext" type="event" x="64" y="128" width="16" height="16">
+   <properties>
+    <property name="act1" value="translated_dialog cash_register"/>
+    <property name="cond1" value="is player_at"/>
+    <property name="cond2" value="is player_facing left"/>
+    <property name="cond3" value="is button_pressed K_RETURN"/>
```
bitcraft: this pattern of "player at, player facing, button pressed" is out of date.  we have a new one that works much better called "player_facing_tile" that basically does all these checks for you.  since this works fine as-is, you don't need to change it.  just an FYI

also a feature we are experimenting is another feature that adds the correct conditions for interacting with objects.  change the object type to "interact", and only include the actions.  The correct conditions will be added for you and you can interact with the tile.

bitcraft:
Looks good to me.  I added a comment that might help you save some time.  Not needed to merge, just a FYI.  Let me know if you want to make changes, or to merge as-is.

yairm210:
I'll make the changes, no point in merging something that will be out-of-date immediately =)

bitcraft:
Cool, looks good.  Looking forward to more flavor text.  :smile:

PR #645, opened 2020-02-04
Taba town map cleanup

yairm210:


yairm210:
Hm
I'm not sure why the first 5 commits are even considered here =/
I'll try to 'reset' my develop branch to Tuxmon's, I'm probably working wrong

bitcraft:
@yairm210 don't worry about the commits.  i squash before merging so only a single commit is added with all the changes.

mods/tuxemon/maps/map1.tmx
```diff
@@ -277,20 +277,18 @@
     <property name="cond2" value="is player_at"/>
    </properties>
   </object>
-  <object id="78" name="Player's House Sign" type="event" x="624" y="736" width="16" height="16">
+  <object id="78" name="Player's House Sign" type="event" x="624" y="720" width="16" height="16">
```
bitcraft: Why is the Y value changing here (and in other events)?

yairm210: Relevant to the next time - it's because we're changing from 'player is here' position to 'object is here' position, which is 1 up

PR #646, opened 2020-02-04
Taba town map cleanup

yairm210:
I have learnt never to touch the develop branch on my own and to always create individual branches for PRs 🤕


bitcraft:
Cool thanks for updating the map.

PR #647, opened 2020-02-04
Renamed downstairs bedroom

yairm210:
This is also the place for discussion on how renaming maps affects currently saved games
Once we manage to 'successfully' migrate one map, the other renames will probably be en-masse (depends on y'all though)


bitcraft:
Looks good.  I want to make sure the other map editors do not have unpushed changes before merging this.

kerizane:
Do we want the game to be able to handle these changes? 
Bad user experience if it crashes. Annoying for us to explain to people how to fix.

How about this? 
https://github.com/Tuxemon/Tuxemon/pull/648

If people want to rename a map they need to add it to MAP_RENAMES.

yairm210:
I'm in favor of this!

Qiangong2:
the thing about map1 is that it includes both taba and paper town in one. As well as route 20

yairm210:
Wait, the second town is called paper town?
There's no sign around, and the nurse is called taba town nurse, so... =/

Qiangong2:
Yeah, it's called paper town. It's taken from the wiki

bitcraft:
Are these areas part of a region?  Can we call this "region01.tmx?"

Qiangong2:
> Are these areas part of a region? Can we call this "region01.tmx?"

Well, as with all pokemon-esque games, the entire game is called a "region". The region we are building is called Fondent region.

I guess taba_town.tmx would work fine

PR #648, opened 2020-02-04
Add map renaming to save upgrade logic

kerizane:


kerizane:
If people want to rename a map they need to add it to MAP_RENAMES.

kerizane:
Or we could just default to the starting location

tuxemon/core/save_upgrader.py
```diff
@@ -0,0 +1,52 @@
+"""
+This module is for handling breaking changes to the save file.
+
+Renaming maps:
+  - Increment the value of SAVE_VERSION (e.g. from 1 to 2)
+  - Add an entry to MAP_RENAMES consisting of:
```
kerizane: Went into lots of detail here as map editors aren't expected to know python.

bitcraft: good idea

tuxemon/core/save_upgrader.py
```diff
@@ -0,0 +1,83 @@
...
+        if new_name:
+            save_data['current_map'] = new_name
+
+
+def upgrade_save(save_data):
+    version = save_data.get("version", 0)
+    for i in range(version, SAVE_VERSION):
+        update_current_map(i, save_data)
+        if i == 0:
+            def fix_items(data):
```
bitcraft: Is it possible to move this to module scope?

kerizane: fix_items?

I don't think it's that reusable, but sure.

bitcraft: its more about readability.  i do like nested functions but it was buried inside a conditional and when i wanted to see where `fix_items` was defined, it wasn't immediately obvious where it was.  putting nested funcs directly after the docstrings is fine, imo.

bitcraft:
This looks like a good start, thanks!  I like the change to its own module.

PR #649, opened 2020-02-05
Bug/#638 battle music

xirsoi:
fixes the battle music not being referenced by slug and adds a debug log when slugs don't have a file record in the db.


bitcraft:
thanks for the quick fix

PR #650, opened 2020-02-05
Feature/#313 capture device breaks

xirsoi:
Adds a config that determines if items are consumed when use fails.
Also adds a config for a global modifier to the encounter rate.


tuxemon/core/config.py
```diff
@@ -160,6 +164,10 @@ def get_defaults():
             ("locale", "en_US"),
             ("dev_tools", False),
         ))),
+        ("gameplay", OrderedDict((
```
bitcraft: awesome, thanks for doing this!

bitcraft:
looks great, thank you

PR #654, opened 2020-02-06
Collisions NPC filter

MirceaKitsune:
Now that NPC wandering has been implemented, it's important to have the ability to customize collisions as to only filter certain types of entities. Mappers may want to block just NPC's from wandering in a given area, or just the player from doing so. This PR implements that ability.

Collisions may contain an optional "blocks" parameter. The collision will of course block everyone if the parameter isn't set. It can take a list of values which may contain the following entries:

- The slug of an NPC. This is checked first and applies to both players and other characters. It's used to create a collision that interacts with only specific NPC definitions.
- The keyword "players". If present, the collision will block all players but let NPC's pass.
- The keyword "npcs". If present, the collision will block all NPC's but let the player through.

I tested this on my maps, in combination with other parameters such as `enter_from` or `exit_to`: Results indicate that everything is functioning as intended and there are no crashes either. Further testing is recommended for certainty.

### DIFF
--- a/tuxemon/core/map.py
+++ b/tuxemon/core/map.py
@@ -451,6 +451,8 @@ class Map(object):
                                 tile_conditions['exit'] = exits
                             if "continue" in key:
                                 tile_conditions['continue'] = collision_region.properties[key]
+                            if "blocks" in key:
+                                tile_conditions['blocks'] = collision_region.properties[key]
                         collision_map[collision_tile] = tile_conditions
 
         # Similar to collisions, except we need to identify the tiles
--- a/tuxemon/core/map.py
+++ b/tuxemon/core/map.py
@@ -452,7 +452,7 @@ class Map(object):
                             if "continue" in key:
                                 tile_conditions['continue'] = collision_region.properties[key]
                             if "blocks" in key:
-                                tile_conditions['blocks'] = collision_region.properties[key]
+                                tile_conditions['blocks'] = collision_region.properties[key].split()
                         collision_map[collision_tile] = tile_conditions
 
         # Similar to collisions, except we need to identify the tiles
--- a/tuxemon/core/map.py
+++ b/tuxemon/core/map.py
@@ -41,6 +41,7 @@ import pytmx
 from pytmx.util_pygame import load_pygame
 
 from tuxemon.core import prepare
+from tuxemon.core.db import db
 from tuxemon.core.euclid import Vector2, Vector3, Point2
 from tuxemon.core.event import EventObject
 from tuxemon.core.event import MapAction
@@ -452,7 +453,14 @@ class Map(object):
                             if "continue" in key:
                                 tile_conditions['continue'] = collision_region.properties[key]
                             if "blocks" in key:
-                                tile_conditions['blocks'] = collision_region.properties[key].split()
+                                # make sure each block parameter is valid before adding it
+                                tile_conditions['blocks'] = []
+                                for block in collision_region.properties[key].split():
+                                    if block == "players" or block == "npcs" or block in db.database['npc']:
+                                        tile_conditions['blocks'].append(block)
+                                    else:
+                                        logger.error("invalid collision block {}".format(block))
+                                        raise ValueError
                         collision_map[collision_tile] = tile_conditions
 
         # Similar to collisions, except we need to identify the tiles
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -619,18 +619,12 @@ class WorldState(state.State):
         """
         if entity and tile and 'blocks' in tile:
             for block in tile['blocks']:
-                if block == 'players':
-                    if entity.isplayer:
-                        return True
-                elif block == 'npcs':
-                    if not entity.isplayer:
-                        return True
-                elif block in self.npcs:
-                    if block == entity.slug:
-                        return True
-                else:
-                    logger.error("invalid collision block {}".format(block))
-                    raise ValueError
+                if block == 'players' and entity.isplayer:
+                    return True
+                if block == 'npcs' and not entity.isplayer:
+                    return True
+                if block == entity.slug:
+                    return True
             return False
         return True
 
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -618,13 +618,19 @@ class WorldState(state.State):
         :rtype: bool
         """
         if entity and tile and 'blocks' in tile:
-            for block in tile['blocks'].split():
-                if block == 'players' and entity.isplayer:
-                    return True
-                if block == 'npcs' and not entity.isplayer:
-                    return True
-                if block == entity.slug:
-                    return True
+            for block in tile['blocks']:
+                if block == 'players':
+                    if entity.isplayer:
+                        return True
+                elif block == 'npcs':
+                    if not entity.isplayer:
+                        return True
+                elif block in self.npcs:
+                    if block == entity.slug:
+                        return True
+                else:
+                    logger.error("invalid collision block {}".format(block))
+                    raise ValueError
             return False
         return True
 
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -607,10 +607,10 @@ class WorldState(state.State):
     def tile_blocks(self, entity, tile):
         """ Verifies if a collision tile is supposed to block a given type of NPC
 
-        Three checks are preformed in the following order:
-        If the slug of the NPC is present in the blocks string, return True
-        If the NPC is a player, return True if "players" exists in the blocks string
-        If the NPC isn't a player, return True if "npcs" exists in the blocks string
+        Three checks are preformed:
+        If the NPC is a player, return True if "players" exists in the blocks
+        If the NPC isn't a player, return True if "npcs" exists in the blocks
+        If the slug of the NPC is present in the blocks, return True
 
         :param entity: the player or NPC being checked
         :param tile: the data of the tile being checked
@@ -618,12 +618,14 @@ class WorldState(state.State):
         :rtype: bool
         """
         if entity and tile and 'blocks' in tile:
-            if entity.slug in tile['blocks']:
-                return True
-            elif entity.isplayer:
-                return 'players' in tile['blocks']
-            else:
-                return 'npcs' in tile['blocks']
+            for block in tile['blocks'].split():
+                if block == 'players' and entity.isplayer:
+                    return True
+                if block == 'npcs' and not entity.isplayer:
+                    return True
+                if block == entity.slug:
+                    return True
+            return False
         return True
 
     def get_exits(self, position, collision_map=None, skip_nodes=None):
--- a/tuxemon/core/states/world/worldstate.py
+++ b/tuxemon/core/states/world/worldstate.py
@@ -605,11 +605,25 @@ class WorldState(state.State):
             pass
 
     def tile_blocks(self, entity, tile):
+        """ Verifies if a collision tile is supposed to block a given type of NPC
+
+        Three checks are preformed in the following order:
+        If the slug of the NPC is present in the blocks string, return True
+        If the NPC is a player, return True if "players" exists in the blocks string
+        If the NPC isn't a player, return True if "npcs" exists in the blocks string
+
+        :param entity: the player or NPC being checked
+        :param tile: the data of the tile being checked
+
+        :rtype: bool
+        """
         if entity and tile and 'blocks' in tile:
-            if entity.isplayer and not 'player' in tile['blocks']:
-                return False
-            if not entity.isplayer and not 'npc' in tile['blocks']:
-                return False
+            if entity.slug in tile['blocks']:
+                return True
+            elif entity.isplayer:
+                return 'players' in tile['blocks']
+            else:
+                return 'npcs' in tile['blocks']
         return True
 
     def get_exits(self, position, collision_map=None, skip_nodes=None):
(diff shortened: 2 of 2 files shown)


Qiangong2:
Can you give an action example?

MirceaKitsune:
@Qiangong2

```
blocked: players // Only NPC's can pass through this collision
blocked: npcs // Only players can pass through this collision
blocked: players npcs // Neither can pass (default, same as not setting the property)
blocked: npc_wife // Only the professor's wife is blocked by this collision
blocked player npc_wife // Neither the player or professor's wife can walk through
```

tuxemon/core/states/world/worldstate.py
```diff
@@ -502,13 +502,19 @@ def get_collision_map(self):
         # TODO: overlapping tiles/objects by returning a list
         collision_dict = dict()
 
+        # tile layout must be set first
```
bitcraft: why is this changed?

tuxemon/core/states/world/worldstate.py
```diff
@@ -634,7 +666,8 @@ def get_exits(self, position, collision_map=None, skip_nodes=None):
                 ("left", (position[0] - 1, position[1])),
         ):
             # if exits are defined make sure the neighbor is present there
-            if exits and not neighbor in exits:
+            self_blocks = self.tile_blocks(entity, collision_map.get(position))
```
bitcraft: The collision checking code must be very fast because it is used for pathfinding and player movement.  The reason why the collision map exists is so that complex collision logic is handled only once when the map is loaded.  As I understand, you are creating tiles which conditionally block depending on the string value assigned to them.  Since that is the case, you should move the tile text parsing to the code which builds the collision map.  Its not a good practice to use things like `text.split` in a tight loop.  Instead, you should break down the values and store in a list, which can be more quickly compared against.  I understand that this is a small change overall, but i want to make sure you understand that we cannot keep adding inefficient code into the collision detection code.  Its already a bottleneck for pathfinding, and I don't want to watch it get worse.

tuxemon/core/states/world/worldstate.py
```diff
@@ -598,6 +604,30 @@ def get_explicit_tile_exits(self, position, tile, skip_nodes):
...
+        :param entity: the player or NPC being checked
+        :param tile: the data of the tile being checked
+
+        :rtype: bool
+        """
+        if entity and tile and 'blocks' in tile:
+            for block in tile['blocks'].split():
+                if block == 'players' and entity.isplayer:
+                    return True
+                if block == 'npcs' and not entity.isplayer:
```
bitcraft: we should not allow unexpected values in the map code.  if the only expected values are "players", "npcs" & a slug, then we should be raising an error here if the value is not valid.

tuxemon/core/states/world/worldstate.py
```diff
@@ -598,6 +604,30 @@ def get_explicit_tile_exits(self, position, tile, skip_nodes):
...
+                    return True
+                if block == entity.slug:
+                    return True
+            return False
+        return True
+
     def get_exits(self, position, collision_map=None, skip_nodes=None):
```
bitcraft: This is going to break pathfinding.  Consider the situation where the player (or different NPC) is standing in a position which conditionally blocks the NPC in a different way than the one pathfinding (or vice-versa).  When the pathfinding code runs, the path will be different depending on what npc is in that spot with the blocking rules.  So a path that might be perfectly valid for one npc will not be found because while looking for paths, the code uses blocking rules that apply for a different NPC.

bitcraft:
Im not sure if this logic works well.  IMO, id rather see this implemented as "fences" using Tiled object such as Rectangles, Circles, Polygon, etc.

Qiangong2:
This still has an issue with leaving the professor's house, right?

bitcraft:
Does the NPC leave the house?  Like, just off the map, or does it teleport?

Qiangong2:
It teleports, but then immediately crashes with an error here:
`if exits and not neighbor in exits`

Qiangong2:
I mentioned it in the discord the other day

bitcraft:
wow, im actually impressed that it tries to teleport.  that would be super cool if it "just worked" honestly.  i remember you mentioning the error, but i didn't connect it to the wife teleporting.  imo NPCs should be able to automatically teleport the way the player does.  ill open a ticket #655

Qiangong2:
Actually, the error is with the player teleporting from inside maple_house.tmx to map1.tmx and then the game immediately crashes. However, it also crashes when the npc tries to teleport at all.

MirceaKitsune:
@Qiangong2 The exits crash should have been fixed yesterday. Please make sure you're running the latest development version. If it still occurs it's another issue.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: I discovered a silent bug in this part of the code as it was interfering with this implementation, was going to comment on it. The tile data contains an NPC stored here as `entity`, in the same data field with parameters such as `enter_from` `exit_to` `continue` as well as the new `blocks` in this PR. The issue was that when the NPC entity is set here, it gets erased by the next call. So if the old `collision_dict[pos]` contained:

`{entity: <npc_object>}`

It was immediately overridden to something like:

`{exit_to: up, enter_from: down}`

My change was to ensure they are mixed instead so no properties are lost:

`{entity: <npc_object>, exit_to: up, enter_from: down}`

From what I gathered this is why it was happening: The `collision_dict.update` call is the one refreshing this field which overrides the old data. It only seems to do that in some circumstances though. But when it does, the `collision_dict[pos] = {"entity": npc}` set just a few lines above is completely replaced, unless that's done afterward and mixed with the old dict.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: That's very good to know, didn't think about it. I'll see how I can do that so the splitting is done only once. Will need to store the result in the collision system but I'm sure I'll find a way.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: Correct, will fix in the next commit.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: Doesn't the pathfinding code use the `get_exits` function where the check is implemented to get its rules? So if a path is invalid for a specific NPC, the pathfinding code would know and go for a different path instead, accounting only the collisions that affect that NPC?

Need to wrap my mind around around exactly how this is the case first. If it's so I'll think of what can be done about it, this would indeed be a problem.

tuxemon/core/states/world/worldstate.py
bitcraft: ok fair enough.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: The solution was very easy, will be fixed in the next commit.

tuxemon/core/states/world/worldstate.py
bitcraft: no.  get_exists only knows about the current state of the map.  with these changes get_exits changes depending on what NPC is in a tile position.  since all NPCs will pathfind using this method, they will now be using the rules of the tiles where some NPC is sitting.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: There's actually a problem with this: I can't find a way to determine if a value is invalid here. The parameter may contain any number of NPC slugs, but the scan only occurs for the NPC trying to move onto that tile: It's impossible to check if the value might refer to the slug of another NPC and is thus value, as we're only working with the NPC moving in this call.

I could of course add a loop to scan all NPC's on the map and make sure the parameter matches at least one. But would the potential performance impact be worth it only to raise an exception for an incorrect value? If I can get a list of all map NPC's maybe I can use `if x in y:` which I understand is a lot cheaper so this might do.

bitcraft: ok.  if i were doing this, i would make sure this block data is parsed when the map is loaded.  i would use the npc database and check if the slug is there, then check against the list ["players", "npcs"] and raise a ValueError if its not there.  the list would be checked again at runtime, but the database check is not needed at that time.

MirceaKitsune: Issue is that both the object I have access to (`self.npcs`) and the database would require a loop to extract the slugs. I can't directly do `if slug in npcs` as "npcs" will always contain an object and not a list of slugs. Only solution seems to be with this addition:

```
# create a list of npc slugs, used for checking parameter validity
npcs = []
for npc in self.get_all_entities():
    npcs.append(npc.slug)
```

tuxemon/core/states/world/worldstate.py
MirceaKitsune: I see. Is there an easy way to fix this? If not I could go with a simple solution for now: Adding a TODO note and in the meantime, disable this filter entirely for pathfinding. IIRC there is a `npc.pathfinding` boolean which is True when the NPC is navigating that way. This might however be a bad idea, since if the NPC stops pathfinding inside a collision it might get stuck there forever.

tuxemon/core/states/world/worldstate.py
```diff
@@ -618,13 +618,19 @@ def tile_blocks(self, entity, tile):
...
-                if block == entity.slug:
-                    return True
+            for block in tile['blocks']:
+                if block == 'players':
+                    if entity.isplayer:
+                        return True
+                elif block == 'npcs':
+                    if not entity.isplayer:
+                        return True
+                elif block in self.npcs:
```
bitcraft: ok, this will check for slugs, which is great, but its still in "the hot path" for collision detection

tuxemon/core/states/world/worldstate.py
bitcraft: no hacks, please.  i don't have an easy solution right now.  essentially, all npc movement and pathfinding need to carry "context" with them because the exits change depending on who is looking that them.

bitcraft: i would accept disabling teleports for npcs and an alternative

tuxemon/core/states/world/worldstate.py
```diff
@@ -502,13 +502,19 @@ def get_collision_map(self):
...
         # Get all the NPCs' tile positions
         for npc in self.get_all_entities():
             pos = nearest(npc.tile_pos)
-            collision_dict[pos] = {"entity": npc}
 
-        # tile layout takes precedence
-        collision_dict.update(self.collision_map)
+            # make sure the data for this position exists
+            if not pos in collision_dict or not collision_dict[pos]:
+                collision_dict[pos] = {}
```
MirceaKitsune: Just in case the necessity of this check is questioned: Sometimes collision_dict doesn't contain the pos parameter. At other times it does but its value is None. In both cases it must be initialized as a dict, so that below we either add the entity to the existing parameters or this empty dict instead if none were found here.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: IIRC 'if x in y' is a cheap check in Python, so hopefully this doesn't add any extra load. If I checked the NPC database here, it would include all NPC's defined in the game not just those present on the map... the check would thus be more expensive as there are many extra keys to needlessly scan through.

It should be kept in mind that those checks are only preformed if a player or NPC is walking near a collision object that has the `block` parameter set: None of this is executed if the right conditions aren't met. So if the mapper doesn't add both tons of wandering NPC's plus lots of filtered collision objects that they'll practically run into, this should be reached very rarely.

tuxemon/core/states/world/worldstate.py
MirceaKitsune: Ok. I'll need to wait for you or another dev to provide a solution to this one: It delves into more difficult aspects that I'm not sure how to approach.

One clue, even if this is probably obvious: The new `def tile_blocks(self, entity, tile)` function can be reused anywhere, given we have access to the NPC entity we're checking and the tile data of the spot we're moving to. If there's a specific point in the pathfinding code where the same checks would be needed, it can probably be called there as well.

tuxemon/core/states/world/worldstate.py
bitcraft: im really not comfortable with incremental performance losses in pathfinding.  we already have tickets out because it is slow enough.  if you are not going to implement the load-time checks for slugs, then drop this change and add a TODO

MirceaKitsune: I'd like to do this in map.py load if I could. The problem is that's called pretty early, there doesn't seem to be a way to fetch all NPC's present on the map at that time. Even if there was, what about NPC's that aren't added instantly but by an event after the map loaded?

I'll look a bit more if there's another way, but with the current implementation I just don't see one. I'll probably add that TODO instead.

bitcraft: thats a good point, but i think that the databases are loaded before maps, in which case all the slugs should be available.  im trying to deprecate the ability to create custom npcs in the maps because we have some issues with it in the past, so we should rely on the slugs being available.

MirceaKitsune: Managed to fix this the right way! The validity of block parameters is now checked just once, each parameter being added to the list only if valid while an exception is raised otherwise. Let me know if this looks good in the latest commit.

Funny fact: My idea to check only the NPC's available on the map wouldn't have worked anyway. That's because an NPC might be added after the collision object is loaded. This would result in a crash as it won't find the NPC and there's no way of knowing what future events will do. So just making sure the NPC exists in the database was the right way to go all along.

tuxemon/core/map.py
```diff
@@ -451,6 +452,15 @@ def loadfile(self):
                                 tile_conditions['exit'] = exits
                             if "continue" in key:
                                 tile_conditions['continue'] = collision_region.properties[key]
+                            if "blocks" in key:
+                                # make sure each block parameter is valid before adding it
+                                tile_conditions['blocks'] = []
+                                for block in collision_region.properties[key].split():
+                                    if block == "players" or block == "npcs" or block in db.database['npc']:
+                                        tile_conditions['blocks'].append(block)
```
bitcraft: perfect :smile:

MirceaKitsune:
Is there currently a consensus on how to work around the pathfinding issue, and any advice on how to solve that part? I believe that's the last blocker for this feature.

kerizane:
I don't understand how this would break pathfinding. Can somebody explain the issue?

Issue #636, opened 2020-02-03
Include context around map event crashes


bitcraft:
When debugging map event crashes, its difficult to know what part of the map files caused the crash.  Tracebacks are very helpful to find the cuase, but its difficult to find the line in the script which caused the error.  Ideally, the following information would be available when a crash happens:
* map name
* script/event name
* line number in the script
* the actual line text

Feel free to add info.  In order for this to work better, we will need more context added to actions/events internally that include the information above.

kerizane:
Working on this

PR #656, opened 2020-02-09
Show line that caused the crash from tmx script

kerizane:
#636

### DIFF
--- a/tuxemon/core/event/eventengine.py
+++ b/tuxemon/core/event/eventengine.py
@@ -31,6 +31,9 @@ from __future__ import unicode_literals
 
 import logging
 import os.path
+from contextlib import contextmanager
+from lxml import etree
+from textwrap import dedent
 
 from tuxemon.constants import paths
 from tuxemon.core import prepare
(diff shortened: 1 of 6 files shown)


tuxemon/core/event/eventengine.py
```diff
@@ -449,3 +457,47 @@ def process_event(self, event):
...
+    :type item: core.event.MapCondition or core.event.MapAction
+    :type game: core.control.Control
+    :rtype None
+    """
+    try:
+        yield
+    except Exception:
+        file_name = game.get_map_filepath()
+        tree = etree.parse(file_name)
+        event_node = tree.find("//object[@id='%s']" % event.id)
```
bitcraft: When i opened this ticket i was thinking that we would store the text of each map event in the map event object, along with the map name and line number (if possible).  I think this approach of reading the source tmx file after the exception is raised is good enough for now, but in the future, i would also like this info on each event.  there is already a debug mode which lists events, and it might be useful to have the event info before crashing.

PR #658, opened 2020-02-09
Show line that caused the crash from tmx script

kerizane:
#636


kerizane:
Tested with python2 and python3

Quick testing run

bitcraft:
Looks good, thanks for doing the basic checks.

PR #659, opened 2020-02-09
Feature/#541 learn techniques

xirsoi:
First pass at items that teach techniques, also creates support for items having conditions for use. 

The "learn" effect accepts a single technique name.

The conditions implementation is dead simple atm; a dictionary with the key being a property of the monster and the value being acceptable values. You can treat it as an OR by separating acceptable values with commas.

See the tm_avalanche and tm_blossom items for examples of how this can work.


tuxemon/core/item.py
```diff
@@ -187,6 +222,35 @@ class with the name of the effect you define in item.db.
...
+        **Examples:**
+        >>> tm = Item("tm_blossom")
+        >>> tm.learn(player1, bulbatux, "blossom")
+        """
+        
+        try:
+            tech = Technique()
+            tech.load(slug)
+            target.learn(tech)            
+        except RuntimeError:
```
bitcraft: when is a RuntimeError raised?  What circumstances does it happen? I would rather allow that exception to bubble up than to cover it here.

tuxemon/core/item.py
```diff
@@ -145,6 +147,33 @@ def advance_round(self):
...
         return
 
+    def validate(self, target):
+        """Ensures that the target meets all conditions that the
+        item has on it's use.
+
+        :param user: The monster or object that is using this item.
+        :param target: The monster or object that we are using this item on.
+
+        :type user: Varies
```
bitcraft: s/Varies/Any to conform with the `typing` module

tuxemon/core/item.py
```diff
@@ -145,6 +147,33 @@ def advance_round(self):
         """
         return
 
+    def validate(self, target):
+        """Ensures that the target meets all conditions that the
+        item has on it's use.
+
+        :param user: The monster or object that is using this item.
```
bitcraft: I don't see the 'user' parameter

tuxemon/core/states/monster/__init__.py
```diff
@@ -118,12 +121,12 @@ def refresh_menu_items(self):
                 monster = self.game.player1.monsters[index]
             except IndexError:
                 monster = None
+                continue
             item.game_object = monster
-            item.enabled = monster is not None
+            item.enabled = monster is not None and self.is_valid_entry(monster)
```
bitcraft: Should we be checking the user?  Does the user have any condition that would disable the item?

tuxemon/core/item.py
```diff
@@ -172,8 +201,14 @@ class with the name of the effect you define in item.db.
 
         # Loop through all the effects of this technique and execute the effect's function.
         for effect in self.effect:
-            last_effect_name = str(effect)
-            result = getattr(self, last_effect_name)(user, target)
+            effect_lines = str(effect).split()
```
bitcraft: I don't like string operations done at runtime that should already be loaded from disk.  I'm not going to block the PR because of this though, since its not in a hot loop, and I don't think this code will be long lived.  Eventually, the effects/actions will be refactored into something better, so its fine here.  An alternative to this would be to create a lightweight "dataclass" style object.  Again, probably not worth it here, but just something to think about.

tuxemon/core/item.py
```diff
@@ -172,8 +201,14 @@ class with the name of the effect you define in item.db.
...
+            effect_name = effect_lines[0]
+            effect_param = effect_lines[1] if len(effect_lines) > 1 else None
+
+            if effect_param:
```
bitcraft: Please add a `TODO`; something like `TODO: refactor effects to remove the need of getattr`

tuxemon/core/states/monster/__init__.py
xirsoi: I had initially intended to support that, but there are currently no conditions to check.
It would also be bad UX to have that check happen in the monster menu; it should be handled in the Item menu.

tuxemon/core/item.py
bitcraft: Please update the docstring by removing unused parameters.

tuxemon/core/item.py
bitcraft: please change Varies to Any to conform with the typing module.  https://docs.python.org/3/library/typing.html#typing.Any

tuxemon/core/item.py
bitcraft: please respond before resolving discussions

tuxemon/core/states/monster/__init__.py
bitcraft: fair enough

tuxemon/core/item.py
xirsoi: Just going to actually do the thing, rather than leave a crappy implementation.

tuxemon/core/item.py
xirsoi: I resolved it because I had fixed it locally, I just hadn't pushed it up yet.
There's no need to re-review a PR if no changes have been pushed up.

tuxemon/core/item.py
xirsoi: Saw no need to respond, I just made the change you requested.

mods/tuxemon/db/item/apple.json
```diff
@@ -4,7 +4,7 @@
   "use_success": "generic_success",
   "use_failure": "generic_failure",
   "effects": [
-    "heal"
+    "heal 20"
   ],
   "power": 20,
```
bitcraft: what use now is the `power` attribute?  if its part of the effects, i think we should remove it to prevent headaches down the road because it looks like it does something.

xirsoi: I have no use for it, but I didn't want to just remove it without discussion. I'll get rid of it now.

bitcraft:
moved: https://github.com/Tuxemon/Tuxemon/pull/684

PR #664, opened 2020-02-15
Development rubberduck

Sanglorian:
This adds three maps prepared by Rubberduck.


bitcraft:
Thanks for packaging this into a PR.

PR #665, opened 2020-02-15
WIP Client View

bitcraft:
Refactor the world state in such a way that the game doesn't require a display.  It also means that the world view is independent of the game, allowing for a clear path to multiplayer and special actions, like changing the controlled npc at runtime.

### DIFF
--- a/tuxemon/compat/__init__.py
+++ b/tuxemon/compat/__init__.py
@@ -1,0 +1,4 @@
+try:
+    from pygame.sprite import Rect
+except ImportError:
+    from tuxemon.compat.rect import Rect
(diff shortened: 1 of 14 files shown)


kerizane:
  File "tuxemon.py", line 46, in <module>
    from tuxemon.core import prepare, main
  File "/home/k/Tuxemon/tuxemon/core/main.py", line 38, in <module>
    from tuxemon.core.player import Player
ModuleNotFoundError: No module named 'tuxemon.core.player'

bitcraft:
thanks for the report.  i still have changes locally that i need to push (including some files, too)

bitcraft:
closing because github is having a hard time with the diffs

PR #673, opened 2020-02-29
Healing items blocked if your Tuxemon is fainted

### LINKED ISSUE #672: Potions are usable on fainted Tuxemon but don't revive them
If you use a potion on a Tuxemon after it faints in battle, the "It worked!" message appears and heals it up but you still are unable to start battle or encounter wild mons.

pinsonc:
Closes #672 

Healing items are no longer usable on fainted Tuxemons. Previously, using healing items would increase their health points but not remove the "fainted" status, meaning they would still be unusable in battles.

### DIFF
--- a/tuxemon/core/item.py
+++ b/tuxemon/core/item.py
@@ -213,7 +213,6 @@ class Item(object):
             return {"success": False}
 
         # don't heal if fainted
-        print(check_status(target, 'status_faint'))
         if check_status(target, 'status_faint'):
             return {"success": False}
 
--- a/tuxemon/core/item.py
+++ b/tuxemon/core/item.py
@@ -211,6 +211,10 @@ class Item(object):
         if target.current_hp == target.hp:
             return {"success": False}
 
+        # don't heal if fainted
+        if target.current_hp <= 0:
+            return {"success": False}
+
         # Heal the target monster by "self.power" number of hitpoints.
         target.current_hp += self.power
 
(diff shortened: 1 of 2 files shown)


Qiangong2:
Is this just a blanket ban on using items on a Tuxemon if they are fainted? Or does it look for a specific attribute?

bitcraft:
Is the healing item lost if you use it on a fainted monster?

pinsonc:
It only bans items on Tuxemon whose HP is 0 or less. Right now whether the item is lost is based on the `items_consumed_on_failure` setting in the config imported in `prepare.py`. So if that is `true`, the item will be consumed.

I'll look into checking if the Tuxemon is actually `FAINTED`. Thanks!

pinsonc:
New commit now looks for `status_faint` rather than `HP <=0`

tuxemon/core/item.py
```diff
@@ -211,6 +212,11 @@ def heal(self, user, target):
         if target.current_hp == target.hp:
             return {"success": False}
 
+        # don't heal if fainted
+        print(check_status(target, 'status_faint'))
```
bitcraft: please remove this print statement.  you can change it to a logger if you would like to keep it.

bitcraft: ill remove it

bitcraft:
Merged.  Thank you!

PR #675, opened 2020-03-01
Made it impossible to use capture devices in trainer battles

### LINKED ISSUE #674: Able to catch other trainer's Tuxemon
When battling with other trainers, you can still use capture devices and successfully capture their Tuxemon. Current behaviors:
1. The battle is over but not considered a win after they losing their last Tuxemons
2. They still have it when you try battling again.
3. The caught pokemon keep the same amount of hp as when you caught it.

pinsonc:
Closes #674 

Found some interesting stuff in this one. Taking a look at `MainCombatMenuState`,
```python
# master branch code
def choose_item():
            # open menu to choose item
            menu = self.game.push_state("ItemMenuState")

            # set next menu after after selection is made
            menu.on_menu_selection = choose_target # overloads on_menu_selection in items/__init__.py

def choose_target(menu_item):
            # open menu to choose target of item
            item = menu_item.game_object
            self.game.pop_state()   # close the item menu
            # TODO: don't hardcode to player0
            combat_state = self.game.get_state_name("CombatState")
            state = self.game.push_state("CombatTargetMenuState", player=combat_state.players[0],
                                         user=combat_state.players[0], action=item)
            state.on_menu_selection = partial(enqueue_item, item)
```
By redefining the action `on_menu_selection`, the default `on_menu_selection` in the `ItemMenuState` is not executed at all. This means that using an item from combat currently bypasses all the conditions set in the `is_usable` field in the item json files.

I initially tried fixing it within the `Item` class because I thought that would be cleanest but I couldn't think of a succinct way to get whether or not the current battle was a trainer battle within that class. So the quickest fix I could think of was changing the choose target function to both ensure the item could be used in battle and that the item was not a capture device if it was a trainer battle.

Right now the only issue with this fix is that you don't lose your turn if you try to use an item that cannot be used in battle. Nor do you lose the item. I tried forfeiting the turn but it seems the battle phase system is based on how many actions are enqueued and I couldn't make a "dummy" action to meet the threshold since I don't know the code base well enough yet.

### DIFF
--- a/tuxemon/core/item.py
+++ b/tuxemon/core/item.py
@@ -211,6 +211,10 @@ class Item(object):
         if target.current_hp == target.hp:
             return {"success": False}
 
+        # don't heal if fainted
+        if target.current_hp <= 0:
+            return {"success": False}
+
         # Heal the target monster by "self.power" number of hitpoints.
         target.current_hp += self.power
 
--- a/mods/tuxemon/db/inventory/basic_store.json
+++ b/mods/tuxemon/db/inventory/basic_store.json
@@ -2,8 +2,7 @@
   "slug": "inv_basic_store",
   "inventory": {
     "capture_device": null,
-    "potion": 20,
-    "apple": null
+    "potion": 20
   }
 }
 
--- a/mods/tuxemon/db/inventory/basic_store.json
+++ b/mods/tuxemon/db/inventory/basic_store.json
@@ -2,7 +2,8 @@
   "slug": "inv_basic_store",
   "inventory": {
     "capture_device": null,
-    "potion": 20
+    "potion": 20,
+    "apple": null
   }
 }
 
--- a/tuxemon/core/states/combat/combat_menus.py
+++ b/tuxemon/core/states/combat/combat_menus.py
@@ -113,12 +113,11 @@ class MainCombatMenuState(PopUpMenu):
             combat_state = self.game.get_state_name("CombatState")
             self.game.pop_state()   # close the item menu
             # check to see if item can be used inside a trainer battle
-            if (combat_state.is_trainer_battle and item.slug == 'capture_device'):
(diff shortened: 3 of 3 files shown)


tuxemon/core/item.py
```diff
@@ -211,6 +211,10 @@ def heal(self, user, target):
         if target.current_hp == target.hp:
             return {"success": False}
 
+        # don't heal if fainted
```
bitcraft: This is a commit that belongs to another PR.

bitcraft:
I'm hesitant to merge this because there is currently a large refactor in progress on how items work right now.  I think that this approach is probably fine in the short term, but in general, I don't want item-specific behaviors coded into game components.  You'll notice, for example, that the combat state `states/combat/py` doesn't have code handling items or capture devices (except for animations).  This is by design, to keep related behaviors abstract and separated.  In general, I'd like to keep that trend, and placing this restriction in a menu goes against that.

If i were to implement this, then I would have the capdev check if the monster is owned by a trainer, then fail to operate, or tell the player it cannot be used on trainer-owned monsters.  This would mean there is no behavior logic related only to the capture device leaking outside of its class.

@xirsoi is currently working on the item changes.

tuxemon/core/item.py
pinsonc: 🤦‍♂ I branched off of my other branch lol. Thanks for catching this.

pinsonc:
Yeah I agree this fix is hack-y and I felt gross mixing the behaviors. I didn't realize there was an item refactor in the works. I wouldn't merge this either lol

bitcraft:
hahaha cool.  thanks for understanding!  i hate turning down PRs, especially when its from a new contributor.  thank you though!

kerizane:
What's going on in the item refactor?

xirsoi:
Item refactor is complete. An ItemCondition needs to be written to check that the target isn't an owned monster to satisfy this particular issue.

Issue #678, opened 2020-03-01
Make rumble support optional


jonadem:
I haven't installed libShake as it was referred as optional in the README.md (on Debian 10).
It makes me unable to run Tuxemon, I got : 
```
$ python tuxemon.py 
pygame 1.9.6
Hello from the pygame community. https://www.pygame.org/contribute.html
Enabling logging of all modules.
Traceback (most recent call last):
  File "tuxemon.py", line 62, in <module>
    main.main(load_slot=args.slot)
  File "/mnt/data/Documents/Softwares/Tuxemon/tuxemon/core/main.py", line 57, in main
    control = Control(prepare.ORIGINAL_CAPTION)
  File "/mnt/data/Documents/Softwares/Tuxemon/tuxemon/core/control.py", line 121, in __init__
    self.rumble_manager = rumble.RumbleManager()
  File "/mnt/data/Documents/Softwares/Tuxemon/tuxemon/core/rumble/__init__.py", line 63, in __init__
    from .libshake import LibShakeRumble
  File "/mnt/data/Documents/Softwares/Tuxemon/tuxemon/core/rumble/libshake.py", line 10, in <module>
    from tuxemon.core.tools import Rumble
ImportError: cannot import name 'Rumble' from 'tuxemon.core.tools' (/mnt/data/Documents/Softwares/Tuxemon/tuxemon/core/tools.py)
```

You should either document that libShake is requested or make the code able to run without the support for Rumble.

Qiangong2:
I don't have libshake installed and it works fine for me. Which python version are you using (and which version of pip)?

bitcraft:
Rumble support is optional, this is a different error.  I've been moving the codebase to total py3 compliance and futurize made this mistake, which for whatever reason was not a problem for my system after verification.

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/rumble/libshake.py#L10

This `Rumble` object is no longer in `tools.py`.

jonadem:
So, it can be fixed with : 
```
$ git diff
diff --git a/tuxemon/core/rumble/libshake.py b/tuxemon/core/rumble/libshake.py
index 38a32832..dd64c4e0 100644
--- a/tuxemon/core/rumble/libshake.py
+++ b/tuxemon/core/rumble/libshake.py
@@ -7,7 +7,7 @@ from ctypes import *
 from threading import Thread
 from time import sleep
 
-from tuxemon.core.tools import Rumble
[...]

PR #681, opened 2020-03-04
Fix wrong path for module import #678

jonadem:


bitcraft:
Looks good, thank you!

PR #683, opened 2020-03-06
New clock and rect module

bitcraft:
Adding a couple files to break up the changes with the client view.

Rect: used when pygame is not available
Clock: Experimental scheduler which is very accurate and easy to use

### DIFF
--- a/tuxemon/compat/__init__.py
+++ b/tuxemon/compat/__init__.py
@@ -1,0 +1,4 @@
+try:
+    from pygame.sprite import Rect
+except ImportError:
+    from tuxemon.compat.rect import Rect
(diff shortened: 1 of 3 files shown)


xirsoi:
What benefits does the pure Python Rect class confer?

bitcraft:
So that pygame is not needed as a dependency.  I’d like to be able to run the server with zero pygame imports. This is a step towards that.

xirsoi:
This looks good to me then.

Qiangong2:
So would rect give Tuxemon the ability to run on platforms without pygame?

bitcraft:
The immediate goal of things like the Rect and Clock module are to allow the server portion to run without pygame.  A long term, unrealistic goal would be to allow running the graphical parts on using another graphics library.  Im not committed to it or endorsing it though.  Pygame is is already well supported on a lot of systems and will be more-so when the 2.0 version is released.

Qiangong2:
Okay, thanks for clarifying :)

PR #690, opened 2020-03-11
Split audio/graphics functions from tools.py into own files

bitcraft:
* graphics.py contains pygame related graphics things
* audio.py contains pygame related audio things

Part of a series of changes to support a headless server that doesn't need pygame to run.  Tools.py is still required for many operations, but eventually graphics and sound can be optional.


xirsoi:
This looks good to me.

PR #695, opened 2020-03-13
Sessions

bitcraft:
Introduce "Session" object

A Session contains references to:
* Player
* World that the player is part of
* Control

Far reaching changes unfortunately, so a large PR:
* Remove the `player1` reference from the `Control` object
* Rename the `Control` object to `Client`, since it is the local game client
* Remove methods that manage players from `Control`
* Where possible, replace references to the player or game client with the `local_session` object

This is a simple object now and really just supports local single player.  The api will change in the future, but this is the foundation and part of a series of changes to decouple the game engine from the single player client.

The Player object is the NPC controlled by the client.  The world object is not implemented yet, and will be a refactor of the world state, which contains map, events, and npcs.  It will be the WorldState, but not as a state.

The Control object will eventually be called the "client".  It manages inputs and game states.  It will also provide a time source for the world.

Eventually, a Session object will support multiplayer and split screen.  One human player will have a single session, but the world and control objects are shared.

This is a part of a series of changes to enable a headless server and a stand alone game client.


bitcraft:
Tested for regressions and so far seems good.  I will take another close look at the changes and make sure there are no obvious errors.  The next steps will be to merge in the map loader changes (another pr)...then client view can be merged in.

bitcraft:
Regression testing has come up clear.  Please take a look if you have a moment

PR #696, opened 2020-03-16
Issue #573 - Grey font color on moves

### LINKED ISSUE #573: -1 appears next move if it was used previously
image

We should implement a form of move uses in the future

vch9:
Fixes #573 

Since shadow_text uses self.font_color to draw text,  
I put the font_color is a temporary variable, change to grey,
then put back the font_color.

I would personally made font_color as an optional parameter,
with self.font_color as default.

### DIFF
--- a/tuxemon/core/states/combat/combat_menus.py
+++ b/tuxemon/core/states/combat/combat_menus.py
@@ -147,7 +147,10 @@ class MainCombatMenuState(PopUpMenu):
                 if tech.next_use <= 0:
                     image = self.shadow_text(tech.name)
                 else:
+                    tmp = self.font_color
+                    self.font_color = (220, 220, 220)
                     image = self.shadow_text("%s -%s" % (tech.name, tech.next_use))
+                    self.font_color = tmp
                 item = MenuItem(image, None, None, tech)
                 menu.add(item)
 
--- a/tuxemon/core/states/combat/combat_menus.py
+++ b/tuxemon/core/states/combat/combat_menus.py
@@ -147,10 +147,7 @@ class MainCombatMenuState(PopUpMenu):
                 if tech.next_use <= 0:
                     image = self.shadow_text(tech.name)
                 else:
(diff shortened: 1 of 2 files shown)


vch9:
Here's how it looks like:
tuxemon

xirsoi:
Nice!

xirsoi:
I think the negative sign is confusing, could you maybe change it so the number is in parens?

vch9:
Would look like this:
image

xirsoi:
but multiply the number by -1 for display purposes, so it's positive and we don't see the negative sign.

tuxemon/core/states/combat/combat_menus.py
```diff
@@ -147,7 +147,10 @@ def choose_technique():
                 if tech.next_use <= 0:
                     image = self.shadow_text(tech.name)
                 else:
+                    tmp = self.font_color
+                    self.font_color = (220, 220, 220)
                     image = self.shadow_text("%s -%s" % (tech.name, tech.next_use))
```
bitcraft: I'm not a huge fan of this. `shadow_text` already has an optional kwarg for the background color, lets add one for the foreground color.  If the foreground color is passed, use that, if not, use `self.color`.  We should move this gray color with the rest so that colors are consistent.  https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/menu/menu.py#L51

vch9: I agree with that

xirsoi:
Looks good to me 👍

tuxemon/core/states/combat/combat_menus.py
vch9: @bitcraft you should check:
5d5824e

bitcraft: looks perfect

bitcraft:
looks great!  thank you for following the changes and your contribution!

PR #697, opened 2020-03-18
adjustment for early encounters

xirsoi:
I replaced the default encounter list with a more selective one and altered the level range from 1-6 to 2-4. I only added base forms and tried to get a good spread of types for variety and testing.

I also modified another encounter list, but only for levels, as the monster selection seemed deliberate. All other encounter lists looked like they were more carefully curated than the default one, so I left them alone.

This should make testing smoother.


bitcraft:
@Qiangong2 please take a look at the changes

Qiangong2:
Seems fine to me. I guess I'm just confused why the default encounter needed to be changed?

xirsoi:
Because it's used in the early maps and the experience is terrible? If I've gotta spend 2 hours testing and re-testing by going through the early-game every time i want to submit a PR it's gonna have to stop sucking.

Qiangong2:
Is saving broken then? Because we should fix that immediately if it is

xirsoi:
Continuing from a save is how I missed multiple bugs in the Item refactor.

bitcraft:
looks good to me.  a bit more balanced.

PR #701, opened 2020-03-19
676: Stop softlock from player closing monster menu when tuxemon faints

### LINKED ISSUE #676: Softlock after pressing escape after Tuxemon fainted in combat
After a Tuxemon faints and you have more Tuxemon in your bag, you are prompted to choose another Tuxemon. However, if you press the escape button at that state, the game would go back to the battle screen saying "{Tuxemon Name} has fainted.." and doesn't allow any more input (including using escape button). Currently, the only way out is to close this game.

kerizane:
Fixes #676


bitcraft:
looks good, short and simple, ty

PR #702, opened 2020-03-19
Fix 690 regression

kerizane:
Game crashes when you try to select a fainted tuxemon after your tuxemon has just fainted.

Caused by #690


bitcraft:
thanks for cleaning up my mess~~

PR #703, opened 2020-03-19
Stop game crashing when you sell an item

kerizane:


bitcraft:
cool, fix looks good, ty

PR #706, opened 2020-03-21
Corrected few errors with print statement and changed execfile function to exec 

Styne001:


bitcraft:
Can you confirm this works on python 3 and 2.7?  AFAIK the android builds are only supported on on python 2.7.  Also, we are not deprecating python 2.7 yet.  Please add `from __future__ import print_statement` to the top.

also according to this page https://stackoverflow.com/questions/436198/what-is-an-alternative-to-execfile-in-python-3 there is special handling on exec/execfile.  please make sure your changes do not break python 2.7 use.

build_android.py
```diff
@@ -16,10 +16,11 @@
 SIX_ARCHIVE = "six-1.9.0.tar.gz"
 SIX_LINK = "https://pypi.python.org/packages/source/s/six/%s#md5=476881ef4012262dfc8adc645ee786c4" % SIX_ARCHIVE
 
+
 def download_rapt():
     # First, download rapt
-    print "  Downloading rapt..."
-    headers = { 'User-Agent' : 'Mozilla/5.0' }
+    print("  Downloading rapt...")
```
bitcraft: changing to print statements requires an import from future to work on py2.7

build_android.py
```diff
@@ -104,49 +105,48 @@ def set_default_config(file_name):
...
     for item in os.listdir('.'):
         if 'android-sdk-' in item:
             sdk = True
 
     if not sdk:
-        print "  SDK not found."
+        print("  SDK not found.")
         sys.argv.append('installsdk')
-        execfile('android.py')
+        exec('android.py')
```
bitcraft: I believe that python3 expects the parameter to for exec to be a string.  In this case, this will not work by passing a filename; the contents of the file must be read into memory first.  We still support python 2.7, so any changes must work with both versions

bitcraft:
obsolete

PR #707, opened 2020-03-23
Fix capture devices catching trainer monsters

bitcraft:
Fixes 674

Attempting to use the capture device on an owned (not wild) monster will result in a dialog saying "this item cannot be used on this monster".  It is generic now because we cannot set per-item error messages at the moment.

* "owner" attribute on monsters
* Owner is set when monster is added to npc/player, removed when they are released (possible?)
* Implement new condition to check for owner
* Allow combat menus to check for item conditions
* Allow item conditions to have parameter-less conditions

### DIFF
--- a/tuxemon/core/monster.py
+++ b/tuxemon/core/monster.py
@@ -50,7 +50,7 @@ SIMPLE_PERSISTANCE_ATTRIBUTES = (
     'slug',
     'status',
     'total_experience',
-    'flairs',
+    'flairs'
 )
 
 SHAPES = {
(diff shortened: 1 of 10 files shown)


tuxemon/core/monster.py
```diff
@@ -50,7 +50,7 @@
     'slug',
     'status',
     'total_experience',
-    'flairs'
+    'flairs',
```
xirsoi: why the trailing comma? is this not incorrect?

bitcraft: im going to remove this.  it was just a knee-jerk reaction that i made.  i prefer trailing commas on all items because it makes it really easy to add items or rearrange without maintaining the last comma.  eventually, i will run black on the entire code base, which will add trailing commas as part of its style guidelines.

bitcraft: to clarify, i will remove it because its not related to the change of the pr

xirsoi:
Looks good to me!

bitcraft:
Thanks for the review!

Issue #651, opened 2020-02-05
npc_freq request


Qiangong2:
As funny as it is to have an npc move at 0.2 speed, We should really have a movement frequency action to work with npc_speed and wander

It's nearly impossible to talk to the professor's wife

MirceaKitsune:
Have you tried setting npc_speed to a lower value, as well as the npc_wander frequency? I'd rather not change the core functionality personally, but see sense in the defaults being tweaked. If that would be better, the default wander speed could be set to 0.5 instead, by editing this line:

https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/event/actions/npc_wander.py#L74

Qiangong2:
Honestly, I was thinking it would be better as an optional argument for npc_speed

MirceaKitsune:
It could be done that way too? Not sure if a good idea still: NPC speed is different from telling the NPC to wander around.

Qiangong2:
One reason I'd want to set it separately is because at some point we may want a Tuxemon chasing its owner, or vice-versa

bitcraft:
ill weigh in on this one, i don't like overloading actions with lots of arguments, and in this case i prefer setting the wander frequency in an more related action  (`npc_wander`).

it is also a bit more logical to set with `npc_wander` since setting it with npc_speed does nothing until you actually start the wandering action.  Im all for more sensible defaults though, and also changing the speed of the wife so we don't have to "trap her in a corner and smash the enter key rapidly to open a dialog"  :smile:

MirceaKitsune:
I too noticed that catching her to trigger the dialogue is a bit difficult, though for some reason I thought it's just me having bad luck. I agree this can get a little bit annoying with NPC's that have dialogue.

I didn't really think about this when I did the wandering code: The limits and defaults were based on what made most sense to see. At 0.5 (minimum) the NPC walks all over the place as if in a panic, while at 5 (maximum) they only walk occasionally. I defaulted to 1 as it resulted in the most logical looking movement patterns. I'm wondering if it might solve this to use 2 or 3 as the default instead.

There could be another solution too: Allow talking to an NPC while it's moving between tiles, if you're looking and pressing ENTER toward both the old and new position. So if you catch an NPC in-between tiles, interacting with either of the two positions would allow the trigger to fire.

bitcraft:
[...]

PR #713, opened 2020-03-25
Adjust movement speed of mother NPC

Russell-Tran:
Hi! This pull request is rather trivial, so it's more of my way of saying hello to the community for the first time!

I have replaced the mother NPC's movement speed of 0.2 for wandering frequency of 3. I think the person who set the movement speed to less than 1 was intending to set the wandering frequency instead, as moving an NPC at a speed less than 1 makes the NPC look like it's glitchy / gliding. 

This is related in topic to to #651


bitcraft:
A wander freq. of 3 seems fine to me.  Looks good, ty and welcome to the community!

PR #715, opened 2020-03-25
714: Don't call arrange_menu_items if you don't implement it

### LINKED ISSUE #714: Bug: AttributeError: 'MenuSpriteGroup' object has no attribute 'arrange_menu_items'
Currently caused during Tuxemon battles, either through an attack action or throwing a capture device: The game crashed with the following error: `AttributeError: 'MenuSpriteGroup' object has no attribute 'arrange_menu_items'`

Will follow up shortly with more info

kerizane:
Fixes #714

### DIFF
--- a/tuxemon/core/menu/menu.py
+++ b/tuxemon/core/menu/menu.py
@@ -195,7 +195,8 @@ class Menu(state.State):
             for item in items:
                 self.add(item)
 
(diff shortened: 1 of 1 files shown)


bitcraft:
looks good thanks for the fix

Issue #574, opened 2020-01-03
Battle check needs to be added


Qiangong2:
Right now, you can battle Leila continuously in cotton town because it never checks if you battled or not

Russell-Tran:
Hi! I've been working on this, but have been struggling with the implementation a bit. My solution involves editing `npc.py` and `start_battle.py`, and I've gotten it to work within a given (session, map.tmx) of the game, but I can't get it to persist in the save files, nor persist if I leave the `mapA.tmx` for `mapB.tmx` and then come back to `mapA.tmx`.

I think this might be because the states of actually-non-playable NPCs are not actually being kept track of anywhere, but it could be because I'm not familiar with the codebase. It seems that saving and loading the game only triggers the functions `NPC.get_state()` and `NPC.set_state()` on behalf of the playable protagonist (where the player's protagonist is called an "npc" because it's an instance of the NPC class).  @bitcraft  thoughts?

bitcraft:
> perhaps it might also be a good idea to separate the protagonist from the NPC class.

I don't think its a good idea to make the player any different from other NPCs.  There is a huge benefit of treating players and other NPCs the same, since any game actions can apply to both without changing the code.  Could you elaborate?

> but I can't get it to persist in the save files

this is what "game variables" are for.  ill admit i don't work in the game script much, but i would think that what should happen is after the battle, check the game variable `battle_last_result` and set a game variable such as `hasBattledLeila=True`.  @Qiangong2 is there an issue with this approach?  Of course the scripts don't work exactly like that.  You would need a 2nd event to watch the `battle_last_result` value on that map and then set `hasBattledLeila=True` if the battle was won.

Qiangong2:
That's exactly what I was hoping it would be. There would also need to be a check if you lost that would send you directly to the last Tuxemon Center you visited. That way, your Tuxemon can heal

Russell-Tran:
> perhaps it might also be a good idea to separate the protagonist from the NPC class.
>> I don't think its a good idea to make the player any different from other NPCs. There is a huge benefit of treating players and other NPCs the same, since any game actions can apply to both without changing the code. Could you elaborate?
Definitely agree. I think I was more caught up on the naming convention because it seemed silly to have a "player" be a "nonplayable character", and that a player class might need to keep track of more information than an NPC class would. But I actually just discovered that there's a `Player` class which extends `NPC` (in `player.py`), so actually we're all good.

PR #716, opened 2020-03-29
WIP Battle check

Russell-Tran:
To resolve issue #574

This PR prevents a trainer battle from being started if the protagonist has already battled this trainer NPC. It persists across loading and saving, as well as when leaving and returning to a given `.tmx` map.

Please note it does not include a feature to "check if you lost that would send you directly to the last Tuxemon Center you visited. That way, your Tuxemon can heal" per @Qiangong2 from the discussion on issue #574. We'll have to open another issue to implement that feature. We should note that for that feature, the game variable `'battle_last_result'` actually already exists and would help with that issue.

This PR does establish a new game variable  `'battle_history'`, which is a dictionary that maps npc's names to the outcome of the battle with the player (no key implies a battle hasn't happened yet). Hopefully this variable will help for future improvements to the game where we want the npc to have different dialogues post-battle depending on whether the player won or lost.

Please squash commits when merging this PR. Thank you

### DIFF
--- a/tuxemon/core/event/actions/start_battle.py
+++ b/tuxemon/core/event/actions/start_battle.py
@@ -56,22 +56,31 @@ class StartBattleAction(EventAction):
     ]
 
     def start(self):
+        """
+        Assumes single player against single NPC
+        """
         player = self.game.player1
 
         # Don't start a battle if we don't even have monsters in our party yet.
         if not check_battle_legal(player):
             logger.debug("battle is not legal: won't start because we don't have enough monsters in our party")
-            return False # don't start battle
+            return False
 
-        # Fetch the antagonist NPC from the world state
+        # Fetch the world state
         world = self.game.get_state_name("WorldState")
         if not world:
             logger.debug("battle could not be started because WorldState was not obtained")
-            return False # don't start battle
+            return False 
+
+        # Fetch the antagonist NPC from the world state
         npc = world.get_entity(self.parameters.npc_slug)
-        if npc.battled_already:
+
+        # Don't start a battle if we've already battled this npc before
+        if npc.name in player.game_variables['battle_history']:
             logger.debug("battle did not start because this NPC battled the protagonist already") 
-            return False
+            return False 
+
+        # Load the party of this npc from their npc.json entry.
         npc.load_party()
 
         # Stop movement and keypress on the server.
@@ -84,11 +93,10 @@ class StartBattleAction(EventAction):
             env_slug = player.game_variables['environment']
         env = db.lookup(env_slug, table="environment")
 
-        # Add our players and setup combat
+        # Setup combat with our player against the npc
         logger.debug("Starting battle!")
         self.game.push_state("CombatState", players=(player, npc), combat_type="trainer",
                              graphics=env['battle_graphics'])
-        npc.battled_already = True
 
         # Start some music!
         filename = env['battle_music']
--- a/tuxemon/core/event/actions/start_battle.py
+++ b/tuxemon/core/event/actions/start_battle.py
@@ -60,19 +60,23 @@ class StartBattleAction(EventAction):
 
         # Don't start a battle if we don't even have monsters in our party yet.
         if not check_battle_legal(player):
-            logger.debug("battle is not legal, won't start")
-            return False
+            logger.debug("battle is not legal: won't start because we don't have enough monsters in our party")
+            return False # don't start battle
 
+        # Fetch the antagonist NPC from the world state
         world = self.game.get_state_name("WorldState")
         if not world:
+            logger.debug("battle could not be started because WorldState was not obtained")
+            return False # don't start battle
+        npc = world.get_entity(self.parameters.npc_slug)
+        if npc.battled_already:
+            logger.debug("battle did not start because this NPC battled the protagonist already") 
             return False
+        npc.load_party()
 
         # Stop movement and keypress on the server.
         if self.game.isclient or self.game.ishost:
             self.game.client.update_player(player.facing, event_type="CLIENT_START_BATTLE")
-
-        npc = world.get_entity(self.parameters.npc_slug)
-        npc.load_party()
 
         # Lookup the environment
         env_slug = "grass"
@@ -90,5 +94,7 @@ class StartBattleAction(EventAction):
         self.game.event_engine.execute_action("play_music", [filename])
 
     def update(self):
+        # Check if the CombatState no longer exists (hence combat is over)
         if self.game.get_state_name("CombatState") is None:
+            npc.battled_already = True 
             self.stop()
--- a/tuxemon/core/event/actions/start_battle.py
+++ b/tuxemon/core/event/actions/start_battle.py
@@ -88,13 +88,12 @@ class StartBattleAction(EventAction):
         logger.debug("Starting battle!")
         self.game.push_state("CombatState", players=(player, npc), combat_type="trainer",
                              graphics=env['battle_graphics'])
+        npc.battled_already = True
 
         # Start some music!
         filename = env['battle_music']
         self.game.event_engine.execute_action("play_music", [filename])
 
     def update(self):
-        # Check if the CombatState no longer exists (hence combat is over)
         if self.game.get_state_name("CombatState") is None:
-            npc.battled_already = True 
             self.stop()
--- a/tuxemon/core/player.py
+++ b/tuxemon/core/player.py
@@ -48,4 +48,17 @@ class Player(NPC):
         self.isplayer = True
 
         # Game variables for use with events
-        self.game_variables = {}
+        self.game_variables = {
+        	'battle_history' : {} # Maps the names of NPCs the Player has battled to the outcome of their battle
+        						  #    any NPC name not present in the dictionary implies the Player has never 
+        						  #    battled them before
+        }
+
+    def set_state(self, game, save_data):
+    	super(Player, self).set_state(game, save_data)
+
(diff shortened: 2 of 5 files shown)


bitcraft:
@Qiangong2 i need your input on this one
@Russell-Tran if i am reading the changes correctly, once a battle between the player and another npc has happened, another battle will never be possible.  im not a pokemon guy, i don't know how things are supposed to work, but just from a first glance, this seems limiting.  if this is indeed how it is supposed to work, then sure it looks good, but if not, we will need additional changes to allow rematches or battles later.

tuxemon/core/player.py
```diff
@@ -48,4 +48,17 @@ def __init__(self, npc_slug):
...
 
         # Game variables for use with events
-        self.game_variables = {}
+        self.game_variables = {
+        	'battle_history' : {} # Maps the names of NPCs the Player has battled to the outcome of their battle
+        						  #    any NPC name not present in the dictionary implies the Player has never 
+        						  #    battled them before
+        }
+
+    def set_state(self, game, save_data):
```
bitcraft: good idea adding this check.  we have some save game migration code in https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/save_upgrader.py i would prefer that it be moved here.  i won't reject this pr though, i feel this is reasonable.  if you choose not to move it, please add a comment:
TODO: remove in future or move to save game migrations

tuxemon/core/event/actions/start_battle.py
```diff
@@ -56,31 +56,44 @@ class StartBattleAction(EventAction):
...
         if not world:
-            return False
+            logger.debug("battle could not be started because WorldState was not obtained")
+            return False 
+
+        # Fetch the antagonist NPC from the world state
+        npc = world.get_entity(self.parameters.npc_slug)
+
+        # Don't start a battle if we've already battled this npc before
+        if npc.name in player.game_variables['battle_history']:
```
bitcraft: I'd prefer to use slugs here because they are less likely to change compared to the name.  If a map designer changes a NPC's name during development, this value will break.  Slugs are our internal IDs for objects and they don't change often.

Qiangong2:
@Russel-Tran I believe this shouldn't be merged in until there's a check to make sure that if you lose, you can fight them again. That's a critical part of the story going forward. Plus, it won't let you battle anyone or anything if your Tuxemon have 0 health, so you could skip over parts.

tuxemon/core/player.py
Russell-Tran: Thanks for finding this for me! I'll definitely move this check over so the code is cleaner.

tuxemon/core/event/actions/start_battle.py
Russell-Tran: Thanks. Let's go with slug

Russell-Tran:
> @Russell-Tran if i am reading the changes correctly, once a battle between the player and another npc has happened, another battle will never be possible. im not a pokemon guy, i don't know how things are supposed to work, but just from a first glance, this seems limiting. if this is indeed how it is supposed to work, then sure it looks good, but if not, we will need additional changes to allow rematches or battles later.

and

> @Russel-Tran I believe this shouldn't be merged in until there's a check to make sure that if you lose, you can fight them again. That's a critical part of the story going forward. Plus, it won't let you battle anyone or anything if your Tuxemon have 0 health, so you could skip over parts.

Good catch here. Made the silly assumption that you only battle the NPC once, regardless of whether you win or lose. Let me go ahead and make sure that if you lose, you can fight them again.



> we will need additional changes to allow rematches or battles later.

Usually I think it's if you win, then you can't battle them immediately again to encourage you to keep going forward in the game and to keep the storyline logical for beating a boss, etc. Rematches do exist and tend to be enabled after the equivalent of "some time passing", which is simulated by the player progressing through the story / hitting milestones through the story such as "completed half the game", etc. The rematch usually involves the NPC having new dialogue and a new (stronger) party of monsters to represent that time has passed, that the NPC has been preparing for a rematch, or that the Player's monsters have gotten stronger and so the original NPC's party needs to be upgraded to provided a sufficient challenge.

So, absolutely, additional changes will be needed to allow rematches. But as a feature it might be a bit more involved as it requires adding a "second party" attribute to the npc `.json` files as well as logic to switch the NPC between "original dialoge" and "rematch dialogue."

Russell-Tran:
> we will need additional changes to allow rematches or battles later.

Usually I think it's if you win, then you can't battle them immediately again to encourage you to keep going forward in the game and to keep the storyline logical for beating a boss, etc. Rematches do exist and tend to be enabled after the equivalent of "some time passing", which is simulated by the player progressing through the story / hitting milestones through the story such as "completed half the game", etc. The rematch usually involves the NPC having new dialogue and a new (stronger) party of monsters to represent that time has passed, that the NPC has been preparing for a rematch, or that the Player's monsters have gotten stronger and so the original NPC's party needs to be upgraded to provide a sufficient challenge.

So, absolutely, additional changes will be needed to allow rematches. But as a feature it might be a bit more involved as it requires adding a "second party" attribute to the npc `.json` files as well as logic to switch the NPC between "original dialoge" and "rematch dialogue."

Qiangong2:
The ability to have a rematch should not be based on time. It should be a condition

Russell-Tran:
> The ability to have a rematch should not be based on time. It should be a condition

Exactly, such as hitting a milestone.

bitcraft:
> do you know if we already have code for teleporting the player to

yes.  please check the events/actions folder.  you can get a reference to the game's event engine and call teleport to another map.  i think "transition_teleport" will be the best choice.

> well as logic to switch the NPC between "original dialoge" and "rematch dialogue."

I'd kinda like to draw a line here for now.  You may have noticed, or will eventually, that I/we are trying to implement most of the game itself in map events (actions & conditions).  Hardcoding rematch/original dialogs into json would be going a bit too far into "implementing game/story specific design into the game engine".  Game designers should not have to box their dialogs into a predetermined format.  What if we want an alternative "rematch" dialog?  Maybe they want to say something else if they lost with a specific monster, or whatever.  Maybe we don't want a rematch, but they will make a comment about losing or winning last time, etc.

I'm fine with "battle_history" because i think there is value to it, but I don't think we need to build constraints into the game ("rematch dialogs").  Tuxemon is built for pokemon-like games, but its not a clone and a goal is to allow developers to do interesting things without just writing everything in python in the game engine.

> So, absolutely, additional changes will be needed to allow rematches

If you go through our tickets, you'll see that we try to make plans and break issues up.  Some are pretty detailed.  I'd like to encourage you to follow that pattern.  In general, its best to make a plan in the issue discussion before coding and opening a pr....there is less friction in making the changes if everyone knows what to expect.

In this case, I am thinking of additional functionality that needs to be done for this to be complete:

- We need an event condition that checks if a battle has ever won, lost, draw, run away, etc
- Not starting a battle because a battle happened before is a design choice, not a technical one.  I'd like to see battle check implemented as a map condition, not hardcoded in the game.

If you are not going to add a condition check to this PR, I would like you to open a new issue describing an event condition to check the last battle against an npc.  We need a condition for map designers to make their own choices about starting battles or not, since the existing game variable checking code will not be able to read this nested dictionary structure.

tuxemon/core/player.py
```diff
@@ -47,5 +47,7 @@ def __init__(self, npc_slug):
         super(Player, self).__init__(npc_slug)
         self.isplayer = True
 
-        # Game variables for use with events
-        self.game_variables = {}
+        # Game variables for use with events:
+        self.game_variables['battle_history'] = {} # Maps the slugs of NPCs the Player has battled to the outcome of their battle
```
kerizane: The comments all the way on the right is a bit ugly,  I'd prefer this.
 \# Maps the slugs of NPCs the Player has battled to the outcome of their battle
 \# any NPC slug not present in the dictionary implies the Player has never 
 \# battled them before
 self.game_variables['battle_history'] = {}

tuxemon/core/player.py
```diff
@@ -47,5 +47,7 @@ def __init__(self, npc_slug):
         super(Player, self).__init__(npc_slug)
         self.isplayer = True
 
-        # Game variables for use with events
-        self.game_variables = {}
```
kerizane: Maybe you shouldn't have removed this line?
        self.game_variables = {}

kerizane:
game_variables are simple variables used by the maps.

e.g.
        "thewifedidspeak": "yes",
        "goodbyeoldknight": "finallydone",
        "foome": "ye",
        "talkpls": "gone",
        "tuxchoice": "end",

I think battle_history should be it's own attribute, not nested withing game_variables.

tuxemon/core/player.py
xirsoi: Definitely shouldn't have removed that line.

Russell-Tran: Thank you for pointing this out. Absent the other things I need to fix in this PR, can someone please clarify why this line shouldn't be removed? game_variables is already an attribute of tuxemon.core.npc.NPC, so if someone decided to modify the init value in there, wouldn't that get lost by overwriting in the child class `Player` here?

xirsoi: Hmm... you make a good point, perhaps @bitcraft can weigh in on this?

tuxemon/core/player.py
xirsoi: it's also technically a pep8 style violation, pushing the line length past 120 characters

tuxemon/core/player.py
bitcraft: `game_variables` is defined in the NPC class, so its fine to remove.

Issue #718, opened 2020-04-02
Quantity Menu expands with each keystroke when buying or selling


xirsoi:

kerizane:
Weird, doesn't happen for me.

xirsoi:
Do you have the latest from development?
When I pulled latest I got this behavior.

kerizane:
Thought I did, can see it now. 
Embarrassing.

kerizane:
Caused by latest commit:

4838529964407dc3478a78fddc9cb896b2368246

xirsoi:
Yeah, and it's not clear how. It's some strange interaction between different parts of the menu code.

kerizane:
Yeah, very weird interaction. Got a fix though.

PR #719, opened 2020-04-02
Fixes bug where quantity menu was expanding when you pressed buttons

kerizane:
#718


kerizane:
I've tested this. Item cursor still skips invalid targets (e.g. non-fainted tuxemons when using revive, tuxemons that can't learn a TM)

xirsoi:
This is a truly bizarre interaction.

bitcraft:
¯\_(ツ)_/¯

PR #720, opened 2020-04-02
updates the party strip to check for mons with status effects

xirsoi:
Also changes the code to display the capdevs in order. Mostly. The active tuxemon always comes first, not sure why. Don't particularly care.

### DIFF
--- a/tuxemon/core/states/combat/combat_animations.py
+++ b/tuxemon/core/states/combat/combat_animations.py
@@ -370,24 +370,24 @@ class CombatAnimations(Menu):
             centerx = home.left + scale(13)
             offset = -scale(8)
 
-        count_fainted = 0
-
-        for monster in player.monsters:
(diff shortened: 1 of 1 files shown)


tuxemon/core/states/combat/combat_animations.py
```diff
@@ -432,12 +432,12 @@ def animate_update_party_hud(self, player, home):
 
         for index in range(player.party_limit):
             sprite = None
-            if len(player.monsters) - count_fainted > index: # render alive Tuxemon balls
+            if len(player.monsters) - count_fainted > index:  # render alive Tuxemon balls
```
xirsoi, line 435: I have autopep installed in PyCharm and it made these changes.

bitcraft: thats fine.  minor linting is not a problem

bitcraft:
looks good, ty

Issue #613, opened 2020-01-20
Attack animations not displaying


MirceaKitsune:
Attack animations no longer show up. When a monster attacks the other monster, you see the monster sprites move and hear the attack sound, but the graphical overlay of the attack no longer appears. I don't know when the issue was introduced, just that it's been this way for at least several weeks.

Sanglorian:
At least part of the problem must be that some animations are missing. I see that the Pound animation is there, but most others aren't (e.g. Whirlwind's Volume Tornado or Boulder's Rockshield.)

Could someone test by using one of the Pound animation techniques? 
* Wing Tip
* Sting
* Peck
* Thunderball
* Ram

If that works, then it might just be that the other animations haven't been added to the database.

kerizane:
Pound would work, but it's entered incorrectly in the technique jsons.

Should be "animation": "pound", not "animation": "Pound"

MirceaKitsune:
I don't currently have time to do a thorough test as I'm working on my own game ATM. Did want to mention a few details: I took a look and from what I'm seeing, this isn't an engine bug but indeed an asset issue. In my test I have Fruitera as my starter, and none of its attacks display the animation.

One example is the technique fluff_up.json. I looked it up: It has the field `"animation": "Smokebomb"`. However animations/technique does not seem to contain such a graphics set.

The sting technique also doesn't work. In this case the sprites seem to be there though. Perhaps this is related to the animation definition being written in uppercase (`"animation": "Pound"`) while the files are lowercase (`pound01.png`)?

I'm assuming a cleanup of the asset graphics was done at some point, but the attack json definitions weren't updated to reflect it. If it's only that I can look into fixing them later.

kerizane:
I added those techniques a year ago, but never implemented the animations. 

Note that the animations are present on the wiki as GIFs, but in the game we use multiple PNGs as frames.

Sanglorian:
Are we able to automate importing techniques from the wiki, like we import monsters at the moment? Even if the answer is yes, it might be more difficult to import the technique animation GIFs and convert them to multiple PNGs, of course.

Otherwise it looks like fixing this would be a matter of someone manually downloading the GIFs, renaming them and converting them to PNG.

kerizane:
Could be done.

https://www.imagemagick.org/discourse-server/viewtopic.php?t=22597

Also worth considering if we can just get the game to support GIFs

bitcraft:
[...]

PR #725, opened 2020-04-06
update all monster techniques by ToLower-ing the animation strings

xirsoi:
This was a problem mentioned in #613 . Just used a simple C# program to to lower the animation property.

We'll need to have them be updated in the wiki as well or the problem will just keep coming back.


bitcraft:
I think that we can fix the wiki import script rather than the wiki itself

bitcraft:
This looks like a good start to fixing this issue.  There are two other issues involved:
* animation names contain a space, and the name is used for the filename, which don't contain spaces
* use of `startswith` in `technique.py` when finding filenames could have ambiguous results
I can merge as is.  let me know if you want to tackle the other problems

xirsoi:
Lets merge this now, i'll tackle the other issues in a second PR.
This change alone fixes over a dozen techniques, which improves the experience quite a bit.

I've also consulted with a friend who is a bit of a shell-script wizard and he helped me put together something to convert all the gifs in the wiki into single frame png's for use by the engine.

I'll include that as a third PR, since it'll be huge. It will also be best to have already figured out the solution to spaces and whatnot *before* i do a bunch of file generation/renaming.

bitcraft:
cool looks good to me, ty.  any code that you want to leave on the repo is fine in the "scripts" folder.  i encourage people to add things there and also won't scrutinize them heavily.  the gif->png stuff might be better in the wiki import script though, using Pillow.

Issue #728, opened 2020-04-10
After buying an item, the cursor moves to the top of the list


bitcraft:
Cursor should stay on the last item purchased, if possible.

Flydx1234:
If the item sells out, should the cursor return to the top or stay in the same position?
I may have a fix if we don't need the cursor to reset.

bitcraft:
I don't think it should reset.  Suppose you want to buy a couple potions by rapidly tapping tapping a button, its not currently possible because the cursor will jump to the top of the list.

kerizane:
This fix doesn't work... 

Look at healing tuxemon with potions.

kerizane:
The cursor was working recently, look for the commit that broke it (e.g. with git bisect).

bitcraft:
As with this #735, menu/cursor related things seem to get fixes and regressions often.  Thats partly on me for not following a consistent test structure.  I want to hold off on any fixes to the menu/cursor handling until we get unit tests on it to prevent regressions.  The code is too fragile.

bitcraft:
To clarify, this was an issue on the store menu, and the PR fixed it, at least in my testing.  i have not tried from the monster menu.  But I am still going to wait to fix it until we have unit tests around it.

Flydx1234:
I see what is happening.
This is caused by commit 4838529. 
On the monster menu, when reloading the cursor goes to the first "enabled" monster such as when using an item.

bitcraft:
Thanks for digging out the source of the issue.  Ill make sure this code is covered by tests and work on a fix.

kerizane:
Can we close this?

kerizane:
Fixed.

PR #733, opened 2020-04-11
Makes cursor stay in place after buying an item from shop.

Flydx1234:
This should fix issue #728


bitcraft:
short and simple, looks great, ty!

PR #739, opened 2020-04-14
Record number of steps

### LINKED ISSUE #710: Record number of steps taken by the player
for #700, there should be a game variable that
* is incremented each time the player moves one tile
* is not incremented when pushing a wall
* is not incremented when the player is pushed or moved by another object, npc, or action
* must be tried with the controls; ie the player intended to move the npc, and the npc moved, so gain one step

vch9:
At every move, the difference between the old and current position is calculated.
Introduces a new game variable: `steps`.
Since the game variable is initialized in `player.__init__`, it's introduced in the
`save_upgrader.py`

Closes #710


tuxemon/core/save_upgrader.py
```diff
@@ -57,6 +57,9 @@
 
 
 def upgrade_save(save_data):
+    if 'steps' not in save_data['game_variables']:
```
bitcraft, line 60: perfect

tuxemon/core/player.py
```diff
@@ -49,3 +49,28 @@ def __init__(self, npc_slug):
 
         # Game variables for use with events
         self.game_variables = {}
+
+        # Number of steps
+        self.game_variables['steps'] = 0
+
+    def move(self, time_passed_seconds):
```
bitcraft, line 56: this method is called at least once per a frame, and is responsible for any and and all movements from the player and from external events.  this means that if an external action moves the player additional "steps" are created.  since this is not a huge feature yet, leaving the logic here is fine, but please add this comment:
```
TODO: this will also record involuntary steps.  refactor so that only movements from the player are recorded.
```

vch9: I'll add this comment, but are external actions uses the npc.path to move the player ? I guess it does, but if it's not the number of steps could be incremented when the player appends his path.

bitcraft: internally, all movements use path.  even pressing a direction key will create a path to the next tile.  external actions (typically pathfind) creates a path.  at this time, there is no way to know what created the path...an external action or a player key, making this ticket complicated.  so, im ok with making it a `TODO` for now.

bitcraft:
Looks great thank you!

PR #741, opened 2020-04-17
Fixes syntax warning upon launch using Python 3.8

Flydx1234:
Since Python 3.8 produces syntax warnings when using is/is not to compare against literals,  SyntaxWarning: "is not" with a literal. Did you mean "!="?
  assert mouse_pos is not 0
is printed upon launch.
Instead == or != should be used.


bitcraft:
Looks good to me. Thanks for the fix.

PR #745, opened 2020-04-18
Fix teleport from 37707 to the player's bedroom.

vnmabus:
The teleport was using the old name of the map.


bitcraft:
This can be merged.  For reference though, we have a way to track map name changes so it’s not a huge priority to fix map names in the map actions.

PR #747, opened 2020-04-19
added controls as setting in config; default config adjusted to match

austintian03:


tuxemon/core/config.py
```diff
@@ -171,6 +186,17 @@ def get_defaults():
             ("debug_logging", True),
             ("debug_level", "error")
         ))),
+        ("controls", OrderedDict((
```
bitcraft, line 189: this should be reversed to allow many keys to do the same thing (lshift and rshift):
```python

        ("controls", OrderedDict((
            ("lshift", "player_b"),
            ("rshift", "player_b"),
            ...
```

tuxemon/core/config.py
```diff
@@ -115,7 +119,18 @@ def __init__(self, config_path=None):
         # input config (None means use default for the platform)
         self.gamepad_deadzone = .25
         self.gamepad_button_map = None
-        self.keyboard_button_map = None
+        self.keyboard_button_map = {
```
bitcraft, line 122: this is a great start.  it looks correct, but maybe we could make this a bit more DRY by refactoring into a single method?  And it also looks like it is going to change the defaults in `PygameKeyboardInput.default_input_map`.  hope you don't mind this suggestion, but maybe we could do something like
```python
class TuxemonConfig(object):
    def __init__(self, config_path=None):
      ...
      self.keyboard_button_map = get_custom_pygame_keybaord_controls(cfg)
      ...

def get_custom_pygame_keybaord_controls(cfg):
    import pygame.locals
    from tuxemon.core.platform.platform_pygame.events import PygameKeyboardInput

    custom_controls = PygameKeyboardInput.default_input_map.copy()
    for key, value in cfg.items("controls"):
        key = key.lower()
        value = value.upper()
        pygame_value = getattr(pygame.locals, "K_" + key, None)
        button_value = getattr(buttons, value, None)
        if pygame_value is not None and button_value is not None:
            custom_controls[pygame_value] = button_value
    return custom_controls
```

tuxemon/core/config.py
bitcraft: config would be:
```
[controls]
W = UP
S = DOWN
```
I'm note sure if including "player_"... is important or not, but the code is more simple if removed.  your call there.

tuxemon/core/config.py
austintian03: Yeah, this idea looks great. I'll definitely get started on this.

tuxemon/core/config.py
austintian03: Yeah, this makes sense. I'll remove "player_" because it makes more sense for a user to simply type in the button they want rather than player_"<insert button>". Thanks for the suggestions!

austintian03:
So I incorporated the suggestions you made, but the only issue I had was that whenever I wanted to change the controls, I couldn't replace the default controls, only add new ones.

Whenever I changed in the config (for ex.):
` up = up`      ->      'w = up'
The game ends up allowing me to use both 'up' and 'w' as the up button. This is due to the generate_default_config always repopulating the `controls` section with the defaults.

tuxemon/core/config.py
```diff
@@ -115,11 +119,38 @@ def __init__(self, config_path=None):
...
         # not configurable from the file yet
         self.mods = ["tuxemon"]
 
+def get_custom_pygame_keyboard_controls(cfg):
+    import pygame.locals
+    from tuxemon.core.platform.platform_pygame.events import PygameKeyboardInput
+
+    custom_controls = PygameKeyboardInput.default_input_map.copy()
+    for key, value in cfg.items("controls"):
+        key = key.lower() if len(key) == 1 else key.upper()
```
bitcraft, line 144: please add this comment above `key = key.lower() ...`
```
# pygame.locals uses all caps for constants except for letters
```

bitcraft:
> the only issue I had was that whenever I wanted to change the controls, I couldn't replace the default controls, only add new ones.

Yeah, I've noticed that too.  I really like these changes here though.  A follow up can be done that overrides that defaults, instead of just adding additional mappings.

PR #748, opened 2020-04-20
fix technique animations, status, new download script and animations

bitcraft:
Combat animations are once again visible.  Verified status effects work, but I couldn't find any moves which apply them (only tested poison...).  Verified by editing a save file.

* Animations show during combat
* Added many previously unused animations
* Fixed various techniques to use the correct animation
* Script which downloads gifs from the wiki and creates png frames
* Logger message when animations cannot be loaded from db
* Verified poison status works


bitcraft:
thanks for the review @kerizane

Issue #711, opened 2020-03-25
Option to change back sprite of the trainer during combat


bitcraft:
Its possible to change overworld sprites for the trainer, but the battle sprites won't change.  We should ideally have a db entry that defines some basic trainer sprite options.  Each entry would define the overworld sprite, battle sprite, etc.

Any ideas on where to store the sprite name?  Game variables?  Attribute on the NPC class?

xirsoi:
Currently there is a sprite_name attribute the npc db entries. We could have the back sprite be based on that, like all the other sprites.

Sanglorian:
My preference would be for each [NPC](https://wiki.tuxemon.org/index.php?title=Category:Character) to belong to a [Class](https://wiki.tuxemon.org/index.php?title=Category:Trainer), which would define the sprites (front, back and overland) for all NPCs belonging to that class.

But happy to be flexible!

bitcraft:
Sounds like the consensus is it put it on the NPC objects.  Sounds good to me.  We will need a new entry on the NPC db items for combat front and back sprites, as well as changing the hard coded values in the combat state.

xirsoi:
It's funny, I was thinking the same thing as you, sanglorian. I feel making a new table for a "sprite class" (or something) that would define the set of sprites and be referred to by the npcs would work best.

This would also allow us to remove hard-coded file paths from the combat_animations class.

bitcraft:
Thanks for the link.  Its cool to see that there are already so many usable sprites for the game.  I agree that we should get this into the game and into the db so our NPCs can be customized more.

bitcraft:
Just a small point.  I think that there are two issues here though.  NPCs should still be able to provide their own unique images for sprites; defined and stored on the NPC as well.  Classes would be a template for NPCs at creation.  I don't see it happening right away, but eventually changeable outfits or customization would be nice, and storing that info on the class only would get in the way of that goal.

Qiangong2:
Should we have changeable outfits? Or just at the beginning the ability to choose from say, 3 or 4 styles? I think it would be great to have multiple hats, shirts, shoes and pants/shorts/skirts. Although, that didn't get added in until Pokemon X/Y. Changing haircuts would be neat too.

Wouldn't we need to create sprites with every single combination for this to work?

xirsoi:
We could eventually support that level of customizability, and it would not require every possible combo to be sprited out ahead of time. That would be the naive approach and a very bad idea.

[...]

PR #750, opened 2020-04-20
Adds back and front combat sprite support to NPCs.

Flydx1234:
Issue #711 
Currently, the only NPC's with a back/front sprite are red and maple/femaletrainer. 
Although setting the back/front sprite in this way may not be necessary for the non-player characters.


tuxemon/core/npc.py
```diff
@@ -101,13 +101,24 @@ def __init__(self, npc_slug, sprite_name=None):
         # use 'animations' passed in
         # Hold on the the string so it can be sent over the network
         self.sprite_name = sprite_name
-
+        self.combat_front = combat_front;
```
bitcraft, line 104: remove the semicolons

tuxemon/core/npc.py
```diff
@@ -101,13 +101,24 @@ def __init__(self, npc_slug, sprite_name=None):
...
+        self.combat_back = combat_back;
         if self.sprite_name is None:
             # Try to use the sprites defined in the JSON data
             try:
                 self.sprite_name = npc_data["sprite_name"]
             except KeyError:
                 logger.error('Cannot find sprite for {}'.format(npc_slug))
+        if self.combat_front is None:
```
bitcraft, line 112: if these are None, then the combat state will not work at all.  in this case, since we are aware that this will cause an issue if there is None here we should let the game crash.  i would remove the try/except and let it crash with `KeyError`.  also, rewrite the `sprite_name` getter to be the same.  these are essential values and should always be filled.

mods/tuxemon/db/npc/red.json
```diff
@@ -1,4 +1,6 @@
 {
     "slug": "npc_red",
-    "sprite_name": "red"
+    "sprite_name": "red",
+    "combat_front": "red_front.png",
+    "combat_back": "red_back.png"
```
bitcraft, line 5: this is good.  please go through the other npcs and add front/back sprites.  i think if there is no good default, you should use red.

bitcraft:
just a quick comment for context on adding combat sprites.  one issue that im slowly fixing is allowing bad data to propagate through the code.  its better to enforce essential values from data files.  right now i think that combat sprites are essential.....i guess if they are not then you don't have to change this...just let the game crash during the combat state.  sorry for being wishy-washy on this one.

tuxemon/core/npc.py
Flydx1234: Makes sense.

mods/tuxemon/db/npc/red.json
Flydx1234: For all NPCs, I have added either red or maple front/back sprites.
Knight and professor both have their own front sprites so those are used there.

bitcraft:
Looks great!  Thank you for making those changes.

PR #754, opened 2020-04-22
Use Unicode arrows and musical note.

vnmabus:
Currently the font allows the use of proper arrow characters and the musical note symbol, so I think it is preferable to use them instead of ugly workarounds such as <=, ^ or v.


bitcraft:
This is awesome!  I didn't know this was possible.  Thank you!

PR #764, opened 2020-04-23
check for polyline objects based on points and closed attributes

bitcraft:
* Make check for collision objects less restrictive; just look for work "collision"
* If is for collisions, then check if it is closed or not
* Closed objects are regions/polygons/rects
* Open objects are lines

Also fixed a clif in route2.tmx


bitcraft:
Thanks for the review

Issue #729, opened 2020-04-10
Trainer battles with multiple txmn has broken UI


bitcraft:
Noticed the following issues:

- [x] First enemy txmn name card comes in with trainer
- [x] First enemy txmn name card comes in *again* after released
- [x] First enemy txmn name card is still visible after defeat
- [x] First enemy txmn name card HP is filled after defeat
- [x] The final TXMN capdev icon is not greyed out after defeat
- [x] Various UI elements are duplicated (maybe they are created per monster?)
- [x] The monster name is lower case is messages, maybe it is using the slug?

When all are fixed, this issue may be closed.

Flydx1234:
The last issue with the monster name in lower case, unless mistaken, only happen with tuxemon that are not defined in the locale file.
Do they need to be manually added?

bitcraft:
> Do they need to be manually added?
No, I dont think so.  If its just a missing translation then its not a big deal.  Ill open a new issue to get some logging around missing translations, so its more obvious.

There is still an issue with the tuxeball display.  In a trainer battle, after you defeat the 1st monster, the black arrow thing and the tuxeballs are duplicated.  If you watch closely, you will see them come on the screen again, but very quickly.  Looks great so far, i love the changes to make it look perfect.

PR #768, opened 2020-04-25
Fixing opponent battle tuxemon

Flydx1234:
Fixes the following issues from #729 :
> 
> * [X]  First enemy txmn name card comes in with trainer
> * [X]  First enemy txmn name card comes in _again_ after released
> * [X]  First enemy txmn name card is still visible after defeat
> * [X]  First enemy txmn name card HP is filled after defeat
> * [ ]  The final TXMN capdev icon is not greyed out after defeat
> * [ ]  Various UI elements are duplicated (maybe they are created per monster?)
> * [ ]  The monster name is lower case is messages, maybe it is using the slug?
> 
> When all are fixed, this issue may be closed.

### DIFF
--- a/tuxemon/core/states/combat/combat.py
+++ b/tuxemon/core/states/combat/combat.py
@@ -162,7 +162,7 @@ class CombatState(CombatAnimations):
         self._layout = dict()  # player => home areas on screen
         self._animation_in_progress = False  # if true, delay phase change
         self._round = 0
-
+        
         super(CombatState, self).startup(**kwargs)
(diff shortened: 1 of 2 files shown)


Flydx1234:
Not sure what the duplicated UI elements in combat are.
Only really noticed the first enemy TXMN name.

bitcraft:
If you watch closely, it looks like all the capture device ball sprites are duplicated, as well as the black line/arrow.  I will need to gather screen shots to confirm though.

bitcraft:
Short and sweet.  Ty!

PR #778, opened 2020-05-01
Fixes duplicated ui elements.

Flydx1234:
#729
Elements appear to duplicate due to the redrawing of the hud taking too long.
Decreased duration to remove the effect.

### DIFF
--- a/tuxemon/core/sprite.py
+++ b/tuxemon/core/sprite.py
@@ -31,6 +31,7 @@ from __future__ import unicode_literals
 
 import logging
 import math
+from functools import partial
 
 import pygame
 from pygame.transform import rotozoom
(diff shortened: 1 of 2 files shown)


Flydx1234:
I see. I have removed the black arrow's animation completely, since there's no reason to update it. 
To remove loading the arrow sprite again, I added two variables to self, one for the opponent sprite and player sprite that are both set upon when the battle starts.

I am not sure how to go about removing the old cap dev sprites. Doesn't pygame remove them when drawn over?

bitcraft:
@Flydx1234 We are using a "sprite and groups" rendering system.  This means when sprites are added to the screen, they are on a list of on screen elements are are drawn each frame.  In pygame, if you only blit the surface once, then yes, it is "lost" when drawn over, but we are using sprites, which handling drawing each frame.

I would make a new `CaptureDeviceSprite`, which is a subclass of our `tuxemon.sprite.Sprite`.  These objects would hold a reference to a monster and the state of the capture device display.  They should also have a method to change their state.  Create the sprites, and keep in a list (or not, you can search the state's sprites later).  When updating the display, go over each sprite and if the current monster status (fainted or not) is different from the capdev sprites internal state, then change the state and animate it.

Flydx1234:
I gave it a shot and I think this fixes the issues of the multiple arrow and capdevs being rendered on the screen.

Issue #779, opened 2020-05-01
Fainted tuxemon status doesn't stay between saves.


Flydx1234:
After loading a save when you enter combat, all capdevs show up as available even if you have say, 1/5 tuxemon alive.
Only after fainting in a battle first does the next battle show the capdev as grayed out.

kerizane:
Is this just a display issue, or are you able to send out fainted tuxemon?

Flydx1234:
This is just a display issue.

kerizane:
Issue isn't with saving, when you win a battle the tuxemon loses "status_faint" somehow.

Flydx1234:
Ah yes, I see.

kerizane:
Fix coming

PR #780, opened 2020-05-01
#779: Stop status_faint from being lost between battles

kerizane:


bitcraft:
Thanks for the simple fix!

Issue #742, opened 2020-04-17
Potion and Revive Have Same Icon


ryanhanli:
I noticed that the potion and the revive icon in the store have the same icon.  I can make a different icon along with other icons for future potions/elixirs that possibly deal with poison, burn, etc.

Sanglorian:
That would be great - thank you! At the moment we have an item that deals with any condition, called Restoration: https://wiki.tuxemon.org/index.php?title=Restoration

But we can always use more pixel art of various kinds!

PR #783, opened 2020-05-08
Added New Pixel Art(Potion Icons)

ryanhanli:
[#742](https://github.com/Tuxemon/Tuxemon/issues/742) Added several new potion icons and gave the revive icon one of them.  All of the new unused potions have the suffix "un" which stands for unused, but could be useful for the future when antidotes and other status effect remedies are introduced to the game.


bitcraft:
Thanks for the graphic content.  Personally, I don't think its needed to names the files anything in particular if they are unused.  I'm not going to block the PR because of the "un" suffixes, but I don't think its needed in the future.

bitcraft:
@ryanhanli are these items your own artwork?  we should be added the artist's name and the license that they are released under in the `CREDITS.md` file.  please update that file with the artist (or you) and the license they are released under.

ryanhanli:
Yes they are but I will make a new pull request and with all the changes you asked for including removing the "un" parts.

PR #785, opened 2020-05-11
Key bindings in config now override defaults (Tuxemon#767)

austintian03:
This fix reverses the key/value pairs of the `controls` section of the config. Now the key is the pygame internal action, and the value is a comma separated list of strings for your keyboard buttons. Since the keys in the config section never get deleted or modified, you can modify the controls without `generate_default_config()` always adding back the default controls.

### DIFF
--- a/tuxemon/core/config.py
+++ b/tuxemon/core/config.py
@@ -53,20 +53,15 @@ class TuxemonConfig(object):
     """
 
     def __init__(self, config_path=None):
-        import os.path
-        
         # load default config
-        cfg = configparser.ConfigParser()
+        cfg = generate_default_config()
+        self.cfg = cfg
 
         # update with customized values
-        if os.path.isfile(config_path):
+        if config_path:
             temp = configparser.ConfigParser()
             temp.read(config_path)
             populate_config(cfg, temp._sections)
-        else:
-            populate_config(cfg, get_defaults())
-
-        self.cfg = cfg
 
         # [display]
         resolution_x = cfg.getint("display", "resolution_x")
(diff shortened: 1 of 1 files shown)


Qiangong2:
How would this work in the Android version of the game? Do the controls still work? Since it uses an overlay, this would affect the configuration wouldn't it?

bitcraft:
@Qiangong2 is is not related to the android build.  controls are no longer bound to the keyboard and multiple virtual controllers are available to use at any time.  the overlay needs to be fixed though.

tuxemon/core/config.py
```diff
@@ -126,18 +126,20 @@ def get_custom_pygame_keyboard_controls(cfg):
...
-        key = key.lower() if len(key) == 1 else key.upper()
-        value = value.upper()
-        pygame_value = getattr(pygame.locals, "K_" + key, None)
-        button_value = getattr(buttons, value, None)
-        event_value = getattr(events, value, None)
-        if pygame_value is not None and button_value is not None:
-            custom_controls[pygame_value] = button_value
-        elif pygame_value is not None and event_value is not None:
-            custom_controls[pygame_value] = event_value
+    #custom_controls = PygameKeyboardInput.default_input_map.copy()
```
bitcraft, line 129: please remove this comment

bitcraft, line 129: please remove this comment

Qiangong2:
@austintian03 Please remove the comment so that we may get this merged in.

bitcraft:
Thank you @austintian03 !

Issue #749, opened 2020-04-20
Combat animations only face one direction


bitcraft:
For some animations, direction is not important.  For others, such as a bullet, or water shot, the direction matters.  We should flip animations when needed.

- [ ] Audit all animations and verify that they all move from left to right
- [ ] Flip any animations files that are not correct
- [ ] Update contribution guidelines such that artists are aware of the left-to-right convention
- [ ] Display the correct orientation of animations by determining which direction they move
- [ ] In the combat code, modify `TechniqueAnimationCache` so that the `get` method can flip animations

bitcraft:
@fieldchicken will take this

AndyMender:
I see quite some progress was made on this already so it would a shame to drop it entirely. I would like to continue the ticket.
Some minor changes on the Tuxemon wiki would also be required:
- changing the direction of breath_blue and breath_fire from left-facing to right-facing.
- shield_turtle_left and snake_left_128 are no longer needed

kerizane:
Note that some animations must not be flipped, e.g. https://wiki.tuxemon.org/File:Buff_armour.gif

AndyMender:
> Note that some animations must not be flipped, e.g. https://wiki.tuxemon.org/File:Buff_armour.gif

Yup, good point! There might be some more in the ones that are currently not used so I listed only the ones in the "used" category and which are also checked into the git tree.

AndyMender:
@bitcraft Should technique JSONs have an extra field indicating whether an animation should be flipped or not? The technique download script would need to populate that value as well and/or we add an extra field in the wiki (for instance, `allow_flipping? yes` and `allow flipping? ` (no value)).

kerizane:
I think we should introduce an animation JSON file to store this, otherwise there will be too much duplication.

Ideally they would be on the wiki, but you could just config it within the download script for now.

AndyMender:
> I think we should introduce an animation JSON file to store this, otherwise there will be too much duplication.

That would definitely be a follow-up effort. For now I think we can leave things as they are or add a field to the technique JSONs if there is currently a need to flip animations in more than one axis/direction.

kerizane:
It would be helpful if you created a list of animations and which way they need to be flipped. The list should be posted as text within this issue.

It's currently not clear how many need flipping, and if it's a simple diagonal flip, or if some will only need flipping in the y axis.

AndyMender:
Here's the list of animations which would need to be flipped and their axes:
```
[...]

PR #788, opened 2020-05-15
Animations audit

fieldchicken:
Issue #749 
1. Corrected all of the current animations so that they move from left to right
2. Updated the CONTRIBUTING file to include the new animation guidelines
3. Updated the combat file to mirror the images of animations that need to move from right to left


bitcraft:
@fieldchicken thanks for thesea changes.  there are currently merge conflicts that need to be resolved.

tuxemon/core/event/actions/start_battle.py
```diff
@@ -64,10 +64,12 @@ def start(self):
             return False
 
         world = self.session.client.get_state_name("WorldState")
+        print(self.parameters)
```
xirsoi, line 67: this is debug code and should be removed

tuxemon/core/event/actions/start_battle.py
```diff
@@ -64,10 +64,12 @@ def start(self):
...
         if not world:
             return False
 
         npc = world.get_entity(self.parameters.npc_slug)
+        print(npc)
```
xirsoi, line 72: same here

tuxemon/core/event/actions/start_battle.py
xirsoi: removing this, and the one below, would also resolve your merge conflict, btw.

tuxemon/core/states/combat/combat.py
```diff
@@ -96,6 +97,8 @@ def load_technique_animation(technique):
         images = list()
         for fn in technique.images:
             image = graphics.load_and_scale(fn)
+            if direction:
```
bitcraft, line 100: The `PygAnimation` class can handle transformations.  Please use that instead of the low level flip here.  Reason is that I am trying to keep pygame from polluting the entire code base.

should be
```python
tech = PygAnimation(images, False)
tech.flip(True, False)
```
See pyganim.py#455

CONTRIBUTING.md
```diff
@@ -37,6 +37,28 @@ be under a free license, such as CC0.  For questions about the license, open a g
 PLEASE NOTE!  All contributions must be submitted with LOWER CASE FILENAMES only!  PRs which
 do not follow this format may be rejected until filenames are renamed.
 
+Additionally, for combat animation contributions please ensure that techniques that move along the x-axis by default do so from left to right. There is code that will determine the correct orientation that the animation should be displayed.
```
bitcraft, line 40: Perfect, ty!

tuxemon/core/states/combat/combat.py
bitcraft: also, remove the pygame import after this is done.

bitcraft:
closing because it appears to be abandoned

PR #797, opened 2020-08-30
Spyder in the Cathedral campaign - all Maps redone, NPCs, plot up to Wayfarer Inn

Sanglorian:
Hi folks,

Hope I'm doing this right! This is a big pull request, containing a number of different elements. I know the preference is for small, contained pull requests but in this case it wasn't really possible as each part was connected to other parts and it needed to be updated and tested as a whole. 

Essentially, this pull request creates the first part of the [Spyder in the Cathedral campaign](https://wiki.tuxemon.org/index.php?title=The_Spyder_in_the_Cathedral). It also rewrites the entire map to be circular rather than straight - the final map is adjacent to the first map.  

Throughout, I've used the "spyder_" prefix to clearly identify the maps and NPCs that belong to the Spyder campaign. I guess in the long term there'll be different mod folders for each campaign. 

The game has been tested to the end of Route 3. However, I can only get the pre-packaged Windows version from earlier this year working, not the current version, so that's what I tested it in. 

To play, simply rename the current player_house_bedroom.tmx file to something else, and then rename spyder_player_house_bedroom.tmx to player_house_bedroom.tmx. 

A couple things to note: 

- I can't get the logic of the Leather Town museum entry fee to work (obviously it doesn't charge at this point, but even the entry/exit isn't working as I planned). If someone could take a look that'd be much appreciated! 
- Since combat conditions aren't implemented yet, I set the starter to level 50 so people can forge through the combats and test them all out easily. 

Thank you for taking a look! Happy to make any changes etc.


xirsoi:
Just wanna note your comment about the campaigns being separate: It's true that eventually we'll want to separate the game into two (maybe 3) mods, but currently the engine doesn't allow some references to cross the "mod boundary", so to speak. So for now it's fine.

mods/tuxemon/db/encounter/route5.json
```diff
@@ -0,0 +1,19 @@
+{"id": "route5","monsters": [
```
xirsoi, line 1: id should be slug

mods/tuxemon/db/encounter/routeb.json
```diff
@@ -0,0 +1,19 @@
+{"id": "routeb","monsters": [
```
xirsoi, line 1: this should be slug too

mods/tuxemon/db/npc/spyder_barmaid.json
```diff
@@ -0,0 +1,4 @@
+{
+    "slug": "spyder_barmaid",
```
xirsoi, line 2: npcs now require combat_front and combat_back sprites to be defined.

xirsoi (changes requested):
you have a conflict in base.po, but it's super easy to fix, just take both changes.
there may be other issues, but i don't have time atm. i thought i'd let you know about these so you have some time to fix them.

xirsoi:
The collision for the orange statue in Cotton Town, next to Omnichannel HQ, is one tile above where it should be. This means i can walk through the statue, but not enter the space above. I can also go to the map past the statue where no collisions are setup, so i can walk across the river and through the trees 🤣

xirsoi:
Some of the music is not referenced by the correct name in script. Check the music folder of the db for correct slugs.

xirsoi:
The player's name isn't being inserted correctly, it just shows up as NAME in dialogue.

mods/tuxemon/db/encounter/route5.json
Sanglorian: Thanks - will fix.

mods/tuxemon/db/encounter/routeb.json
Sanglorian: Thanks - will fix

mods/tuxemon/db/npc/spyder_barmaid.json
Sanglorian: Thanks - will fix.

Sanglorian:
> you have a conflict in base.po, but it's super easy to fix, just take both changes.

Is that just this? If so, I'll add it to the end of the base.po file :)

> =======
> msgid "cannot_use_item_monster"
> msgstr "This item cannot be used on this monster."
> >>>>>>> development

bitcraft (approved):
thank you!

PR #810, opened 2020-09-24
Remove python2 compatability fixes

bitcraft:
* remove future statements
* remove `(object)` from class defs

I used a tool called "pyupgrade" (https://github.com/asottile/pyupgrade) and it seemed to work well.  annoyingly, it "upgraded" some print statements in small ways, but seems to be working fine still.

@kerizane @xirsoi do you mind taking a look here?  i think this is all really safe, I played through and didn't see any issues.


kerizane (approved):
Looks fine.

A few of the docstrings have r in front of them. Up to you if want to keep that or not.

xirsoi (approved):
looks good to me, if you want i can pull it to my local and try to run it

bitcraft:
...i noticed the "r" adds on the docstrings.  ill clean those up (why were they even added??) then merge it.  thanks for looking.

PR #812, opened 2020-10-15
add the taba town greeter

xirsoi:
Adds a new npc, Greeter Jaime, who is a greeter standing outside the Taba Town Tuxemart.
This npc, and the corresponding dialog scripts, are meant to showcase the functioning of the has_item condition.
Once you have at least one monster in your party and step in front of the door to the mart, Jaime will talk to you.
The event diverges here; if you have capture devices in your inventory already he'll note this and wish you well.
If you don't, he'll tell you the shop is having a promotion right now and give you 5.
He won't greet you again after that.

I've also added a secret cache of 5 tuxeballs. If you follow the southern side of the fence just below the mart to the east and "activate" the tree at the end you'll get 5 tuxeballs. This makes it easier to test this particular event, in the case you are already outside of the mart.

I put a little effort into this so it's good enough to just be left in the game, in my opinion.
Oh, while I was at it, I updated add_item to accept an optional quantity parameter.

### DIFF
--- a/mods/tuxemon/db/npc/npc_taba_greeter.json
+++ b/mods/tuxemon/db/npc/npc_taba_greeter.json
@@ -1,0 +1,6 @@
+{
+    "slug": "npc_taba_greeter",
+    "sprite_name": "tuxemartemployee",
+    "combat_front": "red_front.png",
+    "combat_back": "red_back.png"
+}
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -1956,3 +1956,15 @@ msgstr "@*!^: 0H... @ n3wc0m3r! H0vv p1d y0v g3t he7e? <NONE> h3llen\"T $33N %^&
 
 msgid "cannot_use_item_monster"
 msgstr "This item cannot be used on this monster."
+
+msgid "taba_greeting1"
+msgstr "Greeter Jaime: Hey there ${{name}}! Heard you got a Tuxemon, congrats! If you wanna grow your party, you'll need Tuxeballs."
+
+msgid "taba_greeting2a"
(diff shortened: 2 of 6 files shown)


mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -1956,3 +1956,15 @@ msgstr "@*!^: 0H... @ n3wc0m3r! H0vv p1d y0v g3t he7e? <NONE> h3llen\"T $33N %^&
 
 msgid "cannot_use_item_monster"
 msgstr "This item cannot be used on this monster."
+
+msgid "taba_greeting1"
+msgstr "Greeter Jaime: Hey there ${{name}}! Heard you got a Tuxemon, congrats! If you wanna grow your party, you'll need Tuxeballs."
```
Qiangong2, line 1961: How does the greeter know your name? Maybe something more like, "Congratulations on being the one-thousandth customer!"

mods/tuxemon/db/npc/npc_taba_greeter.json
```diff
@@ -0,0 +1,6 @@
+{
+    "slug": "npc_taba_greeter",
```
Qiangong2, line 2: Except for legacy npcs, all now no longer have npc_ at the beginning. Is there any reason to having it here? npc slugs are only called in npc-specific actions and conditions

Qiangong2 (changes requested):

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
xirsoi: I think I had it in my head that they would know you because this is the shop your family would use, so he would have seen you on the regular.

mods/tuxemon/db/npc/npc_taba_greeter.json
xirsoi: Right, I'll get that fixed.

bitcraft (approved):
thank you!

PR #816, opened 2020-10-28
optimize sprites size

MasterOChaos:
Optimized the sprites size


MasterOChaos:
Removed unnecessary metadata from files

PR #818, opened 2020-10-29
Fix minor spelling issue

MasterOChaos:
Fixes the spelling issue in en_US pertaining to revive_description


Qiangong2 (changes requested):
Shouldn't it be "Revives a knocked out Tuxemon and restores 20 HP" ? The way you wrote it doesn't make sense

MasterOChaos:
Whoops
You are right

MasterOChaos (commented):
Fixed my bad English

Qiangong2:
Thanks. I'm mainly concerned that it's such a small edit that a full pr may not be required at all. Maybe @bitcraft can weigh in. Should this PR be accepted, or is it too small to just have only this?

PR #823, opened 2020-11-09
Weblate tuxemon tuxemon

bitcraft:


kerizane:
image

bitcraft:
@kerizane yeah that looks bad.  i don't want to block merging it right now though.   i think it could be fixed later with better automatic splitting for dialogs

vnmabus:
> 
> 
> @kerizane yeah that looks bad. i don't want to block merging it right now though. i think it could be fixed later with better automatic splitting for dialogs

Yes, this is something I noticed with some dialogs. In Spanish some sentences tend to be longer, and this is difficult to take into account. In the past I changed some dialogs, inserting additional newlines in the translation to force splitting, but I am not sure this is a good practice (Weblate complains when the number of newlines differ between original and translated dialog).

An additional problem is that when I use Weblate, I do not test the translations right away, so I do not notice this errors until later.

bitcraft:
i think that we should do our best translations on weblate right now, even if they don't fit.  i think there is an open issue to automatically "paginate" the dialogs.  i don't want to force translators to make short or awkward translations just because we need a feature for longer dialogs.  so newlines, splitting, etc is fine, but not needed really since its going to change whenever automatic new dialogs are implemented.

this goes for english too.  eventually when automatic new dialogs/pagination is complete, i will likely remove all the newlines that are not needed.

some new lines are needed, but i suspect most are not.

PR #826, opened 2020-11-20
Currency placeholder

vnmabus:
Adds currency placeholder, as discussed in Discord.

In the future, we should replace the dollar with a custom currency symbol. With this change, this can be done in a centralized way. It also opens the door to translate the currency symbol in different languages or to have different currencies for different campaigns and reusing the dialogs between them.


bitcraft (approved):

bitcraft:
looks great, thank you!

PR #855, opened 2021-03-11
Feature/add monster to npc

xirsoi:
adds optional trainer slug parameter to <add/remove>_monster. Defaults to player (to preserve existing functionality).
This allows a script to alter an npc's party, but only for the current session. Changes are not preserved in the save.

### DIFF
--- a/tuxemon/core/event/actions/add_monster.py
+++ b/tuxemon/core/event/actions/add_monster.py
@@ -24,11 +24,12 @@ from tuxemon.core.event.eventaction import EventAction
 from tuxemon.core.event import get_npc
 
 class AddMonsterAction(EventAction):
-    """Adds a monster to the current player's party if there is room.
+    """Adds a monster to the specified trainer's party if there is room.
+    If no is trainer specified it defaults to the current player.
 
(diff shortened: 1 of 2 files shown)


tuxemon/core/event/actions/add_monster.py
```diff
@@ -21,6 +21,7 @@
 
 from tuxemon.core import monster
 from tuxemon.core.event.eventaction import EventAction
+from tuxemon.core.event import get_npc
 
 
 class AddMonsterAction(EventAction):
```
bitcraft, line 27: could you update the doc string for this and the other action?

bitcraft (commented):
looks great, please update the docstrings to reflect that they work on players and npcs

bitcraft:
There are merge conflicts, too

tuxemon/core/event/actions/add_monster.py
xirsoi: whoops, will do

bitcraft:
Thank you!

bitcraft (approved):

PR #857, opened 2021-03-23
add back controller overlay

bitcraft:
adds the controller overlay with no additional features.  works with mouse but not tested with multi touch.  i suspect that there will be bugs with that, but i need a device to test with.


tuxemon/core/client.py
```diff
@@ -51,14 +52,14 @@ def __init__(self, caption):
         :rtype: None
         """
         # Set up our game's configuration from the prepare module.
-        self.config = prepare.CONFIG
+        self.config = config = prepare.CONFIG
```
xirsoi, line 55: why do we have a local instance *and* a class instance of config?

xirsoi (dismissed):
LGTM

tuxemon/core/client.py
bitcraft: reduce the amount to times i need to type "self."

PR #858, opened 2021-03-24
Add Rich Presence

ghost:


tuxemon/core/client.py
```diff
@@ -38,6 +41,11 @@
 
 logger = logging.getLogger(__name__)
 
+#Rich Presence Implementation Part 1
+client_id = '824398106182811698'  # Put your Client ID here, this is mine lol
```
ghost, line 45: Possibly needs changed have not tested with an separate account

bitcraft: lets move this to the config file

tuxemon/core/client.py
```diff
@@ -261,7 +269,10 @@ def update(self, time_delta):
         # Android-specific check for pause
         # if android and android.check_pause():
         #     android.wait_for_resume()
-
+
```
bitcraft, line 272: add this to the main loop here: https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/client.py#L241
follow the same format as `time_since_draw` and `frame_length` and make a timer for 60 seconds so that it is not making this call every frame.

tuxemon/core/client.py
bitcraft: how is the secret handled?  can we move that to the config as well?

bitcraft (changes requested):

PR #860, opened 2021-03-27
this should fix the issue

xirsoi:
There was an issue with start_battle reloading the npc party from the database, rather than using the in-memory data.
To resolve this I had create_npc load the npc party from data and removed the bit where start_battle reloads the party.

I also fixed an issue in load_party where, if the trainer lacked a party it would cause a crash.


bitcraft (approved):
looks good ty

PR #863, opened 2021-04-12
Add eric6 support

ghost:
Remove eric files with gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,10 @@ Thumbs.db
 *.iml
 .idea/*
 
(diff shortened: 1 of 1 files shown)


bitcraft:
i don't think we should be editing this.  this would allow eric6 files to be committed to the repo, which would be the config for your ide.  anyone else using the eric will have their own preferences for this repo wiped out with these changes

ghost:
It should only stop the files from getting sent to the repo
similar to the pycharm support

PR #864, opened 2021-04-15
Quick fix to screenshot not displaying.

magmaus3:
Currently, the tuxemon website is down. This prevents the image from displaying.
I used the archived version from archive.org.


Qiangong2:
Wouldn't it be better to add the image directly to the repo? That way, we can just link to it directly.

bitcraft:
the image should be in the repo, not hotlinking the web archive

PR #872, opened 2021-04-27
Get Route 1 set for code freeze

Qiangong2:
This should get merged in before the code freeze. You can technically continue with the story now, although it makes little sense since I am rewriting most of it.

This PR adds/removes the following:

- Removes giant line of xero grunts
- Adds first battle area (all subsequent battle areas will look the same, with different color schemes)
- Various small Quality of Life fixes

I'd normally do more for a pr, but since the code freeze is happening, I thought it best to get this stuff in


bitcraft (commented):
looks good to me

bitcraft (approved):

PR #875, opened 2021-05-01
Introduce different catch rates for Tuxemon based on rarity and update capture system 

### LINKED ISSUE #874: Capture system heavily influenced by capture device power and Tuxemons HP
The catch probability is mainly influenced by the Tuxemons HP and the power of the capture device. This gives the feeling that all Tuxemon have the same probability of being caught (because they have) and doesn't give the sense of rarity of a Tuxemon which is key to games like Tuxemon.  
I'm currently working on a PR in this matter which will try to give more influence to the catch_rate of tuxemon in order to give more depth to the game. 
For more information check https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/item/effects/capture.py and
https://bulbapedia.bulbagarden.net/wiki/Catch_rate#Capture_method_.28Generation_VI.29

GDMarkou:
Fix #874 
This PR aims to:

-  Remove magic numbers and improve documentation of capture.py
-  Reduce the Influence of Capture device, Tuxemon HP and status_modifier in the capture calculation
-  Add catch rates based on rarity and evolution of Tuxemon

### Capture.py
Reduce the influence of the Capture device from 100 power to 1. This is done to the calculation and not the original source of the capture device. Since there is only one capture device, its power should not influence the capture of a Tuxemon so greatly. Also the status modifier was reduced from 1.5 to 1.2 because it greatly increased all Tuxemon catch probability. Lastly, the shake mechanism has been moved from generation VI to generation III-IV. After a lot of testing the gen III-IV shake mechanic produced more accurate results.

### Catch_rate
All of the above changes happen to support the introduction of the catch_rate variable, as described in the original calculation instructions found at  https://bulbapedia.bulbagarden.net/wiki/Catch_rate#Capture_method_.28Generation_VI.29. Catch_rate ranges from 1-255, and the probability of a Tuxemon being caught is approximately catch_rate/255 (Approximately because other factors like Tuxemon HP and status modifiers affect the capture). 

Introduce catch_rate values to all of the Tuxemon in the Tuxemon DB. The values were assigned based on the following model:

- Evolution 1 - 170 catch_rate
- Evolution 2 - 85 catch_rate
- Evolution 3 - 45 catch_rate
- Super Rare - 3 catch_rate

There were also many Tuxemon that there evolution or rarity could not be estimated. To those Tuxemon the value 120 catch_rate was given. The model used to distribute catch_rate values to the Tuxemon is not final and should go under further investigation, but its introduction, even in this form should be a motivation for the community and will give a better feel of the game.


xirsoi:
Before I begin my review I just want to comment on the drastic reduction in the influence of the capture device.
I just want to point out that it's likely more types of capture devices will be introduced that will have a greater odds of capture, our own versions of the Great and Ultra Ball. I hope your updates can accommodate that.

tuxemon/core/monster.py
```diff
@@ -229,8 +229,13 @@ def __init__(self, save_data=None):
 
         self.weight = 0
 
-        # the multiplier for checks when a monster ball is thrown.
-        self.catch_rate = 1
+        # The multiplier for checks when a monster ball is thrown this should be a value betwen 0-255 meaning that
+        # 0 is 0% capture rate and 255 has a very good chance of capture. This numbers are based on the capture system calculations.
+        # This is based on the pokemon calculation and can be found at https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_by_catch_rate
+        self.catch_rate = 125
```
xirsoi: Perhaps the default value for this could be in the game config?

xirsoi (dismissed):
LGTM, thanks for the contribution!

tuxemon/core/monster.py
GDMarkou: Yes of course. If you can, send me the location of the file and I will assign the default value there

GDMarkou:
This system will be able to support more Capture devices with different powers, but it will need testing to see how different power values influence the probability of a Tuxemon being caught.

tuxemon/core/monster.py
xirsoi: /tuxemon/core/config.py

You'll need to update both the __init__ and the get_defaults()

xirsoi: I suggest putting it in the gameplay section.

Qiangong2:
Are these values hard set? Or are they default and we can potentially edit them in an encounter json?

GDMarkou:
@Qiangong2 I created a small script that reads json files and inputs a certain value to each selected json file. The way I categorized the Tuxemon to categories was through the tuxepedia, because I couldn't find some other variable to categorize them by, for example by evolution. So yes this values are hard set, but changing a particular json shouldn't be a huge issue. If it helps, I can also include the code that I used to write into the json files.

tuxemon/core/monster.py
xirsoi: Still wanna see this go to the config.

GDMarkou: I made the changes to config.py

xirsoi: Great, but you also need to make the code *use* that config value :P

xirsoi (changes requested):

tuxemon/core/monster.py
GDMarkou: Yeah, sorry my bad XD.  Is <code>self.catch_rate = TuxemonConfig().default_monster_catch_rate</code> ok? If there is another better way to do it please inform me. Also do you want me to also move the comments to the config or leave them there?

GDMarkou:
I'm having some trouble with the merge conflict and I created some issues in the PR. I will try to resolve them and when I do I will ask for a new review. 

edit: I solved the issues

tuxemon/core/monster.py
GDMarkou: @xirsoi I have implemented the change.

tuxemon/monster.py
```diff
@@ -297,6 +303,8 @@ def load_from_db(self, slug):
                 self.type2 = results["types"][1].lower()
 
         self.weight = results["weight"]
+        self.catch_rate = results["catch_rate"]
```
xirsoi: The correct way to handle this would be to use 
results.get("catch_rate", TuxemonConfig().default_monster_catch_rate)

This ensures that we get the default when a record is missing the catch_rate instead of just crashing.

xirsoi (changes requested):
This is super close!
That one little change is all that's needed.

Thanks for the contribution!

tuxemon/monster.py
GDMarkou: I have implemented the change you proposed. Thank you for pointing it out!

xirsoi (approved):
Looks good, thanks for the PR!

bitcraft (approved):

PR #879, opened 2021-05-18
Fix locale loading.

### LINKED ISSUE #878: Locales not working
It seems that locales do not work after commit 00b94d97798c45214f9531126c963a19fbeba14c.

The condition for selecting a particular locale is never true, as the `domain` parameter is compared against `info.locale` instead of `info.domain`:

https://github.com/Tuxemon/Tuxemon/blob/fccad7a593f91e03ed919abdacfeb308a2facd73/tuxemon/locale.py#L112

vnmabus:
Fixes #878.


bitcraft (approved):
looks good, thank you

PR #885, opened 2021-05-23
Create modify_json.py

GDMarkou:
This script allows the user to add or change a certain value to all json files included in a folder. This script was used to introduce the catch_rate attribute to the DB of Tuxemon and has been modified to support any new introduction of a new variable, or modification of an already existing one with ease.


bitcraft (approved):
looks great!  thanks for including the usage at the top

PR #887, opened 2021-05-28
Cleanup

bitcraft:
There is a new file in scripts called "fmt.sh".  I'm testing it now and will make it part of a CI/CD thing at some point.  I've run it over the code base and made a few small fixes and cleanups.


xirsoi (approved):
Looks great!
I love string interpolation 😄

bitcraft:
im a huge fan of string interpolation, too.  it looks so clean.  there is a pyflakes migration that will update them automatically.  really handy.  thanks for the review!  @xirsoi

Issue #888, opened 2021-05-28
Doc strings and type hinting


bitcraft:
If you are looking to contribute, but don't know where to start, this issue is for you!

The tuxemon code base started before type hinting was a thing, and the docstrings are a bit verbose in places and can be made redundant with type hinting.

We should make an effort for add type hints where possible and remove unneeded or incorrect docstrings.

PR #892, opened 2021-06-11
Refactor the documentation.

vnmabus:
- Regenerate the autogenerated documentation each time calling
sphinx-apidoc so it is always up to date.
- Use Napoleon extension so that Sphinx understand Google style
docstrings.
- Set option to the doc config to add the type information in the
description of each parameter/return instead of polluting the signature.
- Add types in main just to see that everything works.


vnmabus:
This is the first step towards #888 and ##772.

bitcraft (approved):
Thank you for cleaning up the docs

Issue #801, opened 2020-09-08
Catch Resistance (monster species property)


Sanglorian:
I have added a new property to tuxemon species on the wiki: "Catch Resistance". 

The idea is that Catch Resistance, if it exists, is a multiplier for the Catch Likelihood of a monster. If it is set to 1 (or not set at all), then the monster has an unmodified Catch Likelihood. A monster with a Catch Resistance of 0.5 has its Catch Likelihood halved, and a monster with a Catch Resistance of 0 cannot be caught.

This serves two functions: we can introduce monsters that cannot be caught at all, and we can have some monsters (e.g. rare or unique monsters) that are possible to catch but more difficult than normal. 

[Forum discussion](https://forum.tuxemon.org/thread.php?pid=2437#p2437)

xirsoi:
There's a property for catch_rate already in the Monster class. It's even used in catch calculations. It's just that currently no monsters use the property. It works exactly as you describe.

GDMarkou:
Hello I'd like to work on that issue.

xirsoi:
Go for it, if you have questions you can ask here or in the discord.

xirsoi:
Make sure to support values above 1 as well. It should be possible to have some monsters that are *easier* to catch, like the commons found in starting areas, to give the player a leg up in the beginning.

GDMarkou:
I was thinking of giving monsters in starting zones catch_rate = 1 so you can catch them 100% of time. Does it make any difference if the catch_rate is >1 ?

edit: I read a little more on the forums and I think I understand what you meant. Also I wanted to ask, where are the catch calculations being executed?

bitcraft:
https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/item/effects/capture.py

GDMarkou:
[...]

PR #895, opened 2021-06-24
Introduce Catch resistance mechanic

GDMarkou:
Fix issue (#801)

This PR aims to:
*  Introduce catch resistance mechanic for the capture system
*  Add variety to the capture system, making it more interesting and complex

### Catch_resistance 
The concept behind the catch_resistance as discussed in issue( #801 ), is to have a multiplier for the catch probability of a Tuxemon, in order to increase or decrease the chances of a Tuxemon being caught. Since the catch_rate was introduced recently, I propose that the catch resistance should not be a single value for every Tuxemon, but rather a randomly generated number within a defined span. The span is defined by the upper_catch_resistance and lower_catch resistance that have been assigned to each Tuxemon following this model:
- Evolution 1 - lower_catch resistance = 0.95 , upper_catch_resistance = 1.25
- Evolution 2 - lower_catch resistance = 0.9 , upper_catch_resistance = 1.15
- Evolution 3 - lower_catch resistance = 0.85 , upper_catch_resistance = 1.1
- Super Rare - lower_catch resistance = 0.8 , upper_catch_resistance = 1.05

For Tuxemon that their evolution/rarity was not defined, the values 0.95 and 1.05 were given.
This way for example a evolution 1 type Tuxemon will have more chances of being caught giving the players a head start in the game. 

### Calculation

The catch_resistance is calculated every time a capture device is used, and is a randomly generated number in the span of lower_catch resistance, upper_catch_resistance. This value is then multiplied by the already predefined probability to give us the final rate.
  
## Note

For the introduction of the new values in the Tuxemon DB the script modify_json.py was used.

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -21,5 +21,7 @@
         "Earth"
     ],
     "weight": 25,
-    "catch_rate": 85.0
+    "catch_rate": 85.0,
+    "lower_catch_resistance": 0.9,
+    "upper_catch_resistance": 1.15
 }
--- a/mods/tuxemon/db/monster/aardorn.json
+++ b/mods/tuxemon/db/monster/aardorn.json
@@ -21,5 +21,7 @@
         "Earth"
     ],
     "weight": 25,
-    "catch_rate": 170.0
+    "catch_rate": 170.0,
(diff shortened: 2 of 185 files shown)


bitcraft (approved):
looks great, thank you!

PR #898, opened 2021-06-29
Typing and docs for Client, State and StateManager

vnmabus:
Adding types and improved the docs for the aforementioned classes.

Part of  #888 and ##772.

### DIFF
--- a/tuxemon/client.py
+++ b/tuxemon/client.py
@@ -88,7 +88,6 @@ class Client(StateManager):
         self._state_dict = dict()
         self._state_stack = list()
         self._state_resume_set = set()
-        self._remove_queue = list()
 
         keyboard = PygameKeyboardInput(config.keyboard_button_map)
         gamepad = PygameGamepadInput(config.gamepad_button_map, config.gamepad_deadzone)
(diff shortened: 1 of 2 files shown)


tuxemon/client.py
```diff
@@ -181,24 +189,33 @@ def draw_event_debug(self):
...
 
-        States can "keep" events by simply returning None from State.process_event
+        States can "keep" events by simply returning None from
+        State.process_event
 
-        :param events: Sequence of pg events
-        :returns: Iterator of game events
-        :rtype: collections.Iterable[pg.event.Event]
+        Parameters:
+            events: Sequence of Pygame events]
```
bitcraft: trailing `]`?

bitcraft (dismissed):
Looks good, but there may be a typo

tuxemon/client.py
vnmabus: In fact, I have a question here: should the type of the events be `PlayerInput` instead? It seems that is the case in `State`.

bitcraft: Yeah, good point.  We are no longer using pygame events in the states.  The player events abstract that away so they do not need to know if it was a keyboard, joystick, or anything different.

bitcraft (approved):
Looks good

PR #899, opened 2021-06-30
Typing animation

vnmabus:
Add typing and docs for the animation module.

### DIFF
--- a/tuxemon/animation.py
+++ b/tuxemon/animation.py
@@ -121,6 +121,11 @@ class Task(TaskBase):
     below use milliseconds.  However, you are free to use whatever
     time unit you wish, as long as it is used consistently.
 
+    Parameters:
+        callback: Function to execute each interval.
+        interval: Time between callbacks.
+        times: Number of intervals.
+
     Examples:
         >>> task_group = pygame.sprite.Group()
 
@@ -159,15 +164,7 @@ class Task(TaskBase):
         interval: float = 0,
         times: int = 1,
     ) -> None:
-        """
-        Task constructor.
 
-        Parameters:
-            callback: Function to execute each interval.
-            interval: Time between callbacks.
-            times: Number of intervals.
-
-        """
         if not callable(callback):
             raise ValueError
 
@@ -354,6 +351,22 @@ class Animation(pygame.sprite.Sprite):
     The 'round_values' parameter will be set to True automatically
     if pygame rects are used as an animation target.
 
+    Parameters:
+        targets: Any valid python objects.
+        delay: Delay time before the animation starts.
+        round_values: Wether the values must be rounded to the nearest
+            integer before being set.
+        duration: Time duration of the animation.
+        transition: Transition to use in the animation. Can be the name
+            of a method of :class:`AnimationTransition` or a callable
+            with the same signature.
+        initial: Initial value. Can be numeric or a callable that
+            returns a numeric value. If ``None`` the value itself is used.
+        relative: If the values are relative to the initial value. That is,
+            in order to find the actual value one has to add the initial
+            one.
+        kwargs: Properties of the ``targets`` to be used, and their values.
+
     """
 
     default_duration = 1000.0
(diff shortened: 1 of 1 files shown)


tuxemon/animation.py
```diff
@@ -16,54 +21,68 @@
...
-        """Schedule a callback during operation of Task or Animation
+        self._callbacks: DefaultDict[
+            str,
+            List[ScheduledFunction],
+        ] = defaultdict(list)
+
+    def schedule(
+        self,
+        func: ScheduledFunction,
+        when: Optional[str] = None,
```
bitcraft, line 68: Im not sure how this got to be a string, but probably should be `Union[int, float]` or something like that.  For tasks, the time unit is technically unitless; determining to execute or not is a comparison between two objects, but i think for clarity, an int or float is more likely to be used.  Again, nothing is enforced here, because the class doesn't matter.... `Any` would probably be most accurate, even if confusing.

tuxemon/animation.py
```diff
@@ -1,13 +1,18 @@
...
 
 import pygame
+from typing import Any, Optional, Callable, Sequence, DefaultDict, List, Tuple,\
+    Mapping, Union
 
 __all__ = ("Task", "Animation", "remove_animations_of")
 
 from tuxemon.compat import Rect
 
+ScheduledFunction = Callable[[], Any]
```
bitcraft, line 14: Why is `Any` here?  Im thinking a non-callable object would cause an exception when the callback are executed.

tuxemon/animation.py
```diff
@@ -74,7 +93,7 @@ def schedule(self, func, when=None):
             raise ValueError
         self._callbacks[when].append(func)
 
-    def _execute_callbacks(self, when):
+    def _execute_callbacks(self, when: str) -> None:
```
bitcraft, line 96: See other comment about the time unit's type.

tuxemon/animation.py
```diff
@@ -440,36 +535,37 @@ def finish(self):
...
-
-        Update callback will not be executed
-        Final callback will be executed
-        Values will not change
-        Animation will be removed from group
+    def abort(self) -> None:
+        """Force animation to finish, without any cleanup.
 
-        Will raise RuntimeError if animation has not been started
+        Update callback will not be executed.
```
bitcraft: minor: this would be better formatted as a list, imo

tuxemon/animation.py
```diff
@@ -413,18 +512,14 @@ def update(self, dt):
...
-    def finish(self):
-        """Force animation to finish, apply transforms, and execute callbacks
+    def finish(self) -> None:
+        """Force animation to finish, apply transforms, and execute callbacks.
 
-        Update callback will be called because the value is changed
-        Final callback ('callback') will be called
-        Final values will be applied
-        Animation will be removed from group
+        Update callback will be called because the value is changed.
```
bitcraft: minor: this would be better formatted as a list, imo

bitcraft (dismissed):
Looks good as-is, but a few comments are there if you want to address them

tuxemon/animation.py
vnmabus: It is a callable that can return anything. I could enforce that the return type is ``None`` if you prefer.

tuxemon/animation.py
vnmabus: `when`  can be `"on interval", "on finish", "on abort"`.

vnmabus: Maybe I should say "event" instead of "time" but I also see the confusion potential with that.

tuxemon/animation.py
bitcraft: i might just be unfamiliar with the syntax.  if this is just how to type a generic callable and its return values, then thats fine with me.

tuxemon/animation.py
bitcraft: maybe just change it to `Any` since that is the most correct.  any confusing part of it could be clarified in a comment if needed.

vnmabus: But it cannot be a number. Just a string.

bitcraft: it can be any hashable object.  the mechanism relies on a dictionary.  
```
callbacks = self._callbacks[when]
```
Is `hashable` a type?

bitcraft: numbers are hashable and would work with this object
```
>>> d[0] = None
>>> d[1.0] = None
>>> d.keys()
[0, 1.0]
```

vnmabus: Yes, it is a type.

However, `when` is a member of `self._valid_schedules` and the elements of that are always strings.

bitcraft: honestly, im not really sure what to do.  `str` is fine.  lets just leave it, sorry for the confusion.

bitcraft (approved):

PR #905, opened 2021-07-01
yaml event loader

bitcraft:
Optional support to put events into an external file.  YAML for now, may change later.  Only one map is changed currently, the bedroom, as a demonstration.  More maps can be converted later.


xirsoi (approved):
Hot damn, this looks good!
I'm excited for the feature and i'm getting all the good feels from the slickness of the implementation!

PR #910, opened 2021-07-02
Refactor actions parameters

vnmabus:
Refactor the actions.

Improvements:
* The parameters are now specified creating a NamedTuple, instead of using a list as a class argument.
* The casting of the arguments is now done extracting the information from the NamedTuple.
* Mypy can now typecheck the parameters, as the NamedTuple is known at import time.
* The class `EventAction` is now abstract, as well as the `start` method and the `param_class` attribute. If one of those is not defined in a `final` subclass, Mypy would warn about it (otherwise it will fail at runtime).

Caveats:
* Uses `Protocol`, `get_origin` and `get_args`, that are only available in Python 3.8 or above. I could make it to work in previous versions using the module `typing_extensions` if that is necessary.
* The namedtuple used for the parameters has to be passed twice to the action, one as a generic parameter (for typechecking) and one as a class attribute (for constructing it). This is because currently there is no portable way to recover the class from the typing information (see https://github.com/python/typing/issues/629).
* The plugin subsystem logs warnings like `ERROR - found incomplete actions: AddItemActionParameters`. This is because it assumes that the only classes available in the module are the intended ones. This limitation is easy to fix (just pass the intended superclass to the loading routine, and check the classes against it. It also will allow to typecheck the return type), and will be addressed later.

### DIFF
--- a/tuxemon/event/actions/add_item.py
+++ b/tuxemon/event/actions/add_item.py
@@ -36,7 +36,6 @@ class AddItemAction(EventAction):
     """
 
     name = "add_item"
-    valid_parameters = [(str, "item_slug"), ((int, None), "quantity")]
     _param_factory = AddItemActionParameters
 
     def start(self):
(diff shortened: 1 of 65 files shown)


bitcraft (approved):
thank you for these changes!

PR #912, opened 2021-07-04
Tilesheet consolidation

Sanglorian:
I've created three "core" tilesheets that consolidate a bunch of overlapping tilesheets. I've also deleted a bunch of tilesheets that I didn't think were in use.


bitcraft:
@Qiangong2 could i get your opinion on these changes?

bitcraft (approved):
thank you!  this will help me develop the feature to read these tiles with exits

PR #913, opened 2021-07-04
Fix a typo in setup_cx_freeze.py

magmaus3:
Accidentally used a period instead of a comma


xirsoi (approved):
Thanks for fixing this!

PR #914, opened 2021-07-08
Typing plugins

vnmabus:
Typing the plugins module.

It adds a new optional parameter `interface` to the `plugin.load_plugins` function that serves two purposes:
* It restricts the loaded classes to those that conform to a particular superclass or runtime checkable protocol. Thus, auxiliary classes, such as action parameters, can be used in plugin files without errors.
* It tells Mypy that the returned classes conform to that interface, so the code that uses them can typecheck.

In the road several limitations of Mypy have been found. Some have been worked around and the others will (hopefully) be solved by the Mypy/Python team in the future.


bitcraft (approved):
Looks good, ty

PR #916, opened 2021-07-13
Add more (user friendly) commands to the cli

magmaus3:
Currently, the cli mostly relies on the python interactive console. This pr adds easier and more convenient to use commands, like add_monster or random_encounter.


bitcraft (approved):
Thanks for these great features!

PR #921, opened 2021-07-22
Typing rect.

vnmabus:
Provides a Protocol that can be used to type functions that accept either a Pygame Rect or a compat Rect. The protocol is minimalistic, and only accepts a few methods, read only attributes and one-parameter constructors (the things that were implemented in compat Rect).

Most of the implementation of the compat Rect has been moved to the protocol in order to reduce the code.

The Pygame Rect class is compatible with the protocol **in the Pygame development branch**. In the current released version of Pygame some methods do not have the right (bool) return type.

I have tested that the game still works BUT I have not tested only with the compat Rect, as I am not sure of which parts do not require Pygame at all. I guess that in the future, either we will have to expand the protocol to cover every usecase or we will need to explicitly import the Pygame Rect in parts that require it, or otherwise they will not typecheck.

By the way, Mypy does not currently typecheck the distinction in the constructor of compact Rect of the tuples based on their length. I hope this will change with https://github.com/python/mypy/pull/10367, so I did not bother to cast or anything like that.

### DIFF
--- a/tuxemon/compat/__init__.py
+++ b/tuxemon/compat/__init__.py
@@ -1,4 +1,9 @@
+from tuxemon.compat.rect import ReadOnlyRect
+from typing import Type
+
+Rect: Type[ReadOnlyRect]
+
 try:
-    from pygame.sprite import Rect
+    from pygame.rect import Rect
 except ImportError:
     from tuxemon.compat.rect import Rect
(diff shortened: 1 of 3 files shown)


bitcraft (approved):
Thanks for these changes!

PR #922, opened 2021-07-26
Typing tools and prepare modules.

vnmabus:
Typing the tools and prepare modules.

The functions `new_scaled_rect `, `check_parameters` and `scale_rect` were not necessary and so they are removed.


bitcraft (approved):
Looks good thank you!

PR #927, opened 2021-08-05
Allow the translation of the input alphabet.

vnmabus:
The "\0" character represent an empty space, often necessary to align letters (for example in German, where the uppercase ß is not available).

Example alphabets are provided for English, Spanish, German and Esperanto.


bitcraft (approved):
great feature.  i like the use of the translation files for the alphabet.  ty!

PR #957, opened 2021-08-27
Flying Type and Overfeed Move/Effect

SandytheLizard:
Adds a flying type, nothing has the type yet, its just a placeholder.
Adds the move overfeed that heals the hp of the target to full but halves their speed.

### DIFF
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -49,10 +49,9 @@ TYPES = {
     "normal": TypeChart(None, None, None, None),
     "wood": TypeChart("earth", "fire", "metal", "water"),
     "fire": TypeChart("metal", "earth", "water", "wood"),
-    "earth": TypeChart("water", "metal", "flying", "fire"),
+    "earth": TypeChart("water", "metal", "wood", "fire"),
     "metal": TypeChart("wood", "water", "fire", "earth"),
     "water": TypeChart("fire", "wood", "earth", "metal"),
-    "flying": TypeChart("earth", "metal", "metal", "wood"),
 }
 
 
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -49,9 +49,10 @@ TYPES = {
     "normal": TypeChart(None, None, None, None),
     "wood": TypeChart("earth", "fire", "metal", "water"),
     "fire": TypeChart("metal", "earth", "water", "wood"),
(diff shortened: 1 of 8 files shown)


tuxemon/technique.py
```diff
@@ -395,7 +410,18 @@ def lifeleech(self, target: Monster) -> EffectResult:
...
+    def overfeed(self, target: Monster) -> EffectResult:
+        user = self.link
+        assert user
+        target.current_hp = target.hp
+        target.speed = formula.simple_overfeed(self, user, target)
+            
+                
+        return {
+            "damage": damage,
+            "should_tackle": bool(damage),
```
xirsoi: Because damage will always be 0 in this case, the technique will always report as failed.

tuxemon/technique.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         link: Optional[Monster] = None,
     ) -> None:
         self._combat_counter = 0  # number of turns that this technique has been active
+        self._life_counter = 0
```
xirsoi, line 87: I don't see this being used anywhere.

tuxemon/formula.py
```diff
@@ -49,9 +49,10 @@ class TypeChart(NamedTuple):
     "normal": TypeChart(None, None, None, None),
     "wood": TypeChart("earth", "fire", "metal", "water"),
     "fire": TypeChart("metal", "earth", "water", "wood"),
-    "earth": TypeChart("water", "metal", "wood", "fire"),
+    "earth": TypeChart("water", "metal", "flying", "fire"),
```
xirsoi: Lets not add a sixth type without further community discussion.

xirsoi (changes requested):
Good work so far, but needs a couple changes.

tuxemon/technique.py
SandytheLizard: the game was crashing saying that it wasn't defined, so i added that and it fixed it.

tuxemon/formula.py
SandytheLizard: ok

tuxemon/technique.py
SandytheLizard: fixed

SandytheLizard (commented):
Fixed what you requested.

xirsoi (approved):

PR #960, opened 2021-08-28
Cleaning up missing tilesets in Spyder

Sanglorian:
There were a few tilesets we cleared out in the latest update.  This fixes up the maps from the Spyder campaign that were missing tiles or had broken tilesheet connections.


Sanglorian:
Have updated this PR with a couple of other map fixes, as identified in the Element chat.

bitcraft (dismissed):

bitcraft:
lgtm, but there are merge conflicts

Sanglorian:
Didn't mean to dismiss your review, but I think this should have fixed that conflict. Have re-opened the edited map in Tiled and it works.

bitcraft (approved):

PR #966, opened 2021-08-29
Wip code for allowing stats

SandytheLizard:
Allows for stats in JSON technique files

### DIFF
--- a/tuxemon/technique.py
+++ b/tuxemon/technique.py
@@ -45,7 +45,6 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-
 class EffectResult(TypedDict, total=False):
     damage: int
     element_multiplier: float
@@ -76,7 +75,6 @@ class Technique:
     """A technique object is a particular skill that tuxemon monsters can use
     in battle.
     """
-
     def __init__(
         self,
         slug: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


tuxemon/technique.py
```diff
@@ -239,9 +254,21 @@ def use(self, user: Monster, target: Monster) -> TechniqueResult:
             "capture": False,
             "statuses": [],
         }
-
         # TODO: handle conflicting values from multiple technique actions
         # TODO: for example, how to handle one saying success, and another not?
+        statsmaster = [self.statspeed, self.stathp, self.statarmour, self.statmelee, self.statranged, self.statdodge]
+        statslugs = ['speed', 'hp', 'armour', 'melee', 'ranged', 'dodge']
+        print(statsmaster)
+        if None not in statsmaster:
```
vnmabus, line 262: This is always true.

tuxemon/technique.py
```diff
@@ -239,9 +254,21 @@ def use(self, user: Monster, target: Monster) -> TechniqueResult:
...
+            for stat in statsmaster:
+                for paramater in range(len(stat)):
+                    print(stat)
+                    value = stat['value']
+                    dividing = stat['dividing']
+                    override = stat['overridetofull']
+                    if value > 0 and override == False:
+                        statvalue = getattr(target, statslugs[paramater])
+                        print(statvalue)
```
vnmabus, line 271: I am not sure what do you want to do here, but probably you want to loop for similar things together (e.g. `self.statarmour` and `"armour"`), right?
```suggestion
            for stat, slug in zip(statsmaster, statslugs):
                print(stat)
                value = stat['value']
                dividing = stat['dividing']
                override = stat['overridetofull']
                if value > 0 and override == False:
                    statvalue = getattr(target, slug)
                    print(statvalue)
```

PR #972, opened 2021-08-31
Improve typing of get_state_by_name.

vnmabus:
Now you can pass the class instead of the class name. In that case Mypy recognizes that the returned type is an instance of that class.


tuxemon/event/actions/spawn_monster.py
```diff
@@ -58,15 +60,15 @@ class SpawnMonsterAction(EventAction[SpawnMonsterActionParameters]):
...
         if not world:
-            return False
+            return
 
         npc_slug = npc_slug.replace("player", "npc_red")
         trainer = world.get_entity(npc_slug)
         if trainer is None:
             logger.error(f"Could not find NPC corresponding to slug {npc_slug}")
-            return False
+            return
```
xirsoi, line 71: Wait, I thought returning true or false was how we knew if the event action failed?

vnmabus: I didn't see the return value of that method used anywhere. Do you know where is it checked?

bitcraft: no i don't think it was ever like that.  scripts don't have flow control, and every action is executed in order until the end, regardless of the status of another.

bitcraft (approved):

PR #973, opened 2021-08-31
Remove duplicate play music event (spyder_leather_town.tmx)

### LINKED ISSUE #971: Spyder Campagin Leather Town: Error playing music
Currently, when visiting the spyder_leather_town.tmx map, there is an issue where the music does not play properly. It sounds like drilling or buzzing, almost as if the first note or two are skipping and looping over and over.

There does not appear to be anything unusual about the event which starts the music, but the terminal is filled with repeated instances of the following errors:

```
tuxemon.event.actions.play_music - ERROR - 'JRPG_town_loop.ogg'
tuxemon.event.actions.play_music ERROR - unable to play music
```
This appears to indicate that the except clause in tuxemon.event.actions.play_music is being tripped:

```
except Exception as e:
            logger.error(e)
            logger.error("unable to play music")
```

The file name itself appears to match that in the mods/music/ directory (JRPG_town_loop.ogg) and I can confirm that this file appears to play correctly when played directly on my machine, as well as in other Tuxemon maps.

One possibility is that there are duplicate play music events on the same map, which are causing interference (as the condition for playing the music is that that particular song is not playing already).

freshreplicant:
Fixes #971. 

The errors were most likely caused by the `duplicate tuxemon.event.play_music` events, plus their associated conditions (not `music_playing music_town_theme`).

Music now appears to play normally.


bitcraft (approved):
thank you for this fix!

PR #975, opened 2021-09-01
Split state manager and game client, tests

bitcraft:
Now made with 100% REAL TESTS!


bitcraft:
@vnmabus please review

tuxemon/main.py
```diff
@@ -55,10 +55,9 @@ def main(
     config = prepare.CONFIG
 
     import pygame
-    from tuxemon.client import Client
+    from tuxemon.client import LocalPygameClient
 
-    client = Client(config.window_caption)
-    client.auto_state_discovery()
```
vnmabus, line 61: Why is this no longer necessary?

tuxemon/state.py
```diff
@@ -60,13 +72,6 @@ class should be created. Update must be overloaded in the child class.
        pause         - Called when state is no longer active
        shutdown      - Called before state is destroyed
 
-    Parameters:
```
vnmabus, line 63: I thought we preferred to document constructor parameters in the class definition instead of documenting the `__init__` method.

tuxemon/state.py
```diff
@@ -265,27 +280,32 @@ def shutdown(self) -> None:
 
 
 class StateManager:
-    """Mix-in style class for use with Client class.
+    """Allows game states to be managed like a queue"""
 
-    This is currently undergoing a refactor of sorts, API may not be stable.
+    def __init__(
+        self, package: str, on_state_change: Optional[callable] = None
```
vnmabus: I usually put either everything in one line or one parameter per line. I started doing that following Wemake-python-styleguide and I found it way more legible.

tuxemon/state.py
```diff
@@ -354,6 +390,34 @@ def collect_states_from_path(
             logger.error(template.format(folder))
             raise
 
+    def update(self, *args, **kwargs) -> None:
```
vnmabus: I would copy the `State` signature for typechecking. Unfortunately, the pattern of accepting args and kwargs and delegate to other functions does not have a standard way to be typed effectively.

tuxemon/main.py
bitcraft: moved to state manager

tuxemon/state.py
bitcraft: sorry, i don't recall that, but i see other classes have them documented there, so i will make that change

tuxemon/state.py
bitcraft: i agree, but im using black to format which will only do that if the line is too long.  i will do one-parameter-per-line, but black tends to not follow that convention.

tuxemon/state.py
bitcraft: thats fair.  a bit cumbersome, but fine.

tuxemon/main.py
vnmabus: Ok

PR #978, opened 2021-09-01
Status Moves V2

SandytheLizard:
1. A stat that does not need modified does not need to be in the JSON file
2. Instead of divide, it accepts an iterator of +, -, /, *, =
3. you may specify a maxvalue,if you do it randomly picks between the minvalue and maxvalue. value has been replaced by minvalue


tuxemon/db.py
```diff
@@ -106,9 +106,10 @@ class JSONMonster(JSONMonsterOptionalFields):
     weight: float
     sprites: JSONMonsterSprites
 
-class JSONStat(TypedDict):
-    value: int 
-    dividing: bool
+class JSONStat(TypedDict, total=False):
+    minvalue: int
+    maxvalue: int
+    iterator: str
```
vnmabus: Why `iterator` instead of `operation`?

tuxemon/db.py
```diff
@@ -106,9 +106,10 @@ class JSONMonster(JSONMonsterOptionalFields):
     weight: float
     sprites: JSONMonsterSprites
 
-class JSONStat(TypedDict):
-    value: int 
-    dividing: bool
+class JSONStat(TypedDict, total=False):
+    minvalue: int
```
vnmabus: I would prefer `value` and `max_deviation` instead, as it is more clear in the fixed damage case, and it can be generalized to the normal distribution it necessary.

mods/tuxemon/db/technique/amnesia.json
```diff
@@ -3,21 +3,33 @@
...
+    "statchange"
   ],
   "healing_power": 0,
   "icon": "",
   "is_area": false,
   "is_fast": false,
   "potency": 1,
   "power": 2.25,
+  "statdodge": {
+    "minvalue": 0, "maxvalue": 3, "iterator":"-", "overridetofull":false
```
vnmabus: Remove `iterator` and `overridetofull` when they have their default values, as you already made them optional. It makes the JSON easier to read.

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -171,8 +171,13 @@ msgstr "{user} is leeching health from {target}."
 
 msgid "combat_state_overfeed_get"
 
-msgstr "{user} Overfed {target}. {target}'s speed was halved, {target}'s hp was fully restored"
+msgstr "{user} Overfed {target}. {target} 's speed was halved, {target}' s hp was fully restored"
 
+msgid "combat_amnesia_get"
+msgstr "{user} 's dodge fell. \n {user} 's attack damage types rose."
```
vnmabus: Maybe the "attribute x increased/decreased/restored" could be separate strings, so that we only have to translate them once.

tuxemon/technique.py
```diff
@@ -297,25 +297,38 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
         '''
         statsmaster = [self.statspeed, self.stathp, self.statarmour, self.statmelee, self.statranged, self.statdodge]
         statslugs = ['speed', 'hp', 'armour', 'melee', 'ranged', 'dodge']
+        newstatvalue = 0
         for stat, slug in zip(statsmaster, statslugs):
             if not stat:
                 continue
-            value = stat['value']
-            dividing = stat['dividing']
+            minvalue = stat['minvalue']
```
vnmabus: As the values of the stat are now optional (because `total=False`), you need to use the `get` method with a default value in case they don't exist. If you configure Mypy in your IDE it should flag this as an error.

tuxemon/technique.py
```diff
@@ -297,25 +297,38 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+                    setattr(target, slug, newstatvalue)
+                elif iterator == '+':
+                    newstatvalue = basestatvalue + value
+                    setattr(target, slug, newstatvalue)
+                elif iterator == '-':
+                    newstatvalue = basestatvalue - value
+                    setattr(target, slug, newstatvalue)
+                elif iterator == '*':
+                    newstatvalue = basestatvalue * value
+                    setattr(target, slug, newstatvalue)
```
vnmabus: This could be done in a more compact way:
```python
ops_dict = {
    "+": operator.add,
    "-": operator.sub,
    "*": operator.mul,
    "/": operator.floordiv,
}

newsatvalue = ops_dict[iterator](basestatvalue, value)
setattr(target, slug, newstatvalue)
```

tuxemon/technique.py
```diff
@@ -297,25 +297,38 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+            maxvalue = stat['maxvalue']
+            iterator = stat['iterator']
             override = stat['overridetofull']
             basestatvalue = getattr(target, slug)
-            if value > 0 and override == False:
-                if dividing == False:
-                    newstatvalue = basestatvalue + value
-                else:
+            if maxvalue:
+                value = random.randint(minvalue,maxvalue)
```
vnmabus: Using random directly makes everything non deterministic. Maybe we should discuss having something similar to numpy `RandomState` objects, in order to be able to reproduce combats, for networking, debugging or rewatching purposes.

tuxemon/technique.py
```diff
@@ -297,25 +297,38 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+                    newstatvalue = basestatvalue * value
+                    setattr(target, slug, newstatvalue)
             if override == True and slug == 'hp':
-                success = True
                 target.current_hp = target.hp
+                newstatvalue = target.current_hp
+                setattr(target, slug, newstatvalue)
         return {
-            "success": bool(stat)
+            "success": bool(newstatvalue)
```
vnmabus, line 334: I do not understand this success value.

mods/tuxemon/db/technique/amnesia.json
SandytheLizard: Removed override to full, Iterator does not have default value.

tuxemon/db.py
SandytheLizard: fixed

tuxemon/db.py
SandytheLizard: done

tuxemon/technique.py
SandytheLizard: going to into a later PR

tuxemon/technique.py
SandytheLizard: fixed

tuxemon/technique.py
SandytheLizard: i did this because the prorgam returns that an attack failed, even if it hit,  when even one stat is 0. By putting newstatvalue=0 right before the loop, and newstatvalue in the loop once a stat is above 0, it checks for at least one stat being successful rather than all of them.

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
SandytheLizard: im not sure i did it right, please take a look

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,30 @@ msgstr "{name} took poison damage!"
 msgid "combat_state_lifeleech_get"
 msgstr "{user} is leeching health from {target}."
 
+msgid "dodge_increased"
+msgstr {user} 's dodge increasedd"
```
vnmabus: I hope the messages can be made generic:
```suggestion
msgstr {user} 's {stat} increased"
```

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,30 @@ msgstr "{name} took poison damage!"
...
+
+msgid "speed_halved"
+msgstr {user} 's speed was halved"
+
+msgid "hp_restored"
+msgstr {user} 's hp was restored"
+
 msgid "combat_state_overfeed_get"
+msgstr "{user} Overfed {target}" + speed_halved + hp_restored
```
vnmabus: Is this syntax even legal?

tuxemon/technique.py
```diff
@@ -297,25 +297,32 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
-                else:
-                    newstatvalue = basestatvalue // value
-                success = True
+            if maxvalue:
+                value = random.randint(minvalue,maxvalue)
+            else:
+                value = minvalue
+            if value > 0 and  override == False:
+                ops_dict = {
+                "+": operator.add,
```
vnmabus, line 318: You need to import the operator module.

tuxemon/technique.py
```diff
@@ -297,25 +297,32 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+                "-": operator.sub,
+                "*": operator.mul,
+                "/": operator.floordiv,
+            }
```
vnmabus: Bad identation.

tuxemon/technique.py
```diff
@@ -297,25 +297,32 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
         statsmaster = [self.statspeed, self.stathp, self.statarmour, self.statmelee, self.statranged, self.statdodge]
         statslugs = ['speed', 'hp', 'armour', 'melee', 'ranged', 'dodge']
+        newstatvalue = 0
         for stat, slug in zip(statsmaster, statslugs):
             if not stat:
                 continue
-            value = stat['value']
-            dividing = stat['dividing']
+            minvalue = stat['minvalue']
+            maxvalue = stat['maxvalue']
```
vnmabus: You need to compute them from `value` and `max_deviation`.

tuxemon/technique.py
vnmabus: I still do not get it. What happens if an attack reduces a stat to 0?

SandytheLizard: then it fails because a stat cant be 0

vnmabus: What about hp?

tuxemon/technique.py
```diff
@@ -297,25 +297,32 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+                value = minvalue
+            if value > 0 and  override == False:
+                ops_dict = {
+                "+": operator.add,
+                "-": operator.sub,
+                "*": operator.mul,
+                "/": operator.floordiv,
+            }
+
+                newstatvalue = ops_dict[operator](basestatvalue, value)
```
vnmabus: I have just noticed that there is a name clash between the variable `operator` and the module of the same name. The variable should be `operation` instead.

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
SandytheLizard: no, i was still testing stuff and didnt mean to update that yet. its fixed now,

tuxemon/technique.py
SandytheLizard: should be fixed

tuxemon/technique.py
SandytheLizard: fixed

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
 msgid "combat_state_lifeleech_get"
 msgstr "{user} is leeching health from {target}."
 
+msgid "dodge_increased"
+msgstr {user} 's {stat} increased"
```
bitcraft: why is there a space between the name and the `'s`?

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+
+msgid "speed_halved"
+msgstr {user} 's speed was halved"
+
+msgid "hp_restored"
+msgstr {user} 's hp was restored"
```
bitcraft: missing `.` or `!`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
 msgid "combat_state_lifeleech_get"
 msgstr "{user} is leeching health from {target}."
 
+msgid "dodge_increased"
+msgstr {user} 's {stat} increased"
+
+msgid "speed_halved"
+msgstr {user} 's speed was halved"
```
bitcraft: punctuation

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+
+msgid "attacks_rose"
+msgstr {user} 's attack damage types rose."
 
-msgstr "{user} Overfed {target}. {target}'s speed was halved, {target}'s hp was fully restored"
+msgid "combat_levitate_get"
+msgstr "{user} rose higher into the air."
 
+msgid "combat_amnesia_get"
+msgstr "{user} has gotten amnesia"
```
bitcraft: punctuation

tuxemon/technique.py
```diff
@@ -214,7 +215,11 @@ def full_recharge(self) -> None:
     def reset_combat_counter(self) -> None:
         """Reset the combat counter."""
         self._combat_counter = 0
-
+    def keep_old_stats(self):
+        for player in self.active_players:
```
bitcraft: technically, there could be more than oner player, or players could leave.  maybe not now, but in the future, maybe.  this is going to cause problems.  old_stats should keep a reference to the monster.  this should be a dictionary

tuxemon/technique.py
bitcraft: yeah, current_hp is a problem.  if it is set to 0, then it is a "success".  you should check if there is a change from the old stat to the new one, then check for success.  as is, if a stat change KOs a monster, then the dialog will say it failed.

bitcraft (changes requested):

tuxemon/technique.py
```diff
@@ -297,25 +302,34 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
-                else:
-                    newstatvalue = basestatvalue // value
-                success = True
-                setattr(target, slug, newstatvalue)
-            if override == True and slug == 'hp':
-                success = True
-                target.current_hp = target.hp
+            value = stat.get('value')
+            max_deviation = stat.get('max_deviation')
+            operation = stat.get('operation')
```
vnmabus: `operation` and `overridetofull` should have default values different from `None`.

tuxemon/technique.py
```diff
@@ -297,25 +302,34 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+            override = stat.get('overridetofull')
+            basestatvalue = getattr(target, slugdata)
+            if max_deviation:
+                value = random.randint(value,max_deviation)
```
vnmabus: The `max_deviation` is the maximum deviation from the mean value.
```suggestion
                value = random.randint(value - max_deviation, value + max_deviation)
```

tuxemon/technique.py
SandytheLizard: fixed by checking if the stat is hp and if so the newstatvalue is equal to 1 which is always true

tuxemon/technique.py
vnmabus: It is easier than that: you just have to pass the default value as the second parameter of `get`.

mods/tuxemon/db/technique/amnesia.json
```diff
@@ -3,21 +3,33 @@
...
   "icon": "",
   "is_area": false,
   "is_fast": false,
   "potency": 1,
   "power": 2.25,
+  "statdodge": {
+    "minvalue": 0, "max_deviation": 3, "operation":"-"
+  },
+  "statmelee": {
+    "value": 2"operation":"+"
```
bitcraft: colon in the wrong spot

tuxemon/technique.py
```diff
@@ -297,25 +300,38 @@ def changestats(self, user: Monster, target: Monster) -> EffectResult:
...
+                
+            if value > 0 and override != True:
+                ops_dict = {
+                "+": operator.add,
+                "-": operator.sub,
+                "*": operator.mul,
+                "/": operator.floordiv,
+                }
+
+                newstatvalue = ops_dict[operation](basestatvalue, value)
```
bitcraft, line 323: need to try/except for divide by zero errors.  though unlikely, if the value is zero, and operation is 0, there will be a crash here

tuxemon/technique.py
bitcraft: what about the other stats?  why are you so sure they will never be 0?

tuxemon/technique.py
```diff
@@ -214,7 +215,9 @@ def full_recharge(self) -> None:
     def reset_combat_counter(self) -> None:
         """Reset the combat counter."""
         self._combat_counter = 0
-
+    def keep_old_stats(self):
+        self.old_stats_data.append([mon.speed, mon.hp, mon.armour, mon.melee, mon.ranged, mon.dodge])
```
bitcraft, line 220: this won't work because `mon` is not in this scope.  please playtest before asking for reviews.

bitcraft (changes requested):

mods/tuxemon/db/technique/amnesia.json
SandytheLizard: already fixed that

tuxemon/technique.py
SandytheLizard: fixed

tuxemon/technique.py
SandytheLizard: i dont think thatll happen because the condition above it says to only happen if value >0

tuxemon/technique.py
SandytheLizard: im honestly unsure why i thought that, i added a check that makes the value always at least 1 before it returns the value

tuxemon/technique.py
SandytheLizard: ok i added a line to make mon work again

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+msgstr {user}'s speed was halved."
+
+msgid "hp_restored"
+msgstr {user}'s hp was restored."
+
 msgid "combat_state_overfeed_get"
+msgstr "{user} Overfed {target}."
+
+msgid " dodge_fell"
+msgstr "{user}'s dodge fell.
```
bitcraft: missing the closing `"`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+msgstr {user}'s hp was restored."
+
 msgid "combat_state_overfeed_get"
+msgstr "{user} Overfed {target}."
+
+msgid " dodge_fell"
+msgstr "{user}'s dodge fell. 
+
+msgid "attacks_rose"
+msgstr {user}'s attack damage types rose."
```
bitcraft: missing opening `"`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
 msgid "combat_state_lifeleech_get"
 msgstr "{user} is leeching health from {target}."
 
+msgid "dodge_increased"
+msgstr {user}'s {stat} increased."
```
bitcraft: missing`"`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+
+msgid "speed_halved"
+msgstr {user}'s speed was halved."
```
bitcraft: missing `"`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+
+msgid "hp_restored"
+msgstr {user}'s hp was restored."
```
bitcraft: missing `"`

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -169,10 +169,29 @@ msgstr "{name} took poison damage!"
...
+
 msgid "combat_state_overfeed_get"
+msgstr "{user} Overfed {target}."
+
+msgid " dodge_fell"
```
bitcraft: extra space before "dodge"

tuxemon/technique.py
bitcraft: i don't understand.  where is `mon` in this function?

SandytheLizard: sorry, for whatever reason I typed the new line in but it didnt save the changes.

tuxemon/technique.py
bitcraft: ok, just calrifing then, you are aware that this will never be false?  i see that `statdodge` has a min of 0 in the JSON.  consider if the dodge value is 3 at the start, then we get 0 for the diff.  the final result will be 3, which your function will call a success.  is that what you intend to do?  ideally, you would take the difference of the old and new, so you can reliably tell if the technique did anything or not.  because as-is, it ill always register a success, even if there was no change to the stats.

bitcraft (changes requested):

tuxemon/technique.py
SandytheLizard: i think itll be fine as-is, as unless the move is coded badly and somehow affects stats but has no stat changes, theres no reason as to why the stats wouldn't change

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
bitcraft: still missing

bitcraft (changes requested):

bitcraft (approved):

PR #980, opened 2021-09-03
Use Zopfli algorithm to shrink images further

ghost:
I did it again I know there is no reason to but I did and yes this does not lower quality at all.


bitcraft:
what are the json files?
can you include a script with the command(s) you used so this is reproducible?
have you played the game to make sure they load correctly?

ghost:
> what are the json files?
> can you include a script with the command(s) you used so this is reproducible?
> have you played the game to make sure they load correctly?

darn it these json files always plague me
https://github.com/amadvance/advancecomp is the tool i used it must be compiled separately by the user
Game loads correctly

ghost:
Just adding script instead

PR #982, opened 2021-09-03
Replace references to intellectual property with Tuxemon references

freshreplicant:
This PR does three things:

1. Removes potentially problematic references to existing intellectual property and replaces it with references to similar Tuxemon.

2.  As the test is obsolete, this PR also comments out any previously uncommented lines, making it clear that this entire test is obsolete and therefore most likely non-functional (the files this test originally depended on do not appear to be present anymore). 

3. Also added a comment to the top of the file to explain this status.

### DIFF
--- a/obsolete_tests/interactive/test_fusion.py
+++ b/obsolete_tests/interactive/test_fusion.py
@@ -4,23 +4,26 @@ from fusion.py
 
 if __name__ == "__main__":
     #### EXAMPLES ####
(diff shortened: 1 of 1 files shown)


bitcraft (approved):

bitcraft:
good catch, and thank you for removing the references to other IP

freshreplicant:
Thanks for the quick review, much appreciated.

PR #983, opened 2021-09-03
Stats Reset File

SandytheLizard:


tuxemon/states/combat/combat.py
```diff
@@ -870,13 +874,22 @@ def trigger_player_run(self, player):
     def end_combat(self):
         """End the combat"""
         # TODO: End combat differently depending on winning or losing
+        old_stats_data = Technique.keep_old_stats(self)
         for player in self.active_players:
-            for mon in player.monsters:
+            for mon, i in zip(player.monsters, range(len(player.monsters))):
```
vnmabus: Use `enumerate` instead.

tuxemon/states/combat/combat.py
```diff
@@ -870,13 +874,22 @@ def trigger_player_run(self, player):
...
                 mon.end_combat()
-
-        # clear action queue
+                statslugs = ['speed', 'hp', 'armour', 'melee', 'ranged', 'dodge']
+                for stat in range(len(self.old_stats_data[i])):
```
vnmabus: Use `enumerate` here too.

tuxemon/states/combat/combat.py
```diff
@@ -163,7 +163,10 @@ def startup(self, **kwargs):
         self.show_combat_dialog()
         self.transition_phase("begin")
         self.task(partial(setattr, self, "phase", "ready"), 3)
-
+        self.old_stats_data = []
```
bitcraft, line 166: i don't think this should be here.  there is a concept of "a single source of truth", and this is spreading the "truth" around.  since techniques already have a copy of the old stats, then it doesn't need to be repeated here.  i would like to see this removed, since it is redundant.

bitcraft: another issues is active_players.  it may change in a battle if there are multiple trainers, which is supported.  so that being a list of values with no reference to monsters/trainers makes it impossible to make it correct if trainers change.

tuxemon/states/combat/combat.py
```diff
@@ -663,6 +666,7 @@ def perform_action(self, user, technique, target=None):
         """
         technique.advance_round()
 
+
```
bitcraft, line 669: please avoid changing formatting unless it is more make it conforming.  PEP8 and other standards state the code should have no more than 2 newlines between code/comments.  adding this make 3 lines of whitespace, which is not conforming.

tuxemon/states/combat/combat.py
```diff
@@ -870,13 +874,22 @@ def trigger_player_run(self, player):
...
+            for i, mon in enumerate(player.monsters):
                 mon.end_combat()
-
-        # clear action queue
+                statslugs = ['speed', 'hp', 'armour', 'melee', 'ranged', 'dodge']
+                for j, stat in enumerate(self.old_stats_data[i]):
+                    if  self.old_stats_data[i][stat] > 0:
+                            setattr(mon, statslugs[stat], self.old_stats_data[i][stat])
+                            print(getattr(mon, statslugs[stat]))
+                # clear action queue
```
bitcraft, line 886: remove this print statement

tuxemon/states/combat/combat.py
```diff
@@ -870,13 +874,22 @@ def trigger_player_run(self, player):
...
         self._action_queue = list()
-
+        oldstatsdata = []
```
bitcraft, line 888: i don't understand what this is doing.

tuxemon/states/combat/combat.py
```diff
@@ -870,13 +874,22 @@ def trigger_player_run(self, player):
     def end_combat(self):
         """End the combat"""
         # TODO: End combat differently depending on winning or losing
+        old_stats_data = Technique.keep_old_stats(self)
```
bitcraft, line 877: this code is hard to follow, and made worse because the old stats values are spread around the techniques and now the combat class.  it would be more clear if the monsters were checked for statuses, and if the status has changed their stats, then revert the stats using the old stats value from the technique.

bitcraft (changes requested):

xirsoi:
I have concerns about this implementation, please do not merge until I've had a chance to do some tests with it tonight.

tuxemon/states/combat/combat.py
SandytheLizard: if you guys have a more efficient way of doing it thats fine, i admit this code turned out a bit hard to follow as i kept having issues and making changes. it does work though

bitcraft: this is going to be a source of problems in the future.  please add a comment so that future developers can narrow down issues in battles with multiple trainers and strange stats changes.
```
# The values here are set in the technique and do not contain reference to trainers.  This code needs to be updated to support multiple trainers.
```

tuxemon/states/combat/combat.py
bitcraft: can you explain what this is doing, or remove it if its not doing anything?

tuxemon/states/combat/combat.py
bitcraft: i believe this is redundant now?

bitcraft (changes requested):

kerizane:
Why is the old stats stored on the technique class? Why not just attach it to the monster itself.?

kerizane (changes requested):
Why are the old stats stored on the technique class? Why not just attach it to the monster itself?

PR #986, opened 2021-09-03
Add compressed saves

ghost:


tuxemon/lib/compress_json.py
```diff
@@ -0,0 +1,204 @@
+# -*- coding: utf-8 -*-
```
bitcraft: is this directly pulled from his github?   i checked and its MIT licensed, so to be compliant, it will need a copy of it in the header https://en.wikipedia.org/wiki/MIT_License

tuxemon/config.py
```diff
@@ -88,6 +88,8 @@ def __init__(self, config_path: Optional[str] = None) -> None:
         self.dev_tools = cfg.getboolean("game", "dev_tools")
         self.recompile_translations = cfg.getboolean("game", "recompile_translations")
         self.skip_titlescreen = cfg.getboolean("game", "skip_titlescreen")
+        self.compress_save = cfg.getboolean("game", "compress_save")
```
bitcraft: what happens if both are true?  might be better to make is a string value, so that "gzip", or "lzma" can be specified

tuxemon/config.py
```diff
@@ -200,6 +202,8 @@ def get_defaults() -> Mapping[str, Any]:
                         ("locale", "en_US"),
                         ("dev_tools", False),
                         ("recompile_translations", True),
+                        ("compress_save", True),
```
bitcraft: we should not be saving compressed saved files by default

tuxemon/save.py
```diff
@@ -52,127 +54,256 @@
...
-    elif save_data is None:
-        # File not found; it probably wasn't ever created, so don't panic
-        return None
-    else:
-        save_data["error"] = "Save file corrupted"
-        save_data["player_name"] = "BROKEN SAVE!"
-        logger.error("Failed loading save file.")
+config = prepare.CONFIG
+
+if config.compress_save:
```
bitcraft: why is all this code duplicated?  i would think that the original json saves could be used (pass them into the compress method?) instead of duplicating the code?

tuxemon/states/persistance/load_menu.py
```diff
@@ -6,7 +6,6 @@
 
 logger = logging.getLogger(__name__)
 
-
 class LoadMenuState(SaveMenuState):
```
bitcraft: please don't change formatting unless it is makes the code comply with PEP8 and other standards.  PEP8 states that there should be two lines between code blocks at the root level.  removing this will make the code non-compliant,

tuxemon/states/persistance/save_menu.py
```diff
@@ -32,15 +33,37 @@ def initialize_items(self):
         slot_rect = Rect(0, 0, rect.width * 0.80, rect.height // 6)
         for i in range(self.number_of_slots):
             # Check to see if a save exists for the current slot
-            if os.path.exists(prepare.SAVE_PATH + str(i + 1) + ".save"):
-                image = self.render_slot(slot_rect, i + 1)
-                item = MenuItem(image, T.translate("menu_save"), None, None)
-                self.add(item)
+            if cfgcheck.compress_save:
```
bitcraft: please try to reduce all the duplicated code.

bitcraft (changes requested):

tuxemon/lib/compress_json.py
ghost: fixed

tuxemon/states/persistance/load_menu.py
ghost: fixed

tuxemon/lib/compress_json.py
```diff
@@ -0,0 +1,214 @@
...
+    compression_kwargs = {} if compression_kwargs is None else compression_kwargs
+    json_kwargs = {} if json_kwargs is None else json_kwargs
+    compression = infer_compression_from_filename(path)
+    mode = get_compression_write_mode(compression)
+
+    directory = os.path.dirname(path)
+    if directory:
+        os.makedirs(directory, exist_ok=True)
+
+    if compression == "json":
```
vnmabus: Maybe something like this could be used?
```python
if compression is None:
    compression = "gzip"

if compression is "json":
    open_function = open
else:
    compression_tool = importlib.import_module(compression)
    open_function = compression_tool.open

file = open_function(
    path,
    mode=mode,
    encoding="utf-8",
    **compression_kwargs,
)
```

tuxemon/save.py
```diff
@@ -52,127 +54,172 @@
...
 
-    """
-    screenshot = pygame.Surface(client.screen.get_size())
-    world = client.get_state_by_name(WorldState)
-    assert world
-    world.draw(screenshot)
-    return screenshot
+def open_save_file(save_path: str) -> Optional[Dict[str, Any]]:
 
+	if config.compress_save:
```
bitcraft: PEP8

tuxemon/save.py
```diff
@@ -52,127 +54,172 @@
...
+	Saves the current game state to a file using gzip compressed JSON.
+
+	Parameters:
+		save_data: The data to save.
+		slot: The save slot to save the data to.
+
+	"""
+	# Save a screenshot of the current frame
+	# if/else paradise
+	if config.compress_save:
```
bitcraft: PEP8.  Indentation should be 4 spaces, no tabs.

tuxemon/save.py
```diff
@@ -52,127 +54,172 @@
...
+		if config.csave_lzma:
+			save_path = prepare.SAVE_PATH + str(slot) + ".csave.lza"
+			text = compress_json.dump(
+			save_data,
+			save_path,
+			compression_kwargs = {},
+			json_kwargs = { "indent": 4, 'separators': (",", ": ") }
+		)
+		else:
```
bitcraft: the only thing changing here is the filename and compression kwargs.  this should be refactored to remove the redundant code in these if/else blocks

tuxemon/save.py
```diff
@@ -52,127 +54,172 @@
...
+	Loads game state data from a save file.
+
+	Parameters:
+		slot: The save slot to load game data from.
+
+	Returns:
+		Dictionary containing game data to load.
+
+	"""
+	if config.compress_save:
```
bitcraft: PEP8

tuxemon/save.py
```diff
@@ -52,127 +54,172 @@
...
-        save_data = open_save_file(save_path)
-        if save_data is not None:
-            time_of_save = datetime.datetime.strptime(save_data["time"], TIME_FORMAT)
-            times.append((slot_index, time_of_save))
-    if len(times) > 0:
-        s = max(times, key=itemgetter(1))
-        return s[0]
-    else:
-        return None
+	times = []
```
bitcraft: PEP8 -- please avoid changing formatting unless it is to bring existing code into compliance

tuxemon/states/persistance/save_menu.py
```diff
@@ -32,15 +33,37 @@ def initialize_items(self):
...
+                if cfgcheck.csave_lzma:
+                    if os.path.exists(prepare.SAVE_PATH + str(i + 1) + ".csave.lza"):
```
bitcraft: the only change here is the filename, it appears.  i don't see a need for duplicating everything

bitcraft (changes requested):

tuxemon/save.py
ghost: ?

tuxemon/save.py
ghost: ack

tuxemon/save.py
ghost: ack

tuxemon/states/persistance/save_menu.py
ghost: best way I know how to do this if you know a better way I am all ears

vnmabus: It would be something like:
```python
# save_format should be a string containing the name of the compression module or ``None`` for no compression

# I propose to use as the extension the compression module name, so that no mapping "compression" <-> "extension" is required

save_suffix = "save" if save_format is None else f"csave.{save_format}"

save_name = f"{prepare.SAVE_PATH}{i + 1}.{save_suffix}"

if os.path.exists(save_name):
    image = self.render_slot(slot_rect, i + 1)
    item = MenuItem(image, T.translate("menu_save"), None, None)
    self.add(item)
else:
    if not empty_image:
        empty_image = self.render_empty_slot(slot_rect)
    item = MenuItem(empty_image, "SAVE", None, None)
    self.add(item)
```

tuxemon/config.py
```diff
@@ -200,6 +202,8 @@ def get_defaults() -> Mapping[str, Any]:
                         ("locale", "en_US"),
                         ("dev_tools", False),
                         ("recompile_translations", True),
+                        ("compressed_save", False),
+                        ("compressed_save_use_lzma", False),
```
vnmabus, line 206:
```suggestion
                        ("compression_method", "lzma"),
```

vnmabus: I would have a single key with possible values `None` (the default) and a string with the name of a compression module "lzma", "zipfile", whatever.
```suggestion
                        ("compression", None),
```

tuxemon/save.py
bitcraft: 4 spaces, not 8.  do not use tabs

tuxemon/save.py
```diff
@@ -52,127 +54,174 @@
 
 slot_number = None
 TIME_FORMAT = "%Y-%m-%d %H:%M"
+config = prepare.CONFIG
 
+def capture_screenshot(client: LocalPygameClient) -> pygame.surface.Surface:
```
bitcraft, line 72: Please rebase or merge, because this PR is out of date

tuxemon/save.py
```diff
@@ -52,127 +54,174 @@
...
+		if config.csave_lzma:
+			save_path = f"{prepare.SAVE_PATH}{slot}.csave.lza"
+		else:
+			save_path = f"{prepare.SAVE_PATH}{slot}.csave.gz"
+	else:
+		save_path = f"{prepare.SAVE_PATH}{slot}.save"
+
+	save_data = open_save_file(save_path)
+
+	# DEBUG STUFF HERE
```
bitcraft: please remove the debug comments.  if its important, you can change to `logger.debug`

bitcraft (changes requested):

tuxemon/save.py
```diff
@@ -56,172 +56,176 @@
...
-					logger.error("cannot decode save CBOR: %s", save_path)
-
-				return {}
-
-		except OSError as e:
-			logger.info(e)
-			return None
+    if config.compress_save:
+        try:
+            DATA = compress_json.load(
```
bitcraft: pep8: variable names should be lower case

tuxemon/save.py
```diff
@@ -56,172 +56,176 @@
...
+        if config.csave_lzma:
+            save_path = f"{prepare.SAVE_PATH}{slot}.csave.lza"
+        else:
+            save_path = f"{prepare.SAVE_PATH}{slot}.csave.gz"
+    else:
+        save_path = f"{prepare.SAVE_PATH}{slot}.save"
+
+    save_data = open_save_file(save_path)
+
+    # DEBUG STUFF HERE
```
bitcraft: please remove debug comments

bitcraft (approved):
looks good, thank you for cleaning it up

PR #987, opened 2021-09-03
Replaces more references to intellectual property

freshreplicant:
These commits remove further references to existing intellectual property (IP).

It also edits some docstrings to clarify that while inspiration was drawn from the genre as a whole, the code and assets within the project are original and not derived from any other source (unless appropriately licensed).

The only remaining tangential mentions of existing IP should now be when it is a direct reference to the name of an external appropriately licensed source. If necessary, these can be removed by a future commit.


Sanglorian:
Thank you! You could actually remove the part about the damage formula being inspired by other popular games of the genre altogether as it's no longer true -- we've got an original formula now.

freshreplicant:
> Thank you! You could actually remove the part about the damage formula being inspired by other popular games of the genre altogether as it's no longer true -- we've got an original formula now.

Ah ok! Cool. Should we just remove it totally, or replace it with some detail about our own damage formula? If we just take it out then there won't be much in that particular docstring anymore.

Sanglorian:
I think take it out altogether at this point -- and then at a later date we could go through and provide more details about all the formulas and calculations.

bitcraft:
LGTM, but ill leave it open if you want to make more change

freshreplicant:
@Sanglorian and @bitcraft, the latest commit removes the unnecessary lines from the damage docstring.

xirsoi (approved):
LGTM

PR #989, opened 2021-09-05
Mod managment class

magmaus3:
This PR adds mod downloader and manager class.

Features:
- Ability to use multiple repositories
- Dependency support
- Caching packages

You can currently use it by copying `test_deps.py` from `tests/mod_manager` to the root of the repo, and running it.

[Content server's repo](https://github.com/vXtreniusX/TuxemonContentServer)
[Documentation](https://github.com/vXtreniusX/TuxemonContentServer/blob/master/USAGE.md)


magmaus3:
Ready for review, but I'll do some changes later, including adding compatibility to [ContentDB](https://github.com/minetest/contentdb) and changing the dependency installing.

test_deps.py
```diff
@@ -0,0 +1,7 @@
+from tuxemon.mod_manager import Manager
```
bitcraft: for tests, please read https://github.com/Tuxemon/Tuxemon/tree/development/tests#readme

bitcraft: if you are including scripts for testing things, but it is not a unittest, you can put into scripts for now.  these files don't belong in the root of the project.

tests/mod_manager/test_deps.py
```diff
@@ -0,0 +1,7 @@
+from tuxemon.mod_manager import Manager
```
bitcraft: Only tests that can be evaluated by a test runner should be in the tests folder.

tests/mod_manager/test_symlinks.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+
+
+class TestSymlinks(unittest.TestCase):
+    def test_linking(self):
+        from tuxemon.symlink_missing import symlink_missing
+        path = "/home/vxtreniusx/Tuxemon/"
+        breakpoint()
```
bitcraft: im not sure, but i think this would break a test runner.  is `breakpoint()` required?

tests/mod_manager/test_symlinks.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+
+
+class TestSymlinks(unittest.TestCase):
+    def test_linking(self):
+        from tuxemon.symlink_missing import symlink_missing
+        path = "/home/vxtreniusx/Tuxemon/"
```
bitcraft: please don't hardcode paths from your personal computer

tuxemon/mod_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        """Downloads the specified package"""
+        if repo is None:
+            repo = self.get_package_repo(name)
+
+            # Remove trailing slash
+            if repo[-1] == "/":
+                repo = repo[:-1]
+
+        url = str(repo) + f"/packages/{name}/releases/{str(release)}/download"
+        filename = os.path.join(paths.CACHE_DIR + f"/downloaded_packages/{name}.{release}.zip")
```
bitcraft: this uses `+` inside of `os.path.join`.... i think you should be using a comma instead?

tuxemon/mod_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        Same as the download_package(), but it includes dependency installing.
+        When symlink is True, dependency's files will be linked.
+        """
+        # Get info
+        meta = self.get_package_info(name, repo)
+        if "dependencies" in meta:
+            dependencies = meta["dependencies"]
+
+            for pack in dependencies:
+                self.download_package(pack, release, repo, dont_extract=dont_extract, install_deps=False)
```
bitcraft: doesn't look like it can detect cyclic redundancies?  a -> b -> c -> a would create a loop, it seems

tuxemon/mod_manager.py
```diff
@@ -0,0 +1,190 @@
...
+
+            to_append = {name: path_to_folder}
+            after = {**before, **to_append}
+            file.write(
+                json.dumps(after, indent=4)
+            )
+
+    def read_package_from_list(self, name):
+        """Reads path of the specified mod"""
+        #try:
```
bitcraft: please avoid leaving commented code, if possible.  there is a bit in Tuxemon already, for legacy code, but we should avoid adding new commented code lines.

tests/mod_manager/test_symlinks.py
```diff
@@ -0,0 +1,12 @@
...
+        breakpoint()
+        symlink_missing(path + "mods/1", path + "mods/2")
```
bitcraft: how is the test result evaluated?

tuxemon/mod_manager.py
bitcraft: also, there is no sanitation of the `name` value here.  a malformed package could be named something like `../foo` and it might be able to write files outside of the cache folder.

tuxemon/mod_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        """Extracts the specified zip archive to the mods directory"""
+        with zipfile.ZipFile(file) as zip_:
+            zip_.extractall(path=os.path.join(paths.BASEDIR, "mods", outfolder))
+
+    def get_package_info(self, name, repo):
+        """Get specified package info. Always downloads the info from the server."""
+        with urllib.request.urlopen(repo + f"/api/packages/{name}") as packages:
+            return json.loads(packages.read().decode("UTF-8"))
+
+    def get_package_repo(self, name):
```
bitcraft: if we are allowing mods to be hosted by anyone, then we should assume all servers are malicious.  this means the client  should be checking for broken or malicious values from anything returned by the mod server.  in the case of `name`, it is used to make file paths for writing files.  we should make sure that `name` cannot be used to escape from the tuxemon cache.

tuxemon/mod_manager.py
```diff
@@ -0,0 +1,190 @@
...
+                json.dumps(before, indent=4)
+            )
+
+
+    def remove_package(self, name):
+        """Removes the local package"""
+        # Get the path
+        path = self.read_package_from_list(name)
+
+        shutil.rmtree(path, ignore_errors=True)
```
bitcraft: this needs more checking so the name isn't something like `../../../../../home/`

tuxemon/symlink_missing.py
```diff
@@ -0,0 +1,49 @@
+"""
+Module mainly used for the mod_manager.py
+"""
+import os
+
+
+def symlink_missing(target, *sources):
```
bitcraft, line 7: will this work on windows?

bitcraft (changes requested):

tests/mod_manager/test_symlinks.py
magmaus3: I forgot to remove it, I used it for debugging.

tuxemon/symlink_missing.py
magmaus3: Yes, it works on Windows.

tests/mod_manager/test_symlinks.py
magmaus3: It's not, it was mostly for debugging

tuxemon/mod_manager/__main__.py
```diff
@@ -0,0 +1,205 @@
...
+        else: pass
+
+    def extract(self, file, outfolder):
+        """Extracts the specified zip archive to the mods directory"""
+        with zipfile.ZipFile(file) as zip_:
+            zip_.extractall(path=os.path.join(paths.BASEDIR, "mods", outfolder))
+
+    def get_package_info(self, name, repo):
+        """Get specified package info. Always downloads the info from the server."""
+        for char in '/\\?%*:|"<>.,;= ':
```
bitcraft: it would be nice to see this sanitize code in a function since its duplicated so much

tuxemon/mod_manager/__main__.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+    def remove_package(self, name):
+        """Removes the local package"""
+        # Get the path
+        path = self.read_package_from_list(name)
+        for char in '?%*:|"<>.,;= ':
+            if char in path:
+                raise ValueError(f"Detected incorrect character ({char})")
+        shutil.rmtree(path, ignore_errors=True)
```
bitcraft: what would a typical path be?  it looks like it is supposed to be relative to the folder of whatever is running the script?  i don't see where the values in the json are checked.  would it then be possible for the name/path value here to start with a `/`?  that is not being sanitized, so one could put in `/home/` in there and the home folder would be removed, or something similar.

tuxemon/mod_manager/tuxemoncontentserver.py
```diff
@@ -0,0 +1,81 @@
...
+import requests
+import json
+
+def update(url):
+    """Returns the response from the server"""
+    packages = requests.get(url=url)
+    print(packages.text)
+    return packages.json()
+
+def download_package(self, name, release, repo=None, dont_extract=False, install_deps=True, installed=None):
```
bitcraft, line 16: this looks like it is largely duplicated in another file.  could they be refactored to remove it?

test_manager.py
```diff
@@ -0,0 +1,25 @@
+from tuxemon.mod_manager import Manager
```
bitcraft: please move interactive tests to the scripts folder for now.

bitcraft (changes requested):

tuxemon/mod_manager/tuxemoncontentserver.py
magmaus3: Yes, I'm planning to remove the update and download stuff from the original file

bitcraft (approved):
lgtm

PR #995, opened 2021-09-12
Fix spyder start

freshreplicant:
This PR removes the teleport workarounds in place for the Spyder and the Cathedral campaign. Changes include:

1. spyder_bedroom.tmx and spyder_downstairs.tmx are properly connected via teleport. No more redirect to the default campaign bedroom when going back upstairs (this was a workaround in the event that you renamed the default spyder_bedroom.tmx to player_house_bedroom.tmx to start in the Spyder campaign). 

2. SPYDER_README.md now mentions the tuxemon.cfg option for changing the starting map.

3. Removed the renaming workaround from the SPYDER_README.md as it is no longer needed (see above).

4. Removes spyder_player_house_bedroom.tmx (@Sanglorian indicated this was an outdated version of spyder_bedroom.tmx).


bitcraft (approved):
looks good thank you

PR #996, opened 2021-09-15
Typing menu elements.

vnmabus:


tuxemon/menu/interface.py
```diff
@@ -1,33 +1,50 @@
+from __future__ import annotations
 import pygame
 
 from tuxemon import tools, graphics
 from tuxemon.ui.draw import GraphicBox
+from typing import ClassVar, Optional, Callable
 
 
-class Bar:
+class Bar():
```
bitcraft: why are the parens added?

vnmabus: Sorry, I always use them as for me that syntax is more consistent with other class and function definitions. I can remove them if you prefer.

bitcraft: the need for parens was removed with python3 for new-style classes and its pretty unusual to see in modern code.  i can't find anything official that says to leave them out, but the docs do show the class def using the form without parens.
https://docs.python.org/3/tutorial/classes.html#class-definition-syntax

vnmabus: Removed.

bitcraft (approved):

PR #997, opened 2021-09-16
Specify map_name at WorldState startup time.

vnmabus:
This guarantees that `current_map` always have a valid value.


bitcraft (approved):

bitcraft:
Thank you!

PR #998, opened 2021-09-16
Typing events

vnmabus:


tuxemon/platform/events.py
```diff
@@ -19,59 +19,118 @@
...
-    default_input_map = None
+    default_input_map: Mapping[Optional[int], int]
 
-    def __init__(self, event_map=None):
+    def __init__(
+        self,
+        event_map: Optional[Mapping[Optional[int], int]] = None,
+    ) -> None:
         if event_map is None:
-            event_map = self.default_input_map.copy()
```
bitcraft, line 54: this copy should be kept.  if the event map is changed, then the default would be changed for any other devices that are using it.  the config file can reconfigure controls, so this means it won't be possible to have different controls for other players....think split screen.

bitcraft (changes requested):

tuxemon/platform/events.py
vnmabus: The event map is not changed anywhere. It is only used in the files modified in this PR.

Also, I have used an immutable type, so Mypy should flag as an error any attempt to change it.

bitcraft: fair enough

bitcraft (approved):

PR #1008, opened 2021-09-21
Fix in remove_monster - trainer is not a parameter

ultidonki:
I haven't created an issue for this, just a small fix - error message is as follows when correctly using the remove_monster action:
'RemoveMonsterActionParameters has no variable trainer'
Crashes on the following line:
     trainer_slug = self.parameters.trainer

To test it, I created the following events:

1:
act1 get_player_monster var1
act2 set_variable xoxo:yes
cond1 not variable_set xoxo:yes

2:
remove_monster var1
cond1 variable_set xoxo:yes

This will crash with error 'RemoveMonsterActionParameters has no variable trainer'
Once fixed it correctly removes the chosen tuxemon


bitcraft (approved):
perfect, thank you!  as an FYI since you mentioned it, its not necessary to open an issue before a PR, so no problem.  for larger fixes or changes though, better to open an issue to avoid duplication of work.  in any case, thanks again!

PR #1009, opened 2021-09-23
Scripting docs

vnmabus:
Creates a new reference section for scripting in the docs, which currently has the list of conditions and their parameters.

I have never worked with Sphinx and autodoc internals. Thus, it is possible that there is a better/easier solution than this one.

In the ideal world the file condition_list.rst should be autogenerated, but I have neither the knowledge or the heart to do that right now.


bitcraft (approved):

bitcraft:
yeah, im sure there is a way to script it with a plug in or some code somewhere, but im also not familiar enough to start with it right now.  ill open an issue for it, but in the mean time, this is a great start, thank you!

PR #1012, opened 2021-09-28
Fix typing errors in set monster level action

ultidonki:
Firstly, we need to add a len() around the player.monsters list, as you can't compare a list and 0 and it was giving an error when the action was used. 
(I guess you could replace it with:
`if not self.session.player.monsters:`
I can change it if that's more Python-y? )

Secondly, add typing around the params, as you're allowed to pass None for both params, eg the action:
set_player_monster ,
Will set all player's monsters to level 1. 
Previously this was giving errors if you tried to pass None as a param.

### DIFF
--- a/tuxemon/event/actions/set_monster_level.py
+++ b/tuxemon/event/actions/set_monster_level.py
@@ -26,14 +26,21 @@ from typing import NamedTuple, final, Union
 
 class SetMonsterLevelActionParameters(NamedTuple):
     slot: Union[int, None]
-    level: Union[int, None]
+    level: int
 
 
 @final
 class SetMonsterLevelAction(EventAction[SetMonsterLevelActionParameters]):
     """Changes the level of a monster in the current player's party. The action parameters
     may contain a monster slot and the amount by which to level. If no slot is specified,
-    all monsters are leveled. If no level is specified, the level is reverted to 1.
+    all monsters are leveled. The level parameter can be negative, which decreases
+    the monster's level.
+
+    Examples:
+    set_player_monster 0,5 # Increases the monster in the first slot's level by 5
+    set_player_monster ,1  # Increases all player's monsters by 1 level
+    set_player_monster 4,-100 # Decreases the monster in the fifth slot's level
+                # by 100 levels
 
     Valid Parameters: slot, level
     """
@@ -53,13 +60,7 @@ class SetMonsterLevelAction(EventAction[SetMonsterLevelActionParameters]):
                 return
 
             monster = self.session.player.monsters[int(monster_slot)]
-            if monster_level:
-                monster.level = max(1, monster.level + int(monster_level))
-            else:
-                monster.level = 1
+            monster.level = max(1, monster.level + int(monster_level))
         else:
             for monster in self.session.player.monsters:
-                if monster_level:
-                    monster.level = max(1, monster.level + int(monster_level))
-                else:
-                    monster.level = 1
+                monster.level = max(1, monster.level + int(monster_level))
--- a/tuxemon/event/actions/set_monster_level.py
+++ b/tuxemon/event/actions/set_monster_level.py
@@ -21,12 +21,12 @@
 
 from __future__ import annotations
 from tuxemon.event.eventaction import EventAction
-from typing import NamedTuple, final
+from typing import NamedTuple, final, Union
 
 
 class SetMonsterLevelActionParameters(NamedTuple):
-    slot: int
-    level: int
+    slot: Union[int, None]
+    level: Union[int, None]
 
 
 @final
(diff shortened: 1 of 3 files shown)


xirsoi:
Why the heck would you wanna set all the player's monsters to level 1?
I question the validity of `set_monster ,`

ultidonki:
Lol, I agree, I thought it was weird. I've looked up the original pull request and there's not much explanation there either. 

However, I can think of a possible use-case:
If you want to set all player monster's to level 50, you need to do:
set_monster_level ,
set_monster_level ,49
To first set all levels to 1, and then bump up all levels to 50. 

As far as I know no-one's using this, but it could be useful to someone in the future, maybe.

tuxemon/event/actions/set_monster_level.py
```diff
@@ -21,12 +21,12 @@
...
 from __future__ import annotations
 from tuxemon.event.eventaction import EventAction
-from typing import NamedTuple, final
+from typing import NamedTuple, final, Union
 
 
 class SetMonsterLevelActionParameters(NamedTuple):
-    slot: int
-    level: int
+    slot: Union[int, None]
```
bitcraft, line 28: I think this is better typed as `Optional[int]`

vnmabus: Usually yes, but here the type is being used to parse the parameter by trying to cast to each type sequentially in case of a Union (so the order matters). Maybe support for Optional could also be added, but for now, it is not.

bitcraft: So this is fine as is?

vnmabus: It should be, yes. We could try to accept also `Optional` in the parsing function, in a separate PR, and change every occurrence of `Union[whatever, None]` in event parameters. It is very low priority, though.

bitcraft: ty for walking me through that vnmabus

bitcraft:
> if not self.session.player.monsters:

i would prefer it written like this.  the previous way of `if not self.session.player.monsters > 0` is kinda confusing.


> Will set all player's monsters to level 1.

i kinda agree with xirsoi here.  leaving out the param and then setting all the monster levels seems like unexpected behavior.  thinking about it...  i don't think any of the params should be optional, as any default there is kinda hard to reason with.  explicit is better than implicit, especially since the scripts are written and read by non-programmers and should be easy to understand without knowing too much.

ultidonki:
OK, I've added a couple of things to this PR - though I can break them into seperate PR's if that helps.

- I've changed the complicated expression to this one now:
`if not self.session.player.monsters:` 

- I've removed the ability to pass a blank parameter for 'level', so it no longer resets the monster's level to 1. (It turns out you can do this anyway by passing in negative numbers, eg. -999 will clamp it to 1, always.) Neither of the 2 official campaigns were using this functionality anyway, so nothing needs to be changed in those mods. 

-  I noticed this event actually allows you to go above MAX_LEVEL, so I pushed some logic into monster.py to clamp the new level between 1 and MAX_LEVEL, and made a call to that function from this action instead. 

- I also added some unit tests to the set_level function in monster.py just to make sure it always clamps between 1 and MAX_LEVEL now.

tuxemon/event/actions/set_monster_level.py
ultidonki: I've done some tests with Optional[int] based on this convo - in this action, the behaviour before and after was the same, both for valid params (int and none) and invalid params (float and str - they were accepted without raising an error in both the Union and Optional cases). 

I can make the change here if you want, as it seems the same, but I'll raise an issue seperately to track changing all the others.

vnmabus: It seems that `Optional[whatever]` is translated at runtime to `Union[whatever, None]`. That is why it works. In our case, the order in which each type appears matters, as each type is tried sequentially, so we have to be 100% sure that future versions of Python do not change the order, and `None` is still the last parameter. If that is the case we could change them safely.

ultidonki: I think you're right - I've looked at that code now (cast_values in tools.py) and if Optional was implemented in any other way, that code would fail. I don't think we want to rely on a quirk of the python implementation, so I'll keep it as Union and close the other issue I raised. Thanks for checking that!

vnmabus: The official documentation says literally "Optional[X] is equivalent to Union[X, None]". I think that is unlikely that the behaviour changes in future versions, so if you want to use optional instead, it is ok for me.

ultidonki: I'm happy either way - Union[X, None] reminds users that they need to handle both the X and None cases, which is good. But if Optional is used everywhere else in the code base then it might be better to be consistent. So I don't mind.

It's good to know that the choice is just a style choice, and both will have the same effect behind the scenes, I think that's the important thing.

bitcraft (approved):

bitcraft:
Thank you!

PR #1020, opened 2021-10-12
Basic support for ContentDB

magmaus3:
This PR adds basic ContentDB support. It's not perfect, but it should allow for simple downloads.


TestClient.py
```diff
@@ -0,0 +1,48 @@
+# Should the download_package function avoid extracting the zip file?
```
bitcraft: please put interactive tests in the scripts folder

TestClient.py
```diff
@@ -0,0 +1,48 @@
...
+
+print(
+    "",
+    "Test package downloader.",
+    "Check tuxemon/mod_manager/__init__.py for module.",
+    "Downloads are located at ~/.tuxemon/cache/downloaded_packages",
+    sep="\n"
+)
+
+if not os.path.exists(os.path.expanduser("~/.tuxemon/cache/downloaded_packages")):
```
bitcraft: this would be better written as
`os.makedirs(foo, exists_ok=True)`
https://docs.python.org/3/library/os.html#os.makedirs

TestClient.py
```diff
@@ -0,0 +1,48 @@
...
+    os.mkdir(os.path.expanduser("~/.tuxemon/cache/downloaded_packages"))
+
+ip = input("Enter the server URL with format: http://ip:port/\n> ")
+man = Manager(ip)
+print("Updating the package list...")
+man.update(ip)
+print("Writing to cache...\n")
+man.write_to_cache()
+print("Press [CTRL] + [C] or [CTRL] + [D] to exit")
+while True:
```
bitcraft: since this is an infinite loop, you should add a way to exit, or print a reminder to quit by pressing "CTRL+C"...since there are people who don't know that.

tuxemon/mod_manager/__init__.py
```diff
@@ -56,8 +56,8 @@ def cache_to_pkglist(self):
 
     def update(self, url):
         """Returns the response from the server"""
-        with urllib.request.urlopen(url + "/api/packages") as packages:
-            return json.loads(packages.read().decode("UTF-8"))
+        packages = requests.get(url + "/api/packages")
```
bitcraft, line 62: :+1:

tuxemon/mod_manager/__init__.py
```diff
@@ -86,73 +86,98 @@ def download_package(self, name, release, repo=None, dont_extract=False, install
...
 
         self.write_package_to_list(os.path.relpath(outfolder), name)
 
         if not dont_extract:
-            self.extract(filename, outfolder)
-            with open(f"{outfolder}/meta.json", "w") as metafile:
-                meta = self.get_package_info(name, repo)
-                metafile.write(json.dumps(meta, indent=4, sort_keys=False))
+            with zipfile.ZipFile(filename) as zip_:
+                print(zip_.namelist())
```
bitcraft: all these print statements would be better as logging statements, or removed.  most appear to be used for debugging

tuxemon/mod_manager/__init__.py
```diff
@@ -86,73 +86,98 @@ def download_package(self, name, release, repo=None, dont_extract=False, install
...
-        with zipfile.ZipFile(file) as zip_:
-            zip_.extractall(path=os.path.join(paths.BASEDIR, "mods", outfolder))
-
-    def get_package_info(self, name, repo):
+    def install_dependencies(self, author, name, repo, symlink=True, **args):
+        """Recursively resolve dependencies and symlink them"""
+        print(author, name, repo)
+        # Request dependencies for specified package
+        r = requests.get(f"{repo}/api/packages/{author}/{name}/dependencies/?only_hard=1")
+        if r.status_code == 404:
```
bitcraft: instead for checking for an error, you should be checking for success.  ie, don't check for 404 and quit, check for 200 and continue (or not 200 and then quit)

tuxemon/mod_manager/__init__.py
```diff
@@ -198,12 +223,24 @@ def remove_package_from_list(self, name):
...
         # Get the path
         path = self.read_package_from_list(name)
+        if os.path.isabs(path):
+            raise IOError("Path is absolute")
         if path != sanitize_paths(path):
             raise ValueError("Detected incorrect characters in path")
         shutil.rmtree(path, ignore_errors=True)
         self.remove_package_from_list(name)
+
+    def install_local_package(self, filename, name=None, download_deps=False, link_deps=False):
```
bitcraft, line 239: is this used anywhere?

magmaus3: `install_local_package` is supposed to be used separately, for just installing the packages

tuxemon/mod_manager/__init__.py
magmaus3: Didn't realise that `requests` module is already in requirements.txt. I didn't want to add additional dependencies.

tuxemon/mod_manager/__init__.py
magmaus3: I think that it would be better to replace the unpacking part in `download_package` with this function

tuxemon/mod_manager/__init__.py
```diff
@@ -86,73 +89,97 @@ def download_package(self, name, release, repo=None, dont_extract=False, install
...
-            zip_.extractall(path=os.path.join(paths.BASEDIR, "mods", outfolder))
-
-    def get_package_info(self, name, repo):
+    def install_dependencies(self, author, name, repo, symlink=True, **args):
+        """Recursively resolve dependencies and symlink them"""
+        logger.debug(author, name, repo)
+        # Request dependencies for specified package
+        r = requests.get(f"{repo}/api/packages/{author}/{name}/dependencies/?only_hard=1")
+        if r.status_code != 200:
+            return
```
bitcraft: Should an error be raised here?  Or is it ok if its not installed?

tuxemon/mod_manager/__init__.py
bitcraft: ok.  is this finished then, or will be used in the future?  generally, better to avoid adding code if its not being used.

tuxemon/mod_manager/__init__.py
magmaus3: I think that it should be raised

tuxemon/mod_manager/__init__.py
```diff
@@ -75,9 +78,9 @@ def update_all(self):
 
     def list_packages(self):
         """Returns package dictionary, either from the server or the cache"""
-        return self.packages
+        return self.packages # why
```
bitcraft: what does this comment mean?  `# why?`

tuxemon/mod_manager/__init__.py
```diff
@@ -86,73 +89,102 @@ def download_package(self, name, release, repo=None, dont_extract=False, install
...
+        # Request dependencies for specified package
+        r = requests.get(f"{repo}/api/packages/{author}/{name}/dependencies/?only_hard=1")
+        if r.status_code != 200:
+            raise ValueError(f"Requested {r.url}, received status code {r.status_code}")
+        logger.debug(r.text, author, name)
+        dep_list = r.json()
+        # Resolve dependencies
+        for dependency in dep_list:
+            for entry in dep_list[dependency]:
+                for package in entry["packages"]:
```
bitcraft, line 147: since the string "package" is returned from an untrusted source and we are using ti to manipulate the file system, it should be sanitized to make sure it cannot do anything unexpected or malicious.

tuxemon/mod_manager/__init__.py
```diff
@@ -198,12 +230,24 @@ def remove_package_from_list(self, name):
...
+
+    def install_local_package(self, filename, name=None, download_deps=False, link_deps=False):
+        """
+        Installs local packages.
+        Based on the download_package function, but without the downloads
+        """
+        # TODO: Write an TODO comment
+        outfolder = os.path.join(paths.BASEDIR, "mods", name)
+        self.write_package_to_list(os.path.relpath(outfolder), name)
+        with zipfile.ZipFile(filename) as zipf:
```
bitcraft, line 246: the zip files are also untrusted, we should make sure there is enough disk space before decompression.  also should be checking for nested zip files.  these two problems are exploits to fill the disk or slow the cpu during decompression.

bitcraft (dismissed):
looks good.  just a couple more comments and it will be ready.

tuxemon/mod_manager/__init__.py
magmaus3: The function it is in returns the contents of an variable, that you can access in an easier way.
I'll remove it

bitcraft (approved):

bitcraft:
looks great!  thank you for the feature and working with me to get it ready to merge

tuxemon/mod_manager/__init__.py
bitcraft: looks like this is the last thing...  i can merge after this

PR #1032, opened 2021-10-22
Fixes empty status. Closes #1013

### LINKED ISSUE #1013: Set Monster Status action no longer cures fainted tuxemon
**Describe the bug**
When passing no status to the Set Monster Status action, it's expected that all statuses will be removed.
Eg. nurses use the action:
`set_monster_status ,`
to cure all statuses on all monsters.

Recently, instead of doing this, it will append the empty string, '', to a monster's list of status.
Eg. after using the action their status will become: 
`monster.status = ['status_fainted', '']`

**To Reproduce**
1. Let one tuxemon faint in battle. 
2. Get the tuxemon healed at a nurse. 
3. In the status screen, the Tuxemon will look OK, as they will have full health, but if you look at their status via debug, eg. player.monsters[].status, you'll see the new status is ['status_fainted', '']
4. Go into battle, and you'll see you have a blanked-out capture device, indicating that one of your tuxemon is fainted.
5. If you throw that tuxemon into battle, it'll cause all sorts of bugs, like fainting immediately and then the game gets confused and can't progress the battle. 

**Expected behavior**
Passing nothing to the 2nd param of set_player_status should remove all statuses. 

**Additional context**
Using git bisect, this started happening after the parameter's type was changed to Union[str, None]
When passing empty parameters to Union[int, None] or Union[float, None], None is correctly passed. So it's probably a problem with Union[str, None]'s. 
I don't know how you'd fix this.

danvilar:


bitcraft (approved):

bitcraft:
Thank you!

PR #1033, opened 2021-10-23
Typing combat

vnmabus:
First version of typing of the combat state. There are still typing errors, but most of them require changing other parts, and will be changed later.


bitcraft (approved):
Thanks again for this improvement

PR #1036, opened 2021-10-24
Always return a state in get_state_by_name

vnmabus:
Make `get_state_by_name` always return a state. This removes the need for some asserts.

In case the state is not present an exception will be raised.


bitcraft (approved):
LGTM

PR #1039, opened 2021-10-26
Simplify argument parsing.

vnmabus:
Simplifies the logic of the parser (and found a missing type as a result). I played the Omnichannel campaign (until arriving to the village for the second time, as both roads were blocked: is that the end for now?) without problems.


bitcraft (approved):
thank you for doing the long play test

PR #1041, opened 2021-10-31
Typing improvements

vnmabus:
Some typing improvements.

In order to be able to pass the world at NPC creation, the Player creation had to be moved to the WorldState startup function, which is an horrible hack. This should be changed in the future.


bitcraft:
horrible hacks are ok if there is a goal to fix them, and the hack is documented

bitcraft (approved):

PR #1044, opened 2021-11-06
Pyganim refactor

vnmabus:
- Rename PygAnim objects to more understandable names.
- Remove PygConductor for map animations (which have only one animation).
- Make PygAnim objects use the game clock, instead of calling `time` on their own.


bitcraft (approved):
looks good!  the new class names are a good pick.

PR #1045, opened 2021-11-08
added a way to go back to a state for the input menu

andrewhong04:


tuxemon/client.py
```diff
@@ -544,6 +544,9 @@ def pop_state(self, state: Optional[State] = None) -> None:
         """Pop current state, or another"""
         self.state_manager.pop_state(state)
 
+    def remove_state(self, state: State) -> None:
+        self.state_manager.remove_state(state)
```
bitcraft, line 549: missing docstring

tuxemon/menu/menu.py
```diff
@@ -17,6 +17,7 @@
 from tuxemon.graphics import ColorLike
 from tuxemon.platform.events import PlayerInput
 from tuxemon.animation import Animation
+from tuxemon.states.transition.fade import FadeInTransition
```
bitcraft: what is `FadeInTransition` used?

tuxemon/states/start/__init__.py
```diff
@@ -121,4 +122,4 @@ def exit_game() -> None:
             label = T.translate(key).upper()
             image = self.shadow_text(label)
             item = MenuItem(image, label, None, callback)
-            self.add(item)
+            self.add(item)
```
bitcraft: please avoid changing formatting on source files

tuxemon/menu/menu.py
```diff
@@ -67,6 +68,7 @@ class Menu(Generic[T], state.State):
     default_character_delay = 0.05
     shrink_to_items = False  # fit the border to contents
     escape_key_exits = True  # escape key closes menu
+    back_state = None  # if escape_key_exits is false, then the escape key can change the state
```
bitcraft: i don't see why the esc key and also back_state logic needs to be linked.  why should it not be possible to press escape and also use back_state?

tuxemon/menu/menu.py
```diff
@@ -495,6 +497,10 @@ def process_event(self, event: PlayerInput) -> Optional[PlayerInput]:
             if event.pressed and self.escape_key_exits:
                 self.close()
 
+            if event.pressed and self.back_state != None and not self.escape_key_exits:
```
bitcraft: minor: `None` is a singleton, so the best comparison to use is `is not None`

tuxemon/menu/menu.py
```diff
@@ -495,6 +497,10 @@ def process_event(self, event: PlayerInput) -> Optional[PlayerInput]:
...
+                self.close()
```
bitcraft: is close needed here?  the code for menu will manipulate the state stack by calling `pop_state`.  having to quick stack operations could lead to unexpected behavior.  like popping the current state (the menu), then replacing the one right after it (the start state?)
https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/menu/menu.py#L673

bitcraft (commented):
looks good, bu a few comments for clarification

tuxemon/menu/menu.py
andrewhong04: Sorry, I tried doing a fade transition when the player goes back to the start menu. It didn't work out. I'll change it in the next commit.

tuxemon/states/start/__init__.py
andrewhong04: Yeah, I don't know what happened there

tuxemon/menu/menu.py
andrewhong04: They should be linked since I programmed the menu to go to the back_state using the escape key, which escape_key_exits also uses as well.
Should I make it that when the `back_state is not None`, it'll go to the back_state no matter what value escape_key is?

tuxemon/menu/menu.py
andrewhong04: When I remove `self.close()`, it shows this. I couldn't really think of a better solution to this.
image
```
            if event.pressed and self.back_state is not None:
                self.client.pop_state()
                self.client.replace_state(state_name=self.back_state)
```
Using this block of code has the same solution since it calls pop_state() directly if you want to use this instead.

tuxemon/states/start/__init__.py
bitcraft: some editors seem to hate trailing newlines.  its not a big deal, but since this it your first contribution, i just wanted you to know

tuxemon/menu/menu.py
bitcraft: For fade states, push the next state, then push the fade state.

tuxemon/menu/menu.py
bitcraft: have you considered the discord comment i left?  we can actually do this without even adding the back_state
```
diff --git a/tuxemon/states/start/__init__.py b/tuxemon/states/start/__init__.py
index ebbb2b66..a9d574fc 100644
--- a/tuxemon/states/start/__init__.py
+++ b/tuxemon/states/start/__init__.py
@@ -90,17 +90,18 @@ class StartState(PopUpMenu[StartGameObj]):
             )
 
         def set_player_name(text: str) -> None:
+            map_name = prepare.fetch("maps", prepare.CONFIG.starting_map)
+            self.client.push_state("WorldState", map_name=map_name)
+            self.client.push_state(FadeInTransition)
             local_session.player.name = text
 
         def new_game() -> None:
             # load the starting map
-            map_name = prepare.fetch("maps", prepare.CONFIG.starting_map)
-            self.client.replace_state("WorldState", map_name=map_name)
             self.client.push_state(
                 state_name=InputMenu,
                 prompt=T.translate("input_name"),
                 callback=set_player_name,
-                escape_key_exits=False,
+                escape_key_exits=True,
             )
             self.client.push_state(FadeInTransition)
```
the benefit here is no new added features, which, honestly we kinda try to avoid since it just adds a maintenance burden.
to answer your question though, i don't see why they should be linked.  the back key should only close the state.  adding a conditional to it (close the state?  close and use the back state?) is confusing because in both instances, the menu state is closed anyway.  and the existence of a back state will also enable the back state.  to put it another way, it wouldn't make sense to have a situation where there is a back_state and also disabling the escape key.

tuxemon/menu/menu.py
bitcraft: So this is going to be really confusing when i explain this but i will try anyway....
the state stack "remembers" what the previous states were.  in the old code, the WorldState is added to the stack (replaced) before the InputMenu, so closing the input menu goes "back" to the WorldState, by default.  So, the problem becomes..."if i close the input state, how do i also close the world state?".  

the reason why the extra close is required with this implementation is:
* close () will pop the current state (InputMenu)
* the current state is now world state
* then replace the current state (world state), with the start state

```python
        def new_game() -> None:
            # load the starting map
            map_name = prepare.fetch("maps", prepare.CONFIG.starting_map)
            self.client.replace_state("WorldState", map_name=map_name)
            self.client.push_state(
                state_name=InputMenu,
                prompt=T.translate("input_name"),
                callback=set_player_name,
                escape_key_exits=False,
            )
            self.client.push_state(FadeInTransition)
```

tuxemon/menu/menu.py
andrewhong04: Sorry, I didn't see the negatives of adding a new feature. I added your pieces of code to the pull request and resolved some of your other comments.

tuxemon/menu/menu.py
```diff
@@ -494,7 +494,7 @@ def process_event(self, event: PlayerInput) -> Optional[PlayerInput]:
             handled_event = True
             if event.pressed and self.escape_key_exits:
                 self.close()
-
+
```
bitcraft: formatting :D

tuxemon/menu/menu.py
andrewhong04: I think I sort of understand the code a little better. Thanks for the clarification.

tuxemon/menu/menu.py
andrewhong04: Hopefully, this latest commit fixes the issue, lol.

bitcraft (approved):
looks great!  thank you for this fix!  i know that this problem has come up more than once.

PR #1049, opened 2021-11-09
fixed a bug where the save slot menu would show up 

andrewhong04:
...when you would specify a save slot to load via argparser

also removed an unused import

### DIFF
--- a/tuxemon/states/persistance/load_menu.py
+++ b/tuxemon/states/persistance/load_menu.py
@@ -32,8 +32,7 @@ class LoadMenuState(SaveMenuState):
                 self.client.pop_state(old_world)
             except ValueError:
                 # when game is loaded from the start menu
-                self.client.pop_state()  # close this menu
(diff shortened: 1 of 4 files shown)


tuxemon/states/persistance/load_menu.py
```diff
@@ -32,8 +32,7 @@ def on_menu_selection(self, menuitem: Optional[MenuItem[None]]) -> None:
                 self.client.pop_state(old_world)
             except ValueError:
                 # when game is loaded from the start menu
-                self.client.pop_state()  # close this menu
```
bitcraft, line 35: why are these removed?

andrewhong04: > why are these removed?

The game seemed to work without those functions when I added the new solution. There wasn't any difference with having them removed, so I'll add them back right now.

bitcraft (approved):
thank you!

PR #1053, opened 2021-11-12
initial commit for new cli

bitcraft:
New CLI interface:
* Commands are plugins (add mod support later)
* Commands support help
* Able to directly execute actions while game is running
* Able to test conditions while game is running
* Somewhat robust error handling and reporting
* Commands support sub-commands
* Some support for argument introspection, more to add later

Related changes:
* Quieted down some logging that made the CLI interface to chatty
* Added commands to event engine to get actions and conditions

Autocomplete is not ready.  I've been sitting on it for too long and i need to push this out.  In the future, autocomplete can be added using `prompt_toolkit` and the `get_parameters` method on commands.


bitcraft:
@vnmabus please take a look :D

tests/tuxemon/test_parser.py
```diff
@@ -0,0 +1,123 @@
...
+        result = parse_action_string("spam eggs, ")
+        self.assertEqual(("spam", ["eggs", ""]), result)
+
+    def test_no_space_between_comma(self):
+        result = parse_action_string("spam,eggs")
+        self.assertEqual(("spam,eggs", []), result)
+
+    def test_enclosed_space(self):
+        result = parse_action_string("   spam   ")
+        self.assertEqual(("", ["spam"]), result)
```
vnmabus, line 81: Honestly I didn't expect this. Why do we want this behaviour?

tests/tuxemon/test_parser.py
```diff
@@ -0,0 +1,123 @@
...
+    def test_space_in_arg1(self):
+        result = parse_action_string("spam ex parrot")
+        self.assertEqual(("spam", ["ex parrot"]), result)
+
+    def test_space_in_arg2(self):
+        result = parse_action_string("spam eggs,  ex parrot")
+        self.assertEqual(("spam", ["eggs", "ex parrot"]), result)
+
+
+class TestParseConditionString(unittest.TestCase):
```
vnmabus, line 96: I don't get the results of these. They seem a bit counter intuitive IMHO.

tuxemon/cli/clicommand.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    name: ClassVar[str] = "command name"
+    description: ClassVar[str] = "command description"
+    usable_from_root: ClassVar[bool] = True
+    example: ClassVar[str] = ""
+
+    def invoke(self, ctx: InvokeContext, line: str) -> None:
+        """
+        Called when command input contains this command name as first term
```
vnmabus: Don't you like a dot at the end of sentences?

tuxemon/cli/clicommand.py
```diff
@@ -0,0 +1,101 @@
...
+    Base class for CLIOptions
+
+    """
+
+    name: ClassVar[str] = "command name"
+    description: ClassVar[str] = "command description"
+    usable_from_root: ClassVar[bool] = True
+    example: ClassVar[str] = ""
+
+    def invoke(self, ctx: InvokeContext, line: str) -> None:
```
vnmabus, line 25: Shouldn't this be an `abstractmethod`? Is there a case where we don't want to override that?

tuxemon/cli/clicommand.py
```diff
@@ -0,0 +1,101 @@
...
+        raise CommandNotFoundError
+
+    def resolve(self, ctx: InvokeContext, path: str) -> Tuple[CLICommand, str]:
+        """
+        Resolve a path into command and remaining text
+
+        * Split the command into tokens
+        * Starting from the first token, search it for the next token
+        * If the next token is a subcommand, repeat
+        * If the next token is not a subcommand, return command and the rest of the input
```
vnmabus: You are exceeding 80 chars in several lines. This is against PEP8 and makes it more difficult to see in Github side by side comparison.

tuxemon/cli/commands/action.py
```diff
@@ -0,0 +1,77 @@
...
+        Parameters:
+            ctx: Contains references to parts of the game and CLI interface.
+            line: Input text after the command name.
+
+        """
+        line = f"{self.name} {line}"
+        name, args = parse_action_string(line)
+        try:
+            ctx.session.client.event_engine.execute_action(name, args)
+            print("ok.")
```
vnmabus: CLI utilities are usually silent when everything goes well.

tuxemon/cli/commands/action.py
```diff
@@ -0,0 +1,77 @@
...
+        except Exception as exc:
+            traceback.print_exc()
+            print("Cannot execute action. Check the input and try again.")
```
vnmabus: Maybe print in standard error output?

tuxemon/cli/commands/help.py
```diff
@@ -0,0 +1,42 @@
...
+            line: Input text after the command name.
+
+        """
+        command, tail = ctx.root_command.resolve(ctx, line)
+        if command is not ctx.root_command:
+            print(command.description)
+            print()
+        if command.example:
+            print("Example:")
+            print(f"  > {command.example}")
```
vnmabus, line 33: Why the arrow? Are you sure examples will be one liners?

bitcraft:
error when game is not loaded, but you try to add an item
```
> action add_item player,potion
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/cli/commands/action.py", line 73, in invoke
    ctx.session.client.event_engine.execute_action(name, args)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventengine.py", line 292, in execute_action
    return action.execute()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventaction.py", line 184, in execute
    with self:
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventaction.py", line 148, in __enter__
    self.start()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/actions/add_item.py", line 53, in start
    if self.parameters.quantity is None:
AttributeError: 'NoneType' object has no attribute 'quantity'
Cannot execute action. Check the input and try again.
```

bitcraft:
error on bad input
```
> test add_item player,potion
need more arguments or syntax error
```

bitcraft:
if the action is permissive and doesn't raise an exception, the response is ambiguous
```
> action add_item amogus
2021-11-12 12:08:39,006 - tuxemon.item.item - ERROR - Failed to find item with slug amogus
ok.
```

bitcraft:
error when given bad input for an action
```
> action add_monster rockitten,
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/cli/commands/action.py", line 73, in invoke
    ctx.session.client.event_engine.execute_action(name, args)
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventengine.py", line 292, in execute_action
    return action.execute()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventaction.py", line 184, in execute
    with self:
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/eventaction.py", line 148, in __enter__
    self.start()
  File "/home/ltheden/PycharmProjects/Tuxemon/tuxemon/event/actions/add_monster.py", line 59, in start
    monster_slug, monster_level, trainer_slug = self.parameters
TypeError: cannot unpack non-iterable NoneType object
Cannot execute action. Check the input and try again.
```

tuxemon/cli/commands/test.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        line = f"is {self.name} {line}"
+        try:
+            opr, typ, args = parse_condition_string(line)
+            cond = MapCondition(typ, args, 0, 0, 0, 0, opr, "USERINPUT")
+        except ValueError:
+            raise ParseError
+        try:
+            result = ctx.session.client.event_engine.check_condition(cond)
+            print(str(result))
```
vnmabus: Doesn't `print` already calls `str`?

tuxemon/cli/parameter.py
```diff
@@ -0,0 +1,14 @@
...
+
+@dataclasses.dataclass
+class Parameter:
+    """
+    Used to categorize and manage parameters for commands
+
+    """
+
+    name: str
+    klass: str
```
vnmabus: Ahh, the classical German fallback XD. Is this actually used anywhere?

tuxemon/cli/processor.py
```diff
@@ -0,0 +1,128 @@
...
+    def invoke(self, ctx: InvokeContext, line: str) -> None:
+        """
+        Default when no arguments are entered
+
+        Parameters:
+            ctx: Contains references to parts of the game and CLI interface.
+            line: Input text after the command name.
+
+        """
+        print("???")
```
vnmabus: Maybe print to stderr?

tuxemon/cli/processor.py
```diff
@@ -0,0 +1,128 @@
...
+                except EOFError:
+                    break
+
+                if line:
+                    try:
+                        command, tail = self.root_command.resolve(ctx, line)
+                        ctx.current_command = command
+                        command.invoke(ctx, tail)
+                    except ParseError:
+                        print(f"Unknown syntax: {line}")
```
vnmabus: All errors should be printed to stderr.

tuxemon/client.py
```diff
@@ -134,12 +136,16 @@ def __init__(self, config: TuxemonConfig) -> None:
...
-        self.exit = False  # Allow exit from the CLI
         if self.config.cli:
-            self.cli = cli.CommandLine(self)
+            # TODO: There is no protection for the main thread from the cli
+            # actions that execute in this thread may have undefined
+            # behavior for the game.  at some point, a lock should be
+            # implemented so that actions executed here have exclusive
+            # control of the game loop and state.
+            self.cli = CommandProcessor(local_session)
+            thread = Thread(target=self.cli.run)
```
vnmabus, line 146: We are playing with fire (although the GIL will prevent memory corruption, and most actions are unlikely to run while state is being changed: humans are not fast). How we did that before without creating a thread?

vnmabus (changes requested):

tests/tuxemon/test_parser.py
bitcraft: we don't!  but at least we have some good reasons to fix it.  To make the tests, i just put together these cases and recorded the input.  at some point we should fix it

tests/tuxemon/test_parser.py
bitcraft: i agree.  its just what was made a really long time ago and it was never covered with tests

tuxemon/cli/clicommand.py
bitcraft: i found a valid case where you don't need to override it.  if you command only has subcommands, then you would never invoke the parent command.

vnmabus: That makes sense.

tests/tuxemon/test_parser.py
vnmabus: Maybe we should put the desired result and mark the test as `@unittest.expectedFailure`?

tuxemon/cli/clicommand.py
bitcraft: hahaha, no not at all.  i know you've added a lot of new docstrings with the trailing punctuation, so i will make it the same, though.  my reasoning, is that often in docstring, technical information, etc. the "word phrases" are not valid sentences, meaning they might not have a subject and a verb, so would be a "noun phrase" or "verb phrase" and in those cases, its not a sentence, and doesn't require punctuation.  outside of code, headings typically don't have periods either.

its also a style i see often, i guess msft uses it (just learned that when i looked it up

https://docs.microsoft.com/en-us/style-guide/punctuation/periods

tuxemon/cli/clicommand.py
bitcraft: the black formatter uses 88 by default, which is what i've been using. pep8 actually define the limit as 79, but that silly.  i don't mind bringing it down, as long as we agree on *something*

tuxemon/cli/clicommand.py
vnmabus: The official docstring PEP recommends adding a period:
https://www.python.org/dev/peps/pep-0257/

tuxemon/cli/commands/action.py
bitcraft: fair enough

tuxemon/cli/commands/action.py
bitcraft: sure

tuxemon/cli/commands/help.py
bitcraft: the arrow is just for decoration.  it looks like a prompt, and you often see that in webpages with a `$` for instance.  the cli doesn't support multiline input, so there would be no example of a command with more than one line

tuxemon/cli/commands/test.py
bitcraft: ill check

tuxemon/cli/clicommand.py
vnmabus: I would rather follow PEP8 as it is the official styleguide and most IDEs have native support for checking that. We should also change the config in `setup.cfg` for flake8, which has that limit as 100 IIRC.

tuxemon/cli/parameter.py
bitcraft: no not at the moment.  it was being used for autocomplete, and will be used to provide suggestions.  i think it can be removed

tuxemon/cli/processor.py
bitcraft: sure

tuxemon/cli/processor.py
bitcraft: ok

tuxemon/client.py
bitcraft: this is how it was run before.  so this isn't a change, im just documenting the problem

tests/tuxemon/test_parser.py
bitcraft: idk.  the spec for the text format, as far as i am concerned is still not set and id rather not make any assumptions of what it should be.  id like to leave it like this so if the parsing code is touched or refactored, we will know if it has deviated from the current [not ideal] spec

tuxemon/cli/commands/test.py
bitcraft: removed

tuxemon/cli/clicommand.py
bitcraft: fair enough

tuxemon/cli/parameter.py
bitcraft: removed

vnmabus (approved):
LGTM

PR #1058, opened 2021-11-12
Basic Crash Reporter

magmaus3:
This PR adds an crash reporter + saving logs to files.

 - Log uploads are using transfer.sh for storage.
 - When the app crashes, an popup (currently using tkinter) appears, asking if log should be sent.
   Then it prints out the log url in command line.
 - Logs on the user's machine are automatically archived on launch.


run_tuxemon.py
```diff
@@ -52,4 +55,16 @@
...
-    main.main(load_slot=args.slot)
+    try:
+        main.main(load_slot=args.slot)
+    except Exception as ex:
+        from pygame import quit as pg_quit
+        pg_quit()
+        
+        logging.error(ex, exc_info=True)
+        if ex != KeyboardInterrupt and ex != EOFError:
+            #if input("Send logs? (Y/N)\n> ").upper() == "Y":
```
bitcraft: is this needed?

tuxemon/config.py
```diff
@@ -119,6 +119,12 @@ def __init__(self, config_path: Optional[str] = None) -> None:
         self.debug_logging = cfg.getboolean("logging", "debug_logging")
         self.debug_level = cfg.get("logging", "debug_level")
 
+        
+        self.log_host_url = cfg.get("crash_reporting", "log_host_url")  # URL to transfer.sh compatible
```
bitcraft: please put the comments above the line, it looks much cleaner

tuxemon/log.py
```diff
@@ -32,8 +32,34 @@
...
+from os.path import exists
+from datetime import datetime
+# For messagebox
+import tkinter as tk
+from tkinter import messagebox
 
+# For gathering system information
+import platform
 
+def archive_log():
```
bitcraft, line 114: this is fine now, but just a FYI, python has a log rotation, compression, etc.  if this gets more complex, we should probably use what is in the stdlib
https://stackoverflow.com/questions/8467978/python-want-logging-with-log-rotation-and-compression

tuxemon/log.py
```diff
@@ -72,8 +98,78 @@ def configure():
...
+
+    try:
+        logging.info(f"pygame_sdl_version\t{__import__('pygame').get_sdl_version()}")
+    finally:
+        pass
+
+    # Get file
+    file = open(f"{USER_LOG_DIR}/latest.log")
+    send_files = {"tuxemon_log.txt": file}
+    r = requests.post(config.log_host_url, files=send_files, headers={"Max-Days": config.log_storage_max_days, "User-Agent": "curl/7.80.0"})
```
bitcraft: im guessing that the header here is specific to transfer.sh?  to prevent leaking logs without proper expiration, lets either make a generic class that handles sending logs to a provider, or simply raise an error if its not going to transfer.sh, and put a TODO here about allowing other providers

tuxemon/config.py
```diff
@@ -242,6 +248,16 @@ def get_defaults() -> Mapping[str, Any]:
                     )
                 )
             ),
+            (
+                "crash_reporting",
+                OrderedDict(
+                    (
+                        ("log_host_url", "https://transfer.sh/"),
```
bitcraft: could we add an option to enable the popup?  during development, its going to crash often, and the popup will get really annoying.

bitcraft (commented):
looks great!  looking forward to seeing the logs being generated.  few comments

tuxemon/config.py
magmaus3: I was thinking about adding an option to the config with `None` as the default value (and creating the popup), and bool values for automatically sending/not sending logs.

run_tuxemon.py
magmaus3: The logging.error part is for printing the traceback, so it isn't "eaten" by the except function.

The commented if is not needed, I'll remove it later.

magmaus3 (commented):
‮

magmaus3:
The generic class is done

tuxemon/log.py
```diff
@@ -32,10 +32,94 @@
...
+    def __init__(self, use_local_url=False, storage_time=1):
+        if use_local_url:
+            self.url = self.local_url
+        else:
+            self.url = self.remote_url
+
+        self.log_storage_max_days = storage_time
+
+
+
```
bitcraft: PEP8: too many lines between method blocks.  should be exactly one line

tuxemon/log.py
```diff
@@ -46,34 +130,123 @@ def configure():
...
+            logging.info(f"{i}\t{__import__(lib_entries[i]).__version__}")
+        except:
+            logging.info(f"{i}\tunknown")
+
+    try:
+        logging.info(f"pygame_sdl_version\t{__import__('pygame').get_sdl_version()}")
+    finally:
+        pass
+
+    # Get file
```
bitcraft: please avoid including commented code, especially if there is no note about what it is.

tuxemon/log.py
```diff
@@ -46,34 +130,123 @@ def configure():
...
+#    file = open(f"{USER_LOG_DIR}/latest.log")
+#    send_files = {"tuxemon_log.txt": file}
+
+    # TODO: Add other providers support
+    try:
+        provider = eval(f"{config.log_host}(storage_time={config.log_storage_max_days})")
```
bitcraft: what is eval being used here for?  can we avoid using it?

tuxemon/log.py
```diff
@@ -46,34 +130,123 @@ def configure():
...
+
+def popup_send_log_consent():
+    """
+    Popup requesting consent for sending crash report.
+    Returns bool value with the choice.
+    """
+    if prepare.CONFIG.popup:
+        return True
+        
+    elif not prepare.CONFIG.popup:
```
bitcraft: why not if/else?

tuxemon/log.py
```diff
@@ -46,34 +130,123 @@ def configure():
...
+            __import__("webbrowser").open_new_tab(response[0])
+        else: 
+            pass
+
+    except NameError:
+        logging.error(f"Attempted to load a non-existent provider {config.log_host}")
+    except requests.exceptions.ConnectionError:
+        logging.error("Connection to the provider failed.", exc_info=True)
+    
+
```
bitcraft, line 258: PEP8: should be 2 lines between functions/classes defined in the global scope (not indented)

tuxemon/log.py
```diff
@@ -32,10 +32,94 @@
...
+        """
+        Sends the specified file to the remote log/file storage.
+        Returns the text response and status code.
+        
+        Parameters:
+            file_obj: The file object to send
+        """
+        files = {"tuxemon_log.txt": file_obj.read()}
+        r = requests.post(
+                            self.url,
```
bitcraft: excessive indentation

tuxemon/log.py
```diff
@@ -46,34 +130,123 @@ def configure():
...
+    # Get file
+#    file = open(f"{USER_LOG_DIR}/latest.log")
+#    send_files = {"tuxemon_log.txt": file}
+
+    # TODO: Add other providers support
+    try:
+        provider = eval(f"{config.log_host}(storage_time={config.log_storage_max_days})")
+        response = provider.send_log()
+        print(f"Report URL: {response[0]}")
+        if config.auto_open_link.upper()[0] == "T":  # Using the standard "if config" didn't work for me
```
bitcraft: we're trying to keep the lines short.  please avoid in line comments, especially when they create very long lines.  this one is 104 chars.

tuxemon/config.py
```diff
@@ -119,6 +119,14 @@ def __init__(self, config_path: Optional[str] = None) -> None:
         self.debug_logging = cfg.getboolean("logging", "debug_logging")
         self.debug_level = cfg.get("logging", "debug_level")
 
+        # [crash_reporting]        
+        self.log_host = cfg.get("crash_reporting", "log_host")  # The log storage type.
```
bitcraft: please move the comments to the line above the code

bitcraft (approved):
logic looks great, thank you for the changes.  i added more comments, mostly about formatting.  we are going to try to tighten up the formatting requirements.  
if you would like to have it done automatically, you can use 
```shell
black -t py38 -l 79
```
which is what our "official" formatter does.

bitcraft (changes requested):
see comments

tuxemon/log.py
magmaus3: I used it for "spawning" a class object.

magmaus3:
Ready for review, note that I've encountered some merge conflicts, that I wasn't able to fix before. Not sure if I resolved them correctly now.

Sorry for the delay.

PR #1059, opened 2021-11-13
It seems like the new commit required a new library to be installed via pip, I've added the library in requirements.txt

andrewhong04:
Updated requirements.txt


bitcraft (approved):
thank you!

PR #1060, opened 2021-11-14
Item condition eval

vnmabus:
This is a demo of a possible scripting change. Currently it is only implemented for item conditions, as it was the easiest scripting context to modify.

There are two main changes:
- Instead of specific parameters, the condition receives a `ScriptContext` object. This object is intended to contain variables (and maybe methods in the future) usable directly from scripts. It can also have a `parent` context, so that, for example, scripts inside a "Combat" context can access the "World" context variables too.
- A very general `eval` condition is implemented. I have implemented two variants, and we must choose between the two:
    - A very general variant that calls Python `eval` using the context variables as local variables. This would allow scripts to use the full power of Python, with all the upsides and downsides. It is also way easier to maintain as it only requires three lines of code.
    - A restricted variant, which only allows expressions in the form `<value> <operator> <value>`, where `value` can be a Python string, number, variable (from the local context), or even a set of primitive values, and `operator` can be a logical, bitwise, comparison or in operator. This requires way more code (and regex), and is more restrictive, but prevents scripters from ~~removing your disk files~~ shooting themselves in the foot.

The conditions for actual items have been rewritten to use this condition: no other types of conditions are necessary for now! The syntax used is compatible with both variants.

Please, use this PR to discuss this functionality. Do you like it, or maybe is too much flexible or difficult to use for scripters? Opinions from actual mapmakers are VERY welcomed.


Qiangong2:
I guess I'm a bit confused as to its use. Does it make changes to how checking items are done? Or does this just allow scripts to be called directly from conditions?

vnmabus:
> 
> 
> I guess I'm a bit confused as to its use. Does it make changes to how checking items are done? Or does this just allow scripts to be called directly from conditions?

The idea is to reduce the number of conditions that need to be implemented by providing more broadly applicable conditions. So, instead of having conditions compare_target_type, compare_target_current_hp, compare_target_hp, etc, a context including the available objects such as `target` is passed to each condition, and a general `eval` condition can compare whatever two values you want. This could be further generalized to allow calling arbitrary Python code in conditions, or restricted only to simple comparisons.

In the future the same idea could be applied for actions and event conditions.

Qiangong2:
So if I'm understanding it correctly, it will consolidate similar conditions into a single one? Sorry if that's an oversimplification :P

vnmabus:
> So if I'm understanding it correctly, it will consolidate similar conditions into a single one? Sorry if that's an oversimplification :P

Yeah, it will. You can see that for example the `current_hp` condition has been removed. Instead of:
```
current_hp target,>,0
```
you could do
```
eval target.current_hp > 0
```
But now, without having to add new conditions, you can also do:
```
eval target.speed > 10
```
and it will work.

IF in addition we decide to allow arbitrary Python expressions you could even do more complicated things, like:
```
eval (target.speed > 10) or (target.current_hp + 5 < target.hp)
```
for example.

bitcraft:
i like the idea of eval, but i do know that our entire system for events is pretty limited.  setting aside the obvious problem of running any arbitrary code, i wonder if people would abuse the eval code to do arbitrary things in the game?....and rely on that?  that said, i think its probably fine if we make some effort to restrict it to the absolute minimum functionality required.

we would also want to craft a special dict for the globals because eval will allow the use of `__builtins__`, which we may want to avoid allowing them all.
https://docs.python.org/3/library/functions.html#eval

then, we should lock only allow ust a few names for use
https://github.com/Tuxemon/Tuxemon/pull/1060/files#diff-958abd56c44740884036624300430046aa90c4042dc53dc302879a82442f8151R139

its simple, and would be roughly:
```
condition_globals= dict()
condition_globals['__builtins__'] = None

condition_globals['foo'] = bar

result = eval(expression, condition_globals, condition_locals)
```

vnmabus:
It was my understanding that in the future we wanted to allow modders to create custom conditions and actions. That already runs arbitrary code, so the problem will still exist if we do that.

I think it is NOT possible at all to call `eval` and `exec` without the user being able to run arbitrary code (there are plenty of tricks to do that, and several Python sandboxing projects have failed). Maybe it is possible to do something inspecting the AST, but that is error prone and not very maintenable.

Thus, if we allow custom code execution, either by a special condition/action like this one or by allowing users to create custom conditions and actions, we will need to install Tuxemon in a sandboxed environment (Docker, non classic Snap, or the equivalent in other systems), or to warn users that non whitelisted mods can be dangerous and pass the ball to them. I think we should decide what is the right course of action for this.

PR #1062, opened 2021-11-15
Fix groups

vnmabus:
Simplifies a lot of group-related code:
- `draw` code that only has a few changes w.r.t. Pygame has been simplified
- `determine_cursor_movement` has been simplified, especially for the bidimensional movement case. Now up/down movement is done moving to a top-bottom indexing, increasing or decreasing that index, and going back to left-right indexing.

`MenuSpriteGroup` now takes into account disabled `MenuItem`'s.


bitcraft (approved):
great improvement, thank you!

PR #1066, opened 2021-11-17
fixed a mistake i made with newlines

andrewhong04:


bitcraft (approved):
thank you for this follow up fix!

PR #1068, opened 2021-11-18
Allow changing of battle backgrounds via json

ultidonki:
In the wiki, we have about <s>20</s> 16 cool battle BG images, already sized to the game's resolution. 
But, for some reason we haven't put any of them into the game yet.

This PR adds:
* 1 image from the wiki, plus license info in CREDITS.md
* The ability to change the background, in combat_animations.py and combat.py
* * (it falls back to the default if the background isn't specified, similar to how island_front and island_back work)
* An example of changing the grass.json environment to a new BG, using battle_background property in 'graphics' dict

Here's how it looks in the grass environment now: (all other environments look the same as they did previously)
new_battle_bgs


vnmabus:
This is very cool! I think in the future we should allow map areas to have different backgrounds, though.

ultidonki:
Thanks! But I think environments can still be used for different maps? For example, you could have a my-map1.json environment, that changes the background, and battle music, just for my-map? Or are environments intended to be something more general than map stuff?

vnmabus:
Mmmm, I think you are right. It seems the environment can be changed changing a player variable, so it would even be possible to have several ones in the same map.

ultidonki:
I'm glad environments work the way I expected, thanks for checking! Changing the environment inside a map is actually something I hadn't thought of... I could probably use that for a boss battle in my mod, if this gets accepted. Thanks for the idea! ;)

mods/tuxemon/db/environment/grass.json
```diff
@@ -2,7 +2,8 @@
   "slug": "grass",
   "battle_graphics": {
     "island_back":"grass_back_island.png",
-    "island_front":"grass_front_island.png"
+    "island_front":"grass_front_island.png",
+    "background":"Background_240px.png"
```
bitcraft: we've had issues in the past with mixed case filenames.  some OS will allow mixing them and others won't and it has created issues with loading images.  please rename the file to be all lower case.

tuxemon/states/combat/combat_animations.py
```diff
@@ -61,6 +61,14 @@ def startup(
         graphics: Optional[Mapping[str, str]] = None,
         **kwargs: Any,
     ) -> None:
+
+        # Get background image if passed in
+        background_filename_prefix = "gfx/ui/combat/"
+        if "background" in graphics and graphics["background"] > "":
```
bitcraft: can we change this to:
```python
if "background" in graphics and graphics["background"]:
```

bitcraft (changes requested):
this is a great addition!  thank you for this.  i have a couple comments on there, but overall this is great.

tuxemon/states/combat/combat_animations.py
vnmabus: Even better:
```suggestion
        background = graphics.get("background")
        if background:
```

ultidonki:
I've made those 2 changes - and also took the opportunity to rename the file to be more maintainable in the future, ie. forest_background.png, to make it more obvious what it is without having to open it. Thanks both!

bitcraft (approved):
great feature, thank you!

PR #1072, opened 2021-11-20
Fix dialogue appearing before random encounters finish #964

### LINKED ISSUE #964: Random_encounter doesn't block other actions from running
If you have a random_encounter action, followed by some more actions, such as a dialog state, the random_encounter action won't block the dialog state from happening at the same time, so dialog can appear during the battle state.
Currently, the only affected map is spyder_scoop4.tmx, as most random encounters don't have any noticeable actions after them. 

After debugging, this is why it's happening:
In tuxemon/event/actions/random_encounter.py:
The RandomEncounter action queues a BattleState onto the stack using queue_stack(), and then pushes a FlashTransition onto the stack to display that before the battle begins. 
But when checking update() to see if it's done, it looks for the BattleState in the active state stack by using session.client.get_state_by_name() 
But the battle state isn't there, because it's in the _queue_stack, not the active stack. So it immediately sets 'done = True' and remaining actions in the event start being processed, instead of waiting until the BattleState ends. 

There's currently no way to query what's in the queue_stack from tuxemon/client.py, so maybe a new 
StateManager.is_in_queue_stack(state_name): str -> Boolean 
function in tuxemon/state.py and ./client.py would work? Then, actions could query both whether a queued state is active, plus whether it's still in the queue stack, before returning. 

I'm happy to make a PR or let someone else do it if this sounds like a good idea.

ultidonki:
Prepare for some dodgy code... In particular, the types are probably all wrong! However, I've tested and it works. 

This fixes #964, which happens in the Spyder mod, and my mod too - so I think it deserves a fix even the solution isn't that nice. But better suggestions are welcome. 

I've made a .tmx file that reproduces the bug at the bottom of this post. Basically:
If you have a random encounter action, and then some dialogue, the dialogue plays DURING the battle, not waiting until afterwards. 
Then, the game usually crashes in a horrible way. 

The problem is that the random_encounter action waits until the CombatState is not active in the stack, and returns after that. 
However, instead of pushing the CombatState onto the stack, it queues it up for eg. a few seconds while it shows the FlashTransition state. 
So on the first call to update() to see if CombatState is active in the stack, it's NOT active, (because it's in the queued stack), so random_encounter stops itself and the other actions start playing.

This PR adds the ability to query the queued stack just like the active stack, and changes the random_encounter action to make sure the CombatState is not in the queued stack AND not in the active stack before letting other actions run. 

Comments and suggestions are welcome, it seems like a difficult change for what seems like a simple bug. 

[test_dialogue_after_battle.zip](https://github.com/Tuxemon/Tuxemon/files/7573487/test_dialogue_after_battle.zip)

### DIFF
--- a/tuxemon/client.py
+++ b/tuxemon/client.py
@@ -542,6 +542,15 @@ class LocalPygameClient:
         """
         return self.state_manager.get_state_by_name(state_name)
 
+    def get_queued_state_by_name(
+        self,
+        state_name: str,
+    ) -> Tuple[str, Mapping[str, Any]]:
+        """
+        Query the state stack for a state by the name supplied.
+        """
+        return self.state_manager.get_queued_state_by_name(state_name)
+
     def queue_state(self, state_name: str, **kwargs: Any) -> None:
         """Queue a state"""
         self.state_manager.queue_state(state_name, **kwargs)
--- a/tuxemon/event/actions/random_encounter.py
+++ b/tuxemon/event/actions/random_encounter.py
@@ -118,10 +118,14 @@ class RandomEncounterAction(EventAction[RandomEncounterActionParameters]):
             )
 
     def update(self) -> None:
+        # If state is not queued, AND state is not active, then stop.
         try:
-            self.session.client.get_state_by_name(CombatState)
+            self.session.client.get_queued_state_by_name("CombatState")
         except ValueError:
-            self.stop()
+            try:
+                self.session.client.get_state_by_name(CombatState)
+            except ValueError:
+                self.stop()
 
 
 def _choose_encounter(
--- a/tuxemon/state.py
+++ b/tuxemon/state.py
@@ -664,6 +664,17 @@ class StateManager:
         """
         return self._state_stack[:]
 
+    @property
+    def queued_states(self) -> Sequence[Tuple[str, Mapping[str, Any]]]:
+        """
+        Sequence of states that are queued.
+
+        Returns:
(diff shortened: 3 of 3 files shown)


bitcraft:
I think the logic here is good, but the fact that the state is queued should not require an additional way to search for the state.  I think that any code inspecting the state manager should be able to look at the stack and see both active and queued states.  Ill have to think that over, but I can't think of a reason why it shouldn't work.  To clarify, I think a better solution would be to change the state manager to inspect the active states, and the queued states when using `State.get_state_by_name`.  Though, your code in the pr is good.  And thank you for including a test map!  That was very helpful.

ultidonki:
Thanks for looking at it! I'll leave you to think about it, you can leave this PR open or closed, it's fine. 

(I have a workaround for this in my mod so there's no rush for a solution) 

One thing that might not work - you can pass in a type to get_state_by_name, ie. CombatState instead of "CombatState", and it will still find an active state - but in the queue stack it only stores the string "CombatState", and turns it into a type of state later. So you may need to cast the type into a string somehow.  

I don't know enough about the code to say, but I've done a grep for 'queue_state', and only random_encounter.py uses it. So maybe the queue isn't used much? So implementing it the way you say will probably work, if it's only used in that one place. 

Just my thoughts - I'll leave it with you. :)

xirsoi (approved):
I think this fix is important enough to merge, even if the solution isn't necessarily perfect.

PR #1076, opened 2021-11-23
fixed error where the control state display buttons are selectable

andrewhong04:


tuxemon/states/control/__init__.py
```diff
@@ -165,8 +165,7 @@ def change_state(
             label = f"{T.translate(key).upper()}"
             image = self.shadow_text(label)
             item = MenuItem(image, label, None, change_state("SetKeyState", input=input))
-            if input is None:
-                item.enabled = False
+            item.enabled = item is not None
```
bitcraft: the name "input" is reserved for python, and its bad practice to "shadow" an existing global or builtin name ("input").  i would say that replaceing "input", with either "button", or "button_name" on line 164 would be best here.
```python
item.enabled = button is not None
```
(change "button" to whatever you rename "input")

bitcraft (changes requested):

tuxemon/states/control/__init__.py
```diff
@@ -161,11 +161,11 @@ def change_state(
             (display_buttons[buttons.BACK], None)
         )
 
-        for key, input in key_items_map:
+        for key, button in key_items_map:
             label = f"{T.translate(key).upper()}"
             image = self.shadow_text(label)
-            item = MenuItem(image, label, None, change_state("SetKeyState", input=input))
-            item.enabled = item is not None
+            item = MenuItem(image, label, None, change_state("SetKeyState", button=button))
```
bitcraft, line 167: should be `input=button`, unless you want to refactor the SetKeyState

andrewhong04: Yeah, I already did that in line 65. Is there an issue if I refactor it?

bitcraft: i see that now!  no problem then

bitcraft (approved):
thank you!

PR #1093, opened 2021-12-11
Initial draft of monster journal, proof of concept

ultidonki:
This PR adds a new screen, which shows the (translated) name, (translated) description, and front sprite of all monsters. 

It's accessible from the Main Menu, by choosing 'Journal'.  
Using the up and down buttons moves you up and down in the list, and using left and right buttons lets you skip a screen forward or back. 
Clicking on a tuxemon doesn't do anything at the moment - in the future it should maybe open another screen that shows more details of the Tuxemon. 
A lot of the tuxemon names, descriptions, and ID numbers are actually missing, despite being in the wiki - I'll raise an issue seperately saying these need to be scraped/updated from the wiki. (The ID number at the minute is just the number they appear in, there's no order to the tuxemon appearing) 

This PR contains a bit of weird code in places - usually marked with 'TODO' comments. 
It doesn't need to be merged as is - maybe some bits are OK, or some bits can inspire someone on how to do it better - either way is fine, I'm not in a rush to get the functionality merged. 

The screen looks like this:
monster_journal2

The PR also contains code to fix this issue, ie. not being able to show more than 11 items in one list:
too_many_items

It does this by over-riding process_events, on_change_selection, and initialize_items. These 3 functions are probably the hackiest part of the PR, the rest was mostly copied from the Bag state.

### DIFF
--- a/tuxemon/states/monster/__init__.py
+++ b/tuxemon/states/monster/__init__.py
@@ -25,20 +25,30 @@
 #
 #
 # states.MonsterMenuState Handles creating monster menu objects
+# states.JournalMenuState Menu that shows list of all monsters
 #
 from __future__ import annotations
+import logging
 import pygame
 
 from pygame.rect import Rect
 from tuxemon import prepare, graphics
+from tuxemon.locale import T
+from tuxemon.db import db
 from tuxemon import tools
 from tuxemon.menu.interface import HpBar, ExpBar, MenuItem
 from tuxemon.menu.menu import Menu
 from tuxemon.session import local_session
+from tuxemon.sprite import Sprite
 from tuxemon.ui.draw import GraphicBox
 from tuxemon.ui.text import draw_text, TextArea
 from typing import Any, Optional, Generator
 from tuxemon.monster import Monster
+from tuxemon.platform.const import buttons
+from math import floor
+import pygame
+
+logger = logging.getLogger(__name__)
 
 
 class MonsterMenuState(Menu[Optional[Monster]]):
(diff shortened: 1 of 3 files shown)


freshreplicant:
Just a minor thought: is it possible to standardise the monster names with title case?

ultidonki:
Yes, lowercase monster names means there's no translation found in the .po file, I think best way to fix that would be to scrape the wiki for the latest data. 

I'll raise a seperate issue as a reminder to do this.

tuxemon/states/monster/__init__.py
```diff
@@ -25,20 +25,30 @@
...
 # states.MonsterMenuState Handles creating monster menu objects
+# states.JournalMenuState Menu that shows list of all monsters
 #
 from __future__ import annotations
+import logging
 import pygame
 
 from pygame.rect import Rect
 from tuxemon import prepare, graphics
+from tuxemon.locale import T
```
kerizane, line 36: Try to obey https://peps.python.org/pep-0008/#imports

tuxemon/states/monster/__init__.py
```diff
@@ -25,20 +25,30 @@
...
 from tuxemon.menu.menu import Menu
 from tuxemon.session import local_session
+from tuxemon.sprite import Sprite
 from tuxemon.ui.draw import GraphicBox
 from tuxemon.ui.text import draw_text, TextArea
 from typing import Any, Optional, Generator
 from tuxemon.monster import Monster
+from tuxemon.platform.const import buttons
+from math import floor
+import pygame
```
kerizane, line 49: Duplicate

tuxemon/states/monster/__init__.py
```diff
@@ -262,3 +272,231 @@ def on_menu_selection_change(self) -> None:
...
+        We mostly want the cursor to stay where it is, and scroll the list of
+        monsters, except for the beginning or end of the list.
+
+        Therefore, scroll the list by collecting N Tuxemon that we want to
+        display right now, depending on the current monster_index.
+
+        TODO: Maybe make this more efficient in the future, rather than looping
+        through every single tuxemon.
+
+        :return: List(MenuItem)
```
kerizane, line 344: Wrong.

@bitcraft do we need this? typehints are enough

vnmabus: This should be written in Google style ("Returns:" block), and explain the result semantics. As you said, ":type:" and ":rtype:" are redundant with type hints and should be removed.

kerizane:
Can you raise the bugfix as a separate PR?

ultidonki:
Thank you for the review, I agree that should all be fixed!

However, this PR is sort of in limbo. I created it before Pygame_menu was integrated for our GUI, so I feel like a lot of the issues with it would hopefully be fixed if I re-created this menu using pygame_menu. 

I actually tried using pygame_menu to make this menu earlier, before it was integrated - but I think there were some problems with my naive implementation. Adding 183 buttons to a menu using pygame_menu seemed to cause my computer to freeze for 40+ seconds. I have a feeling it could probably be fixed, but until then I don't want to close this PR as this version using the old GUI actually works. But if someone could get it working in pygame_menu then I think that might be better in the long run (or I'll try it again, eventually, but no timeframe for that)

> Can you raise the bugfix as a separate PR?

That's a good idea, but now that pygame_menu has been integrated, I think that bug doesn't happen with pygame_menu, as it can scroll if there's too many items to show in a box. So I think it would probably be better to upgrade the Bag and Buy/Sell menu screens to use pygame_menu, that would fix the problem completely (which is where the problem happens). I think my code to fix it is definitely a bit hacky/unmaintainable, so again I think upgrading to use pygame_menu would be better than my version.

Issue #1082, opened 2021-12-06
Don't allow names longer than 30 characters


ultidonki:
It doesn't actually break anything, but here's Mom calling my name:
![name_too_long](https://user-images.githubusercontent.com/89098769/144911960-b74369cd-6ad3-4d06-a1aa-d0bf67c805ac.png)

Needs to be fixed in both the name state and the rename_player action

novialriptide:
Lol, I can try fixing this.

bitcraft:
lol, nice.  yeah i agree, there should be a limit.  30 seems fine.

andrewhong04:
Is anyone working on this? I feel like I can handle this.

ultidonki:
I'm not looking at it, and I haven't heard anyone else is either. So go for it :smile:

ultidonki:
Closed, thanks @novialriptide !

PR #1095, opened 2021-12-13
Added a limit to player name set and change

andrewhong04:
I'm not sure if this is what you wanted, but basically what I did was I added the limit via the InputMenu instead of the two files (`rename_player.py` and `start/__init__.py`)

Ref to: https://github.com/Tuxemon/Tuxemon/issues/1082

### DIFF
--- a/tuxemon/menu/input.py
+++ b/tuxemon/menu/input.py
@@ -38,6 +38,7 @@ class InputMenu(Menu[InputMenuObj]):
         prompt: str = "",
         callback: Optional[Callable[[str], None]] = None,
         initial: str = "",
+        char_limit: int = None,
         **kwargs: Any,
     ) -> None:
         """
@@ -80,6 +81,7 @@ class InputMenu(Menu[InputMenuObj]):
 
         self.prompt.text = prompt
         self.callback = callback
+        self.char_limit = char_limit
         assert self.callback
 
     def calc_internal_rect(self) -> Rect:
@@ -183,8 +185,9 @@ class InputMenu(Menu[InputMenuObj]):
         self.client.pop_state()
 
     def add_input_char(self, char: str) -> None:
-        self.input_string += char
-        self.update_text_area()
+        if self.char_limit is None or len(self.input_string) <= self.char_limit:
+            self.input_string += char
+            self.update_text_area()
 
     def update_text_area(self) -> None:
         self.text_area.text = self.input_string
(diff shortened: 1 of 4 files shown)


tuxemon/menu/input.py
```diff
@@ -183,8 +185,9 @@ def add_input_char_and_pop(self, char: str) -> None:
         self.client.pop_state()
 
     def add_input_char(self, char: str) -> None:
-        self.input_string += char
-        self.update_text_area()
+        if len(self.input_string) <= self.char_limit or self.char_limit is None:
```
bitcraft: This would fail if char_limit is None because `len(self.input_string) <= self.char_limit` would be evaluated first and would raise a TypeError.  A better way to write this would be to put the `is None` check first.  Python will eval that first and the "short circuit" way conditionals are treated means the latter part `len(self.input_string) <= self.char_limit` would never have have `self.char_limit is None`

```python
if self.char_limit and len(self.input_string) <= self.char_limit:
   ...
```

bitcraft (changes requested):

vnmabus:
Just a little bit of nitpicking: we are limiting here the number of Unicode code points, which is not the same as the number of user-perceived characters, or grapheme clusters (for example emoji or compound characters may consist in more than one code point). Maybe for our purposes there is no problem with that, but I wanted the decision to be explicit.

bitcraft:
> Just a little bit of nitpicking: we are limiting here the number of Unicode code points, which is not the same as the number of user-perceived characters, or grapheme clusters (for example emoji or compound characters may consist in more than one code point). Maybe for our purposes there is no problem with that, but I wanted the decision to be explicit.

ty for this comment.  im not familiar with the details that you are referring to.  do you have a solution that allows us to effectively limit the number of visible characters?

vnmabus:
There is no official (stdlib) solution, as they are still discussing it: https://bugs.python.org/issue30717
There are unofficial packages that support this, for example https://uniseg-python.readthedocs.io/en/latest/
I would say that for our case using code points is probably fine, moreover when the 30 character limit is very generous. Maybe Asian languages where the characters may be more complex have a problem with that, although I am not sure if we currently support them.

bitcraft:
yeah, 30 is a bit high.  im not really concerned about the number at this point, since once this is merged, we can adjust it later.  so without a stdlib way to limit the number of visible characters, i think we should maybe just open an issue about it and figure something out later.  im thinking something along the lines of "determine some size in "character units", that would be roughly "too big" and then check by rendering the input and comparing the pixel width of the string to the "too big" number".  vague, but i don't really have the right words to use to describe it.

so, there is just the one unresolved comment on this pr.  once that is resolved, i would like to merge this.

andrewhong04:
Yeah, I can resolve it today.

bitcraft (approved):

PR #1098, opened 2021-12-20
Fixes crash in worldstate.py

### LINKED ISSUE #1097: Occasional crashes - map_drawing in worldstate.py
**Describe the bug**

Occasionally experience crashes while playing (only tested on Spyder campaign). Happened both on Paper Town and the sea route off to the left which is currently accessible by walking.

**To Reproduce**

Nothing in particular, just walk around the first few maps for a while. It does not seem to happen directly after map transitions.

**Expected behavior**

Not crashing the game.

OS: EndeavourOS Linux x86_64 
Kernel: 5.15.10-arch1-1
Python Version: Python 3.10.1

**Additional context**

Traceback:

```

  File "/home/freshreplicant/Applications/Tuxemon/run_tuxemon.py", line 55, in <module>
    main.main(load_slot=args.slot)
  File "/mnt/hdd/Applications/Tuxemon/tuxemon/main.py", line 115, in main
    client.main()
  File "/mnt/hdd/Applications/Tuxemon/tuxemon/client.py", line 308, in main
    draw(screen)
  File "/mnt/hdd/Applications/Tuxemon/tuxemon/client.py", line 412, in draw
    state.draw(surface)
  File "/mnt/hdd/Applications/Tuxemon/tuxemon/states/world/worldstate.py", line 356, in draw
    self.map_drawing(surface)
  File "/mnt/hdd/Applications/Tuxemon/tuxemon/states/world/worldstate.py", line 521, in map_drawing
    screen_surfaces.append((s, r, l))
UnboundLocalError: local variable 'r' referenced before assignment
```

freshreplicant:
Should fix: #1097

 As explained by @vnmabus:

"The variable r should be initialized even if the if is not entered."

As the crash was only very occasional, this may have to be play tested. Just load the Spyder campaign, walk around the first few maps for a while and see if it crashes still.

### DIFF
--- a/tuxemon/states/world/worldstate.py
+++ b/tuxemon/states/world/worldstate.py
@@ -509,6 +509,7 @@ class WorldState(state.State):
 
             # project to pixel/screen coords
             c = self.get_pos_from_tilepos(c)
+            r = Rect(c, s.get_size())
 
             # TODO: better handling of tall sprites
             # handle tall sprites


bitcraft:
ah hey i see you made this.  sorry i didn't know you did.  i've already pushed a fix, so im going to close this.  just a FYI, the code which makes the rect needs to go after the height check on line 516, because `c` might change because of the hight of the surface, so the rect needs to be made after that check.  but thank you for opening the issue and this pr anyway.

vnmabus:
> ah hey i see you made this. sorry i didn't know you did. i've already pushed a fix, so im going to close this. just a FYI, the code which makes the rect needs to go after the height check on line 516, because `c` might change because of the hight of the surface, so the rect needs to be made after that check. but thank you for opening the issue and this pr anyway.

Here that code is repeated so it would have worked either way (but your version has less repetition so it is better).

PR #1099, opened 2022-01-06
Add random_integer action

### LINKED ISSUE #1074: Add Random_variable action
It would be fun for mappers to have a way to create random variables, and then do different events/actions depending on those random variables. 
They could do things like lotteries, random effects if you drink 'magic water', get a random tuxemon from a vending machine... etc. 

A new action, with syntax similar to:
```random_variable variable_name,integer_number```
That sets variable_name in the player variables dictionary to a random number below integer_number. 

It needs to be in the player variables dictionary, so mappers can check this variable later using cond. (See example .tmx at end of this post)

Instead of a comma in
```random_variable variable_name,number```
possibly use a colon, to make it match the set_variable syntax:
```set_variable variable_name:value```
This might make it easier, or more consistent for mappers to understand. Or it might confuse them? 

Also, consider whether the random number will be between 0, and integer_number - 1, or between 1 and integer number (inclusive). 
I would personally keep it consistent with python, and use 0 and integer_number -1, but that could confuse mappers if they don't know or want to learn programming. I'm happy either way. 

Here's a .tmx file showing the kinds of things mappers could do with it: 
It will say either 'heads' or 'tails' depending on a random number 
(after someone implements random_variable, obviously)
[test_random_variable.zip](https://github.com/Tuxemon/Tuxemon/files/7576139/test_random_variable.zip)

### PR DESCRIPTION
**Pull request by ultidonki**
This action sets a variable to an integer between x and y, eg:
`random_integer xyz,1,6`
Will return either 1, 2, 3, 4, 5, or 6. 
This allows map makers to use randomness in their maps. 

Will close #1074, where it was discussed. 

Here's a test map that uses the action to say 'heads' or 'tails' endlessly:
[test_random_variable.zip](https://github.com/Tuxemon/Tuxemon/files/7825065/test_random_variable.zip)

### DIFF
--- a/tuxemon/event/actions/random_integer.py
+++ b/tuxemon/event/actions/random_integer.py
@@ -64,4 +64,4 @@ class RandomIntegerAction(EventAction[RandomIntegerActionParameters]):
 
         # Append the game_variables dictionary with a random number between
         # upper and lower bound, inclusive:
-        player.game_variables[var] = str(randint(lower_bound, upper_bound))
+        player.game_variables[var] = randint(lower_bound, upper_bound)
--- a/tuxemon/event/actions/random_integer.py
+++ b/tuxemon/event/actions/random_integer.py
@@ -1,0 +1,67 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tuxemon/event/actions/random_integer.py [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+    param_class = RandomIntegerActionParameters
+
+    def start(self) -> None:
+        player = self.session.player
+
+        var, lower_bound, upper_bound = self.parameters
+
+        # Append the game_variables dictionary with a random number between
+        # upper and lower bound, inclusive:
+        player.game_variables[var] = str(randint(lower_bound, upper_bound))
```
ultidonki: I'm casting the random integer to a string. This is because, in the condition variable_set, it does a comparison between the game_variable (which could be a string, integer, or float) to an optional[str] - so I think it always fails if the game_variables value isn't a string. 

I'm not sure if this is a bug in variable_set or if I'm not thinking it through properly... Probably needs thinking about before this gets merged. (This PR isn't high priority anyway, I'm not using random_integer yet)

bitcraft: If you need to compare variables, there is the `variable_is` condition, which implements operators to compare numeric values.  it should work on numbers or strings.  im not sure if i want to play too fast and loose with types with the game variables.  if if should be an int, then it should be an int.

ultidonki: I've left it as an int - that makes sense, and I've tested with variable_is, and that works well! Thanks!

bitcraft (approved):
looks good!

PR #1102, opened 2022-01-09
Add print command to CLI. Closes #1087

### LINKED ISSUE #1087: Add a 'Print' command to the command-line interface
The new command-line feature is really cool! I love it. 

This is low priority, but I wouldn't mind something like this:
``` print game_variable ```
Which will print out the value of game_variable.
``` print ```
With no arguments, it should just print out all of the player.game_variables for the current player.

(And if you're going to implement this, of course, 
``` help print ```
Should print out how to use it.)

**Optional:**
I don't know how hard it would be, but command-line completion for each variable would be very cool as well...
As I often don't remember the names of all my variables. 

**Alternatives:**
I noticed you can do this by running:
``` shell
ctx.session.player.game_variables.game_variable 
```
And if you set logging to 'debug' it'll print out all game_variables each frame. 
So it's not a huge priority.

ultidonki:
I ended up creating this to close my own issue, #1087 - it wasn't that hard actually. 

Auto-completion for available variables would be nice, but I don't think that's implemented by any of the CLI commands, so I didn't do it either?

### DIFF
--- a/tuxemon/cli/commands/print.py
+++ b/tuxemon/cli/commands/print.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from tuxemon.cli.clicommand import CLICommand
+from tuxemon.cli.context import InvokeContext
+
+class PrintCommand(CLICommand):
+    """
+    Print variables set by map event actions.
+
+    """
+
+    name = "print"
+    description = "Print values of variables set using map actions and events."
+    example = "print steps"
+
+    def invoke(self, ctx: InvokeContext, line: str) -> None:
+        """
+        Print values of variables set using map actions and events.
+
+        Parameters:
+            ctx: Contains references to parts of the game and CLI interface.
+            line: Complete text as entered into the prompt.
+
+        """
+        variable = line.strip()
+        if variable > "":
+            if variable in ctx.session.player.game_variables:
+                print(f"{variable}: {ctx.session.player.game_variables[variable]}")
+            else:
+                print(f"'{variable}' has not been set yet by map actions.")
+        else:
+            print(ctx.session.player.game_variables)
--- a/tuxemon/cli/commands/print.py
+++ b/tuxemon/cli/commands/print.py
@@ -23,7 +23,7 @@ class PrintCommand(CLICommand):
 
         """
         variable = line.strip()
-        if variable > "":
+        if variable:
             if variable in ctx.session.player.game_variables:
                 print(f"{variable}: {ctx.session.player.game_variables[variable]}")
             else:


tuxemon/cli/commands/print.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        Print values of variables set using map actions and events.
+
+        Parameters:
+            ctx: Contains references to parts of the game and CLI interface.
+            line: Complete text as entered into the prompt.
+
+        """
+        variable = line.strip()
+        if variable > "":
```
bitcraft: lets change this to just `if variable` because the `>` when used with a string is idk, really not used much and could raise some eyebrows.

bitcraft (changes requested):
small code style ask, otherwise looks perfect.

bitcraft:
The CLI interface is capable of autocompletion, but its hard to use.  I have some existing code that implements it, but its not merged due to it being kinda cumbersome.  It has to be basically perfect, otherwise it easy to get bad suggestions and makes it confusing.  In the future, I think it will be using `CLICommand.get_parameters()` and `get_subcommands` methods to get the suggestions:

```python
    def get_parameters(self, ctx: InvokeContext) -> Iterable[Parameter]:
        """
        Return parameters for use by help or autocomplete.

        Parameters:
            ctx: Contains references to parts of the game and CLI interface.

        """
        for cmd in self.get_subcommands(ctx):
            yield Parameter(cmd.name)

    def get_subcommands(self, ctx: InvokeContext) -> Iterable[CLICommand]:
        """
        Return all subcommands of this command.

        Parameters:
            ctx: Contains references to parts of the game and CLI interface.

        """
        return ()
```

The hope being that this will be enough to implement autocomplete and suggestions without building a full blown grammar.

ultidonki:
Thanks for the quick feedback on the PR's, might be a few days (or weeks) until I can work on them, but all the suggestions sound good, thanks!

And thanks for explaining the autocomplete as well, in that case I won't work on it now, but I'll keep that information for future reference, just in case I want to try adding it in the future. Thanks for that!

bitcraft:
@ultidonki there are sill unresolved comments

ultidonki:
Sorry, I'll try and fix these issues with the PR's soon. They're only small things, so I guess I'll try and make time to look at them soon. Thanks!

tuxemon/cli/commands/print.py
ultidonki: Thank you, I've changed it from if variable > "": to if variable: - I'll try to remember to do that in the future, too. Thanks!

bitcraft (approved):
looks good, thank you

PR #1103, opened 2022-01-09
Add print action. Closes #1100

### LINKED ISSUE #1100: Add a 'print variable' action
Debugging map events is hard, especially when there's loads of events all on the same map. 
Sometimes you don't know *where* an error is being triggered, like if you have 20 actions all on the same event, is it failing on the first one or the last one? 
Being able to print out messages would help, because you could 
It would also be nice to see what 

Something like a new action, like:
``` print ```
Would print out all variable names and their values in the player.game_variables dictionary.
``` print <variable_name> ```
Would print out the value of variable_name, if it's in the player.game_variables dictionary, and say 'Variable <variable_name> not found' or something if it's not found. 

See also #1087 - a request to print out game variables from the command-line interface, which is similar to the request here.
By having it as an action, you can automate the debugging process while testing maps, by making it print variable values before and after an event where you think there's an error.

### PR DESCRIPTION
**Pull request by ultidonki**
This adds a print action, callable from map events. See #1100 for details

The following map tests printing a bunch of variables, some that exist, some that don't:
[test_print_action.zip](https://github.com/Tuxemon/Tuxemon/files/7835012/test_print_action.zip)

### DIFF
--- a/tuxemon/event/actions/print.py
+++ b/tuxemon/event/actions/print.py
@@ -1,0 +1,65 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+from __future__ import annotations
+import logging
+
+from tuxemon.event.eventaction import EventAction
+from typing import NamedTuple, Optional
+
+logger = logging.getLogger(__name__)
+
+
+class PrintActionParameters(NamedTuple):
+    variable: Optional[str]
+
+
+class PrintAction(EventAction[PrintActionParameters]):
+    """
+    Print the current value of a game variable.
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tuxemon/event/actions/print.py [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+from tuxemon.event.eventaction import EventAction
+from typing import NamedTuple, Optional
+
+logger = logging.getLogger(__name__)
+
+
+class PrintActionParameters(NamedTuple):
+    variable: Optional[str]
+
+
```
bitcraft, line 34: Should be decorated by `@final`
```python
from typing import final

@final
class Foo...
```

tuxemon/event/actions/print.py
```diff
@@ -0,0 +1,65 @@
...
+class PrintAction(EventAction[PrintActionParameters]):
+    """
+    Print the current value of a game variable.
```
bitcraft: Please change to "Print the current value of a game variable to the console".  Some people might think it is printed to the game screen.

tuxemon/event/actions/print.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    name = "print"
+    param_class = PrintActionParameters
+
+    def start(self) -> None:
+        player = self.session.player
+
+        variable = self.parameters.variable
+        if variable and variable > "":
```
bitcraft: Please don't use the `>` operator to compare strings.  `if variable` is sufficient to check for empty strings.

tuxemon/event/actions/print.py
```diff
@@ -0,0 +1,65 @@
...
+
+class PrintActionParameters(NamedTuple):
+    variable: Optional[str]
+
+
+class PrintAction(EventAction[PrintActionParameters]):
+    """
+    Print the current value of a game variable.
+
+    If no variables are specified, print out values of all game variables.
```
bitcraft: please change "variables" to "variable", since only one variable is supported.

bitcraft (changes requested):
thank you for adding the documentation to the docs.  please see the comments.

ultidonki:
Thank you - I've made the changes requested in each comment now!

bitcraft (approved):
lgtm

PR #1104, opened 2022-01-09
Fix crash if moving cursor when only 1 item in your bag. Closes #1088

### LINKED ISSUE #1088: Menu crashes if you move cursor when there's only 1 menu item in it
**Describe the bug**
Start the game with CLI enabled, and add 1 item to your bag (it doesn't happen if the bag is empty, or has 2 or more items in it).
``` action add_item potion,1 ```

Now go into Bag, and push up, down, left, or right. 

The game will crash with:
```
File "tuxemon/sprite.py", line 414, in determine_cursor_movement
    raise RuntimeError
```

Caused by the following 2 lines:
```
    if index == original_index:
        raise RuntimeError
```
But in a menu with only 1 item, index will always be equal to original_index. 

**Other info**
If you add 2 items:
``` 
action add_item potion,1 
action add_item revive,1
```
and move around on the Bag screen, the error doesn't happen.

ultidonki:
A quick fix to fix #1088 - see that issue for details on how to replicate, and discussion on how to fix it.

### DIFF
--- a/tuxemon/sprite.py
+++ b/tuxemon/sprite.py
@@ -408,9 +408,6 @@ class MenuSpriteGroup(SpriteGroup[_MenuElement]):
             while seeking_index:
                 index = self._advance_input(index, event.button)
 
-                if index == original_index:
-                    raise RuntimeError
-
                 seeking_index = not self.sprites()[index].enabled
 
         return index


bitcraft (approved):
Looks good, thank you!

PR #1094, opened 2021-12-12
WIP Buying and selling concept 1

ultidonki:
Please don't merge! 

This PR contains all of the functionality of buying and selling that I want to add. And it should be bug-free. 
So feel free to play with it, or code-review it. 

But there's all sorts of weird stuff I want to find a cleaner solution for. (For example, I might split the ShopMenuState into both a BuyShopMenuState, and a SellShopMenuState, because right now I only know whether the player is buying or selling by checking if buyer == None. Which is a bit hacky, but it's how it was working before I started looking at it.)

I might have more time to clean it up in the new year. (Unless someone else wants to clean it up first!)

Here's what it looks like: 
buying_shop_2

Features include:
- You can now create 'economies' in .json, which tell you the sell/buy price of a certain shop. See #429 
- Economies have parent-child relationships, so if a buy/sell price isn't in the current economy, it'll look up the buy/sell price in the parent economy.
- Player now has a game variable of 'money', used as currency in the shop. See #668 
- Buying now shows you: Price of item, how much money the buyer has, how many items in the bag the buyer already has. 
- Selling now shows you: Price you'll gain by selling, how much money the seller has, how many items in the bag the seller already has. 
- Buying and selling now increases/decreases the 'money' in both the buyer and seller's 'money' variable. 
- When buying, you cannot choose a quantity higher than you can afford. 

There's a temp map for testing in there - shops.tmx 
On this map, you gain 1 potion per frame, so the amount of potions you have in your bag will always be wrong, as it increases even when you're on the buying/selling menu. 
If you take 1 step in any direction, you'll gain $100. 
(Which should make testing it easier)

If you sell the item at the bottom of the sell menu, the game crashes with an 'assert selected' error. This happens in development too, I haven't had time to fix it.

Issue #429, opened 2018-07-11
Add buy/sell prices to items "Economies"


bitcraft:
Part of #421 
========

Implement a method of storing and retrieving item prices.  One possible way to make this extensible would be to create a namespace for prices.  Call it "economy". 

* An economy is defined by buy/sell prices of items.
* Alternate economies can be defined
* All NPC's/shops are part of the "default" economy. 
* Economies have a parent/child relationship, children override parents, parents provide value not in children.

A different economy can be used for a different region, affecting prices of all items.  Economies should be able to be defined in JSON.

This issue tracks the creation of Economies.

bitcraft:
Discussion here: https://forum.tuxemon.org/thread.php?pid=2241

ultidonki:
I'm going to start working on this - as I need it for a mod idea I have.
(Unless anyone else has started on this, which would be welcome!) 
I won't be doing any of the other parts of #421, only this one. 

As a proof of concept, the JSON will be:
```
{
  slug: "my_new_economy",
  parent: "default", # If item price is not defined in here, look it up in the 'default' economy 
  items: [
    {
      item_name: "potion",
      price: "20", # Price is what you buy at
      cost: "5" # Cost is what you sell at
    },
   {
      item_name: "revive",
      price: "100",
      cost: "25"
    }
  ]
}
```

There will probably also be these 2 functions:
```get_item_price(item_slug, economy_slug) -> Integer ```
```get_item_cost(item_slug, economy_slug)  -> Integer ```
but I haven't got that far yet.

PR #1105, opened 2022-01-09
Create economies, and functions to lookup item price and cost

ultidonki:
This is part 1 of adding buying/selling to shops in tuxemon, it's split off from PR #1094 - I'm splitting that PR into smaller commits so they're easier to test and understand. (This one is still big, it's about 50% of the lines of code of the original PR!)

This PR only adds economies, as preparation for the buying/selling features which I'll add in later PR's. 
These set buy/sell prices/costs on items, which will be used later by shops to set prices. 
It allows economies to be loaded from .json, and it allows economies to be set on an NPC using a set_economy action.
At the minute, the set_economy action doesn't do anything, but it'll be used later when I change the shops code. 

You can't actually use the economies in shops yet, so I've added some unit tests around the price/cost lookup functions.

I've also generalized the price/cost lookup functions into a lookup_item_field(item_slug, field) function - which was requested on a comment in #1094 
And I've removed the ability to recursively look up parent economies, which was requested in #429 - I want to keep each PR as simple as possible. I can add that again in a later PR.


xirsoi:
I'd like to present an alternative orientation that would require minimal changes to your implementation.
What if an Economy held price *modifiers* rather than prices? These modifiers would be applied to the base price, found in the item definition.

I'd love to get input from @Qiangong2 and @Sanglorian (or others) on this idea.

Otherwise this looks great!

ultidonki:
Sure, that's fine to discuss putting it on items, either way sounds good to me! I can make the changes if we decide to go that way. Thanks for looking at this!

Qiangong2:
In Pokemon, ceetain events or items can lower prices for a short span of time. With modifiers, we could attribute them to items which would give us more flexibility on pricing.

bitcraft (dismissed):
looks good to me.  this is a nice start.

bitcraft (approved):

Issue #790, opened 2020-05-19
Locking controls is not working [untested]


Qiangong2:
Normally, after using player_stop, the player cannot move until player_resume has been called. However, the game acts like player_resume has been called if the event sequence ends. This is problematic for long events as it forces them to be overly long/complicated, or split into multiple tiny ones. player_stop should continue to be enforced across events until player_resume is called

kerizane:
Can you provide an example?

Qiangong2:
> Can you provide an example? 

After you listen to Allie and go to leave, your mom stops you and brings you inside. If you run while the pathfind happens, since that's the last event, pathfind stops and player resume is initiated.

bitcraft:
Stop, as it applies to all on screen sprites (players and NPC) is directed at the current movement, nothing to do with controls specifically (only referred to for the player because reseting controls is required....the underlying changes to the sprite are the same for the player sprite and npcs).  Here is the documentation for it:
```python
    def stop_player(self):
        """ Reset controls and stop player movement at once.  Do not lock controls

        Movement is gracefully stopped.  If player was in a movement, then
        complete it before stopping.

        :return:
        """
```
The player resume function is very old and idk if it has worked for a long time (years...).  I can confirm that right now it doesn't do anything.  It should be removed or fixed.

I don't see a problem with stop at the moment.  If you are having issues because the player can move the char. around between events, you should lock the controls....which i guess is not possible.  We need an event for that.

bitcraft:
> If you run while the pathfind happens, since that's the last event, pathfind stops 

To go a bit further on, there is only one path for any on-screen sprite.  There is no "stack" that allows us to "go back" to previous movements like a pathfind, or anything else.

Sanglorian:
I've corrected the [event reference](https://wiki.tuxemon.org/index.php?title=Event_Reference).

Should we put a request for a way to lock player controls in a separate issue or treat this issue as that request?

bitcraft:
I think that the core issue is with ```lock``` and ```unlock```.  I'll update the title to reflect that.

Jameskmonger:
I've got a WIP fix for this. Should be completed soon https://github.com/Tuxemon/Tuxemon/pull/840

PR #840, opened 2021-02-27
WIP Expose lock_controls as an action

Jameskmonger:

PR #1111, opened 2022-01-23
Add actions to lock and unlock controls.

### LINKED ISSUE #1110: Create sink state to intercept input during cutscenes - Alternative to lock controls
Player can move during cut scenes, often breaking the flow of campaigns and allows you to see NPCs vanishing into thin air if you follow them.

This issue and its contents are based on a discussion on the Discord between @vnmabus and @bitcraft.

From my understanding of the discussion, this implementation would do the following:

1. Intercept input going to the world state
2. The related script action would create or  destroy the sink state (this would require map makers to ensure the sink is not left in the stack after the cutscene)
3. It would still be possible to use the menu or some other permitted action, as long as the sink state never passes the input to the next state (commonly the world state)

#840 

**Additional context**
NA

vnmabus:
This is done by creating a state that discards events going to lower
states in the stack. This allows the user to still interact with dialogs
created after the controls have been locked.

I have tested this adding the actions manually to the initial
taba_town.tmx cutscene.

Closes #1110. This is an alternative to #840 that is more general, as it
will even work with states that are not the WorldState (for example,
minigames).

### DIFF
--- a/tuxemon/event/actions/lock_controls.py
+++ b/tuxemon/event/actions/lock_controls.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+from typing import NamedTuple, final
+from tuxemon.event.eventaction import EventAction
+from tuxemon.states.sink import SinkState
+
+
+class LockControlsActionParameters(NamedTuple):
+    pass
+
+
+@final
+class LockControlsAction(
+    EventAction[LockControlsActionParameters],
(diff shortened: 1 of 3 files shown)


freshreplicant:
Closes #790

bitcraft (approved):
looks great, thank you!

PR #1113, opened 2022-01-30
Locking controls during cutscenes

Sanglorian:
Have set up lock_controls and unlock_controls on Spyder map events where I believe they're needed. Haven't been able to test as I don't have a running version of Tuxemon on my machine at the moment - so have marked as a WIP.


bitcraft:
@Sanglorian looks good to me.  i don't see any problems with it, so i can merge it

bitcraft (approved):

freshreplicant:
I tested these fixes, they work perfectly. 

My only two thoughts are:

1. The hacker should tell you where the cafe is, something like "The cafe is in the brick building just north of here!"
2. Maybe having the hacker walk on **and** off screen from the left is best, especially if the message above is added. That way he draws the player in the direction they must walk to find the cafe.

I can also make these changes separately when this PR is merged.

Sanglorian:
Thank you, @freshreplicant - have removed draft status. Would love if you made those changes once PR is merged.

PR #1114, opened 2022-02-02
Fix NPC appearing on map after random encounter

ultidonki:
This fixes an NPC appearing on the worldstate map after a random encounter, as discussed in Matrix.

This map shows the problem (and can be used to check the problem is fixed): Just enter the grass to start a random encounter, and after the battle the NPC will be in the top-left square of the map. 
[random_npc.zip](https://github.com/Tuxemon/Tuxemon/files/7989664/random_npc.zip)

During random_encounter, an NPC is created as part of the combat state, but not shown during the battle. But recently they started appearing in the top-left corner of the map after the battle was over. This might be because of some updates in the UI, which made NPC's always have a world now, but I haven't checked the root cause, it was suggested to just remove the NPC after the battle which I've done in this PR, and it seems to work. 

Instead of using an NPC called 'maple_girl', I've changed it to an NPC called 'random_encounter_dummy', because if someone used maple_girl on the map, it could've caused problems maybe if we now removed it? There's still a problem of relying on a file in /mods/ which might be removed by a mod creator when they start a new mod, but hopefully the name will make them think twice. For now I'll just leave it like that.

### DIFF
--- a/mods/tuxemon/db/npc/random_encounter_dummy.json
+++ b/mods/tuxemon/db/npc/random_encounter_dummy.json
@@ -1,0 +1,6 @@
+{
+    "slug": "random_encounter_dummy",
+    "sprite_name": "girl1",
+    "combat_front": "femaletrainer_front.png",
+    "combat_back": "maple_back.png"
+}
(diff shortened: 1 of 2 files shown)


xirsoi (approved):
Looks great! Thanks for getting this fixed so quickly :)

PR #1115, opened 2022-02-02
Delete further images featuring IP from docs folder

freshreplicant:
Title is self explanatory, removes two instances of IP from the docs folder. If these images are still important we can recreate them with Tuxemon assets.


bitcraft (approved):
nice catch, thanks!

Issue #1101, opened 2022-01-09
Game crashes if you catch more than 6 tuxemon


ultidonki:
**Describe the bug**
If you have 6 tuxemon, and catch, or are given (add_monster action) a 7th, the game crashes with the following traceback:
  File "/src/games/tuxemon/event/eventengine.py", line 461, in update_running_events
    action.start()
  File "/src/games/tuxemon/event/actions/add_monster.py", line 76, in start
    trainer.add_monster(current_monster)
  File "/src/games/tuxemon/npc.py", line 600, in add_monster
    self.monster_boxes[CONFIG.default_monster_storage_box].append(monster)
KeyError: 'Kennel'

Although it says it's a CONFIG error, I've deleted my tuxemon.cfg and still had the issue, so is there a problem with the default config?
When I printed self.monster_boxes, I think it was an empty dict? So trying to access it is with any key is probably causing the issue. 

**To Reproduce**
Load a new game, and run the following 7 times from the command-line:
``` action add_monster agnite,1 ```
It will crash on the 7th time.
You can also try this map which automates the action to run hundreds of times:
[test_7_tuxemon.zip](https://github.com/Tuxemon/Tuxemon/files/7834931/test_7_tuxemon.zip)
I've also confirmed it happens when you have 6 tuxemon in your party, and catch a 7th in battle. 

**Expected behavior**
The 7th tuxemon should go in a box.

**Comment by freshreplicant** [comment]
I can confirm this, even if you rename the default_monster_storage_box value in tuxemon.cfg, it will crash and give a KeyError for whatever you changed the value to.

The blank dictionary suggests that maybe the monster_boxes dictionary isn't being created properly.

If I try to explicitly create the key and a blank list as an item, it doesn't seem to happen (see code below). This isn't a solution though, because I'm sure that the dictionary should have been created elsewhere already.

```python
        monster.owner = self
        if len(self.monsters) >= self.party_limit:
            self.monster_boxes[CONFIG.default_monster_storage_box] = []
            self.monster_boxes[CONFIG.default_monster_storage_box].append(monster)
        else:
            self.monsters.append(monster)
            self.set_party_status()
```

[...]

PR #1124, opened 2022-02-12
Fix bug in store_monster creating boxes with blank names

ultidonki:
If you don't pass a second parameter to store_monster, it creates a box called None (not the string, the actual None value) and stores the tuxemon in there.
```
get_player_monster temp
store_monster temp
```
Result: `monster_boxes = {None: [monster]}`

If you pass a 2nd parameter to store_monster, but it's a blank string, it creates a box with the empty string, "", and stores the tuxemon in there.
```
get_player_monster temp
store_monster temp,
```
(^ Notice the comma at the end of the 'store_monster' command, which wasn't on the first one)
Result: `monster_boxes = {'': [monster]}`

This PR fixes both bugs by using the default config box name if the box param is left blank.

### DIFF
--- a/tuxemon/event/actions/store_monster.py
+++ b/tuxemon/event/actions/store_monster.py
@@ -24,6 +24,7 @@
 
 from __future__ import annotations
 from tuxemon.event.eventaction import EventAction
+from tuxemon.prepare import CONFIG
 import logging
 import uuid
 from typing import NamedTuple, final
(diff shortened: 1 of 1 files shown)


freshreplicant:
Is this related to #1101?

ultidonki:
> Is this related to #1101?

I noticed it while testing your comment in that issue, but this won't fix #1101.

If you catch a tuxemon or use add_monster, you get the crash in #1101 after catching the 7th tuxemon.
In this problem, if you use store_tuxemon without passing a box parameter, it works (no crash), but the box name is weird, which this PR fixes. 

Rather than add this fix to the discussion in #1101, I decided to just fix it myself, as it's a small fix. :)

xirsoi (approved):
Great catch! Thanks for this.

PR #1125, opened 2022-02-12
Add menu to choose between default, and spyder campaign

ultidonki:
This PR adds a menu, after selecting 'New Game', that allows you to choose between the default and spyder campaign mods.

In the future, we should probably build the menu dynamically depending on what mod manifest files are found in /mods/*, but for now this menu allows you to choose either of the default campaigns without using the command-line, or modifying tuxemon.cfg

If you don't pass any command-line params to change the map, this is what you see:
mod_chooser

If you pass a -s for a custom starting map, or change the default starting map in tuxemon.cfg, you see an extra menu option, 'START':
I wanted to be able to skip the mod menu if you already had passed in a different map to start from, but I couldn't do it. 
Instead, selecting 'START' loads the map you passed in:
mod_chooser_2

This should preserve all functionality the screen used to have - for example, if you click cancel on the Name Input screen, you'll go back to this menu, added in #1045.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -693,6 +693,13 @@ msgstr "Press a key to set your new input key"
 msgid "options_new_input_key1"
 msgstr "Restart your game for changes to take effect!"
 
+# Default mod names on selection menu
+msgid "default_campaign"
+msgstr "Tuxemon"
+
+msgid "spyder_campaign"
+msgstr "Spyder and the Cathedral"
+
 # Multiplayer notifications
 msgid "multiplayer_accept"
 msgstr "Accept"
--- a/tuxemon/states/start/__init__.py
+++ b/tuxemon/states/start/__init__.py
@@ -68,7 +68,6 @@ class BackgroundState(State):
     def draw(self, surface: pygame.surface.Surface) -> None:
         surface.fill((0, 0, 0, 0))
 
-
 class StartState(PopUpMenu[StartGameObj]):
     """The state responsible for the start menu."""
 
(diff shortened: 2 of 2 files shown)


freshreplicant:
This is long overdue, great to see it being tackled.

Some thoughts about this feature and its implementation:

1. The 'default campaign' really needs a more unique name. It's been brought up before, but it feels weird having one called 'Tuxemon' or 'default' because it makes it seem like the Spyder campaign is somehow less official. This PR goes towards addressing that, but I still think they should both be called 'Tuxemon: {Insert Subtitle}'. It would also hit the nostalgia hard.
2. Can the gap between the two campaign names be increased?
3. Currently there's quite a lot of bugs/crashes for the 'default' campaign that still need to be addressed, but this PR at least lets you easily switch back and forth which is a major improvement on the current situation.

For a more long term implementation (maybe once we switch GUI framework) I think it'd be really cool to have them appear as game cartridges you can switch between with left and right, each with their own cover art (we have some concepts somewhere already).

ultidonki:
>     1. I still think they should both be called 'Tuxemon: {Insert Subtitle}'.

If this PR gets accepted, the name can be changed anytime by either of the mod authors, in base.po - I have no opinion on the names, they can be changed to whatever they want. :) 

>     2. Can the gap between the two campaign names be increased?

It looks like this can be done by adding a blank image with a height larger than the font, to space them out. 
I think it's the same spacing as the NEW GAME/LOAD GAME/ start menu, so I don't really see a problem with it. But if people want me to change it then I can. 

> For a more long term implementation (maybe once we switch GUI framework) I think it'd be really cool to have them appear as game cartridges you can switch between with left and right, each with their own cover art (we have some concepts somewhere already).

I think we'd need to wait for the new GUI before attempting something like that. :)
Although, if either mod author wants it, we can put an image next to the text - if you look at the Save/Load game menus, we can have an image similar to the screenshots that appear in there. (This would also increase the spacing between the choices, for your second point)

freshreplicant:
> If this PR gets accepted, the name can be changed anytime by either of the mod authors, in base.po - I have no opinion on the names, they can be changed to whatever they want. :)

No problem, can add it in another PR afterwards. As you might have seen on Matrix, I talked to Zhongtiao and he likes 'Tuxemon: Xero' as the name of his campaign.

> 
> > ```
> > 2. Can the gap between the two campaign names be increased?
> > ```
> 
> It looks like this can be done by adding a blank image with a height larger than the font, to space them out. I think it's the same spacing as the NEW GAME/LOAD GAME/ start menu, so I don't really see a problem with it. But if people want me to change it then I can.
> 

It might be good to do that here, just because we're dealing with more text.

> > **For a more long term implementation (maybe once we switch GUI framework)** I think it'd be really cool to have them appear as game cartridges you can switch between with left and right, each with their own cover art (we have some concepts somewhere already).
> 
> I think we'd need to wait for the new GUI before attempting something like that. :) Although, if either mod author wants it, we can put an image next to the text - if you look at the Save/Load game menus, we can have an image similar to the screenshots that appear in there. (This would also increase the spacing between the choices, for your second point)

Oh yeah, hence the 'long term implementation' part. An image might be neat actually, wonder if we have any existing assets that would work? Either way, that can be done in a later PR.

One last thing that's actually directly relevant to this PR: for the 'Start' option' is there any way to pull in the name from the tuxemon.cfg file? Maybe something like 'Load map from tuxemon.cfg: {starting_map}'

ultidonki:
> for the 'Start' option' is there any way to pull in the name from the tuxemon.cfg file? Maybe something like 'Load map from tuxemon.cfg: {starting_map}'

Yes, that sounds like a good idea - I should be able to do that. I'll try and add that at some point. :smiley:

xirsoi (approved):
This is a great addition, thanks for working on this @ultidonki 👍

PR #1129, opened 2022-03-05
Give both campaigns names & align names in new game menu

freshreplicant:
1. Changes default campaign to 'Tuxemon: Xero' from 'Tuxemon'. 
2. Adds 'Tuxemon' prefix to the 'Spyder and Cathedral' campaign for consistency. 
3. Changes the custom map start button to the name of the non-standard map set in tuxemon.cfg (by default this option is hidden until the default starting map is changed in the config file).
4. Centres the campaign names and puts the longest one on top for better readability.


bitcraft:
@Sanglorian name change looks ok to you?

Sanglorian:
> @Sanglorian name change looks ok to you?

Sure does, thanks for checking.

bitcraft (approved):
looks good, thank you!

PR #1134, opened 2022-03-16
Add 2 more battle backgrounds

ultidonki:


bitcraft (changes requested):
looks like changes made to the ocean background (cropped?)?  small technicality, but we must indicate if changes are made, according to the license 
https://wiki.creativecommons.org/wiki/Best_practices_for_attribution#This_is_a_good_attribution_for_material_you_modified_slightly

ultidonki:
Thank you, good catch! I've added a note that it's been cropped now.

bitcraft (approved):
thank you!

PR #1139, opened 2022-03-27
Add trainer line of sight battles in spyder route 2

ultidonki:
_Note: This is definitely a hacky, and slightly unmaintainable way of getting the battles working automatically... 
It requires 5 new event actions to get all the win/lose/automatic battle events to work.
Eventually, I'd like to create some actions to automate these instead, but for now this seems the best way to do it., but feel free to reject the PR if you think it would be better to create some actions to help instead - Otherwise I'll start adding this to every route/dungeon in the spyder campaign_  

New features:
- If you walk in front of one of the 3 trainers in Spyder's route 2, a battle will now start automatically. 
(Previously, you would need to talk to them)

- Also, if you lose, you will be teleported to the previous town (Spyder's Cotton Town) by using the teleport_faint action, and all your tuxemon will be healed. 
If you talk or walk in front of the trainer again, you will need to battle them again, repeating until you win. 

- If you win, you can't battle them again, or talk to them, just like it worked previously. 
(Previously, winning did nothing, you would stay in the same place with all your monsters defeated)

(If you lose during a random battle in the grass, or during the Billie battle that automatically starts on this map, you're not teleported back anywhere - I'll fix/test this in a future PR) 

(I've also fixed some messed-up tiles on this map, where rocks were appearing as stairs or roofs of houses, which is why some of the tilset stuff has been updated in the diff)

### DIFF
--- a/mods/tuxemon/maps/spyder_route2.tmx
+++ b/mods/tuxemon/maps/spyder_route2.tmx
@@ -1,30 +1,27 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<map version="1.4" tiledversion="1.4.2" orientation="orthogonal" renderorder="right-down" compressionlevel="0" width="40" height="20" tilewidth="16" tileheight="16" infinite="0" nextlayerid="2" nextobjectid="188">
+<map version="1.2" tiledversion="1.3.2" orientation="orthogonal" renderorder="right-down" compressionlevel="0" width="40" height="20" tilewidth="16" tileheight="16" infinite="0" nextlayerid="2" nextobjectid="202">
  <properties>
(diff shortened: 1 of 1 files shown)


ultidonki:
Maybe don't merge this one, actually - I think you could do this a better way using templates, if we get those working in the future. 

We could create some templates for trainer battles, like what to do if you win/lose/talk to them/get seen by them. Then, we could use that template everywhere, and if we ever need to update it, we'd just update the templates. The version in this PR needs to have each battle modified by hand if you want to make a change to all of them.

bitcraft:
closing at your request.  we can reopen it if needed

PR #1154, opened 2022-04-12
add ability to release

### LINKED ISSUE #867: way to abandon tuxemon and get more tuxeballs
I was playing the game and when I started finding tuxemon, they were pretty weak, but I still captured them. After a while I started finding higher level tuxemon, I started to doubt that capturing those weak tuxemon was a good idea. So I was thinking about a way to abandon some tuxemon. And then I started thinking about tuxeballs. That shop guy only gave me 5 tuxeballs, and so far there is no way to get more tuxeballs. Please make some tuxeballs to collect!

ddbazini:
This patch adds the ability for the player to release a Tuxemon from their party, simply by going into the menu screen. Before the Tuxemon is released, a confirmation window will appear that asks the player if they are sure about releasing it. Once the player confirms, the Tuxemon will no longer be in the player's party. I also made sure to add a check for whether or not this is the only Tuxemon in the party, as the player shouldn't be able to have 0 Tuxemon with them (past the beginning of the game).
All of the relevant dialogue strings were put into the relevant .po files, although they are currently all in English. 

In addition to testing with the CLI Interface, I also made some unit tests related to catching and releasing Tuxemon. They weren't there before, but they are still good to have as they account for edge cases, such as adding more than 6 monsters or releasing the last monster in the party.

This pull request should be enough to close Tuxemon#867 .


bitcraft (changes requested):
looks good.  the code is good, there are tests, you've added type hints, and there are comments about how the window is closing.  perfect!  i do have a couple comments --

1) i don't think we need to include the english for other translations.  in fact, i think we should avoid it because weblate will consider them "translated", ~~and it would be possible to mix english with another language during gameplay.  by leaving it out, there should be an error message during play, and the slug/msgid will be displayed instead.~~ by leaving them out the game will use the english translation automatically.

2) please use `black -t py38 -l 79 <filename>` to format the files after making changes.  so run it on the tests,  tuxemon/npc.py, and tuxemon/states/world/world_menus.py

vnmabus:
Actually, English messages (not slugs) are used as a fallback automatically.

bitcraft:
ah nice catch on the english fallback.  so yeah, no reason to put the english in the other languages.

ddbazini:
So just to clarify, I should:
1. remove my changes from all of the .po files (except for the English one)
2. Run the black command on the files
3. Make a new pull request

bitcraft:
>     1. remove my changes from all of the .po files (except for the English one)
yes

>     2. Run the black command on the files
please

>     3. Make a new pull request
no, just push the commits again and this pr will be updated

ddbazini:
Apologies for the multiple commits, the latest one properly only has the English .po file updated.

bitcraft:
multiple commits are not a problem

bitcraft (approved):

PR #1156, opened 2022-04-13
Misa's introduction

Qiangong2:
This PR is for introducing Misa to the player. It also introduces Kyle, an Omnichannel leader.

This PR also fixes:

The tiles in the Professor's Lab that broke due to a Tiled change.
Add sounds when your Tuxemon are healed 
Add sounds when you are given something


bitcraft (approved):
this looks great

PR #1163, opened 2022-04-23
Let own monster attack itself

Finnex:
Tuxemon now opens a dialog with an error msg if the own monster has been chosen for a damage technique.


bitcraft (approved):
looks great, thank you

PR #1164, opened 2022-04-25
Soundeffects now use value in config file for volume.

ScriptLineStudios:
Soundeffects now use value in config file for volume. Also lowered default value.


bitcraft (approved):
perfect, thank you

PR #1166, opened 2022-05-03
Pygame menu choice

vnmabus:
Refactor choice menus to use `pygame_menu`.

Created base class for states that consist in a `pygame_menu` menu.


bitcraft (approved):
more simple than i would have thought!  thank you

PR #1171, opened 2022-05-08
Fix controller input for new menues, and map B button correctly

ultidonki:
This fixes controllers to work with the new pygame menues, and makes B button on controllers work with the new menues too. :smiley:


vnmabus:
I think the patch is OK. We probably should ask the pygame-menu author to provide that attribute as a public attribute/property/init-parameter in order not to mess with the private implementation.

bitcraft (approved):

PR #1172, opened 2022-05-08
Always add black background even if skipping title screen

ultidonki:
The black background state should always be shown, even if you skip the title screen.

Otherwise, when on an indoor map, opening the menu leaves the menu still visible on any black areas, as it's not being re-drawn properly.


bitcraft (approved):
good catch, thank you

PR #1173, opened 2022-05-09
Add trainer battles if static NPC sees player

ultidonki:
@Sanglorian , do you want to make sure you don't have any changes you need to make before this gets merged, as it will be hard to merge our changes if you do? Do you have any changes in the Greenwash dungeon that haven't been pushed yet, for example? (This PR changes almost every map in the spyder campaign)

This PR adds trainer battles for static NPCs in the Spyder campaign, so if you walk 1-4 spaces in front of an NPC, the battle will trigger without you having to talk to them. NPCs that are set to 'wander' using npc_wander aren't affected, you still need to talk to them. 

This PR also contains a bunch of bug fixes I found when going through the spyder campaign, too:
- Some NPC's were spawning in the wrong place, or weren't appearing because there was a syntax error
- Fixed a bug in Nimrod's middle floor, where you can't get past the stairs when you walk onto that map for the first time. 
- The dragonbirth animation on route 3 wasn't appearing because of a syntax error (The animation now crashes when it plays but I'll fix that in another PR)
- The guard blocking the path on route 6, before Candy Town, now pathfinds out of the way after talking to him, so you no longer need to walk through the river to get to Candy Town
- Lots of doors were broken in Greenwash, I've fixed them all
- I added the one-way tile in Greenwash that pushes you back if you try to go in the opposite way to the arrow
- Added collisions to spyder_greenwash_level2.tmx


Sanglorian:
Thanks for checking, @ultidonki -- I don't have any changes.

bitcraft (approved):
i don't see any issues here

PR #1175, opened 2022-05-10
Add spyder shopkeepers

ultidonki:
This PR adds shopkeepers to the spyder campaign for most shops - but not the first one, in paper town (intentionally, as it wouldn't make sense for the story, I think).


bitcraft (approved):
lgtm

PR #1177, opened 2022-05-11
Add player_defeated condition

ultidonki:
This condition checks if the player is defeated, ie. has at least one tuxemon in their party, and all their tuxemon are defeated. 

The typical usage would be to teleport to the nearest health center if defeated:
`act1 teleport_faint
cond1 is player_defeated`

But modders could also do story events if the player was defeated, too. 

The python that does the check could maybe be easier to read/use. Happy to change it if needed.


bitcraft (approved):
looks great.  thank you for documenting everything

PR #1190, opened 2022-05-21
Add/remove strange collisions in spyder route 3, fixes #1189

### LINKED ISSUE #1189: Missing collision in spyder_route3.tmx allows to bypass the bolder
There is a collision missing in spyder_route3.tmx, which allows to bypass the boulder (see attached images)
collision_missing_1
collision_missing_2

ultidonki:
Adds the missing collision mentioned in #1189, and removes a strange collision in the middle of the grass just above the boulder that I noticed, too.


bitcraft (approved):
not really a huge deal, but you tiled version is ancient.  in this case i don't think it matters here, but might be possible to introduce problems with incompatible maps by downgrading the version.  for this change, its not a problem, just something to think about.

PR #1191, opened 2022-05-29
Updated tilesets: core outdoor nature and core outdoor water

rubberduck9999:
it contains the updated tilesets core outdoor nature and core outdoor water, a few sample maps and even an updated prototyping tileset.
I have basically split the current core tileset into nature and water, meant to replace the old one soon, next planned are city props and buildings


bitcraft (approved):
looks great!  thank you!

Issue #1188, opened 2022-05-20
Trying to release a tuxemon crashes the game


rubberduck9999:
Trying to release a tuxemon crashes the game with the following error message (see attached image)

Also I think it is a good Idea to ask the player if he really wants to release a tuxemon, if it is not already so. I just don't know and it crashes right now

![release_1](https://user-images.githubusercontent.com/82282845/169586936-26354061-5bde-4499-9d45-1a734f556b8b.png)
![release_2](https://user-images.githubusercontent.com/82282845/169586941-2849516c-5f0f-4212-a7fd-91e2658b1270.png)

rubberduck9999:
I tested it today and worked, so I think this can be closed now, or is there a reason it is still open?

ultidonki:
Yes, it can be closed - I think the issue is only closed automatically, if someone puts 'Closes #xxx' or 'Fixes #xxx' in the commit message, or the title of the message you write when opening the PR in github. @daniellubovich - just to let you know you can do this for the future, as I think every Github project can do this. It's not necessary though! So yeah, this can be closed. :smiley:

rubberduck9999:
ok, I understand, closing it now

PR #1193, opened 2022-06-01
Fix crash when releasing monsters

daniellubovich:
This is a simple, inelegant fix for the issue described here:
https://github.com/Tuxemon/Tuxemon/issues/1188.

The fix just updates the confirmation submenu to use PygameMenuState and
follow the pattern set by the other monster submenu.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -633,7 +633,7 @@ msgid "release_confirmation"
 msgstr "Are you sure you would like to release {name}?"
 
 msgid "tuxemon_released"
-msgstr "{name} has been released. Refresh menu to view change."
+msgstr "{name} has been released."
 
 # Menu notifications
 msgid "log_off"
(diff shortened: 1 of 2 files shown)


daniellubovich:
Hey! Just saw there was an open bug that didn't seem too crazy and decided to shoot my shot. Any feedback is appreciated - just hoping to get involved wherever help is needed.

ultidonki:
This looks like a good solution - thanks! Looking at git blame, I think the 'release tuxemon' feature was added using the old menu system, but these menues were updated to the new pygame_menu at the same time - by the time both branches were merged, the 'release tuxemon' menu was still using the old menu system, which was causing the crash. 

I think your change to update them to the new pygame menu system is the right one to fix it. 

One small thing I noticed is you're working on your 'development' branch - it should be fine but if you create a new branch for each pull request in the future, it can be easier - for example, you can raise several pull requests at the same time, each one from a different branch. :smile:

bitcraft (approved):
looks good, ty!

daniellubovich:
Awesome, thanks for the feedback!

PR #1194, opened 2022-06-02
Add more robust validation of various db models

ShadowApex:
Add some additional verification that sprites exist, slug references, and translation references.

### DIFF
--- a/mods/tuxemon/db/technique/boulder.json
+++ b/mods/tuxemon/db/technique/boulder.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://raw.githubusercontent.com/Tuxemon/Tuxemon/development/schemas/technique-schema.json",
   "accuracy": 1,
   "animation": "shield_rock",
   "category": "physical",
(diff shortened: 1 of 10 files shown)


mods/tuxemon/db/technique/boulder.json
```diff
@@ -11,7 +12,7 @@
   "is_fast": false,
   "potency": null,
   "power": 1.0,
-  "range": "special",
+  "range": "ranged",
```
bitcraft: boulder is a "special" technique because it doesn't apply damage, and instead applies the effect "hardshell".  this move in particular always comes up because its assumed to be an attack.  https://wiki.tuxemon.org/index.php?title=Boulder

bitcraft (commented):
looks great.  just the one comment about boulder, because it should probably stay as "special", unless we have some reason to make it ranged

mods/tuxemon/db/technique/boulder.json
ShadowApex: Ah, I see. It looks like we don't yet have an implementation of the "hardshell" effect. I can stub this out and add a TODO to implement the effect logic.

bitcraft (approved):
great changes, thank you

PR #1197, opened 2022-06-08
Shops: Make buying items cost money, and don't allow buying items unless you have enough money

ultidonki:
Currently, opening a shop shows how much everything will cost, but you can still buy everything without paying for it.

This PR stops the player from buying items unless they have enough money. 
It also exchanges money between NPC's during the buying transaction. 

It doesn't change anything for the Selling menu (if you sell something, you still don't get any money), but that should be a small PR in the future that uses the same NPC.buy_transaction function from this PR. 

This is hard to test in the normal mods so you can test on this map:
[shops_test.zip](https://github.com/Tuxemon/Tuxemon/files/8862822/shops_test.zip)

You start with no money, and walking 1 step gets you $100 on this map, so you can quickly make money and test the functions work properly.

### DIFF
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -909,7 +909,7 @@ class NPC(Entity[NPCState]):
 
         # Only players will have money in game_variables, so only update money
         # if it exists:
-        if self.game_variables.get("money"):
+        if self.game_variables.get("money") is not None:
             self.game_variables["money"] += qty * unit_price
 
     def buy_transaction(
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -848,8 +848,8 @@ class NPC(Entity[NPCState]):
         return success
 
     def can_buy_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
-        current_money = self.game_variables["money"]
-        if current_money:
+        current_money = self.game_variables.get("money", None)
+        if current_money != None:
             return current_money >= qty * unit_price
         # If no 'money' variable, must be an NPC. Always allow buying:
         return True
@@ -875,9 +875,9 @@ class NPC(Entity[NPCState]):
 
         Raises an exception if there's not enough money to pay the price."""
 
-        # Only players will have game_variables, so only update money or raise
+        # Only players will have money in game_variables, so only update money or raise
         # exception if it exists:
-        if self.game_variables:
+        if self.game_variables.get("money", None):
             if not self.can_buy_item(item_slug, qty, unit_price):
                 raise Exception(
                     f"Tried to buy item with {self.game_variables['money']} coins "
@@ -907,9 +907,9 @@ class NPC(Entity[NPCState]):
                 f"Tried to sell item of qty {qty}, but not enough available."
             )
 
-        # Only players will have game_variables, so only update money
+        # Only players will have money in game_variables, so only update money
         # if it exists:
-        if self.game_variables:
+        if self.game_variables.get("money", None):
             self.game_variables["money"] += qty * unit_price
 
     def buy_transaction(
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -848,8 +848,8 @@ class NPC(Entity[NPCState]):
         return success
 
     def can_buy_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
-        current_money = self.game_variables.get("money", None)
-        if current_money != None:
+        current_money = self.game_variables.get("money")
+        if current_money is not None:
             return current_money >= qty * unit_price
         # If no 'money' variable, must be an NPC. Always allow buying:
         return True
@@ -877,7 +877,7 @@ class NPC(Entity[NPCState]):
 
         # Only players will have money in game_variables, so only update money or raise
         # exception if it exists:
-        if self.game_variables.get("money", None):
+        if self.game_variables.get("money"):
             if not self.can_buy_item(item_slug, qty, unit_price):
                 raise Exception(
                     f"Tried to buy item with {self.game_variables['money']} coins "
@@ -909,7 +909,7 @@ class NPC(Entity[NPCState]):
 
         # Only players will have money in game_variables, so only update money
         # if it exists:
-        if self.game_variables.get("money", None):
+        if self.game_variables.get("money"):
             self.game_variables["money"] += qty * unit_price
 
     def buy_transaction(
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -856,20 +856,29 @@ class NPC(Entity[NPCState]):
 
     def can_sell_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
         current_item = self.inventory.get(item_slug)
-        if item_slug:
+        if current_item:
             current_qty = current_item["quantity"]
             return current_qty >= qty or current_item["infinite"]
         # If they don't have the item, then they can't sell the item:
         return False
 
-    def buy_item(
-        self, session: Session, item_slug: str, qty: int, unit_price: int
+    def buy_decrease_money(
+        self,
+        session: Session,
+        seller: NPC,
+        item_slug: str,
+        qty: int,
+        unit_price: int,
     ) -> None:
-        """Decreases current money, and adds the items to the inventory.
+        """Decreases current money during a buy transaction, but doesn't change
+        an item's quantity (use NPC.give_item too after this)
 
         Raises an exception if there's not enough money to pay the price."""
+
+        # Only players will have game_variables, so only update money or raise
+        # exception if it exists:
         if self.game_variables:
-            if self.game_variables["money"] < qty * unit_price:
+            if not self.can_buy_item(item_slug, qty, unit_price):
                 raise Exception(
                     f"Tried to buy item with {self.game_variables['money']} coins "
                     f"but not enough money:\n"
@@ -878,30 +887,28 @@ class NPC(Entity[NPCState]):
 
             self.game_variables["money"] -= qty * unit_price
 
-        current_item = self.inventory.get(item_slug)
-        if current_item:
-            current_item["quantity"] += qty
-        else:
-            self.inventory[item_slug] = {
-                "item": Item(session, self, item_slug),
-                "quantity": qty,
-            }
-
-    def sell_item(
-        self, session: Session, item_slug: str, qty: int, unit_price: int
+    def sell_increase_money(
+        self,
+        session: Session,
+        buyer: NPC,
+        item_slug: str,
+        qty: int,
+        unit_price: int,
     ) -> None:
-        """Increases current money, and removes the items from the inventory.
+        """Increases current money during a sell transaction, but doesn't change
+        an item's quantity (use NPC.give_item too after this)
 
         Raises an exception if there's not enough items in the inventory."""
         current_item = self.inventory.get(item_slug)
-        if not current_item or (
-            current_item["quantity"] < qty and not current_item["infinite"]
+        if not current_item or not self.can_sell_item(
+            item_slug, qty, unit_price
         ):
             raise Exception(
                 f"Tried to sell item of qty {qty}, but not enough available."
             )
 
-        current_item["quantity"] -= qty
+        # Only players will have game_variables, so only update money
+        # if it exists:
         if self.game_variables:
             self.game_variables["money"] += qty * unit_price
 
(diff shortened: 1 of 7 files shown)


tuxemon/npc.py
```diff
@@ -847,5 +847,79 @@ def alter_item_quantity(
...
+    def can_buy_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
+        current_money = self.game_variables["money"]
+        if current_money:
+            return current_money >= qty * unit_price
+        # If no 'money' variable, must be an NPC. Always allow buying:
+        return True
+
+    def can_sell_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
+        current_item = self.inventory.get(item_slug)
+        if item_slug:
```
bitcraft: should this var be `current_item`?

tuxemon/npc.py
```diff
@@ -847,5 +847,79 @@ def alter_item_quantity(
...
+        # If they don't have the item, then they can't sell the item:
+        return False
+
+    def buy_item(
+        self, session: Session, item_slug: str, qty: int, unit_price: int
+    ) -> None:
+        """Decreases current money, and adds the items to the inventory.
+
+        Raises an exception if there's not enough money to pay the price."""
+        if self.game_variables:
```
bitcraft: small suggestion, but not needed.  this check for money, quantity, and price is essentially the same thing as `can_sell_item`.  you could consider using that here as a check instead of implementing it again here.

tuxemon/npc.py
```diff
@@ -847,5 +847,79 @@ def alter_item_quantity(
...
+            if self.game_variables["money"] < qty * unit_price:
+                raise Exception(
+                    f"Tried to buy item with {self.game_variables['money']} coins "
+                    f"but not enough money:\n"
+                    f"price {unit_price} times qty {qty} is {qty * unit_price}"
+                )
+
+            self.game_variables["money"] -= qty * unit_price
+
+        current_item = self.inventory.get(item_slug)
```
bitcraft: i think this could be replaced by `NPC.give_item`, which would handle the details

tuxemon/npc.py
```diff
@@ -847,5 +847,79 @@ def alter_item_quantity(
...
+                "quantity": qty,
+            }
+
+    def sell_item(
+        self, session: Session, item_slug: str, qty: int, unit_price: int
+    ) -> None:
+        """Increases current money, and removes the items from the inventory.
+
+        Raises an exception if there's not enough items in the inventory."""
+        current_item = self.inventory.get(item_slug)
```
bitcraft, line 902: looks like a duplicate of `can_sell_item`

bitcraft (commented):
nice feature!  please see the comments

ultidonki:
Thanks! I've fixed all those suggestions now, they all look good. :smiley:

tuxemon/npc.py
```diff
@@ -847,5 +847,87 @@ def alter_item_quantity(
 
         return success
 
+    def can_buy_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
+        current_money = self.game_variables["money"]
+        if current_money:
+            return current_money >= qty * unit_price
+        # If no 'money' variable, must be an NPC. Always allow buying:
```
kerizane, line 854: Inconsistent with your other comment.

\# Only players will have game_variables, so only update money
        \# if it exists:

ultidonki: I'm not sure what's inconsistent here, sorry - players have a custom 'money' variable, but NPC's don't. 

What I'm trying to explain with the comment is:
NPC's should not have a 'money' variable (for now) so it's not possible to do a check on how much money they have. So I need to always let the buy/sell transaction succeed, if the NPC has no 'money' variable. And that's fine, because they're an NPC, players won't be able to abuse this to get their buy transactions to always succeed. 

Player's override NPC, and always do have a 'money' variable when they're initialized, I've added that variable recently in tuxemon/player.py 
Therefore, if there IS a 'money' variable, then there is a player, and we need to
a) Do a check to see if there's enough money to buy something
b) Increase/decrease the money variable if the buy/sell transaction succeeds. 

I'm happy to rewrite the comment to make more sense (or maybe there is a bug there I haven't seen?)

I guess we COULD actually add a 'money' variable to the NPC's as well, and then you can only sell items to them if they have enough money? It would simplify the code, as both buyer and seller would always be NPC's, and always have a 'money' variable. 
I think there was a discussion about doing something like that. But for now I've only added the 'money' variable to the player. I'm happy to make that change if you think treating NPC and players the same would be better.

kerizane:
```
        current_money = self.game_variables["money"]
        if current_money:
```
vs
```
        # Only players will have game_variables, so only update money or raise
        # exception if it exists:
        if self.game_variables:
```
The logic is wrong in one of these places. 

For NPCs, either game_variables is an empty dict, or it has "money" set to 0, which is it? If it's an empty dict then you will get a key error when you check "money".

Also, this will allow players to buy things if they have 0 money. Probably not what you wanted.

kerizane (changes requested):
Also, what happens with the exceptions you are raising? Do we end up logging an error just because a player tried to buy something that was too expensive? We want some kind of game feedback to the player, but shouldn't clutter the logs with this.

ultidonki:
> Also, what happens with the exceptions you are raising? Do we end up logging an error just because a player tried to buy something that was too expensive? We want some kind of game feedback to the player, but shouldn't clutter the logs with this.

The exceptions are intended for programmers/users of the NPC.buy_transaction API - programmers should always check the transaction is possible before starting it, using can_buy_item and can_sell_item. The exception will crash the game for them to let them know there is a problem, if they haven't already used these functions. I discussed this with Bitcraft who said they preferred an exception to just crash the game instantly, rather than log it or add any complex logic to try and handle it, as it immediately lets developers know what's gone wrong rather than let the game go into an inconsistent state. 

In the 'ShopBuyMenuState', in tuxemon/states/items/__init__.py, I'm already setting the max quantity of item you can buy to be as much as you can afford, or the max quantity the shop has, or 0 if you can't afford anything. So you should never hit the exception normally, the player isn't allowed to select a quantity higher than 0.

tuxemon/npc.py
ultidonki: > For NPCs, either game_variables is an empty dict, or it has "money" set to 0, which is it? If it's an empty dict then you will get a key error when you check "money".

> Also, this will allow players to buy things if they have 0 money. Probably not what you wanted.

Oops, you're right on both counts. This function was only tested on Players, not NPC's - and I'm not sure how I let it happen that you could buy things with 0 money. I'll fix these, and I might add unit tests around the can_buy and can_sell functions to ensure they're being tested properly. Thanks!

ultidonki:
OK, I've made some changes:
1. NPC's do actually always have game_variables set to an empty dict, so I removed checked for game_variables
2. Only players have the 'money' variable, so I used game_variables.get("money", None) to check if money exists safely
3. Fix bug in can_buy_item where if you had 0, it would allow you to buy an item, by replacing the check with `if current_money != None`
4. I've decided not to add unit tests in this PR, because I had trouble creating the NPC in the unit tests without a lot of dependencies, I'll raise an issue for it separately. 

I've tested buying things from the shop in the new menu again and everything's still working fine.

tuxemon/npc.py
```diff
@@ -847,5 +847,87 @@ def alter_item_quantity(
 
         return success
 
+    def can_buy_item(self, item_slug: str, qty: int, unit_price: int) -> bool:
+        current_money = self.game_variables.get("money", None)
```
kerizane: This is picky, but `None` is the default argument for `get` so you don't need to specify it. 
Just do the following:
```
        current_money = self.game_variables.get("money")
```

I checked the rest of the repo and there are only 3 cases where we use .get(key, None) and 63 cases where we use .get(key), so this way is more consistent.

tuxemon/npc.py
```diff
@@ -847,5 +847,87 @@ def alter_item_quantity(
...
+        if current_money != None:
```
kerizane: Should be 
`        if current_money is not None: `

https://peps.python.org/pep-0008/#programming-recommendations
"Comparisons to singletons like None should always be done with is or is not, never the equality operators."

kerizane: (For reference, same applies for `is True` and `is False`, although you would generally be using `if foo` rather than `if foo is True`)

kerizane (changes requested):

tuxemon/states/items/__init__.py
```diff
@@ -393,36 +393,42 @@ def buy_item(quantity: int) -> None:
...
-            None if item_dict.get("infinite") else item_dict["quantity"]
-        )
+
         price = (
             1
             if not self.economy
             or not self.economy.lookup_item_price(item.slug)
             else self.economy.lookup_item_price(item.slug)
         )
+        qty_buyer_can_afford = (
```
kerizane: Shorter
```
        money = self.buyer.game_variables.get("money")
        qty_buyer_can_afford = int(money / price) if money else 0
```

ultidonki:
Thanks for the review, they all look like good changes! - I'll make these changes and test again when I have time in the next few days.

ultidonki:
Thanks - I've made those 3 changes now. :smile:

tuxemon/npc.py
```diff
@@ -847,5 +847,87 @@ def alter_item_quantity(
...
+            item_slug, qty, unit_price
+        ):
+            raise Exception(
+                f"Tried to sell item of qty {qty}, but not enough available."
+            )
+
+        # Only players will have money in game_variables, so only update money
+        # if it exists:
+        if self.game_variables.get("money"):
+            self.game_variables["money"] += qty * unit_price
```
kerizane, line 913: What if we sell an item whilst having no money?

kerizane (changes requested):

kerizane:
Nearly there!

kerizane:
Hmm, how should we enable this? If we turn on this feature, but don't have the ability to get money then this breaks the game. Maybe we should allow the player to have infinite money, similar to items having infinite quantity?

tuxemon/npc.py
ultidonki: > What if we sell an item whilst having no money? 

Good catch - I've fixed this now (I'll admit I haven't tested the player selling anything yet - this PR only adds buying for the player. This will help fix the bugs in the selling when I do that too, though!)

ultidonki:
> Hmm, how should we enable this? If we turn on this feature, but don't have the ability to get money then this breaks the game. 

Good point - I've modified the starting maps for both the Spyder and Xero campaign, so the player starts with 1000000 moneys. This is enough to buy everything in every shop and still have 90% of it left over, which is good (as currently there's no way to display how much money you have left). The modders can change this later when we have more of the system working! :smiley:

tuxemon/states/items/__init__.py
```diff
@@ -393,36 +393,36 @@ def buy_item(quantity: int) -> None:
...
-        )
+
         price = (
             1
             if not self.economy
             or not self.economy.lookup_item_price(item.slug)
             else self.economy.lookup_item_price(item.slug)
         )
+        money = self.buyer.game_variables.get("money")
+        qty_buyer_can_afford = int(money / price) if money else 0
```
kerizane: What if it's free?

kerizane: You kinda try to handle this, but I'd it's free you should be able to buy all of it.

You also need to consider what happens if it's free and infinite.

ultidonki: > If it's free you should be able to buy all of it.
> 
> You also need to consider what happens if it's free and infinite.

I've updated this to handle free finite and infinite items, thanks. I also fixed a bug if the price was == 0, it was raising an error in tuxemon/items/economy.py

tuxemon/states/items/__init__.py
```diff
@@ -393,36 +394,42 @@ def buy_item(quantity: int) -> None:
...
+
         price = (
-            1
+            0
             if not self.economy
             or not self.economy.lookup_item_price(item.slug)
             else self.economy.lookup_item_price(item.slug)
         )
+        money = self.buyer.game_variables.get("money")
+        qty_buyer_can_afford = (
```
kerizane: I think you should split this into two branches.

If price is 0 then max quantity is (some sensible maximum) if it's infinite, else just the amount the seller has.

If price is not zero then it is the minimum of what you can afford Vs how many they have.

kerizane: If they have 10000 items and price is zero you can buy 10000. But if it is infinite and price is zero you can only buy 999. Seems wrong.

ultidonki: By 'splitting it into two branches', do you mean something like the following?
```
        MAX_QTY = 999

        if int(price) == 0:                                                                                                                                                                    
            if item_dict.get("infinite"):                                                                                                                                                      
                max_quantity = MAX_QTY                                                                                                                                                         
            else:                                                                                                                                                                              
                max_quantity =  min(MAX_QTY, item_dict["quantity"])                                                                                                                            
        else:                                                                                                                                                                                  
            qty_can_afford = int(money / price)                                                                                                                                                
            max_quantity = (                                                                                                                                                                   
                min(MAX_QTY, qty_can_afford, item_dict["quantity"])                                                                                                                       
            )                                                          ```

kerizane: Perfect

ultidonki:
Thanks, I've made that change now. I've also changed the label, so if an item doesn't cost anything it says '$ FREE!' instead of $0. :grin: 

free_in_shop

kerizane (approved):

kerizane:
Good work!

Issue #724, opened 2020-04-06
Script Actions to Stash/Unstash party and inventory


xirsoi:
To support the Crawl mechanic, or other story events, we should have actions for temporarily storing the player's party and inventory.

Several options present themselves, each with their own pros and cons:

1) Store them in game variables. Pro is that the script could dictate where it goes, though I'm not sure what good that would do. Con is that we'd have to make sure the Save function would know to serialize monsters and items correctly, as it currently only handles simple values like strings and numbers.

2) Store them in dedicated properties of the Player class. Much the same situation as in game variables, but it would be easier to update the Save function. Con is that we would have no ability to stash multiple parties/backpacks.

3) Simply drop all of it in computer storage. Pro is it's easily the simplest solution with few problems from an engineering perspective. Con is that the UX is not great. The player would have to manually find all their shit and pull it out of storage again. Another Con is that the computer storage mechanics haven't been written yet.

bitcraft:
> Store them in game variables

Save/load should be working just fine.  The simple values in there should be used to create new in-game instances of monsters, items, etc.  So, storing it in game variables seems fine since it will be json text.

> Store them in dedicated properties of the Player class

Minor point, this would be NPC.  I don't want NPCs and "Players" to differ substantially, since it is easier to maintain code that works on both.  I'm not sure that I like this suggestion either though, due to "backpacks", and such thing you mentioned.

> Simply drop all of it in computer storage

Computer storage should be a subset of "multiple inventories".  Ideally, I'd like a medium-sized refactor of inventories by replacing the dictionary with a class that we can serialize and apply attributes to, such as name, size, type of objects, etc.  Then we could develop a class which displays inventories to reuse code.  Way outside the scope of the ticket though.

I like the #1 approach best.  When a crawl starts, the current inventory is serialized then replaced with an empty one.  When the crawl ends, the inventory is dropped again and restored from the saved one.  If there are any deficiencies in the save/load code, they should be fixed.  In the short term, I think this is the best solution if we have momentum on the crawls feature.

xirsoi:
Completely agree with all of that.

I don't know if there's currently momentum on the feature, but I stumbled on the discussion in the forums relatively recently and really liked the idea. I thought at the very least I could introduce a ticket for one of the features required to get it done.

[...]

PR #1201, opened 2022-06-26
Add 'take monster' storage menu in PC

ultidonki:
This PR adds a new menu to the PC, allowing you to take a monster out of storage/a kennel. 

(Sorry for how big the PR and this message is!)

It only allows you to take a monster out of storage - I'll make a new PR in the future allowing you to put monsters into storage, too. 

It uses the new pygame_menu library, but I tried to make it look like this mockup from a seperate issue: #724 

Some screenshots:
storage1
storage2
storage3
storage4

(I'm doing this on a test map that's only 2 squares big, which is why the menu looks like it has a black background - but normally you'd see the tuxemon health centre behind the menu instead)

Here's all the bugs or things that don't work yet, but I think we can fix them in the future, if this PR looks OK: 
1. Doesn't work with different resolutions, only the default 1280x720
2. Mouse can't be used to move scroll bar yet
3. There's about a 1 second pause before entering the screen, with about 30 tuxemon, on my computer, which seems fine once you're used to it. But it might get bigger if more tuxemon are in the kennel?
4. Changing the theme (to make it vertically scrollable) from tuxemon/menu/menu.py:get_theme is a global change, so all menues become vertical scrolling after opening the Take Monster menu (if the items in the menu are big enough). Most noticable if you try to release a tuxemon after taking one from storage.
5. If the number of monsters is 2 less than the number of rows x columns, (like in the test map, pc_test.tmx) the last 2 rows have 2 tuxemon each (you would expect the 2nd-to-last row to have 3 tuxemon, and 1 in the final row). This is because the default pygame-menu Menu adds items vertically, not horizontally. Might be able to fix it by adding a Horizontal Frame, which adds the items horizontally, but when I tried it added more errors so I might try and fix this in the future.
6. The menu of kennel names doesn't scroll vertically, so if you have more than 10 or so the bottoms will start to be cut off, (but this only happens so far if you change the default kennel name in tuxemon.cfg, and reload the game, 10 times, so not a big deal yet)
6. (minor) Scrollbar isn't themed like the rest of tuxemon yet
7. (minor) The columns aren't aligned properly (eg. if all tuxemon in a column have a short, 4-letter name, that column becomes smaller. If one tuxemon has a 10-letter name, it becomes bigger (or normal-sized)
8. (minor) menu could be made prettier in general, maybe center-align the tuxemon faces below each name?

I can try to fix these issues in the future, or now if any of them are big enough. 

Here's a test map, which adds 6 tuxemon into your party and about 25 tuxemon into your default box, and you're standing on a 'PC' event, so just push Enter to be able to test the menu:

[test_pc.zip](https://github.com/Tuxemon/Tuxemon/files/8986581/test_pc.zip)


tuxemon/states/pc/__init__.py
```diff
@@ -188,3 +203,158 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
...
+        rows = int(num_mons * num_widgets_per_monster / columns) + 1
+        # Make sure rows are divisible by num_widgets
+        while rows % num_widgets_per_monster != 0:
+            rows += 1
+
+        super().startup(
+            height=height, width=width, columns=columns, rows=rows, **kwargs
+        )
+
+        def take_monster(monster: monster) -> None:
```
kerizane: Are you sure this typehint is right?

kerizane: Probably need to do the following:

Add `TYPE_CHECKING` to list of typing imports

Then after the imports:

```
if TYPE_CHECKING:
    from wherever import Monster
```

tuxemon/states/pc/__init__.py
```diff
@@ -188,3 +203,158 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
...
+    def add_menu_items(
+        self: MonsterTakeState,
+        menu: pygame_menu.Menu,
+        items: Sequence[Tuple[str, MenuGameObj]],
+    ) -> None:
+
+        width, height = prepare.SCREEN_SIZE
+
+        for monster, callback in items:
+            label = T.translate(monster.name).upper()
```
kerizane, line 236: According to the typehint monster is a str

kerizane (changes requested):
Run mypy or pyright to check your typehints

tuxemon/states/pc/__init__.py
```diff
@@ -188,3 +203,158 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
...
+    Shows all tuxemon currently in a storage kennel, and selecting one puts it
+    into your current party."""
+
+    def add_menu_items(
+        self: MonsterTakeState,
+        menu: pygame_menu.Menu,
+        items: Sequence[Tuple[str, MenuGameObj]],
+    ) -> None:
+
+        width, height = prepare.SCREEN_SIZE
```
kerizane: unused

tuxemon/states/pc/__init__.py
```diff
@@ -188,3 +203,158 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
...
+        super().startup(height=height, **kwargs)
+
+        self.animation_offset = 0
+
+        def change_state(state: str, **kwargs: Any) -> Callable[[], object]:
+            return partial(self.client.replace_state, state, **kwargs)
+
+        player = local_session.player
+        menu_items_map = []
+        for box_name in player.monster_boxes.keys():
```
kerizane:
```
for box_name, monsters in player.monster_boxes.items():
    if not monsters:
        ...
```

kerizane: Iterating over the keys and then immediately getting the value from the dict is unidiomatic.
Using `len(some_iterable) == 0` rather than `not some_iterable` is unidiomatic.

ultidonki:
Thanks for the review - I've made those changes now.

However, when I tried to run mypy on tuxemon/states/pc/__init__.py, it doesn't return any errors, even the ones above. So there's possibly more type errors that aren't being caught. Unfortunately I don't know enough about mypy to get it to work, this is the first time I've used it.

tuxemon/states/pc/__init__.py
```diff
@@ -188,3 +206,156 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
...
+
+
+class MonsterTakeState(PygameMenuState):
+    """Menu for the Monster Take state.
+
+    Shows all tuxemon currently in a storage kennel, and selecting one puts it
+    into your current party."""
+
+    def add_menu_items(
+        self: MonsterTakeState,
```
kerizane: Don't need to typehint self

kerizane:
This is a great! It's an important feature that I'd love to get in. 

I don't want to be too fussy, but I think the following are important:

I think having the name under the image is better.

If you scroll down to the bottom the final row of images aren't shown.

Fix the vertical bar side effect. You could add logic to the start and resume methods on the menu base class to set the appropriate theme. 

Only being able to take Tuxemon out is a bit odd, can you implement storing as well?

ultidonki:
Thanks, I've made the following change:
> I think having the name under the image is better.

Which partly fixes the following:
> If you scroll down to the bottom the final row of images aren't shown.

However, if you scroll up to the top of the screen again, the opposite problem happens - you can see the word 'Rockitten' but not the image. 
I don't think I'll be able to fix this in this PR:
I've tried adding pygame_menu.VFills, adding the image as a decorator of the label (so pygame_menu would think of both the image and label as the same widget), setting padding, setting margin, setting a selection border with 'inflate' (so it draws around both the image and the label). 
Nothing seems to be making any change to how it works, selection seems only to take into account the height of the button's label... Is it OK to leave this bug in for now? Otherwise I might need to dive into the pygame_menu code and see why it's doing what it's doing (which I can do but I don't want to right now, we could leave it for later? Unless you think it's a real blocker, which is fine.)

> Fix the vertical bar side effect. You could add logic to the start and resume methods on the menu base class to set the appropriate theme.

Good idea (I haven't tried this yet)
> Only being able to take Tuxemon out is a bit odd, can you implement storing as well?

Yes, I was going to raise that in a seperate pull request, so it's less code to review for each feature. If you want I can add it to this PR (I've not created it yet but I think it's going to be a lot easier than this PR, I'll re-use the Monster menu and change the callback when it's selected)

kerizane:
Ok, that offset bug can wait. It's annoying but not a deal-breaker.

Yeah, let's add the take menu in this PR. I don't think it's too much code, and I'd rather have the full feature.

ultidonki:
No problem, thanks! I'll add the other storage menu to this PR when I have time, then - and I'll try and fix that vertical scrollbar bug, too.

tuxemon/states/pc/__init__.py
```diff
@@ -88,13 +89,23 @@ def change_state(state: str, **kwargs: Any) -> MenuGameObj:
                 [T.translate("menu_storage_monsters_full")],
             )
         else:
-            storage_callback = change_state("MonsterBoxChooseState")
+            storage_callback = change_state("MonsterBoxChooseStorageState")
+
+        if len(local_session.player.monsters) == 0:
```
kerizane: use not, rather than len == 0

tuxemon/states/pc/__init__.py
```diff
@@ -359,3 +364,59 @@ def animate_close(self) -> Animation:
...
+                )
+            else:
+                menu_callback = self.change_state(
+                    "MonsterTakeState", box_name=box_name
+                )
+            menu_items_map.append((box_name, menu_callback))
+        return menu_items_map
+
+
+class MonsterBoxChooseDropoffState(MonsterBoxChooseState):
```
kerizane: DropOff, not Dropoff

tuxemon/states/pc/__init__.py
```diff
@@ -359,3 +364,59 @@ def animate_close(self) -> Animation:
...
+    """Shows all Tuxemon in player's party, puts it into box if selected."""
+
+    def startup(self, box_name: String, **kwargs: Any) -> None:
+        super().startup(**kwargs)
+
+        self.box_name = box_name
+
+    def on_menu_selection(
+        self,
+        menu_item: MenuItem[Optional[Monster]],
```
kerizane, line 417: Is monster actually optional?

ultidonki: In this menu, it's definitely not Optional, but it's over-riding a method from MonsterMenuState, so I just copied the signature from there, where it has an Optional (although I'm not sure that's right). (In tuxemon/states/monster/__init__.py ) 

Should I remove the Optional for this menu?

kerizane: That will cause mypy to complain about something else. Add `assert monster` to the next line after defining monster

kerizane (changes requested):

ultidonki:
Whoops, sorry, I added 'assert monster' before reading your edited comment with the other option. I'm happy to do either way if someone else prefers, I'll wait for someone else to decide. :smiley:

ultidonki:
> Fix the vertical bar side effect. You could add logic to the start and resume methods on the menu base class to set the appropriate theme.

I forgot to mention - I've fixed this by resetting `theme.scrollbar_area` before calling pop_state on itself, because adding it to the 'shutdown' method didn't work. (I think shutdown wasn't being called because when it was trying to remove itself, the state was in the middle of the state stack or something. I don't want to poke around in the stateManager class too much in case it makes a bug elsewhere so I just added it before calling pop_state itself, which works.)

kerizane:
You have a merge conflict

tuxemon/states/pc/__init__.py
```diff
@@ -73,10 +82,30 @@ def startup(self, **kwargs: Any) -> None:
...
+        if len(local_session.player.monsters) == 6:
+            storage_callback = partial(
+                open_dialog,
+                local_session,
+                [T.translate("menu_storage_monsters_full")],
+            )
+        else:
+            storage_callback = change_state("MonsterBoxChooseStorageState")
+
+        if not len(local_session.player.monsters):
```
kerizane: if not local_session.player.monsters:

kerizane:
I've done a quick test and if you fix that minor issue I merge the feature.

kerizane (approved):

kerizane:
As discussed earlier there are still a few issues relating to this, but it's such an important feature that I'm merging it now. Please create a ticket for the remaining issues.

Thanks for your work on this!

PR #1203, opened 2022-06-27
Allow using items on any monster in your party during combat. Closes #794

### LINKED ISSUE #794: Revive item not usable in combat
Revive requires that the target be unconscious, but using an item in combat requires that you target a monster that is in the combat.

Ideally we'd be prompted to choose from our current team, as we do when using the item out of combat.

ultidonki:
This PR allows you to use items on any monster in your party, during a battle. 

It adds a new optional property to Items - battle_menu. 

If battle_menu is == "monster", then when you use this item during battle, the Monster menu will appear, and you can select any monster in your party to use the item on. 
If battle_menu is == "combat", or not set, then you can choose between your current monster and the enemy monster, as normal. 

I've also added battle_menu to the db.py validation and schemas/item-schema.json, which seems to work (it gets rid of the error messages) but it's my first time updating them so it could be wrong. 

Hopefully adding the battle_menu property was a good idea, but if anyone has any better suggestions I'm happy to change it to something else! 

This should close issue #794

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -35,15 +35,7 @@ import logging
 import os
 from enum import Enum
 from operator import itemgetter
-from typing import (
-    Any,
-    Dict,
-    Literal,
-    Mapping,
-    Optional,
-    Sequence,
-    overload,
-)
+from typing import Any, Dict, Literal, Mapping, Optional, Sequence, overload
 
 from pydantic import BaseModel, Field, ValidationError, validator
 
@@ -71,6 +63,12 @@ class ItemSort(str, Enum):
 class ItemType(str, Enum):
     consumable = "Consumable"
     key_item = "KeyItem"
+
+
+# ItemBattleMenu is which menu you want to use to choose the target.
+class ItemBattleMenu(str, Enum):
+    monster = "monster"
+    combat = "combat"
 
 
 # TODO: Automatically generate state enum through discovery
@@ -112,6 +110,10 @@ class ItemModel(BaseModel):
     )
     effects: Sequence[str] = Field(
         [], description="Effects this item will have"
+    )
+    ## Optional fields:
+    battle_menu: Optional[ItemBattleMenu] = Field(
+        "", description="Which menu should be used to choose the target of the item."
     )
 
     class Config:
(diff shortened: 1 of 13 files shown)


tuxemon/item/item.py
```diff
@@ -154,6 +155,7 @@ def load(self, slug: str) -> None:
         self.usable_in = results["usable_in"]
         self.target = process_targets(results["target"])
         self.effects = self.parse_effects(results.get("effects", []))
+        self.battle_menu = (results["battle_menu"] if "battle_menu" in results else "")
```
kerizane: self.battle_menu = results.get("battle_menu", "")

kerizane (dismissed):

tuxemon/item/item.py
ultidonki: > self.battle_menu = results.get("battle_menu", "")

Thankyou, I've made this change now. Thanks!

tuxemon/db.py
```diff
@@ -35,15 +35,7 @@
...
-from typing import (
-    Any,
-    Dict,
-    Literal,
-    Mapping,
-    Optional,
-    Sequence,
-    overload,
-)
+from typing import Any, Dict, Literal, Mapping, Optional, Sequence, overload
```
kerizane, line 38: Picky, but should this have changed? Could you run isort and black against this file? (Just run script/fmt.sh and only commit the change to files that are currently in this PR).

If this causes lots of unrelated changes then let me know.

kerizane: Looks good other than that.

kerizane (changes requested):

kerizane (approved):

kerizane:
Thanks for the fix.

ultidonki:
Sorry, I only had time to run fmt.sh and isort now, as you requested, but it looks like the PR was fine if it got accepted. Thanks for merging! 

I think the change it made (turning a one-line import into multiple lines), there was talking of possibly changing fmt.sh to stop doing that. Or maybe someone decided it's better on a multi-line, I can't remember, sorry. :smile:

PR #1205, opened 2022-06-28
Small fixes and lock/unlock controls for Xero campaign

ultidonki:
1. Fixes a crash in taba town, after leaving the house after the 2nd announcement - this is because a lock_controls action was appearing after a transition_teleport, so it was never called (so the next unlock_controls caused an exception)

2. Fixes a bug where tweesher and cardiling couldn't be selected in professor's lab, you could only take rockitten

3. Add lock_controls and unlock_controls to a couple more cutscenes, to stop player being able to change maps during cutscenes and possibly causing problems

@Qiangong2 - just checking you don't have any changes in taba_town.tmx, professor_lab.tmx or player_house_downstairs.tmx ? I can wait to make these changes after you've checked anything in if you do. :smiley:

### DIFF
--- a/mods/tuxemon/maps/player_house_downstairs.tmx
+++ b/mods/tuxemon/maps/player_house_downstairs.tmx
@@ -115,6 +115,7 @@
   </object>
   <object id="32" name="announcement" type="event" x="32" y="96" width="16" height="16">
    <properties>
+    <property name="act001" value="lock_controls"/>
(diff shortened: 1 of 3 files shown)


Qiangong2:
@ultidonki I'm not working on those maps at the moment, so if someone else signs off, it will be fine :)

Although, I don't necessarily see anything wrong with the second issue :D

xirsoi (approved):
LG;TFTPR

PR #1207, opened 2022-07-02
Add new tuxemon, Teddisun

ultidonki:
This PR adds the new tuxemon, Teddisun based on the contribution from Leo (from Discord)

I also cleaned up the new_txmn encounter, to add Teddisun, Coproblight, and Pilthropus, so we can see them in the game, and removed a lot of old ones.


kerizane:
Looks good. Why did you remove some of the encounters?

ultidonki:
> Why did you remove some of the encounters?

So, I think the `new_txmn` encounter is only used in one place - when you leave the first village in the Xero campaign, the first grass you come to has the `new_txmn` encounter in there. (The other 2 bits of grass have different encounters)

I think the `new_txmn` encounter is just for testing new tuxemon to the game, so you can quickly start a game and the first bit of grass has them there. I removed a lot of old tuxemon that are no longer 'new' (most of them you can catch elsewhere in the spyder campaign). 

This makes it easier to test the 3 new tuxemon, Teddisun, Coproblight, and Pilthropus, as there's only 5 or so tuxemon in that spot of grass now. Otherwise, when I was testing it was taking me ages to see them.

kerizane (approved):

PR #1214, opened 2022-07-23
Add formatter config files

### LINKED ISSUE #1212: Add config files for CLI tools
# isort
We should start using isort https://pycqa.github.io/isort/ to ensure people are adding imports in the correct order. Include a config file (e.g. isort.cfg) so users don't need to specify command-line arguments.

Recommended settings:
profile = "black"
line-length = 79  # Might be required if we are using black mode

# black
Set the line length via a pyproject.toml file so you don't need to specify it manually.

kerizane:
Fixes https://github.com/Tuxemon/Tuxemon/issues/1212


bitcraft (approved):
looks great, ty

Issue #1211, opened 2022-07-15
Fix typehints


kerizane:
There are currently 155 errors reported by mypy.

Report generated using the following command via a UNIX command line: 

`mypy  tuxemon | cut -d":" -f1 | uniq -c | sort -nr`

     26 tuxemon/sprite.py
     18 tuxemon/states/world/worldstate.py
     13 tuxemon/technique.py
     12 tuxemon/states/combat/combat.py
     10 tuxemon/states/items/__init__.py
     10 tuxemon/animation.py
      9 tuxemon/states/world/world_menus.py
      9 tuxemon/map_loader.py
      8 tuxemon/networking.py
      7 tuxemon/map.py
      6 tuxemon/states/choice/__init__.py
      6 tuxemon/graphics.py
      5 tuxemon/platform/__init__.py
      5 tuxemon/menu/menu.py
      5 tuxemon/config.py
      4 tuxemon/states/combat/combat_animations.py
      3 tuxemon/menu/theme.py
      3 tuxemon/item/item.py
      3 tuxemon/clock.py
      2 tuxemon/session.py
      2 tuxemon/locale.py
      2 tuxemon/item/itemeffect.py
      2 tuxemon/item/itemcondition.py
      2 tuxemon/event/eventengine.py
      2 tuxemon/db.py
      2 tuxemon/cli/processor.py
      1 tuxemon/ui/text.py
      1 tuxemon/tools.py
      1 tuxemon/save.py
      1 tuxemon/rumble/tools.py
      1 tuxemon/rumble/__init__.py
      1 tuxemon/monster.py
      1 tuxemon/fusion.py
      1 tuxemon/event/eventaction.py

kerizane:
Ticket relating to a specific issue:

https://github.com/Tuxemon/Tuxemon/issues/1167

kerizane:
[all_errors.txt](https://github.com/Tuxemon/Tuxemon/files/9122573/all_errors.txt)

kerizane:
tuxemon/rumble/tools.py:4: error: Incompatible types in assignment (expression has type "None", variable has type "LibraryLoader[CDLL]")
tuxemon/clock.py:340: error: Incompatible return value type (got "None", expected "float")
tuxemon/clock.py:409: error: Incompatible types in assignment (expression has type "float", variable has type "int")
tuxemon/clock.py:410: error: Incompatible types in assignment (expression has type "float", variable has type "int")
[...]

PR #1192, opened 2022-05-31
Add pydantic models for input validation

ShadowApex:
This PR refactors `db.py` to use [pydantic](https://pydantic-docs.helpmanual.io/) data models to validate JSON defined in the db directory.

This gives us a few benefits:

* Validates whether or not some given JSON conforms to the correct schema.
* Can generate JSON schemas to allow auto-complete of JSON data in an editor.
* Allows us to use custom validation functions like ensuring sprites exist, translation slugs exist, etc.

I'll be continually improving the validation to provide more stronger validation to prevent crashes due to malformed data.

Issue #1167, opened 2022-05-03
Change `startup` method of states to `__init__`, and use instances instead of classes in `push_state`.


vnmabus:
**Is your feature request related to a problem? Please describe.**
Currently the initialization of a `State` object is done in the `startup` method instead of in the `__init__` method. Although in theory this could allow the reuse of states, using a pool of states, this feature is never used.

Instead, having the initialization in `startup` gives typing problems. As each `State` receives different initialization parameters, the `startup` method should be typed with `kwargs: Any` parameters at the end, to conform to the Liskov substitution principle (`__init__` is special and doesn't participate in LSP). Thus, Mypy does not say anything if you pass wrong arguments to it.

Moreover, as `startup` is called inside `push_state`, which accepts any kind of `State` class, it is not possible to check the number and types of the arguments, period.

**Describe the solution you'd like**
Instead of `startup`, we should use the standard `__init__`. We should pass created instances to `push_state`. That way, we can fully benefit from typing in states.

PR #1216, opened 2022-07-25
Fix type hints

kerizane:
Did some work on #1211, this is it for now.


tuxemon/db.py
```diff
@@ -725,7 +728,7 @@ def lookup(
     ) -> EnvironmentModel:
         pass
 
-    def lookup(self, slug: str, table: str = "monster") -> Mapping[str, Any]:
+    def lookup(self, slug, table="monster"):
```
vnmabus: Why did you remove the return value?

vnmabus (commented):
I also have another suggestion. In Python 3.10 the `|` operator has been overloaded to create union of types (https://peps.python.org/pep-0604/). Mypy supports the syntax also for previous Python versions if the evaluation of the annotations is postponed (with `from __future__ import annotations`). As the syntax is shorter and way clearer, I propose to use it in new type annotations (or when old modules are touched). Thus, instead of `Optional[int]` one would write `int | None`, or instead of `Union[int, str]` one would write `int | str`. Note that this is only for annotations: definition of types or aliases still need to use `Optional` and `Union` until we remove support for Python versions prior to 3.10.

tuxemon/db.py
kerizane: The examples in the PEP that introduced overload didn't have typehints on the actual function, I assumed that was how it worked.

However, I can see that mypy wants it to be typehinted; I will fix it.

https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading

kerizane: Just read the article more carefully... typehints are optional on the implementation. We may as well keep them.

kerizane:
```
    @overload
    def lookup(
        self,
        slug: str,
        table: Literal["environment"],
    ) -> EnvironmentModel:
        pass

    @overload
    def lookup(
        self,
        slug: str,
        table: str,
    ) -> Mapping[str, Any]:
        pass
```
Sigh... mypy doesn't understand this, even though this behaviour is explained in the PEP https://peps.python.org/pep-0586/#interactions-with-overloads

"Overloaded function signatures 11 and 12 overlap with incompatible return types"

kerizane: Simple reproduction of the mypy error:

```
from typing import Literal, Union, overload


@overload
def foo(s: Literal["hi"]) -> int:
    ...


@overload
def foo(s: str) -> str:
    ...


def foo(s: str) -> Union[str, int]:
    if s == "hi":
        return 1
    return s
```
example.py:5: error: Overloaded function signatures 1 and 2 overlap with incompatible return types

___

Will skip this one for now.

kerizane: Actually, I see what's going on here. The fallback option needs to be a superset of the other returning types, there's no option for "string that isn't one of the previous literals"

vnmabus: The REAL problem is that since  #1192 landed, the returning type is not a `Mapping[str, Any]`, but a pydantic `BaseModel`. That is the "incompatible return types" Mypy complains about.

kerizane: There's still one branch where it is a Mapping[str, Any], so it should be a Union of BaseModel and Mapping[str, Any].

https://github.com/Tuxemon/Tuxemon/blob/1c9b59686679a4ca7685027292e0881d1545a55d/tuxemon/db.py#L682

The solution is to either remove this branch, if it's unused, or to make the returning type a union of all the above Models (or just BaseModel) and Mapping[str, Any].

vnmabus: That should be fixed then, to have a common base class.

kerizane: Is the common base class just for the existing models, or do you want the Mapping[str, Any] to be included somehow?

I guess for unknown tables we could wrap the dict in a Model. I don't see much point, except it makes the typehints slightly more concise. We also shouldn't have unknown tables, so don't see much point in keeping this branch around. I guess it could be useful for development when you're implementing a new table and want to use a dict at first, rather than defining the model straight away.

```
class DataModel(BaseModel):
    data: Mapping[str, Any]
```
(Didn't check the syntax, but you know what I mean)

kerizane:
> I also have another suggestion. In Python 3.10 the `|` operator has been overloaded to create union of types (https://peps.python.org/pep-0604/). Mypy supports the syntax also for previous Python versions if the evaluation of the annotations is postponed (with `from __future__ import annotations`). As the syntax is shorter and way clearer, I propose to use it in new type annotations (or when old modules are touched). Thus, instead of `Optional[int]` one would write `int | None`, or instead of `Union[int, str]` one would write `int | str`. Note that this is only for annotations: definition of types or aliases still need to use `Optional` and `Union` until we remove support for Python versions prior to 3.10.

Yeah, that would be nice. The only issue is that the inconsistency might be confusing for some people: e.g. newcomers or people who haven't used typehints much.

I'll save this for another ticket, unless somebody like @bitcraft agrees with you.

kerizane:
Tested the game, seems fine.

kerizane:
Turns out we actually started with 365 (I didn't have all the libraries installed when I did my first scan), I've brought it down to 172.

This will probably do for now.

tuxemon/cli/processor.py
```diff
@@ -64,8 +64,8 @@ class CommandProcessor:
     def __init__(self, session: Session, prompt: str = "> ") -> None:
         self.prompt = prompt
         self.session = session
-        commands = list()
-        folders = list()
+        commands: List[CLICommand] = []
+        folders: List[str] = []
```
vnmabus: Why create it empty and then append a single folder?

tuxemon/cli/commands/whereami.py
```diff
@@ -26,8 +26,12 @@ def invoke(self, ctx: InvokeContext, line: str) -> None:
 
         """
         try:
```
vnmabus: Why the `try` now?

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
+ignore_missing_imports = true
```
vnmabus, line 3:
```suggestion
ignore_missing_imports = true
enable_error_code = ignore-without-code
```

tuxemon/db.py
```diff
@@ -725,18 +752,31 @@ def lookup(
...
-    def lookup(self, slug: str, table: str = "monster") -> Mapping[str, Any]:
+    def lookup(
+        self, slug: str, table: str = "monster"
+    ) -> Union[
+        EconomyModel,
+        EncounterModel,
+        EnvironmentModel,
+        InventoryModel,
+        ItemModel,
+        Mapping[str, Any],
```
vnmabus: Is that mapping REALLY needed? Also, can we use a base class instead of the full list?

tuxemon/event/actions/start_cinema_mode.py
```diff
@@ -46,7 +47,7 @@ class StartCinemaModeAction(EventAction[StartCinemaModeActionParameters]):
     param_class = StartCinemaModeActionParameters
 
     def start(self) -> None:
-        world = self.session.client.current_state
+        world = cast(WorldState, self.session.client.current_state)
```
vnmabus: `assert` is better here. Works for type checking and fails at runtime.

tuxemon/event/actions/stop_cinema_mode.py
```diff
@@ -49,7 +50,7 @@ class StopCinemaModeAction(EventAction[StopCinemaModeActionParameters]):
     param_class = StopCinemaModeActionParameters
 
     def start(self) -> None:
-        world = self.session.client.current_state
+        world = cast(WorldState, self.session.client.current_state)
```
vnmabus: `assert` again.

tuxemon/save.py
```diff
@@ -62,6 +63,14 @@
 config = prepare.CONFIG
 
 
+class SaveData(NPCState):
+    screenshot: str  # screenshot encoded as a string
```
vnmabus: Maybe we should use a different type to prevent mistakes?

tuxemon/states/world/worldstate.py
```diff
@@ -667,7 +669,7 @@ def pathfind(
         self,
         start: Tuple[int, int],
         dest: Tuple[int, int],
-    ) -> Optional[Sequence[Tuple[int, int]]]:
+    ) -> Optional[List[Tuple[int, int]]]:
```
vnmabus: Do we really want to allow modifications?

tuxemon/tools.py
```diff
@@ -382,7 +382,7 @@ def copy_dict_with_keys(
     return {k: source[k] for k in keys if k in source}
 
 
-def assert_never(value: NoReturn) -> NoReturn:
+def assert_never(value: str) -> NoReturn:
```
vnmabus: This change is wrong, there is no warning anymore.

vnmabus (changes requested):

kerizane:
Thanks for the suggestions, they all look good.

tuxemon/tools.py
kerizane: https://peps.python.org/pep-0484/#the-noreturn-type

NoReturn shouldn't be used for arguments. I will doublecheck what this function is doing, but it should probably be Any.

vnmabus: That function is used to detect that every possible value of that variable has been handled at that point. That is why the type is `NoReturn` (which really means "no type allowed").

kerizane: I get this error when using mypy 0.761 on the original code:

> tuxemon/states/combat/combat.py:353: error: Argument 1 to "assert_never" has incompatible type "Union[Literal['begin'], Literal['ready'], Literal['housekeeping phase'], Literal['decision phase'], Literal['pre action phase'], Literal['action phase'], Literal['post action phase'], Literal['resolve match'], Literal['ran away'], Literal['draw match'], Literal['has winner'], Literal['end combat']]"; expected "NoReturn"

NoReturn as a return type makes perfect sense, but you shouldn't use NoReturn as an annotation for "value".

See https://peps.python.org/pep-0484/#the-noreturn-type

> The NoReturn type is only valid as a return annotation of functions, and considered an error if it appears in other positions:

```
from typing import List, NoReturn

# All of the following are errors
def bad1(x: NoReturn) -> int:
    ...
bad2 = None  # type: NoReturn
def bad3() -> List[NoReturn]:
```
___
The signature should be `def assert_never(value: Any) -> NoReturn:` as you want to be able to use assert_never on variables of any type.

vnmabus: Then it means that these literals have not been handled. If you change the type, the function is useless.

There is a similar function in Python 3.11, but we cannot use it yet.

kerizane:
```
from typing import Any, Literal, NoReturn

CombatPhase = Literal[
    "begin",
]


def assert_never(value: NoReturn) -> NoReturn:
    assert False


def outer(phase: CombatPhase):
    if phase == "begin":
        return
    else:
        assert_never(phase)
```

example.py:16: error: Argument 1 to "assert_never" has incompatible type "Literal['begin']"; expected "NoReturn"

kerizane: Oh, you're right.

https://typing.readthedocs.io/en/latest/source/unreachable.html

It seems this was originally a bit of a hack (https://github.com/python/mypy/issues/5818), which is why this surprised me.

It turns out I had a very old version of mypy installed which didn't understand the idiom. Sorry about that.

tuxemon/save.py
kerizane: What do you mean by this?

vnmabus: Using `NewType` for it, to prevent mistakenly assigning a normal string.

tuxemon/db.py
kerizane: I don't think it makes sense to add a base class, as we don't use these models polymorphically.

I could make an alias for the union if you think it's too verbose.

tuxemon/map_loader.py
```diff
@@ -379,9 +387,12 @@ def load_event(
         w = int(obj.width / tile_size[0])
         h = int(obj.height / tile_size[1])
 
+        properties = cast(Dict[str, Any], obj.properties)
```
vnmabus: I would prefer not to cast. Instead, we should properly type `pytmx`.

tuxemon/monster.py
```diff
@@ -691,7 +702,8 @@ def end_combat(self) -> None:
             self.status = []
 
     def speed_test(self, action: EnqueuedAction) -> int:
-        if action.technique.is_fast:
+        technique = cast(Technique, action.technique)
```
vnmabus: Why is the cast needed? In case that is necessary, I would prefer an `assert`.

tuxemon/save_upgrader.py
```diff
@@ -62,7 +65,7 @@
 }
 
 
-def upgrade_save(save_data: Dict[str, Any]) -> Dict[str, Any]:
+def upgrade_save(save_data: Dict[str, Any]) -> SaveData:
```
vnmabus, line 68: Why isn't the input a `SaveData` too?

tuxemon/states/combat/combat_animations.py
```diff
@@ -71,6 +72,10 @@ def startup(
         graphics: Optional[Mapping[str, str]] = None,
         **kwargs: Any,
     ) -> None:
+        # The defaults are a lie to stop mypy complaining about us violating
```
vnmabus, line 75: That should be fixed by doing what is proposed in #1167.

vnmabus (changes requested):

tuxemon/map_loader.py
kerizane: Agreed

kerizane: What about natsorted? Their typehints are also bad.

tuxemon/monster.py
kerizane:
```
    technique: Union[Technique, Item]
    target: Monster
```

tuxemon/save_upgrader.py
kerizane: save_upgrader converts a save file from an earlier format (i.e before refactoring) to the current format. The data that comes in might not match the schema of `SaveData`. Look at `_transfer_storage_boxes` within the same file for an example.

vnmabus: Maybe then we should check that all fields are present and with appropriate types using a `TypeGuard`? We could use it if we add `typing_extensions` as a dependency, at least for type-checking.

tuxemon/monster.py
vnmabus: An `assert` then, better than a cast in my opinion.

tuxemon/map_loader.py
vnmabus: Well, I said that about `pytmx` because its author is a maintainer of this project, so they will certainly be accepted. If we can improve `natsort` we should  also try to, but I think the last commit is from 7 months ago, so maybe the PR will be stalled for a while.

tuxemon/states/combat/combat_animations.py
kerizane: Yes, looking forward to https://github.com/Tuxemon/Tuxemon/issues/1167 being done.

Do you have a suggestion related to this part of the code, or was it just an FYI? I think it's reasonable to explain that the existing typehints are wrong.

tuxemon/map_loader.py
kerizane: Are you happy with me leaving the natsorted cast in? Bitcraft is busy atm (I think), so don't know how quickly a pytmx fix will be accepted. But I'm still happy to remove the pytmx cast.

tuxemon/save_upgrader.py
kerizane: ~I'd rather not make this PR too complicated. Could you raise an issue explaining your idea?~

Actually, I don't think that's too bad. Maybe I will do it.

tuxemon/monster.py
kerizane: Done.

tuxemon/map_loader.py
kerizane: I'm not sure what to do here. natsorted gives a stupid error, but it feels slightly silly adding a pointless assert just to stop mypy complaining.

tuxemon/map_loader.py:395: error: Item "float" of "Union[str, bytes, float, int, PurePath, None, Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]]" has no attribute "startswith"
tuxemon/map_loader.py:395: error: Item "int" of "Union[str, bytes, float, int, PurePath, None, Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]]" has no attribute "startswith"
tuxemon/map_loader.py:395: error: Item "PurePath" of "Union[str, bytes, float, int, PurePath, None, Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]]" has no attribute "startswith"
tuxemon/map_loader.py:395: error: Item "None" of "Union[str, bytes, float, int, PurePath, None, Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]]" has no attribute "startswith"
tuxemon/map_loader.py:395: error: Item "Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]" of "Union[str, bytes, float, int, PurePath, None, Iterable[Union[str, bytes, float, int, PurePath, None, Iterable[Any]]]]" has no attribute "startswith"

tuxemon/save_upgrader.py
kerizane: We could also use pydantic.

tuxemon/states/combat/combat_animations.py
vnmabus: Just a FYI.

kerizane: I could just fix the typehint to make it accurate, even if that makes mypy complain.

It might not be possible though, I've seen exceptions in similar situations.

vnmabus: It is fine to let it as is. We will tackle the root of the issue in #1167.

tuxemon/save_upgrader.py
kerizane: Thought about this some more:

I've chosen to keep this PR simple, and tried to avoid functional changes. I've added some asserts, but they were generally asserting that values that can't be none weren't none, in places where the game would immediately crash anyway.

Validating the save file is a bigger change, and there is some risk to doing it. I'd rather not break somebody's save file just because of some irrelevant type error.

It is a good idea, but I'd prefer to isolate any risk and any technical discussions from this PR.

If you're still interested then please raise a feature request.

bitcraft:
wow, lots of changes.  and its getting old at this point.  i see there are conflicts a different merge here.  in the interest of getting it merged quickly, we can leave the `|` operator changes for later if you prefer.  3.10 is the default download so we should feel free to assume people can/will use that.  as far as the pytmx changes, lmk what i can do to get that compliant.

kerizane:
> as far as the pytmx changes, lmk what i can do to get that compliant.

The method is typed as returning `dict`, which gets interpreted as Dict[Any, Any]. Can you make it more specific? The key is always a str (I think) and the value is a union of types.

kerizane:
I will fix the conflict and see where I got to with this PR. I will do the Union | None in another ticket, should be easy as pyupgrade can do it.

@vnmabus anything that needs before we can merge? Let's try to close this down.

vnmabus:
I don't think so. It is fine for now.

kerizane:
@vnmabus Thanks for all your help with this.

kerizane:
@bitcraft Ready to go

bitcraft (approved):

Issue #1126, opened 2022-02-13
Replace old Travis CI automation with GitHub Actions (or similar)


freshreplicant:
**Is your feature request related to a problem? Please describe.**
Currently there is no automated creation of packages/executables for releases. This means the potential reach and ease of use of Tuxemon is severely limited.

**Describe the solution you'd like**
GitHub Actions seems to be a potential way to addressing this, especially considering the GitHub Marketplace offers lots free [ready-made action templates](https://github.com/marketplace?category=&query=&type=actions&verification=) you can install.

**Describe alternatives you've considered**
[CircleCI](https://github.com/marketplace/circleci), [Azure Pipelines](https://github.com/marketplace/azure-pipelines), etc. Both appear to have limited free tiers. But GitHub Actions might be even more flexible and hackable (with community members freely contributing and maintaining their templates/recipes). 

**Additional context**
There are GitHub actions for every single packaging format that has been discussed as part of our [packaging milestone](https://github.com/Tuxemon/Tuxemon/milestone/8). Flatpak, Arch packages, .debs, Appimage, Android and so on.

There would need to be a way to make sure all the builds are automatically attached to releases, but there appear to be Actions for this too.

Related: #396

kerizane:
Looks like this has been started. We have unittests running and flatpak and windows build.

@Murmele has just implemented automatic artifact publishing for releases as well as a rolling prerelease based on the development branch.

We still have a few jobs to migrate from the Travis yml https://github.com/Tuxemon/Tuxemon/blob/development/.travis.yml

PR #1221, opened 2022-08-05
Improve ci

Murmele:


Murmele:
@kerizane I set the CI up. Do we need the windows binary? because otherwise a step can be removed and the CI will be much faster

Perhabs I am not able to test if the release creation works, because I don't have the permission.

kerizane:
Yeah, I think we need the windows binary. How slow is it?

kerizane:
So, there is a flatpak file and a flatpak yml. What does the yml do?

kerizane:
Not sure what the best practice is here. What about if we only build the artifacts when we're doing a release? The only git actions that would run for PRs would be unittests and code quality. Although, I guess there's no harm in doing the builds and if someone manages to break one of them, it's better to know before we merge it...

I'd like to have a rolling unstable release, as well as the versioned releases. This could be done as a nightly (build run every 24 hours) or whenever we merge to development. 

What do you think?

Murmele:
> Yeah, I think we need the windows binary. How slow is it?

It takes really long to upload. Why the windows installer is not enough?



> Not sure what the best practice is here. What about if we only build the artifacts when we're doing a release?

Having the artifcats has the benefit that also testers can easily install a new version. In Gittyup I am using this feature quite a lot. Just to test I am installing the flatpak version

Murmele:
> So, there is a flatpak file and a flatpak yml. What does the yml do?

The yml is the manifest which was used to build the flatpak package. We can add it to the output files or not.

Murmele:
> I'd like to have a rolling unstable release, as well as the versioned releases. This could be done as a nightly (build run every 24 hours) or whenever we merge to development.

Yes this is done here. Whenever there is a push to development or a tag is created on development, the publish job is executed.

  # a prerelase is created when pushing to master
    # a release is created when a tag will be set
    # last condition is the same as IS_RELEASE,
    # but environment variables cannot be used outside of steps
    # so it was copied to here too
    if: ${{ github.ref == 'refs/heads/development' || (github.event_name == 'push' && github.ref_type == 'tag' && startswith(github.ref_name, 'v'))  }}

It will look like this at the end:
https://github.com/Murmele/Gittyup/releases
image

kerizane:
> It takes really long to upload. Why the windows installer is not enough?

I'm not too fussed, we can leave it out for now.

For reference, how long is "really long"? 10 mins, 1 hour?

kerizane:
Would anyone want the yml file? What's the benefit in having it?

This looks great, thanks for doing this.

Murmele:
> 

I think for now it is only a build artifact, but if someone wants to test something offline this file can be used.

.github/workflows/build.yml
```diff
@@ -0,0 +1,230 @@
...
+        manifest-path: org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml
+        cache: false
+        branch: ${{ steps.flatpak_release_branch.outputs.value }}
+
+    - name: Publish build artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        path: org.tuxemon.Tuxemon/*
+        name: Tuxemon Flatpak
+
```
Murmele:
```suggestion
```
Remove publishing the flatpak manifest in the artifacts

Murmele:
> 
It takes about 14min to upload and 4 minutes to download again

kerizane (approved):

Murmele:
Thanks for merging. Let's see if the publish works as intended :)

kerizane:
It worked, amazing!

If you're interested there are a few more builds that need to be added, see https://github.com/Tuxemon/Tuxemon/issues/1126 and https://github.com/Tuxemon/Tuxemon/milestone/8

PR #1230, opened 2022-08-12
Don't convert pydantic models to dicts

kerizane:
Using classes is much nicer, IMO. It also allows more specific type hints. Many of the type hints specify the Model, but this is wrong because the type was actually a Mapping.

This PR removes all cases where we convert the model into a dict before using it (well, there's one case where a function requires a dict, and it didn't make sense to change it).

It also fixes two bugs introduced by pydantic:
- Combat background not loaded (as it wasn't part of the model).
- Monster sounds defaulting to sound_cry1 instead of sound_{type}_cry1.

### DIFF
--- a/tests/tuxemon/item/test_economy.py
+++ b/tests/tuxemon/item/test_economy.py
@@ -1,5 +1,6 @@
 import unittest
 
+from tuxemon.db import EconomyItemModel
 from tuxemon.item.economy import Economy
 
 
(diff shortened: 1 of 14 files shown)


tests/tuxemon/item/test_economy.py
```diff
@@ -29,10 +30,20 @@ def test_potion_cost(self):
 
     def test_missing_price(self):
         economy = self.economy
-        with self.assertRaises(RuntimeError):
-            price = economy.lookup_item_price("capture_device")
+        price = economy.lookup_item_price("capture_device")
+        self.assertEqual(price, 0)
 
     def test_missing_cost(self):
+        economy = self.economy
```
xirsoi, line 37: ❤️ 
Thank you for adding to test coverage.

xirsoi (approved):

Issue #1204, opened 2022-06-28
Teleport action doesn't remove NPCs from map


ultidonki:
**Describe the bug**
When going from route1.tmx, to taba_town.tmx, these 8 NPC's still exist, even though they should only be in route1.tmx

If you change the event from 'teleport' to 'transition_teleport', it works, and these 8 NPC's disappear:
![route_1_weirdness](https://user-images.githubusercontent.com/89098769/176236694-1bf68338-4f2b-4aa5-9fda-5f0f0072c0c9.png)

**To Reproduce**
1. Start Tuxemon: Xero campaign, or load map taba_town.tmx
2. Continue the game until you can leave the 2nd town in the top-left of the map (or press 'N' with devtools enabled to be able to walk through walls)
3. After going to route1.tmx (you'll see the game pause for a second as you walk onto the next map), go back to taba_town.tmx, and go to the grass inbetween the first 2 villages to see all 8 NPC's from route1

**Expected behavior**
NPC's should be removed if moving between maps
(It works if you use transition_teleport instead of teleport)

kerizane:
The issue is that the create_npc action is "leaking" into the next map and reloading the npcs. 

If you look at the object that creates them, you can see that when you enter the map the conditions are always true. This loop is running repeatedly, but you don't really notice this, because the create_npc action knows not to do anything if the npc already exists.

```
  <object id="111" name="angry populace" type="event" x="432" y="320" width="16" height="16">
   <properties>
    <property name="act10" value="create_npc omnigrunt,31,27,,stand"/>
    <property name="act11" value="create_npc omnigrunt2,32,27,,stand"/>
    <property name="act12" value="create_npc omnigrunt3,33,27,,stand"/>
    <property name="act13" value="create_npc omnigrunt4,34,27,,stand"/>
    <property name="act14" value="npc_face omnigrunt,up"/>
    <property name="act15" value="npc_face omnigrunt2,up"/>
    <property name="act16" value="npc_face omnigrunt3,up"/>
    <property name="act17" value="npc_face omnigrunt4,up"/>
    <property name="act20" value="create_npc liela,31,26,,stand"/>
    <property name="act21" value="create_npc bob,32,26,,stand"/>
    <property name="act22" value="create_npc marissa,33,26,,stand"/>
    <property name="act23" value="create_npc tuxemart_keeper,34,26,,stand"/>
    <property name="cond1" value="not variable_set letusthrough:no"/>
[...]

PR #1233, opened 2022-08-16
Add condition to stop npcs leaking to taba_town

kerizane:
Quick fix of https://github.com/Tuxemon/Tuxemon/issues/1204, we may need further discussion of the root cause.

Race condition was causing the event to be applied to the next map if we get there via a teleport.


kerizane:
Better fix in https://github.com/Tuxemon/Tuxemon/pull/1235

PR #1236, opened 2022-08-16
Handle transition teleport being triggered twice

### LINKED ISSUE #621: transition_teleport triggers two times when teleporting to the same map

kerizane:
Fixes https://github.com/Tuxemon/Tuxemon/issues/621

### DIFF
--- a/tuxemon/event/actions/transition_teleport.py
+++ b/tuxemon/event/actions/transition_teleport.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 from typing import NamedTuple, final
 
 from tuxemon.event.eventaction import EventAction
+from tuxemon.states.world.worldstate import WorldState
 
 
 class TransitionTeleportActionParameters(NamedTuple):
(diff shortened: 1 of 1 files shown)


bitcraft (approved):
thank you!

Issue #1234, opened 2022-08-16
WIP Battle Check - new approach


JaskRendix:
Hi guys. 

This is a continuation of #716 Battle Check. I developed something and I wanted to share with you before PR or trash everything.

So I proceed as follows:

- [ ] 1. Teleporting to the closest Center (updating maps)
- [ ] 2. Creating new dict called "battle history" and editing various files
- [ ] 2.a. combat.py
- [ ] 2.b. worldstate.py
- [ ] 2.c. npc.py
- [ ] 2.d. save_upgrader.py 
- [ ] 3. Creating *event action" called "battles_print"
- [ ] 4. Creating *event condition" called "battle_result"
- [ ] 5. Creating *event condition" called "battle_is"

In details:

**1. Teleporting to the closest Center (updating maps)**
I'm aware there is teleport_faint, but I opted for the classic teleport for two reasons:

- (1) it allows me the condition "is player defeated"
- (2) it allows me to specify the place without being fixed.

So if I'm on Route 3 > Leather Center, etc. By the way the event is active on all the map (width and height), so if someone adventures himself in the route with fainted monsters, then it would be teleported to the center.

This is a piece of the code:
```
  <object id="69" name="Exhausted" type="event" x="0" y="0" width="320" height="640">
   <properties>
    <property name="act1" value="transition_teleport spyder_xxx_center.tmx,4,20,0.3"/>
    <property name="cond1" value="is player_defeated"/>
   </properties>
  </object>
```
Here a list of "where you're going to be teleported":
- Route 1 > Downstairs Home Sweet Home (it can be removed or we can reintroduce the bed for restoring in Paper)
- Route 2 > Cotton Center
- Route 3 trainer > Leather 
- Route 4 > Leather Center
- Route 5 > Flower Center
- Route 6 > Candy Center
- Route A > Downstairs Home Sweet Home
- Route C > Candy Center
- Tunnel > Timber Center
- Tunnel below > Timber Center
- Citipark > Cotton Center
- Omnichannel (all) > Cotton Center
- Nimrod (all) > Flower Center
- Greenwash (all) > Candy Center

Cotton Tunnel, Dragonscave and Dryadsgrove are missing because I'm unsure about the destination.

**2. Creating new dict called "battle history" and editing the following files: combat.py, worldstate.py, npc.py and save_upgrader.py**
Here is a list:

- combat.py
- worldstate.py
- save_upgrader.py

**2.a. combat.py**
As follows.

Datetime
[...]

PR #1237, opened 2022-08-17
player.battle_history

JaskRendix:
Here the PR as described in #1234 

In details:

**1. Teleporting to the closest Center (updating maps)**
I'm aware there is teleport_faint, but I opted for the classic teleport for two reasons:

- (1) it allows me the condition "is player defeated"
- (2) it allows me to specify the place without being fixed.

So if I'm on Route 3 > Leather Center, etc. By the way the event is active on all the map (width and height), so if someone adventures himself in the route with fainted monsters, then it would be teleported to the center.

This is a piece of the code:
```
  <object id="69" name="Exhausted" type="event" x="0" y="0" width="320" height="640">
   <properties>
    <property name="act1" value="transition_teleport spyder_xxx_center.tmx,4,20,0.3"/>
   </properties>
  </object>
```
Here a list of "where you're going to be teleported":
- Route 1 > Downstairs Home Sweet Home (it can be removed or we can reintroduce the bed for restoring in Paper)
- Route 2 > Cotton Center
- Route 3 trainer > Leather 
- Route 4 > Leather Center
- Route 5 > Flower Center
- Route 6 > Candy Center
- Route A > Downstairs Home Sweet Home
- Route C > Candy Center
- Tunnel > Timber Center
- Tunnel below > Timber Center
- Citipark > Cotton Center
- Omnichannel (all) > Cotton Center
- Nimrod (all) > Flower Center
- Greenwash (all) > Candy Center

Cotton Tunnel, Dragonscave and Dryadsgrove are missing because I'm unsure about the destination.

**2. Creating new dict called "battle history" and editing the following files: combat.py, worldstate.py, npc.py and save_upgrader.py**
Here is a list:

- combat.py
- worldstate.py
- save_upgrader.py

**2.a. combat.py**
As follows.

Datetime
```
import datetime
insert the following datetime string:
        self.players[0].game_variables["date"] = dt.datetime.now().strftime("%x")
```
Creation of new variables (game_variables) called:
```game_variables["battle_total"] = number of battles (only NPC, not Random)
game_variables["battle_draw"] = number of battles draw
game_variables["battle_won"] = number of battles won
game_variables["battle_lost"] = number of battles lost
game_variables["percent_draw"] = round(value) (i.e. 42 or 69)
game_variables["percent_won"] = round(value) (i.e. 42 or 69)
game_variables["percent_lost"] = round(value) (i.e. 42 or 69)
```

Reason? Modders. They can create a story, a tournament, etc. where they need this data, because only people with xx% of victories can access.

Creation of new variable (based on battle_history) called:
`self.players[0].battle_history[self.players[0].game_variables["date"]] = self.players[1].slug + str(" - draw")`

Output:
`{'08/16/22': 'spyder_route2_billie - lost'}`

Why the date? I thought about it.
- putting npc.slug as key, it would ended up "updated", probably replaced by "-win"
- no possibility of creating a substantial output

Like:
```
{'08/16/22': 'spyder_route2_billie - lost'}
{'08/17/22': 'spyder_route2_billie - lost'}
{'08/18/22': 'spyder_route2_billie - lost'}
{'08/22/22': 'spyder_route2_billie - lost'}
{'08/31/22': 'spyder_route2_billie - win'}
```

Of course, it's going to be updated if you fight again in the same day. So...

So... why the 08/16/22 format? I wasn't sure at the beginning.

I considered:
- the number of days (i.e. today is 228th day of the year)
- the actual hh:mm:ss, but it was missing the day
- the actual complete format mm/dd/yyyy hh:mm:ss

Considering the above reasons, well, I opted for the 08/16/2022 format.

**2.b. worldstate.py**
An addition related to battle_history()

**2.c. npc.py**
A series of additions related to battle_history()

**2.d. save_upgrader.py**
Updated the savegame.

**3. Developing an event action called: battles_print**
So it's possible to print all the battles (date + npc.slug + result). It's the copy of "print".

Reason? Modders. So they can have a complete map of battle_history.

**4. Developing an event condition called: battle_result**
As you can see below, it checks if you won, lost or draw against a specific npc.slug

```
    Script usage:
        .. code-block::

            is battle_result <character>,<result>

    Script parameters:
        character: Npc slug name (e.g. "npc_maple").
        result: One of "won", "lost" or "draw"
```
Reason? Modders. So they can check if a player won a single time against a NPC (i.e. spyder_route2_billie).

**5. Creating an event condition called: battle_is**
As you can see below, it checks an operation over a battle:

```
    Script usage:
        .. code-block::

            is battle_is <character>,<result>,<operation>,<value1>

    Script parameters:
        character: Npc slug name (e.g. "npc_maple").
        result: One of "won", "lost" or "draw"
        operation: One of "==", "!=", ">", ">=", "<" or "<=".
        value2: A number.
```
Reason? Modders. So they can check how many times the player lost, won or draw against a NPC (i.e. spyder_route2_billie).

Everything wrote above runs. It took me almost a week, but of course when you're inspired things go fast.

In addition:
- small fixes to coordinates inside maps (sometimes there was 32.00442, so I fixed at 32, etc.);
- fixed the bed inside the bedroom, now it can be used, before it wasn't possible;

### DIFF
--- a/tuxemon/event/conditions/battle_is.py
+++ b/tuxemon/event/conditions/battle_is.py
@@ -1,0 +1,95 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+from __future__ import annotations
+
+import logging
+
+from tuxemon.event import MapCondition
+from tuxemon.event.eventcondition import EventCondition
+from tuxemon.session import Session
+
+logger = logging.getLogger(__name__)
+
+
+class BattleIsCondition(EventCondition):
+    """
+    Check an operation over a battle.
+
+    Script usage:
+        .. code-block::
+
+            is battle_is <character>,<result>,<operation>,<value1>
+
+    Script parameters:
+        character: Npc slug name (e.g. "npc_maple").
+        result: One of "won", "lost" or "draw"
+        operation: One of "==", "!=", ">", ">=", "<" or "<=".
+        value2: A number.
+
+    """
+
+    name = "battle_is"
+
+    def test(self, session: Session, condition: MapCondition) -> bool:
+        """
+        Check an operation over a variable.
+
+        Parameters:
+            session: The session object
+            condition: The map condition object.
+
+        Returns:
+            Result of the operation over the variable.
+
+        """
+        player = session.player
+
+        # Read the parameters
+        npc = condition.parameters[0]
+        result = condition.parameters[1]
+        operation = condition.parameters[2]
+        value1 = condition.parameters[3]
+
+        # Union
+        battle_mix = str(npc) + " - " + result
+
+        # Count
+        count = 0
+        for val in player.battle_history.values():
+            if val == battle_mix:
+                count += 1
+
+        # Check if the condition is true
+        if operation == "==":
+            return int(count) == int(value1)
+        elif operation == "!=":
+            return int(count) != int(value1)
+        elif operation == ">":
+            return int(count) > int(value1)
+        elif operation == ">=":
+            return int(count) >= int(value1)
+        elif operation == "<":
+            return int(count) < int(value1)
+        elif operation == "<=":
+            return int(count) <= int(value1)
+        else:
+            logger.error(f"invalid operation type {operation}")
+            raise ValueError
--- a/tuxemon/event/conditions/battle_is.py
+++ b/tuxemon/event/conditions/battle_is.py
@@ -36,13 +36,13 @@ class BattleIsCondition(EventCondition):
     Script usage:
         .. code-block::
 
-            is battle_is <character>,<result>,<operation>,<value1>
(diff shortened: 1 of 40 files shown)


tuxemon/event/conditions/battle_is.py
```diff
@@ -0,0 +1,95 @@
...
+    Script usage:
+        .. code-block::
+
+            is battle_is <character>,<result>,<operation>,<value1>
+
+    Script parameters:
+        character: Npc slug name (e.g. "npc_maple").
+        result: One of "won", "lost" or "draw"
+        operation: One of "==", "!=", ">", ">=", "<" or "<=".
+        value2: A number.
```
kerizane: should be value1? Why not just value?

tuxemon/states/combat/combat.py
```diff
@@ -384,6 +387,10 @@ def transition_phase(self, phase: CombatPhase) -> None:
                 var["battle_last_monster_category"] = monster_record.category
                 var["battle_last_monster_shape"] = monster_record.shape
 
+            # update total number of battles (trainers)
```
kerizane: From the comment above it seems this will run every round of a battle

kerizane:
Please copy your comment over from the issue so that it's easier to refer to here.

tuxemon/event/conditions/battle_is.py
JaskRendix: You're right, it should be value

tuxemon/states/combat/combat.py
JaskRendix: It wasn't my intention. It updates only once each battle. Poor choice of words.

kerizane: This branch is called multiple times a battle so your count will be wrong.

```
elif phase == "housekeeping phase":
            self._round += 1
            # fill all battlefield positions, but on round 1, don't ask
            self.fill_battlefield_positions(ask=self._round > 1)

```

JaskRendix: You''re right, wrong phase, I'm going to move it to another phase.

kerizane:
The teleport is implemented using existing actions and conditions, right?

Please split this into three PRs
1. Map number tweaks plus bed fix
2. Teleport on lose
3. New battle stuff

JaskRendix:
> The teleport is implemented using existing actions and conditions, right?

Yes.

> Please split this into three PRs

Ok

PR #1238, opened 2022-08-17
time of the day

JaskRendix:
While I was working on battle_history, precisely after the datetime string, I realized the absence of time inside Tuxemon. It pushed me to create the following variables (game_variables):

`game_variables["hour"] `
it gives us the HOUR (from 00 to 23)

`game_variables["daynr"]`
it gives us the DAY (001-366) - today is the 229th

`game_variables["b_timeofday"]`
it gives us (day or night) and it's based on hour

`game_variables["c_timeofday"] `
it gives us (dawn, morning, afternoon, dusk or night) and it's based on hour

Reason? Modders.

For example we can:
- decide if some NPC are going to appear or disappear during specific part of the day;
- decide if some location can be accessed or not during some parts of the day;
- set specific days where some events take place (i.e. a different days for each city, holidays, etc.);

Most of it can be done through the following condition:
is variable_set b_timeofday:day
is variable_set c_timeofday:afternoon
is variable_set daynr:333

Next possible developments? 
We can integrate the month, it could give us the variable to set the season (winter 1 - 2 - 3, spring 4 - 5 - 6, summer 7 - 8 - 9 and autumn 10 - 11 - 12)

### DIFF
--- a/tuxemon/player.py
+++ b/tuxemon/player.py
@@ -83,7 +83,7 @@ class Player(NPC):
         """
         var = self.game_variables
         var["hour"] = dt.datetime.now().strftime("%H")
-        var["daynr"] = dt.datetime.now().strftime("%j")
+        var["day_of_year"] = dt.datetime.now().strftime("%j")
 
         # Day and night basic cycle (12h cycle)
         if int(var["hour"]) < 6:
@@ -95,14 +95,14 @@ class Player(NPC):
 
         # Day and night complex cycle (4h cycle)
         if int(var["hour"]) < 4:
-            var["stageofday"] = "night"
+            var["stage_of_day"] = "night"
         elif 4 <= int(var["hour"]) < 8:
-            var["stageofday"] = "dawn"
+            var["stage_of_day"] = "dawn"
         elif 8 <= int(var["hour"]) < 12:
-            var["stageofday"] = "morning"
+            var["stage_of_day"] = "morning"
         elif 12 <= int(var["hour"]) < 16:
-            var["stageofday"] = "afternoon"
+            var["stage_of_day"] = "afternoon"
         elif 16 <= int(var["hour"]) < 20:
-            var["stageofday"] = "dusk"
+            var["stage_of_day"] = "dusk"
         else:
-            var["stageofday"] = "night"
+            var["stage_of_day"] = "night"
--- a/tuxemon/player.py
+++ b/tuxemon/player.py
@@ -86,22 +86,22 @@ class Player(NPC):
 
         # Day and night basic cycle (12h cycle)
         if int(dt.datetime.now().strftime("%H")) < 6:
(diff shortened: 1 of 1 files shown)


kerizane:
Great idea

vnmabus:
Good idea, but I would use more legible names.

tuxemon/player.py
```diff
@@ -76,3 +77,31 @@ def update(self, time_delta: float) -> None:
         diff_y = abs(after.y - before.y)
 
         self.game_variables["steps"] += diff_x + diff_y
+        """
+        %H - Hour 00-23
+        %j - Day number of year 001-366
+        """
+        self.game_variables["hour"] = dt.datetime.now().strftime("%H")
```
kerizane: Save the result of datetime.now to a variable, rather than calling it 10 times in the same method, as it's apparently a bit slow.

Check the FPS of the game (there is a config option for this) and make sure there isn't a performance hit.

kerizane:
Maybe b_timeofday could just be "daytime" with values "true" or "false"?

And then c_timeofday could just be timeofday.

Maybe a bit too similar...

daystage? stageofday? stage_of_day?

kerizane:
Maybe b_timeofday could just be "daytime" with values "true" or "false"?

JaskRendix:
> Maybe b_timeofday could just be "daytime" with values "true" or "false"?

I agree.

> daystage? stageofday? stage_of_day?

Stageofday sounds ok.

tuxemon/player.py
JaskRendix: I checked FPS and I didn't notice a performance hit. I noticed that FPS increases only after enter or exit a location (around 70/80) for a couple of seconds, if not it stays around 60, 59, etc. I tried with and without my commit.

kerizane: extract into variable please int(dt.datetime.now().strftime("%H"))

Even without performance issue, it's more concise

JaskRendix:
```
        var = self.game_variables
        var["hour"] = dt.datetime.now().strftime("%H")
        var["daynr"] = dt.datetime.now().strftime("%j")

        # Day and night basic cycle (12h cycle)
        if int(var["hour"]) < 6:
            var["daytime"] = False
        elif 6 <= int(var["hour"]) < 18:
            var["daytime"] = True
        else:
            var["daytime"] = False
```
You mean something like this? Did I understand?

kerizane:
@vnmabus is stageofday ok, or do we want stage_of_day?

bitcraft:
at the moment, im not too worried about the difference between "stage_of_day" or "stageofday".  we don't have any specific guidelines about names and until we do, im ok as long as its readable.  "daynr" probably should be "dayofyear" though.  cool idea

bitcraft (approved):
LGTM

PR #1239, opened 2022-08-17
Map tweaks plus bed fix

JaskRendix:
Split from #1237 

- small fixes to coordinates inside maps (sometimes there was 32.00442, so I fixed at 32, etc.);
- fixed the bed inside the bedroom, now it can be used, before it wasn't possible.

### DIFF
--- a/mods/tuxemon/maps/spyder_bedroom.tmx
+++ b/mods/tuxemon/maps/spyder_bedroom.tmx
@@ -62,11 +62,12 @@
    </properties>
   </object>
   <object id="26" name="Player Spawn" type="event" x="64" y="80" width="16" height="16"/>
-  <object id="27" name="Resting in Bed" type="interact" x="16" y="48" width="16" height="32">
+  <object id="27" name="Resting in Bed" type="event" x="16" y="48" width="16" height="32">
    <properties>
     <property name="act1" value="translated_dialog spyder_papertown_restinbed"/>
     <property name="act2" value="set_monster_health ,"/>
     <property name="act3" value="set_monster_status ,"/>
+    <property name="cond1" value="is button_pressed K_RETURN"/>
    </properties>
   </object>
  </objectgroup>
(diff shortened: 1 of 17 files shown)


mods/tuxemon/maps/spyder_bedroom.tmx
```diff
@@ -62,11 +62,12 @@
    </properties>
   </object>
   <object id="26" name="Player Spawn" type="event" x="64" y="80" width="16" height="16"/>
-  <object id="27" name="Resting in Bed" type="interact" x="16" y="48" width="16" height="32">
+  <object id="27" name="Resting in Bed" type="event" x="16" y="48" width="16" height="32">
    <properties>
     <property name="act1" value="translated_dialog spyder_papertown_restinbed"/>
     <property name="act2" value="set_monster_health ,"/>
     <property name="act3" value="set_monster_status ,"/>
+    <property name="cond1" value="is button_pressed K_RETURN"/>
```
kerizane, line 70: This looks like it will trigger anywhere in the room if you press A.

JaskRendix: You're right about it, and it's really weird. I have no idea because it happens. I use the same condition in other location and I have never had this problem. If I remove the cond1 I'm stuck close to the stairs and unable to move, because the message keeps pop up. That's the reason why I added the cond1.

kerizane: You need to add a second condition "is player_at" this checks that you are in the right place.

JaskRendix: That's what it was missing. I fix it immediately.

JaskRendix: What I needed as cond2 is this:
`    <property name="cond2" value="is player_facing_tile"/>`
Now the problem is solved.

kerizane (approved):

kerizane:
Thanks!

Issue #1084, opened 2021-12-06
Losing a battle should re-spawn you somewhere else


ultidonki:
I think this has been discussed in Matrix/Discord, but I can't find an issue for it:

**Is your feature request related to a problem? Please describe.**
When you're defeated, nothing happens, you can just keep walking like normal. 
To make it more fun, you should be sent back to eg. a checkpoint when you're defeated. 

**Describe the solution you'd like**
<s>A new action:
``` set_respawn_point map_name.tmx,5,10 ```
When you enter a new map, or reach a checkpoint, this action should fire, setting a spawn point nearby. 
When you're defeated, you should then be teleported back to the most recently set spawn point. 
If no set_respawn_point action has been fired yet, something sensible should happen - like, you don't get teleported anywhere, or the game crashes with an error message. 
The parameters should be exactly the same as a teleport, or transition_teleport:
``` map name,x_pos,y_pos[,maybe_transition_time?] ```
Because, behind the scenes you'll probably do a teleport anyway, so you need all those parameters. </s>
This already exists: Use 
``` set_variable teleport_faint:<map_to_teleport_to.tmx> <x_pos> <y_pos> ```
to set the X, Y co-ordinates and map name of where to teleport, then use
``` teleport_faint ```
with no parameters to teleport back to the map in variable teleport_faint. 

However, ideally I still think this should happen automatically after defeat - and have a yes/no variable teleport_when_defeated to say whether you want the player to be teleported after being defeated or not. 

Optional:
Have a variable 
``` respawn_player_when_defeated ```
set to 'yes' automatically when game starts. 
Only teleport when defeated if this variable is set to 'yes'.
So, modders can turn off this behaviour if they set it to 'no'. 

Optional: 
<s>You might need a condition check to stop this set_respawn_point action from being fired all the time, see alternatives section below, but that makes it more complicated. 
It's probably fine to just keep firing this action for as long you're near the checkpoint, seems harmless. </s>
This seems to work fine, not needed. 

Optional:
<s>``` respawn_player ```
An action to just respawn the player at the nearest checkpoint, without them needing to be defeated. For instant-death situations? Or maybe you could just do that with a transition_teleport? I guess? </s>
This already exists, see the teleport_faint action. 

**Describe alternatives you've considered**
<s>You could have the action set a game variable, like:
``` set_respawn_point taba_town,taba_town.tmx,15,15 ```
This would set the respawn point to the variable 'taba_town', while storing the other parameters somewhere else.
[...]

PR #1240, opened 2022-08-17
teleport on lose

JaskRendix:
Split from #1237 

Teleporting to the last visited Center (updating maps)

**Why?**
Mostly two reasons:
1. Because people can wander around with fainted monsters (missing battles, storylines, etc.)
2. Because if a player loses a fight, then a player needs to be teleported to the last center.

**How?**
In the connection maps (mostly routes or where there are battles / random_encounter), we have added the following code:
```
  <object id="69" name="Exhausted" type="event" x="0" y="0" width="16" height="16">
   <properties>
    <property name="act1" value="teleport_faint"/>
   </properties>
  </object>
```
If someone with fainted monsters tries to access a connection map, it gets immediately teleported.

At the same time in the location with healers (Tuxecenter), we added the following action:

`    <property name="act80" value="set_variable teleport_faint:spyder_healing_center.tmx 7 10"/>`

As you can see, the action updates the variable "teleport_faint", so there is the destination.

**Old savegames? Crash?**
We took care of it by removing the default location from teleport_faint.py; we added a logger with a message.
We took care of starting maps and first fights.

**Which files are involved?**
The maps interested by "teleport_faint": mostly Spyder maps as well as some Xero maps, both for connection maps (routes, etc) and healers (Tuxecenter).

**Next steps?**
There are some things that needs to be investigated separately. 

There are locations with a particular storyline:
```
spyder_dragonscave.tmx (there is a healer in this map)
spyder_scoop1 to 4.tmx
spyder_wayfarer_inn1 & 2.tmx
```
As well as other healers:
```
spyder_cotton_cafe.tmx
spyder_leather_town.tmx 
spyder_wayfarer_inn1.tmx 
```
**How are planned the next step?**
Understanding the logic behind the storylines, so if someone is blocked in a map for a plot, then we can think some specific solution.


kerizane:
Add summary.

kerizane:
Comment above has been edited so most of the following conversation won't make sense....

Original:
___

Don't make the change yet, bit maybe we should automatically heal, to stop the player from leaving with no health.

Qiangong2:
What @kerizane said, plus it should be set so it's the last tuxecenter you visit, not the closest one. Otherwise you could potentially skip certain parts of the story

JaskRendix:
> What @kerizane said, plus it should be set so it's the last tuxecenter you visit, not the closest one. Otherwise you could potentially skip certain parts of the story

It's not the closest, if you look at the scheme it follows the story, always the previous location.

Qiangong2:
Again, I believe it should be last visited, not closest. We want the players to still go into tuxecenters, not to just ignore them unless they die.

kerizane:
Did you see this discussion? https://github.com/Tuxemon/Tuxemon/issues/1084

ultidonki:
I think this PR is fine, personally - the 2 things suggested are nice to have, but this already would be a huge improvement. 

> maybe we should automatically heal, to stop the player from leaving with no health.

I think this could be fixed by just adding an event to each tuxecenter - 'if player_defeated', then set status and heal monster health, just like the nurse does. Just copy + paste that event to each tuxecenter, there's only 5 of them. I can add this if you want, @JaskRendix 

> it should be set so it's the last tuxecenter you visit, not the closest one. Otherwise you could potentially skip certain parts of the story

I don't think this is too important for the Spyder campaign, I agree it would be nice to have, but this already does 90% of that work based on the story. If @JaskRendix can get that working, then that would be better - but otherwise I think this is fine just for the Spyder campaign, I don't think you could skip the story based on this. 

> Did you see this discussion? https://github.com/Tuxemon/Tuxemon/issues/1084

I wrote an idea in there about how we could automatically teleport to the most recent tuxecenter, using teleport_faint - but apart from that, all my other ideas are addressed by JaskRendix's player.battle_history, in fact being able to see a history of all battles against a trainer, including the most recent one, is a good idea I think. :smiley:

JaskRendix:
Ok, the biggest problem about the last visited center is fixed.

I created the next action "teleport_center" based on "teleport_faint", where the focus is the last visited center (game_variables "lastcenter").

The variable gets updated in each center, after healing the monster. So no heal, no update.

This is the reason why I updated the healing centers tmx with a "set variable".

Notice that only in the spyder_bedroom the variable gets updated as the spyder_center in Cotton town.

mods/tuxemon/maps/spyder_bedroom.tmx
```diff
@@ -67,6 +67,7 @@
     <property name="act1" value="translated_dialog spyder_papertown_restinbed"/>
     <property name="act2" value="set_monster_health ,"/>
     <property name="act3" value="set_monster_status ,"/>
+    <property name="act4" value="set_variable lastcenter:spyder_healing_center.tmx"/>
```
kerizane: This is a different map?

JaskRendix: Exactly, if you heal in the bedroom, then it sets the variable in Cotton Town (Paper Town > Route1 > Cotton Town). So if you lose against a wild monster in Route 1, then you'll be teleported to the Cotton Town center. I'm not convinced.

kerizane: Why not teleport back here?

JaskRendix: Yeah, I'm looking into. It's because of this: 
`teleport = [player.game_variables["lastcenter"], 7, 10]`
7,10 doesn't exist in the bedroom, the bedroom is 11x9.
Now I'm checking, I'll update as soon as I fix it.

kerizane:
Do we really need a new action? Can't we just use teleport_faint?

JaskRendix:
> Do we really need a new action? Can't we just use teleport_faint?

Solved.

kerizane:
**Maybe wait until we agree an implementation.**

I don't think the coordinates should be defined within the action, it should be defined in the variable. Modders may want to set the destination to a completely different map.

Use the original implementation of the teleport_faint, use the original variable and add the coordinates to the variable when you are setting it on each map.

I'm also thinking we need automatic healing. Shall we just add that to teleport faint? @ultidonki ?

Would we ever want to use teleport faint without healing?

JaskRendix:
> Maybe wait until we agree an implementation.

Nah, I'm going to close the PR. The point of teleport_faint is to bring the exhausted/fainted character inside the last visited center as well as avoiding that someone can wander around with fainted monsters without incurring in encounters, etc. Never thought about automatic healing, you can do it by yourself.

Now you're telling me about coordinates, then maybe next week there will be a question like "Do we really need a new action? Can't we just use teleport or transition_teleport?". It doesn't make sense.

ultidonki:
@JaskRendix, I'm sorry you've done lots of work but it hasn't been accepted yet. I think you're the only person who's tried to get this working so far, so please keep going because I think it will improve Tuxemon a lot!

@kerizane, I'll raise the automatic healing as a seperate issue, because I'm 99% sure it's fine, but it doesn't need to be added to this PR. 

@JaskRendix - Sorry, I think kerizane confused you with the co-ordinates thing. 
If you do the following I think this PR should be accepted:
1. Delete the changes to teleport_faint.py
2. In each spyder health center map, change `lastcenter` to `teleport_faint`, and add ` 7 10` at the end of the variable:
`- <property name="act90" value="set_variable lastcenter:cotton_cathedral.tmx"/>`
`+ <property name="act90" value="set_variable teleport_faint:cotton_cathedral.tmx 7 10"/>`
3. In the spyder_bedroom change it to `teleport_faint` and ` 7 6` instead:
`- <property name="act4" value="set_variable lastcenter:spyder_bedroom.tmx"/>`
`+ <property name="act4" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"/>`
(Basically, teleport_faint accepts co-ordinates to teleport to as the last 2 numbers in the teleport_faint variable. Doing it this way means modders can change where to teleport to in the .tmx files rather than needing to touch the .py files, which we try to avoid because most of them aren't python programmers!) 

This is probably why you're the first person to try and get this working (because teleport_faint is so confusing) but it should work if you do that! Thank you!

JaskRendix:
@ultidonki  you're right, I was so focused on getting rid of the old "teleport_faint" that I didn't notice the .split(" "), then everything was clear. As you said, it works. I don't know what else to say.

kerizane (dismissed):

kerizane:
Looks good. We can get the healing working in another ticket. 

Thanks for doing this, modifying 35 files can't have been fun.

kerizane:
I will try to do a quick test this evening and will merge it if everything is fine.

mods/tuxemon/maps/spyder_bedroom.tmx
```diff
@@ -67,6 +67,7 @@
     <property name="act1" value="translated_dialog spyder_papertown_restinbed"/>
     <property name="act2" value="set_monster_health ,"/>
     <property name="act3" value="set_monster_status ,"/>
+    <property name="act4" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"/>
```
kerizane, line 70: If you never rest, then when you faint you will be teleported to the default target, which is from a different mod!

Please add a new event within spyder_bedroom to default it. I think this will do:

name="act1" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"
name="cond1" value="not variable_set teleport_faint"

ultidonki:
> @ultidonki you're right, I was so focused on getting rid of the old "teleport_faint" that I didn't notice the .split(" "), then everything was clear. As you said, it works. I don't know what else to say.

No problem, thanks for fixing it so quickly! And thanks again for working on this feature, it should be a really good one when it's merged! :smile:

kerizane:
I was curious as to what would happen if you entered an area without any tuxemon, but the "player defeated" is False if you don't have any tuxemon, so that's fine.

Bad news, you actually missed some maps. There is a bunch of others maps if you play the "Xero" campaign, rather than the "Spyder" campaign. I'm fine with leaving the Xero campaign for another ticket if you like.

I'd also like approval from a map developer who has worked on the spyder campaign, as there might be something I've missed. I'll post on the discord.

ultidonki:
> Bad news, you actually missed some maps. There is a bunch of others maps if you play the "Xero" campaign, rather than the "Spyder" campaign. I'm fine with leaving the Xero campaign for another ticket if you like.

I think it's fine to leave the Xero campaign for another ticket! Getting the Spyder campaign working was probably a lot of work!

> I'd also like approval from a map developer who has worked on the spyder campaign, as there might be something I've missed. I'll post on the discord.

I'll wait for Sanglorian to approve, but it looks good to me. 

I think the change to set the default teleport_faint area is good, so we don't end up teleporting between the other mods!

If we're being perfectionist, I noticed the following, BUT we can look at these in a seperate issue! In fact, I can change them later as they're just small changes. I don't want these problems to slow down the PR as @JaskRendix has already worked so hard on it, so I can change them later myself :smile:
- spyder_Wayfarer_inn 1 and 2 needs 'teleport faint' in there, also there's a healer in wayfarer inn 1. 
- spyder_scoop 1, 2, 3, 4 all need teleport_faint, I think... 
Actually, for both of the above areas, in the story you're quarantined in them, and not supposed to leave. So I'm not sure if teleport_faint-ing out to another town would make sense... I guess that's a question for Sanglorian when I open these in another issue. 
- There's also another 2 healers hidden around - in spyder_leather_town and spyder_cotton_cafe, Team Bazaar will heal you for free. But I'm not sure if we should automatically teleport there after being defeated, or go straight to the normal centres, so I guess that's another question for Sanglorian. 
- You can lose the battle in spyder_paper_town.tmx, so we should automatically heal the player after the first battle. Otherwise, when they step into route 1 they'll immediately teleport back to the player's bedroom. 

But like I said, we should raise this as a seperate issue, I can change them after this PR is merged. In fact, after the battle_history PR is merged too, I think I might play through the entire Spyder campaign again just to see how it plays with these 2 features added in. :smiley:

mods/tuxemon/maps/spyder_bedroom.tmx
JaskRendix: You're right

JaskRendix:
@ultidonki  before the PR I did these checks. I focused on spyder_maps.

Following the missing healers as well as the locations without event. I decided, since there is a healer inside **dragonscave** to remove the "teleport_faint" & "is player_defeated" from dragonscave.tmx, because it needs to be addressed separately as wayfarer_inn and others.

Keyword: "set_monster_health ," (where are the "missing" healers)
```
list integrated in the first post
```
Keyword: start_battle (where are the battles -> chances to faint) -> no presence of teleport_faint & is player_defeated 
```
list integrated in the first post
```

> You can lose the battle in spyder_paper_town.tmx, so we should automatically heal the player after the first battle. Otherwise, when they step into route 1 they'll immediately teleport back to the player's bedroom.

It happens like this, but, yeah, I wouldn't be surprised if in 69 business days someone raises the issue/bug, I saw many weird things around issues/bugs. So, under this aspect, I can consider the auto-healing.

kerizane:
We also need to consider what to do with existing save files that don't have teleport_faint configured, as they would also be teleported to the wrong mod after being defeated.

I think the simplest solution is to just modify the action to do nothing if the variable isn't set. I don't think the default makes sense any more given the two mods. Please check the existing use of teleport_faint to make sure this doesn't cause issues anywhere.

I know we wouldn't have to worry about this if I had just let you have a new action, but it didn't really make sense to have two similar actions. Plus this way is more flexible for modders rather than having coordinates hardcoded within the action.

JaskRendix:
You're right, and about this, so:

**1) Default spyder_bedroom**
Solved.

**2) Lose marion event**
Solved.

**3) No idea about taba_town.tmx**

Here @ultidonki  knows better than me. I have no idea how to approach it, because I have no knowledge about the mod.

**4) Surprise grass in the wrong map**

Doing a 5th check, I noticed a... bug? Something unusual. In spyder_candy_town, someone added grass inside the town. Looking at the code, there are 2 tiles of grass (x=16, y=0, width=2, height=1).

ultidonki:
> Following the missing healers as well as the locations without event. I decided, since there is a healer inside dragonscave to remove the "teleport_faint" & "is player_defeated" from dragonscave.tmx, because it needs to be addressed separately as wayfarer_inn and others.

That's a good point, thank you for checking. Yeah, we can address it in a seperate PR, I can raise an issue of all the things we still need to do after this PR is merged. 

> I think the simplest solution is to just modify the action to do nothing if the variable isn't set.

I think that's a good point, there's probably not many save games out there, but that would be a good change. :smile:

> **1) Default spyder_bedroom**
> 
> I removed the "default" event from spyder_bedroom; I added this: ` <property name="act91" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"/>` in the event "First Fight" in Paper Town.
> 
> This solve the problem of being teleported in the wrong scenario -> Cause: the player loses against Billie.

Sorry, but I'm not 100% sure why you did this? The game will work exactly the same way as it did when there was an automatic action to default the teleport_faint variable to the player bedroom, so I can't see which problem this solves. 
There isn't a teleport_faint in spyder_paper_town.tmx, so if you lose against Billie, nothing happens (which is fine). All you've done is move the initialization of the teleport_faint to after the battle, which is fine because it isn't used until spyder_route1.tmx, so nothing is changed here. 
The problem I was trying to explain was - because there's no 'teleport_faint' in spyder_paper_town, if you're defeated then `is player_defeated` will become true. So, as soon as you go into spyder_route1.tmx, you'll immediately be teleported back to the player bedroom. 
So, if you put a `teleport_faint` into spyder_paper_town, and if you make the change kerizane suggested, about not teleporting if the teleport_faint variable was blank, I think your change would make it work. (You would be teleported after the Billie battle, and not halfway through the cutscene). But at the minute, this change doesn't do anything. 

> **2) Lose marion event**
> 
> I fixed the "Lose marion" event in spyder_route3_trainer by removing: ` <property name="act42" value="teleport_faint"/>`

Whoops... I shouldn't have merged spyder_route3_trainer.tmx in #1173 - This file can just be deleted, it isn't used. Sorry about that. 
 
> **3) No idea about taba_town.tmx**
> 
> Here @ultidonki knows better than me. I have no idea how to approach it, because I have no knowledge about the mod.

We can do the Xero campaign in another pull request, otherwise it's even more stuff you need to do and test. But, if you do want to do it, it's the same basic idea: Set the teleport_faint variable in player_house_bedroom.yaml (notice it's a yaml file, not a .tmx file), and the other healing center. Then, in taba_town.tmx, do teleport_faint if player defeated as normal.
There's another event already in taba_town.tmx, that has 'teleport_faint', but you can just delete this one, it looks like someone was testing it but currently it does nothing. 
You could add teleport_faint to route1.tmx too, but I'd check with Qiangong2 first before making that change, just in case they're working on it (Qiangong2 is in charge of the Xero campaign, Sanglorian in charge of the Spyder campaign) 

> **4) Surprise grass in the wrong map**
> 
> Doing a 5th check, I noticed a... bug? Something unusual. In spyder_candy_town, someone added grass inside the town. Looking at the code, there are 2 tiles of grass (x=16, y=0, width=2, height=1).

> I removed it. It could be dangerous having random_encounters inside the town (2 tiles). If this change is accepted, then someone needs to edit the map and remove the two tiles.

Good catch, that's a good change! :smile:

kerizane:
I'd rather keep the surprise grass, unless you can suggest an actual issue with it. I'm sure sanglorian knew what he was doing when he added this.

kerizane:
Let's keep this to the Spyder campaign. Regarding tabatown, I just want the PR to not break anything. If teleport faint is unused, then we can delete it. @ultidonki are you sure it's unused?

ultidonki:
> Regarding tabatown, I just want the PR to not break anything. If teleport faint is unused, then we can delete it. @ultidonki are you sure it's unused?

Actually, I've just checked, and it IS used, sorry about that. If you're defeated in taba_town.tmx, this event will automatically send you to the default health center - so it's already implemented for this map in the Xero campaign!

So there's no changes needed to the Xero campaign. (However, the `teleport_faint` variable is blank in the Xero campaign, so if the "don't teleport if teleport_faint variable is blank" feature is implemented, we'd need to set teleport_faint in the Xero campaign to get teleporting working again.)

kerizane:
As this event is potentially relying on the default target of "healing_center" let's add an event to this map which sets the variable to the healing center if it's not set.

JaskRendix:
> As this event is potentially relying on the default target of "healing_center" let's add an event to this map which sets the variable to the healing center if it's not set.

As requested. About the savegame, maybe we can fix it setting the default in all the other missing maps (Spyder), like Dojo, ScoopMarket, etc. (where there are no battles or random encounters) pointing to the town center (ex. the art gallery to the center of the same town, etc). So if someone saved there, then the variable is updated.

kerizane:
I don't think it's worth it; we try to avoid having overly complex error handling logic. If we remove the default case from the teleport_faint action then the previous behaviour (not teleporting) will happen until somebody saves for the first time and the new logic will kick in.

As long as the game doesn't explode, or end up in some weird state (teleported to a different mod), then I think we're fine.

kerizane:
Just to make sure we're on the same page:

You're almost done: I just want you to remove the default logic from teleport_faint.

@ultidonki anything else remaining?

JaskRendix:
> You're almost done: I just want you to remove the default logic from teleport_faint.

I removed the default logic.

kerizane:
Now the game will crash if this isn't set.

If it isn't set then log a warning and return.

Test this scenario please.

JaskRendix:
> Test this scenario please.

Yeah, I tested on an old savegame without the variable. It doesn't crash anymore, the terminal starts to show the error. As soon as the player heals the monster in a center, it doesn't happen anymore (logger).

tuxemon/event/actions/teleport_faint.py
```diff
@@ -51,10 +53,12 @@ class TeleportFaintAction(EventAction[TeleportFaintActionParameters]):
...
         player = self.session.player
 
-        # Start with the default value, override if game variable exists
-        teleport = ["healing_center.tmx", 7, 10]
+        # If game variable exists, then teleport:
         if "teleport_faint" in player.game_variables:
             teleport = player.game_variables["teleport_faint"].split(" ")
+        else:
+            logger.error(f"Old savegame. Please heal your monsters in a center to fix it.")
+            return None
```
kerizane: Bit picky, but as the calling function isn't expecting to receive a value from this method, I would use `return` rather than `return None`

JaskRendix: Fair point, I'll fix it

kerizane:
I like that you're thinking about the user with this message, but I would do this differently by making it less implementation specific. 

If a modder is implementing the game from scratch then this message wouldn't make sense to them.

I would just say: "teleport_faint action failed, because the teleport_faint variable has not been set."

kerizane:
If necessary we could provide a message to users via save_upgrader, but as this is new behaviour I don't think we need to warn them about this.

JaskRendix:
Yes, I wanted to give some indication to the user, but I agree with you about "teleport_faint action failed, because the teleport_faint variable has not been set.".

kerizane:
Ok, this looks good. There's been a lot of back and forth on this and I appreciate the work you've put into it. Please write a summary of this PR for Bitcraft so he doesn't have to read through a million comments.

Explain what you've done and why.

JaskRendix:
Sure, I'm already updating and summarizing the first post.

ultidonki:
> @ultidonki anything else remaining?

All looks good to me. Thanks for doing all this work, @JaskRendix !

JaskRendix:
@kerizane do you want that I close the PR, squash everything, and reopen a new one? No problem for me.

I'm going to close and reopen, 16 commits are too many.

PR #1241, opened 2022-08-17
Track battle history 

JaskRendix:
Split from a larger PR - #1237 
This takes over PR #716 and relates to issue https://github.com/Tuxemon/Tuxemon/issues/574

**1. Creating new dict called "battle history" and editing the following files: combat.py, worldstate.py, npc.py and save_upgrader.py**

Here is a list:

- combat.py
- worldstate.py
- save_upgrader.py

**2.a. combat.py**
As follows.

Creation of new variables (game_variables) called:
```game_variables["battle_total"] = number of battles (only NPC, not Random)
game_variables["battle_draw"] = number of battles draw
game_variables["battle_won"] = number of battles won
game_variables["battle_lost"] = number of battles lost
game_variables["percent_draw"] = round(value) (i.e. 42 or 69)
game_variables["percent_won"] = round(value) (i.e. 42 or 69)
game_variables["percent_lost"] = round(value) (i.e. 42 or 69)
```

Reason? Modders. They can create a story, a tournament, etc. where they need this data, because only people with xx% of victories can access.

Creation of new variable (based on battle_history) called:
`self.players[0].battle_history[self.players[1].slug] = "won/lost/draw"`

Output:
```
{'spyder_route2_billie': 'won'}
{'spyder_route2_billie': 'lost'}
{'spyder_route2_billie': 'draw'}
```

**2.b. worldstate.py**
An addition related to battle_history()

**2.c. npc.py**
A series of additions related to battle_history()

**2.d. save_upgrader.py**
Updated the savegame.

**3. Developing an event action called: battles_print**
So it's possible to print all the battles (npc.slug + result). It's the copy of "print".

Reason? Modders. So they can have a complete map of battle_history.

**4. Developing an event condition called: battle_is**
As you can see below, it checks if you won, lost or draw against a specific npc.slug

```
    Script usage:
        .. code-block::

            is battle_is <character>,<result>

    Script parameters:
        character: Npc slug name (e.g. "npc_maple").
        result: One of "won", "lost" or "draw"
```
Reason? Modders. So they can check if a player won a single time against a NPC (i.e. spyder_route2_billie).

**5. Developing an event action called: set_battle**
As you can see below, now you can set new variables.

```
    Set the key in the player.battle_history dictionary.

    Script usage:
        .. code-block::

            set_battle <character>:<result>

    Script parameters:
        character: Npc slug name (e.g. "npc_maple").
        result: One among "won", "lost" or "draw"
```
Reason? Modders. So they can set new variables (i.e. set_battle spyder_route2_billie:lost).

Everything wrote above runs. It took me almost a week, but of course when you're inspired things go fast.


kerizane:
Thanks, please copy the message over again.

kerizane:
Just so you know, I probably won't merge this PR until bitcraft has taken a look at it, and he might not be around for a month or more. 

I will make some suggestions, but you don't need to make the changes until we have agreed a design, as I don't want you to waste your time on something he might disagree with.

kerizane:
I don't think the date as the key works, as this means we can forget that we have defeated / lost to a certain character. Do we even need a key? I think we can handle this as either:
A single list that we append to; or a list per npc, using their slug as the key to the group (I prefer this option).
I think the timestamp could be useful, so I would implement the data like so:
```
"spyder_route2_billie": [
    {"timestamp": "2022-08-17T19:42:15", "outcome": "lost"},
    {"timestamp": "2022-08-17T20:42:15", "outcome": "won"}
]
```
I think this is more future-proof than:
```
"spyder_route2_billie": [
   ("2022-08-17T19:42:15", "lost"),
   ("2022-08-17T20:42:15", "won"),
]
```
And I don't think the space difference is that bad.

ultidonki:
> Just so you know, I probably won't merge this PR until bitcraft has taken a look at it, and he might not be around for a month or more.

I agree! But from my point of view, I think it's very good so far! It can cover most cases I can think of - never battled before, lost a battle, won a battle. Those are the most important ones used 99% of the time so it works from that point of view. 

It's also how I imagined it might work after reading the discussion in #716 - so hopefully it will make everyone happy, when they get a chance to look. :smiley:

bitcraft:
> I don't think the date as the key works, as this means we can forget that we have defeated / lost to a certain character. Do we even need a key? I think we can handle this as either: A single list that we append to; or a list per npc, using their slug as the key to the group (I prefer this option). I think the timestamp could be useful, so I would implement the data like so:
> 
> ```
> "spyder_route2_billie": [
>     {"timestamp": "2022-08-17T19:42:15", "outcome": "lost"},
>     {"timestamp": "2022-08-17T20:42:15", "outcome": "won"}
> ]
> ```
> 
> I think this is more future-proof than:
> 
> ```
> "spyder_route2_billie": [
>    ("2022-08-17T19:42:15", "lost"),
>    ("2022-08-17T20:42:15", "won"),
> ]
> ```
> 
> And I don't think the space difference is that bad.

i agree about this one.  it would be better implemented as a list of objects, rather than a dictionary.  but i don't really think its so bad to reject it.  in general though, i see something like this as a new system that probably deserves its own class to hide the data structure under it.  i think it looks ok as-is, but anything more complicated should be refactored into its own class or module to hide the details of searching for battles, looking up averages, stats, etc.

kerizane:
@bitcraft so what's the plan for this PR?

JaskRendix:
Yes, it doesn't make sense to implement something like this. So, the right way is setting this as a list,then in the db.py setting the NpcBattleHistoryItemModel with the following items. As well as sequence below NPC model, closing with npc-schema.json (time, outcome + slug). And then append in combat.py? Right?

kerizane:
We currently don't save changes to the npc, so it will need to be on the player, not the npc.

kerizane:
Ok, let's avoid overcomplicating it. If you fix the data structure so that we aren't using the date as a key then I will accept it.

JaskRendix:
Fair enough. Now this is the output: {'spyder_route2_billie': 'lost'}
I got rid of the date, now it's simpler.
Now **spyder_route2_billie** (and NPC friends) can support lost, won or draw.
I updated the first post.

kerizane:
So we only track the most recent result for each npc? Probably fine.

JaskRendix:
Exactly, but with set_battle we can set the variable before as well as "reboot" the variable.

Like, if the variable is {'spyder_route2_billie': 'lost'}, then we can set again to "empty" or "lost" without messing the statistics (statistics are on game_variables), so the match can be played again

kerizane:
Thanks, I'll try to take another look over the next few days.

kerizane (approved):

kerizane:
Looks good, thanks!

PR #1242, opened 2022-08-17
Fix 994 - Bad Teleport

### LINKED ISSUE #994: Teleporting to a new map sometimes puts you in the wrong position, if playing in full-screen
**Describe the bug**
Teleporting to a new map sometimes doesn't send you to the correct tile location in the action.
More likely to happen under rare conditions like playing the game in full-screen. 

**To Reproduce**
1. Find a .tmx or .yaml map with a transition_teleport action on it
2. Set the last parameter, the 3rd one, from 0.3 to a lower value, like 0.1
3. Load the map in-game and stand on that tile, to teleport to a new map
4. When the value is 0.1, I can reproduce the problem every time
5. When the value is higher, like 0.5, the problem never happens, even in full-screen
6. The tile position you're standing on is the position from the previous map, as if your position hasn't been updated

**Expected behavior**
Teleporting from one map to another should update your tile position to the x, y position, the first 2 parameters in the teleport_transition action

**Screenshots**
N/A

N/A, but bug was first noticed when we were playing in full-screen, setting the resolution to default made the issue disappear

**Smartphone (please complete the following information):**
Not used

**Additional context**
Noticed by freshreplicant and ultidonki on the Matrix channel
The problem with it happening in full screen at the default value of 0.3, has started happening since commit 'less aggressive loop sleeping', however I did a checkout from August 1st 2021 and the problem still happened then when I changed the teleport_transition value to 0.1, so the bug is probably quite old but just became noticeable recently. 
When trying to debug it, I got to the handle_delayed_teleport method in states/world/worldstate.py, which updates the player position, and confirmed that the self.client.player's position3 and tile_pos were updated to the new one in this method, even when the problem occurred. However, the player's position was still wrong when the next map loaded. So the problem is happening after this point. 
As discussed on the Matrix/Discord bridge, this part of the system might be re-written soon, which might fix the problem, so this is just an issue to remember to test it again if fixed.

kerizane:
Fixes #994 
Teleport would sometimes put the player on the current position on the new map, rather than the expected position defined by the action parameters.

### DIFF
--- a/tuxemon/states/world/worldstate.py
+++ b/tuxemon/states/world/worldstate.py
@@ -264,7 +264,7 @@ class WorldState(state.State):
         self.remove_animations_of(self)
         self.remove_animations_of(cleanup)
 
-        self.stop_player()
+        self.stop_and_reset_player()
 
         self.in_transition = True
         self.trigger_fade_out(duration)


bitcraft (approved):
nice and easy fix. thanks.

Issue #677, opened 2020-03-01
Display image not updated after an item sold out in tuxe_mart


amycqx:
When an item is sold, the `ShopMenuState(Menu)` would `reload_items` and rerender the screen, but the image on the left is not updated when an item is sold out and the cursor is pointing at a new item. 
Currently, the `on_menu_selection` functions are not able to catch that change of item, need to figure out how to change the selected item.

To test: Go to any implemented store and buy all of the potions. When the store updates, you will be pointing at Capture Device but seeing potions and the description of potions.

bitcraft:
Cool, good observation.

kerizane:
This also happens when releasing a monster.

kerizane:
Only happens if you sell your last item.

PR #1243, opened 2022-08-18
Shop menu cursor fix

kerizane:
Fix issue where the game would crash if you sold the last item in the menu. Fixes (https://github.com/Tuxemon/Tuxemon/issues/677) display issue where the image wouldn't be updated when you sold the last item.

### DIFF
--- a/tuxemon/menu/menu.py
+++ b/tuxemon/menu/menu.py
@@ -358,23 +358,32 @@ class Menu(Generic[T], state.State):
         self._needs_refresh = True
         items = self.initialize_items()
 
-        if items:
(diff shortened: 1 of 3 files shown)


bitcraft:
have you done a play test with this change?  i see there was special handling for the monster menu, probably due to it changing when monsters are moved around.  as long as the monster menu works fine in game, we can merge this.

kerizane:
Yes, but I'll do another check to be sure

kerizane:
Yeah it's fine. I added an extra fix: the monster image gets updated after releasing the monster.

bitcraft (approved):

PR #1244, opened 2022-08-18
Allow shops that only buy or only sell

### LINKED ISSUE #1090: Allow shops that only sell, or only buy, not both
In my mod I might have 3 or more shopkeepers in the same room. 
If the player wants to see what they can buy from each one, they'll have to go:
Talk to shopkeeper -> Click 'Buy' -> See buy menu
Each time they want to see what's on offer from each shop. 

I don't really see a need to sell stuff in my mod? But buying will be very important for progressing. 
So it would be 'nice', but not essential, to just have shops that only buy, or only sell, and you can skip the buy/sell menu:
Talk to shopkeeper -> See buy menu

In the action open_shop, have a new parameter, like:
``` open_shop tuxemart_keeper,buy ```
``` open_shop tuxemart_keeper,sell ```
``` open_shop tuxemart_keeper,buy:sell ```
I'm not sure about the last one, if someone wants 'both' menus, if buy:sell is simple enough? Or if there's a better syntax to use. 

I can probably live with having both buy and sell on the menus, maybe I'm over-complicating things. Happy to be over-ruled if people disagree.

kerizane:
Closes #1090

### DIFF
--- a/tuxemon/event/actions/open_shop.py
+++ b/tuxemon/event/actions/open_shop.py
@@ -21,17 +21,19 @@
 
 from __future__ import annotations
 
-from typing import NamedTuple, final
+from typing import NamedTuple, Optional, final
 
 from tuxemon.event import get_npc
 from tuxemon.event.eventaction import EventAction
 from tuxemon.item.economy import Economy
 from tuxemon.states.choice import ChoiceState
 from tuxemon.states.items import ShopBuyMenuState, ShopMenuState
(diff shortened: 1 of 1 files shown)


ultidonki:
Thanks for all these fixes recently - I haven't had time to test them (even though I raised some of them as issues/requests) but I'm sure they work! :smile:

bitcraft (approved):

PR #1246, opened 2022-08-21
Add config option to skip dialog animation

kerizane:
The dialog animation is a bit slow, especially if you're debugging a script and reading the same text over and over.

Add option to config: "dialog_speed"
Options:
slow - (default option, current implementation)
max - no animation

Can add "medium" and "fast" later, but I think this is constrained by the clock speed and will be a bit more work.


bitcraft (approved):
this is a nice option to have, thanks

PR #1248, opened 2022-08-21
1155 npc not found error msg

### LINKED ISSUE #1155: Print useful message when npc is not found
Hard to know why npc lookup failed for map script

might be a good idea to do something like: 
- make custom NPCNotFound exception
- if uncaught prints that it cannot be found
- then prints off npc slugs that are similar
- "did you mean 'npc_red'?

the stdlib has something that can do close matches to strings, so its doable without much effort

kerizane:
Fixes #1155  Implemented for all tables, because why not.

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -30,9 +30,11 @@
 
 from __future__ import annotations
 
+import difflib
 import json
 import logging
 import os
+import sys
 from enum import Enum
 from operator import itemgetter
 from typing import Any, Dict, Literal, Mapping, Optional, Sequence, overload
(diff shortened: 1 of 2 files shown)


kerizane:
"Lookup failed for unknown npc 'npc_mon'. Did you mean 'npc_mom'"?
"Lookup failed for unknown npc 'mom'. Did you mean 'npc_mom'"?
"Lookup failed for unknown npc 'red'. Did you mean 'npc_red'"?
"Lookup failed for unknown npc 're'. No similar slugs. Are you sure it's in the DB?"

bitcraft (approved):
looks great, ty.  very helpful output.

Issue #1257, opened 2022-08-26
Remove unused maps


kerizane:
There are some maps that aren't used as part of the current storyline, and are probably abandoned work. This means unnecessary maintenance work sometimes takes place on these maps.

Maybe we should delete the maps or move them to a different folder.

@JaskRendix please paste in the files names you think aren't used.

JaskRendix:
You anticipated me! Here:


- 37707_tower.tmx (no current storyline)
- 37707_town_missing.tmx (no current storyline)
- 37707_town.tmx (no current storyline)
- bedroom_test.tmx (no current storyline)
- city1_mart_junkyard.tmx (connected to the old tuxe_mart.tmx)
- gadget_store1.tmx (unaccessible from taba_town.tmx)
- gaming_hall.tmx (no current storyline)
- rubberduck_cave_01.tmx (no current storyline)
- rubberduck_city_01.tmx (no current storyline)
- rubberduck_route_01.tmx (no current storyline)
- test_pathfinding_town.tmx
- test_pathfinding.tmx
- test-tileprops.tmx
- test.tmx
- testmap.tmx
- tt_forest_route1.tmx (no current storyline)
- tt_paper_town.tmx (no current storyline)
- tt_searoute1.tmx (no current storyline)
- tt_searoute1.tmx
- tt_testing_2.tmx
- tt_testing.tmx
- tutorial.tmx

No one (Xero or Spyder) can access these maps.

kerizane:
I think moving the unused ones into a new folder should be safe.

JaskRendix:
Subfolder like map/unused?

kerizane:
A subfolder within the current map folder called unused should be fine. @bitcraft ?

JaskRendix:
Yeah bitcraft, what to do? Because I looked into the code, and I'm not sure anymore about the subfolder. I was considering the amount of work if we decide in the future to create the maps_spyder/maps_xero folder too.

Maybe inside mods/tuxemon as maps_unused? I don't know. I'm sharing some thoughts. Especially **tuxemon/states/start/__init__.py** as well as **tuxemon/graphics.py** (starting new) and the savegame load if someone saved the file. 

Update:
I was doing some testing on the parallel folder (sub of mods/tuxemon like maps) and it looks like we avoid the graphics.py issues. I was already considering a strategy on how to deal with: main.py, teleport.py, load_menu.py, __init__.py and worldstate.py

kerizane:
[...]

PR #1258, opened 2022-08-26
Xero maps

JaskRendix:
The idea is renaming all the Xero maps as xero_taba_town.tmx, etc.

Which ones? The ones that the player can access.
How can the player access? Regular teleport, like from bedroom to downstairs, to taba_town, etc. (I mapped, look at the pdf)
Savegame? Yeah, I inserted a replaces for the first locations (from bedroom to route1sanglorian).

The reason: distinguish between the official Xero maps, Spyder maps and work in progress that need to be approved.

About Omnichannel in Xero, there is the building as well as the teleport, but at the moment is missing. This is the reason why I inserted Omnichannels 1234 inside the xero_maps

[xero.pdf](https://github.com/Tuxemon/Tuxemon/files/9433410/xero.pdf)


tuxemon/save_upgrader.py
```diff
@@ -78,6 +78,12 @@ def upgrade_save(save_data: Dict[str, Any]) -> Dict[str, Any]:
     if "steps" not in save_data["game_variables"]:
         save_data["game_variables"]["steps"] = 0
 
+    first_steps = [('taba_town','xero_taba_town'),('player_house_bedroom','xero_player_house_bedroom'),('route1','xero_route1'),
+    ('player_house_downstairs','xero_player_house_downstairs'),('route1_sanglorian','xero_route1_sanglorian'),
+    ('tuxe_mart_taba','xero_tuxe_mart_taba'),('healing_center','xero_healing_center'),('professor_lab','xero_professor_lab')]
+    for k, v in first_steps:
```
kerizane, line 84: This seems dangerous.

route1_sanglorian -> xero_route1_sanglorian -> xero_xero_route1_sanglorian

Just make the original data structure a dict and do a single lookup of the current value. No looping.

If somebody ever reuses one of the current names then it will be converted. Bump the version number and only run this logic for old saves

kerizane (changes requested):

tuxemon/save_upgrader.py
kerizane: Run black on the code.

kerizane:
I appreciate the effort you put into this, and the fact that you mapped it out is very helpful.

In the future it might be better to use folders, but this works for now.  I'll wait and see what other people think.

bitcraft:
yeah idk.  i wish there was a conversation about this before the work was started.  we have two campaigns sharing the same folder but thats actually a problem we need to fix.  it would be a much better use of time working on the game to allow multiple folders here.  tbh, its been a long time since i've personally looked at the issue, so i can't advise it.  i see a lot of conflicts here.  might be a good time to work on splitting the content into folders rather than renaming files.

and because content developers might have unpublished work, we would ideally announce we want to make changes to the content so people can cleanup and push their changes, then have a "code freeze" for a week or so while the content is split.  just splitting everything is going to create chaos to merge everything in while the files are being moved.

kerizane:
What do you think about moving the unused maps into a different folder?

https://github.com/Tuxemon/Tuxemon/issues/1257

Shouldn't cause any issues, and it's still obvious if a map is Spyder or Xero.

JaskRendix:
Yeah, you're right.
I close the PR, the strategy about multiple folders and code freeze for a week sounds better than renaming the files.

PR #1264, opened 2022-08-29
feature teleport on lose

JaskRendix:
Split from #1237 

Closed and reopened because of too many commits #1240 .
Teleporting to the last visited Center (updating maps)

**Why?**
Mostly two reasons:
1. Because people can wander around with fainted monsters (missing battles, storylines, etc.)
2. Because if a player loses a fight, then a player needs to be teleported to the last center.

**How?**
In the connection maps (mostly routes or where there are battles / random_encounter), we have added the following code:
```
  <object id="69" name="Exhausted" type="event" x="0" y="0" width="16" height="16">
   <properties>
    <property name="act1" value="teleport_faint"/>
   </properties>
  </object>
```
If someone with fainted monsters tries to access a connection map, it gets immediately teleported.

At the same time in the location with healers (Tuxecenter), we added the following action:

`    <property name="act80" value="set_variable teleport_faint:spyder_healing_center.tmx 7 10"/>`

As you can see, the action updates the variable "teleport_faint", so there is the destination.

**Old savegames? Crash?**
We took care of it by removing the default location from teleport_faint.py; we added a logger with a message.
We took care of starting maps and first fights.

**Which files are involved?**
The maps interested by "teleport_faint": mostly Spyder maps as well as some Xero maps, both for connection maps (routes, etc) and healers (Tuxecenter).

**Next steps?**
There are some things that needs to be investigated separately. 

There are locations with a particular storyline:
```
spyder_dragonscave.tmx (there is a healer in this map)
spyder_scoop1 to 4.tmx
spyder_wayfarer_inn1 & 2.tmx
```
As well as other healers:
```
spyder_cotton_cafe.tmx
spyder_leather_town.tmx 
spyder_wayfarer_inn1.tmx 
```
**How are planned the next step?**
Understanding the logic behind the storylines, so if someone is blocked in a map for a plot, then we can think some specific solution.

### DIFF
--- a/mods/tuxemon/maps/cotton_cathedral.tmx
+++ b/mods/tuxemon/maps/cotton_cathedral.tmx
@@ -88,6 +88,7 @@
     <property name="act60" value="npc_face cottonnurse,down"/>
     <property name="act70" value="translated_dialog okaythen2"/>
     <property name="act80" value="set_variable healme:none"/>
+    <property name="act90" value="set_variable teleport_faint:cotton_cathedral.tmx 7 10"/>
     <property name="cond10" value="is variable_set healme:yes"/>
    </properties>
   </object>
--- a/mods/tuxemon/maps/healing_center.tmx
+++ b/mods/tuxemon/maps/healing_center.tmx
@@ -72,6 +72,7 @@
     <property name="act60" value="npc_face tabanurse,down"/>
     <property name="act70" value="translated_dialog okaythen2"/>
     <property name="act80" value="set_variable chooses:none"/>
+    <property name="act90" value="set_variable teleport_faint:healing_center.tmx 7 10"/>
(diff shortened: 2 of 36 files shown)


kerizane:
Bitcraft: Yeah, I know you wanted to have global logic that checked for defeat and teleported you straight away, but Jask did it this way. Ultidonki and Qiangong2 were keen to merge this anyway as it's still a big improvement. Even if we refactor, this ticket still does valuable work in defining the spawn points.

JaskRendix:
Bitcraft: Just to be clear. I'm not pushing to merge this. If you don't consider it the right approach, no problem, I'll close the PR.

kerizane (approved):

kerizane:
I've decided to merge this.

@JaskRendix please create an issue with the things that need to be addressed.

PR #1273, opened 2022-08-30
Make spyder scenario's First Fight have a win and lose dialogue

AndyMender:
As it is now, the result message for the `First Fight` with `spyder_billie` seems as if the player actually won even if they lost
battle_encounter_fail

I haven't played through the first battle thoroughly enough, but we could add "success" and "defeat" conditions with separate dialogue slugs. I already added the initial pieces, but am not sure how to set a "player won" condition in the Tiled TMX file.

Any pointers?

---
Current behavior:
1. Start first battle with Spyder Bille as usual via the "First Battle - Start" event
2. If battle is lost, trigger the one-time "First Battle - Lose" event
3. If battle is won, trigger the one-time "First Battle - Win" event

### DIFF
--- a/mods/tuxemon/maps/spyder_paper_town.tmx
+++ b/mods/tuxemon/maps/spyder_paper_town.tmx
@@ -433,7 +433,7 @@
   <object id="425" name="First Fight - Win" type="event" x="384" y="144" width="16" height="16">
    <properties>
     <property name="act1" value="create_npc spyder_billie,26,12"/>
-    <property name="act10" value="npc_face spyder_billie,down"/>
+    <property name="act10" value="npc_face spyder_billie,up"/>
     <property name="act20" value="translated_dialog spyder_papertown_firstfight_win"/>
     <property name="act30" value="pathfind spyder_billie,15,14"/>
     <property name="act40" value="remove_npc spyder_billie"/>
@@ -446,7 +446,7 @@
   <object id="426" name="First Fight - Lose" type="event" x="416" y="144" width="16" height="16">
    <properties>
     <property name="act1" value="create_npc spyder_billie,26,12"/>
-    <property name="act10" value="npc_face spyder_billie,down"/>
+    <property name="act10" value="npc_face spyder_billie,up"/>
     <property name="act20" value="translated_dialog spyder_papertown_firstfight_lose"/>
     <property name="act30" value="pathfind spyder_billie,15,14"/>
     <property name="act40" value="remove_npc spyder_billie"/>
--- a/mods/tuxemon/maps/spyder_paper_town.tmx
+++ b/mods/tuxemon/maps/spyder_paper_town.tmx
@@ -423,13 +423,14 @@
     <property name="act61" value="npc_face player,down"/>
     <property name="act70" value="translated_dialog spyder_papertown_firstfight"/>
     <property name="act80" value="start_battle spyder_billie"/>
-    <property name="act81" value="translated_dialog spyder_papertown_firstfight2"/>
+    <property name="act81" value="translated_dialog spyder_papertown_firstfight_win"/>
+    <property name="cond1" value="is player_defeated"/>
     <property name="act90" value="set_variable firstfightdue:no"/>
     <property name="act91" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"/>
     <property name="act94" value="pathfind spyder_billie,15,14"/>
     <property name="act95" value="remove_npc spyder_billie"/>
     <property name="act96" value="unlock_controls"/>
-    <property name="cond1" value="is variable_set firstfightdue:yes"/>
+    <property name="cond2" value="is variable_set firstfightdue:yes"/>
    </properties>
   </object>
   <object id="428" name="Play Music" type="event" x="0" y="0" width="16" height="16">
--- a/mods/tuxemon/maps/spyder_paper_town.tmx
+++ b/mods/tuxemon/maps/spyder_paper_town.tmx
@@ -411,7 +411,7 @@
     <property name="cond2" value="is party_size less_than,1"/>
    </properties>
   </object>
-  <object id="424" name="First Fight" type="event" x="384" y="144" width="16" height="16">
+  <object id="424" name="First Fight - Start" type="event" x="400" y="160" width="16" height="16">
    <properties>
     <property name="act1" value="lock_controls"/>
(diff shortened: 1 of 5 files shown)


tuxemon/event/actions/start_battle.py
```diff
@@ -68,17 +68,15 @@ def start(self) -> None:
         npc = world.get_entity(self.parameters.npc_slug)
         assert npc
         if len(npc.monsters) == 0:
-            logger.warning("npc has no monsters, won't start")
+            logger.warning(f"npc '{self.parameters.npc_slug}' has no monsters, won't start trainer battle.")
```
AndyMender, line 71: Minor improvements added in an isolated commit. Not critical to this PR so I can remove them.

mods/tuxemon/maps/spyder_paper_town.tmx
```diff
@@ -423,13 +423,14 @@
     <property name="act61" value="npc_face player,down"/>
     <property name="act70" value="translated_dialog spyder_papertown_firstfight"/>
     <property name="act80" value="start_battle spyder_billie"/>
-    <property name="act81" value="translated_dialog spyder_papertown_firstfight2"/>
+    <property name="act81" value="translated_dialog spyder_papertown_firstfight_win"/>
+    <property name="cond1" value="is player_defeated"/>
```
AndyMender: This condition is currently incorrect - it should return `true` if the NPC was defeated, not the player.

kerizane:
I think there is an automatic game_variable called battle_win and battle_lose.

Edit: Maybe this hasn't been merged yet.

kerizane:
I think you will need to split the current event into four:

Start the battle
If win
If lose
Cleanup

You may need your lose logic to have a lower id than the "if player defeated, teleport_faint" (which I think is the last event in the file) in order for it to run first.

ultidonki:
This would be great if you could get it working - <s>but I would wait for #1241 to be merged first - this will implement some new conditions, `battle_is` and `battle_result`, which you can use to check if the player has won or lost against a certain NPC. 

Then you could just use `is battle_result spyder_billie,won` or something to do the win or lose events, as kerizane suggested.  

However, that PR might take a few weeks (or more) to get merged, but once it's done it should be easier for you to add this. </s> 
(Actually, just use `is player_defeated` as suggested in the next comment, that's a better idea) :smile:

kerizane:
You can just use is player_defeated and not player_defeated.

Split it into 4 as I said above

ultidonki:
> You can just use is player_defeated and not player_defeated.

That's a good idea... I hadn't actually thought of that. Ignore my last comment, do it that way. :smile:

mods/tuxemon/maps/spyder_paper_town.tmx
```diff
@@ -423,15 +423,39 @@
...
     <property name="act70" value="translated_dialog spyder_papertown_firstfight"/>
     <property name="act80" value="start_battle spyder_billie"/>
-    <property name="act81" value="translated_dialog spyder_papertown_firstfight2"/>
+    <property name="act81" value="remove_npc spyder_billie"/>
     <property name="act90" value="set_variable firstfightdue:no"/>
     <property name="act91" value="set_variable teleport_faint:spyder_bedroom.tmx 7 6"/>
-    <property name="act94" value="pathfind spyder_billie,15,14"/>
-    <property name="act95" value="remove_npc spyder_billie"/>
-    <property name="act96" value="unlock_controls"/>
+    <property name="act92" value="set_variable firstfightend:yes"/>
```
AndyMender, line 428: This var will control proceeding to the "win" and "lose" checks for the first battle.

mods/tuxemon/maps/spyder_paper_town.tmx
```diff
@@ -423,15 +423,39 @@
...
     <property name="cond1" value="is variable_set firstfightdue:yes"/>
    </properties>
   </object>
+  <object id="425" name="First Fight - Win" type="event" x="384" y="144" width="16" height="16">
+   <properties>
+    <property name="act1" value="create_npc spyder_billie,26,12"/>
```
AndyMender: One of the issues with the event logic flow is that once the event ends, the state is cleaned up so for the "win" and "lose" check we need to recreate the NPC.

mods/tuxemon/maps/spyder_paper_town.tmx
```diff
@@ -423,15 +423,39 @@
...
+    <property name="act10" value="npc_face spyder_billie,down"/>
+    <property name="act20" value="translated_dialog spyder_papertown_firstfight_win"/>
+    <property name="act30" value="pathfind spyder_billie,15,14"/>
```
AndyMender: There is some tweaking to the location of the NPC I would like to do in a follow-up PR perhaps.

mods/tuxemon/maps/spyder_paper_town.tmx
```diff
@@ -423,15 +423,39 @@
...
+    <property name="act40" value="remove_npc spyder_billie"/>
+    <property name="act50" value="set_variable firstfightend:no"/>
```
AndyMender: We use the `firstfightend` variable as a ratchet toggle - once the event is triggered, it won't trigger again.

AndyMender:
I added information about the implementation's behavior to the first PR comment.

Big thanks to both for suggestions! In the end implementing Cleanup was not needed, but relying on `game_variable` conditions helped.

kerizane:
Why are you deleting and immediate recreating the npc?

AndyMender:
> Why are you deleting and immediate recreating the npc?

Now that I have a working version I can tweak the event property blocks, but from what I saw when one tries to `pathfind` or `remove_npc` an npc which was not declared in the event, an assert gets triggered, because `npc = None`.

Regarding deletion, do we have an automatic event cleanup which removes npc objects, etc.?

kerizane:
Something else is going on. You don't need to create the npc in the same event that you use it. If the Npc already existed then it should work.

AndyMender:
Alright, I will take a look at it again and clean up the logic. Big thanks for pointing that out! :)

kerizane:
No automatic deletion. Cleanup is generally just pathfinding the npc offscreen and then removing them.

I suggested a separate event for this to avoid duplication, but as it's just two lines the duplication is fine

AndyMender:
Tested the "lose" and "win" logic paths and it looks like it works as intended now. Ready for final review!

kerizane (approved):

kerizane:
Thanks!

PR #1275, opened 2022-09-01
fix capitalize first letter monster

JaskRendix:
Small fix for a small bug. When you fight against random monsters, then you'll notice that the name is always lowercase. This was because of add_monster. The function pulls the slug, but it's not capitalized the first letter.

### DIFF
--- a/tuxemon/event/actions/add_monster.py
+++ b/tuxemon/event/actions/add_monster.py
@@ -21,6 +21,7 @@
 
 from __future__ import annotations
 
+import string
 from typing import NamedTuple, Optional, Union, final
 
 from tuxemon import monster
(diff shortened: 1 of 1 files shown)


kerizane:
The name won't always be lower case, it's only if the name isn't defined in the locale file. I think these have just been updated and should be fixed. Are you running the latest code, which monsters does this happen with?

kerizane:
Your fix is reasonable, but I think we should allow people to have lowercase names if they wish to.

kerizane:
Are you sure your code works? Maybe you got lucky when you were testing it and found a monster that already had a name defined.

JaskRendix:
I think you're right. I was confused. I thought pairagrin was all lowercase, but I'm running Tuxemon without the fix and it looks correct. I close the PR.

kerizane:
No problem, thanks anyway.

Issue #1279, opened 2022-09-03
Two technique json files with warnings


JaskRendix:
**Describe the bug**
I was doing some testing for #1232 and I discovered the following warning in the code. Looking at the code there is a warning about _null_ (techniques: boulder, feint)

**To Reproduce**
Look at the code and open the boulder.json or feint.json

**Expected behavior**
No warnings

**Cause**
This string:
"$schema": "https://raw.githubusercontent.com/Tuxemon/Tuxemon/development/schemas/technique-schema.json"

**Solution**
If you remove the $schema, then the warning disappears. Can I remove it?

**Desktop (please complete the following information):**
 - OS: Ubuntu 22
 - Python 3.10.4

kerizane:
What is the actual warning?

"Look at the code" ?

JaskRendix:
Boulder
```
[{
	"resource": "/home/robespierre/Tuxemon/mods/tuxemon/db/technique/boulder.json",
	"owner": "_generated_diagnostic_collection_name_#1",
	"severity": 4,
	"message": "Incorrect type. Expected \"number\".",
	"startLineNumber": 13,
	"startColumn": 14,
	"endLineNumber": 13,
	"endColumn": 18
}]
[{
	"resource": "/home/robespierre/Tuxemon/mods/tuxemon/db/technique/boulder.json",
	"owner": "_generated_diagnostic_collection_name_#1",
	"severity": 4,
	"message": "Incorrect type. Expected \"string\".",
	"startLineNumber": 33,
	"startColumn": 18,
	"endLineNumber": 33,
	"endColumn": 22
}]
```
Feint:
```
[{
	"resource": "/home/robespierre/Tuxemon/mods/tuxemon/db/technique/feint.json",
	"owner": "_generated_diagnostic_collection_name_#1",
	"severity": 4,
	"message": "Incorrect type. Expected \"string\".",
	"startLineNumber": 33,
	"startColumn": 18,
	"endLineNumber": 33,
	"endColumn": 22
}]
```

kerizane:
The first warning is complaining that potency is null (line 13 should be a number). Is this a reasonable complaint? What are the other two validation errors?

kerizane:
How many JSON files have schemas defined?

Having JSON schemas and pydantic seems like overkill...

JaskRendix:
> How many JSON files have schemas defined?
[...]

PR #1280, opened 2022-09-03
fix technique warnings

JaskRendix:
Fix for #1279


kerizane:
Issue needs more info, but this doesn't look like the right fix.

kerizane:
Deleting them for now due to https://github.com/Tuxemon/Tuxemon/issues/1283#issuecomment-1236195840 

I will leave NPC alone as this one doesn't use Optional.

kerizane (approved):

Issue #1267, opened 2022-08-29
ERROR - Cannot find animation frames for:


AndyMender:
**Describe the bug**
When starting a fresh game (first scenario), early on there is a series of error logs about missing animation frames:
```
pygame 2.1.2 (SDL 2.0.16, Python 3.8.10)
Hello from the pygame community. https://www.pygame.org/contribute.html
pygame-menu 4.2.8
Enabling logging of all modules.
2022-08-29 23:06:10,905 - tuxemon.rumble - ERROR - No rumble backends available.
2022-08-29 23:08:01,054 - tuxemon.technique - ERROR - Cannot find animation frames for: slash_power
2022-08-29 23:08:01,055 - tuxemon.technique - ERROR - Cannot find animation frames for: darkfire
2022-08-29 23:08:01,055 - tuxemon.technique - ERROR - Cannot find animation frames for: buff_rage
2022-08-29 23:08:09,734 - tuxemon.technique - ERROR - Cannot find animation frames for: buff_rage
2022-08-29 23:08:09,736 - tuxemon.technique - ERROR - Cannot find animation frames for: shot
2022-08-29 23:08:09,737 - tuxemon.technique - ERROR - Cannot find animation frames for: water_spurt

```

**To Reproduce**
1. Do a fresh checkout of the Tuxemon repo
2. Install all Python libs into a venv
3. Start Tuxemon, play until you reach the dumpster area right of the TuxeMart, go through the dialogue with the professor. The errors should appear shortly after.

**Expected behavior**
No errors about missing animation frames.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04 (Xubuntu)
 - Python Version 3.8.10

**Additional context**
Similar to the already reported issue: https://github.com/Tuxemon/Tuxemon/issues/1185

Not sure if it has been mentioned already, but it might be useful to run the animation download pipeline on game start to make sure all local assets are synced. I haven't touched the Tuxemon code in a long while, but can take a look at it.

kerizane:
There is a PR in progress that updates the animations https://github.com/Tuxemon/Tuxemon/pull/1232

kerizane:
I think we should focus on fixing the current manual pipeline before worrying about making it automatic for users. This is a necessary first step, anyway.

There are animations that are on the wiki but are the wrong size.

https://github.com/Tuxemon/Tuxemon/issues/1185#issuecomment-1214347288

[...]

Issue #1185, opened 2022-05-14
ERROR - Cannot find animation frames for: shot, fire lion right, fire lion frong, snake right, icelance, ...


Murmele:
**Describe the bug**
During gameplay a few animations cannot be loaded

**To Reproduce**
1. Load save file: [slot2.zip](https://github.com/Tuxemon/Tuxemon/files/8692898/slot2.zip)
2. go to the right through through the trees
![image](https://user-images.githubusercontent.com/10099533/168425284-bd1e6857-c9be-4678-b42c-060a6e4314e2.png)

```
2022-05-14 13:49:13,974 - tuxemon.technique - ERROR - Cannot find animation frames for: shot
2022-05-14 13:49:14,004 - tuxemon.technique - ERROR - Cannot find animation frames for: fire lion right
2022-05-14 13:49:14,006 - tuxemon.technique - ERROR - Cannot find animation frames for: fire lion front
2022-05-14 13:49:14,028 - tuxemon.technique - ERROR - Cannot find animation frames for: snake right 64px
2022-05-14 13:49:14,045 - tuxemon.technique - ERROR - Cannot find animation frames for: icelance
2022-05-14 13:49:14,059 - tuxemon.technique - ERROR - Cannot find animation frames for: icelance
2022-05-14 13:49:14,078 - tuxemon.technique - ERROR - Cannot find animation frames for: fire lion right
2022-05-14 13:49:14,079 - tuxemon.technique - ERROR - Cannot find animation frames for: fire lion front
```

**Expected behavior**
Should find the animations

**Desktop (please complete the following information):**
 - OS: Manjaro Linux
 - Python Version 3.9

**Smartphone (please complete the following information):**
 - Device: Computer
 - OS: Linux

**Additional context**
Commit id: 4c5549fa3fcf48ed527631213f96d941c9376d8c

bitcraft:
Missing animation frames might be able to be pulled from the wiki.  Some techniques have an animation specified, but they have not been created yet.  Sometimes the animation names don't match what is specified in the json.  For example, "fire lion front", probably refers to the animation "fire lion fron 64px".   The solution here would probably be best solved by renaming the file names, because "64px" in the name in non-conforming.  The other missing animations may be similar.
https://github.com/Tuxemon/Tuxemon/blob/development/mods/tuxemon/animations/technique/fire_lion_front_64px00.png

kerizane:
[...]

PR #1285, opened 2022-09-03
[DRAFT] Support arbitrary dimensions in technique animations

AndyMender:
Split off effort from: https://github.com/Tuxemon/Tuxemon/issues/1267

The goal is to be able to download animations regardless of their size (as long as `height == width`).

What's to be discussed is how the animations should be displayed in-game (scaled to 64x64 to keep the rest of the pipeline as is? centered? zoomed in? other?)

### DIFF
--- a/scripts/download_technique_animations.py
+++ b/scripts/download_technique_animations.py
@@ -44,6 +44,7 @@ def process_filename(filepath: str) -> str:
     """Extract base filename from an animation file path."""
     cleaned_filename = os.path.splitext(os.path.basename(filepath))[0]
     cleaned_filename = cleaned_filename.strip("0123456789_")
+    cleaned_filename = cleaned_filename.split("px_")[-1]
     cleaned_filename = cleaned_filename.lower()
     return cleaned_filename
 
--- a/scripts/download_technique_animations.py
+++ b/scripts/download_technique_animations.py
@@ -44,7 +44,7 @@ def process_filename(filepath: str) -> str:
     """Extract base filename from an animation file path."""
     cleaned_filename = os.path.splitext(os.path.basename(filepath))[0]
     cleaned_filename = cleaned_filename.strip("0123456789_")
-    cleaned_filename = cleaned_filename.split("px_")[-1]
+    cleaned_filename = cleaned_filename.split("px-")[-1]
     cleaned_filename = cleaned_filename.lower()
     return cleaned_filename
 
--- a/scripts/download_technique_animations.py
+++ b/scripts/download_technique_animations.py
@@ -17,7 +17,6 @@ import requests
 from lxml import html
 from PIL import Image
 
-FRAME_SIZE = 64
 WIKI_URL = "https://wiki.tuxemon.org"
 
 TUXEMON_ROOT_DIR = pathlib.Path(__file__).resolve().parent.parent
@@ -94,9 +93,6 @@ def gif_to_frames(filepath: str) -> None:
     with Image.open(filepath) as image:
         if not image.is_animated:
             print(f"{filepath} is not animated, skipped")
-            return
-        if not image.width == image.height == FRAME_SIZE:
-            print(f"{filepath} is not {FRAME_SIZE}x{FRAME_SIZE}, skipped")
             return
 
         base_name = process_filename(filepath)
--- a/tuxemon/graphics.py
+++ b/tuxemon/graphics.py
@@ -229,6 +229,17 @@ def scale_surface(
     )
 
 
+def center_image(surface: pygame.surface.Surface) -> None:
+    # Use a known 64x64 animation frame as reference   
+    ANIM_CENTER_REF = load_and_scale("animations/technique/capture00.png")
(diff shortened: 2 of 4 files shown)


kerizane:
Sanglorian's request sounds good:

https://github.com/Tuxemon/Tuxemon/issues/1185#issuecomment-1214918015

So if an animation is 74 wide, the animation should be centered to have 5 pixels either side of a "normal" 64 wide animation.

There should be no constraints on the size or ratio. The animations should be displayed as is, without any fixes applied.

We may need to have a pydantic check to figure out if an image will be too big for the target (an animation in the middle of the screen can obviously be larger than one in the corner), but not sure how complicated this would be.

AndyMender:
Sounds good! I think the centering logic should be easy to implement.

Do we have any candidate anims I could try this one or should I generate a new one for tests?

kerizane:
Anything from https://wiki.tuxemon.org/Category:Technique_Animation e.g.

https://wiki.tuxemon.org/File:Explosion_dusty_96.gif

AndyMender:
> Anything from https://wiki.tuxemon.org/Category:Technique_Animation e.g.
> 
> https://wiki.tuxemon.org/File:Explosion_dusty_96.gif

I actually realized that the animation download script highlights which GIFs don't fit the format so plenty to work with :).

scripts/download_technique_animations.py
```diff
@@ -45,6 +44,7 @@ def process_filename(filepath: str) -> str:
     """Extract base filename from an animation file path."""
     cleaned_filename = os.path.splitext(os.path.basename(filepath))[0]
     cleaned_filename = cleaned_filename.strip("0123456789_")
+    cleaned_filename = cleaned_filename.split("px-")[-1]
```
kerizane, line 47: Doesn't this require px to be present? You can use partition instead.

AndyMender: Nope, it would work without as well:
```python
>>> a_str = "firelion_back.git"
>>> a_str.split("px-")[-1]
'firelion_back.git'
```
I will give partition a try, though :).

kerizane: Weird, wonder why I thought that. If it works then there's no point.

tuxemon/graphics.py
```diff
@@ -229,6 +229,17 @@ def scale_surface(
...
 
+def center_image(surface: pygame.surface.Surface) -> None:
+    # Use a known 64x64 animation frame as reference   
+    ANIM_CENTER_REF = load_and_scale("animations/technique/capture00.png")
+    ref_rect = ANIM_CENTER_REF.get_rect()
+
+    rect = surface.get_rect()
+    print("Anim dimensions (width, height):", surface.get_size())
+    print("Anim center (x,y):", rect.center)
+    surface.scroll(rect.width - ref_rect.width, rect.height - ref_rect.height)
```
AndyMender: This is a temp solution for testing. Ideally, I would like the `Surface` object for an animation frame to be loaded already centered or centered atter scaling.

Also, the reference 64x64 image should be loaded only once.

AndyMender:
@kerizane I did a deeper dive into the code base and it seems to me like the technique sprites are already centered correctly on top of the enemy tuxemon here: https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/states/combat/combat.py#L1002
```python
            tech_sprite.rect.center = target_sprite.rect.center
```

kerizane:
So do larger animations already work?

AndyMender:
> So do larger animations already work?

To my understanding yes, but feel free to test my branch. I gave Agnite 2 attacks that have animations different than the standard 64x64 - Disintegrate and Breath Fire. The easiest is to start the Spyder scenario and pick Agnite as the starter.

kerizane:
It should be pretty obvious if they work or not! I will trust you.

Do the animations look ok to you?

I will try to take a look, but it would be easier if you could record a couple of the animations and post them here.

AndyMender:
> It should be pretty obvious if they work or not! I will trust you.
> 
> Do the animations look ok to you?
> 
> I will try to take a look, but it would be easier if you could record a couple of the animations and post them here.

They look okay to me, but I agree it makes more sense to get some recordings or at least screenshots for verification. I will update the PR once I have them.

Then, I can set up a follow-up PR to add the missing animations + credits for the ones that don't fit the 64x64 format.

AndyMender:
Here's how the animations look like in-game now
- Breathe Fire - [Wiki entry](https://wiki.tuxemon.org/File:Breath_fire.gif)
breathe_fire
- Disintegrate - [Wiki entry](https://wiki.tuxemon.org/File:Disintegrate_83.gif)
disintegrate

I think maybe the bigger issue is that the grass field for the enemy Tuxemon is partially occluded by the status bar.

kerizane:
I think these look fine, definitely better than nothing. Tidy this up and raise a proper PR please.

AndyMender:
> I think these look fine, definitely better than nothing. Tidy this up and raise a proper PR please.

Clean PR set up 👍🏼

PR #1288, opened 2022-09-03
Fix animations regex

kerizane:
Animations have been broken since 737d8c474, which refactored the animations loading to use a function that expects a different animation format. I have edited the regex to work with both formats.

### DIFF
--- a/tuxemon/graphics.py
+++ b/tuxemon/graphics.py
@@ -271,7 +271,7 @@ def animation_frame_files(
 
     """
     frames = list()
-    pattern = re.compile(rf"{name}\.[0-9]+\.png")
+    pattern = re.compile(rf"{name}\.?[0-9]+\.png")
     # might be slow on large folders
     for filename in os.listdir(directory):
         if pattern.match(filename):


tuxemon/graphics.py
```diff
@@ -271,7 +271,7 @@ def animation_frame_files(
 
     """
     frames = list()
-    pattern = re.compile(rf"{name}\.[0-9]+\.png")
+    pattern = re.compile(rf"{name}\.?[0-9]+\.png")
```
AndyMender, line 274: If I understand correctly, we allow animations to not have the `xx` frame index?

The animation download script will always generate the frame index, so if only 1 frame is present, it will be `animation_name00.png`.

We could also force the regex to take only 2 digs in the frame index, unless it's okay for an animation to have more than 100 frames.

kerizane: The previous regex supported

foo.000.png

It now supports that and:

foo00.png

I just made the first dot optional.

AndyMender: Sounds good!

kerizane:
Not required any more, as I got AndyMender to recreate in his pr.

PR #1289, opened 2022-09-04
fix JSON data tech_id

JaskRendix:
Here the PR for inserting the tech_id field inside technique.json.

Everything is based as starting point on this Excel file: sanglorian.github.io/downloads/Tuxemon/techniques.xlsx

Here at least I had the first 10 numbers, while on Tuxepedia only the number 10.

From this list are missing (7):
overfeed, run, status_faint, status_lifeleech, status_poison, status_recover, swap.

Since I had no idea about the other tech_id, I tried to come up with a valid, but not banal (sort in alphabetical order), logic. If it's not satisfyng, then I can easily revert all the numbers >10 to 0 waiting for the official numeration. If nothing is acceptable, then I'll close the PR.

Logic: if you look at the sheet movesets, then you can see all the monsters up to the last (rockitten, rockat, etc - following the txmn_id). I came up with this method: boulder 11, thunderball 12, glower 13, stampede 14, etc.

A progressive method that follows the Tuxemon order as well as the techniques associated for each monster (look at the screenshot).

It took me some time, but I hope that it shows the respect for the creator.

Attached also a list with all the monsters sorted in both ways (alphabetical and tech_id).

Screenshot from 2022-09-04 14-48-35
[tech_id.txt](https://github.com/Tuxemon/Tuxemon/files/9484783/tech_id.txt)

### DIFF
--- a/mods/tuxemon/db/technique/all_in.json
+++ b/mods/tuxemon/db/technique/all_in.json
@@ -1,4 +1,5 @@
 {
+  "tech_id": 54,
   "accuracy": 0.75,
   "animation": "sparks_gold_alt",
   "category": "physical",
--- a/mods/tuxemon/db/technique/amnesia.json
+++ b/mods/tuxemon/db/technique/amnesia.json
@@ -1,4 +1,5 @@
 {
+  "tech_id": 53,
   "accuracy": 0.75,
   "animation": "sparks_gold_alt",
   "category": "physical",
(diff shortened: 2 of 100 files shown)


kerizane:
@Sanglorian what do you think?

Sanglorian:
Looks great to me! Thanks for this. I've added the ID numbers to the techniques on the wiki now too.

kerizane (approved):

kerizane:
Good work, thanks.

PR #1296, opened 2022-09-06
Validate monster shape

### LINKED ISSUE #1284: Better pydantic validation for monster shape
Pydantic validation should have caught this issue:

https://github.com/Tuxemon/Tuxemon/issues/1278

kerizane:
Fixes #1284

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -25,7 +25,7 @@
         }
     ],
     "evolutions": [],
-    "shape": "Varmint",
+    "shape": "varmint",
     "types": [
         "Earth"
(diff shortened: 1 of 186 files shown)


xirsoi (dismissed):

xirsoi:
@kerizane if you drop your changes to the monster definitions it should resolve the conflict. Since you're tolowering the shape name anyway, it shouldn't matter that they're title cased in the monster definitions.

kerizane:
No, this change is required. The .lower() is within a script which creates the json. I could have done the "lower" within the code, but it makes the pydantic code simpler if  we do it in the json.

xirsoi (approved):

PR #1299, opened 2022-09-07
add evolution JSON monsters

JaskRendix:
Everything is based on Tuxepedia.

I was looking again at the monster JSON and I realized there was something missing: evolutions. Inside the Monster Model there are three fields: path, level_at and monster_slug and I realized that evolution was impossible because we miss the data. So after another deep dive in Tuxepedia, here the PR.

Among our monsters JSON there are 76 evolutions in 59 files.

What I've added (eg. memnomnom.json):
```
    "evolutions": [
        {
            "path": "",
            "at_level": 20,
            "monster_slug": "miaownolith"
        },
        {
            "path": "",
            "at_level": 20,
            "monster_slug": "pyraminx"
        }
    ],
```
I left the path empty, because it wasn't clear what to put there. I can always update everything with the right value. I think the important part is having the right structure.

If there is no evolution: **"evolutions": [],**

@Sanglorian if there is something amiss, I can fix without problems. The value I put inside are only describing the situation, and are temporary as always. I wanted to tell you that now we have the game_variable **daytime** (TRUE and FALSE) as well as **stage_of_day** (dawn, morning, afternoon, dusk, night), so we can make evolve some monster during the night, the day, the dawn, etc.

Some suggestions about the structure:
add another fields for the stage_of_day or daytime;
using path merely for objects;
eventually another field for tech;

I can implement these without problems.

Attached evolutions.pdf that resume the evolution situation, so everyone can understand.
[evolutions.pdf](https://github.com/Tuxemon/Tuxemon/files/9506815/evolutions.pdf)

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -24,6 +24,7 @@
             "technique": "feint"
         }
     ],
+    "evolutions": [],
     "shape": "Varmint",
     "types": [
         "Earth"
--- a/mods/tuxemon/db/monster/aardorn.json
+++ b/mods/tuxemon/db/monster/aardorn.json
@@ -24,6 +24,13 @@
             "technique": "feint"
         }
     ],
+    "evolutions": [
+        {
+            "path": "",
+            "at_level": 20,
+            "monster_slug": "aardart"
+        }
+    ],
     "shape": "Varmint",
     "types": [
         "Earth"
--- a/mods/tuxemon/db/monster/abesnaki.json
+++ b/mods/tuxemon/db/monster/abesnaki.json
@@ -16,6 +16,7 @@
             "technique": "sleep_bomb"
         }
     ],
+    "evolutions": [],
     "shape": "Serpent",
     "types": [
         "Fire"
--- a/mods/tuxemon/db/monster/agnidon.json
+++ b/mods/tuxemon/db/monster/agnidon.json
@@ -16,6 +16,13 @@
             "technique": "energy_claws"
         }
     ],
+    "evolutions": [
+        {
+            "path": "",
+            "at_level": 36,
+            "monster_slug": "agnigon"
+        }
+    ],
     "shape": "Dragon",
     "types": [
         "Fire"
--- a/mods/tuxemon/db/monster/agnigon.json
+++ b/mods/tuxemon/db/monster/agnigon.json
@@ -16,6 +16,7 @@
             "technique": "take_cover"
         }
     ],
+    "evolutions": [],
     "shape": "Dragon",
     "types": [
         "Fire"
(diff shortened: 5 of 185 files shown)


Sanglorian:
Thanks for picking up on the missing evolution entries, @JaskRendix !

The monsters with special requirements like an item aren't supposed to have level requirements too - they should just level up when the item's used on them. 

But you might be adding the levels just as a stop-gap until we implement item-vased evolution? In which case I have no objections to this update.

JaskRendix:
Ok, it makes sense, I need to update the following monsters:
```
botbot -> av8r (Booster Tech)
botbot -> picc (Booster Tech)
botbot -> mrmoswitch (Booster Tech)
botbot -> k9 (Booster Tech)
budaye -> bamboon (Wood Booster or Lucky Bamboo) -> note: two evolution {}
budaye -> frondly (Water Booster or Peace Lily) -> note: two evolution {}
dollfin -> bigfin (Water Booster or Sweet Sand) -> note: two evolution {}
dollfin -> sharpfin (Metal Booster or Sea Girdle) -> note: two evolution {}
grintot -> grinflare (Earth Booster or Flintstone) -> note: two evolution {}
grintot -> grintrock (Metal Booster or Thunderstone) -> note: two evolution {}
ignibus -> eruptibus (Earth Booster or Tectonic Drill) -> note: two evolution {}
ignibus -> embazook (Metal Booster or Stovepipe) -> note: two evolution {}
memnomnom -> miaownolith (Earth Booster or Miaow Milk) -> note: two evolution {}
memnomnom -> pyraminx (Metal Booster or Pyramidion) -> note: two evolution {}
pythwire -> ouroboutlet (Fire Booster)
pythwire -> sockeserp (Metal Booster)

katacoon -> bugnin (if Melee is higher than Armour)
katacoon -> sumchon (if Armour is higher than Melee)
-> maybe putting "stats", so we can associate it with an if
```
I think double {} for the same monster with a different item is the best way, because if not we are forced to add another field.

**UPDATE:**
data structure is updated, katacoon.json  is lv 10, because it evolves (previous form) at lv 9. That's the reason.

Sanglorian:
That's great, thank you! Could you change the two Katacoon evolutions to happen at level 12? (Plus the armor_is_higher_than_melee big). Otherwise looks perfect to me.

JaskRendix:
Sure, I'll replace the **armor_is_higher_than_melee** with a simple stat_armor and stat_melee in the other, so we can remember the evolution is based on it.

kerizane:
Great job! So does evolution work now?
Have you tested that monsters with a special evolution method don't evolve from leveling up?

JaskRendix:
Evolutions don't work, this is only the data structure. Evolve_monster (action and condition) needs to be rewritten from what I understood as well as get_evolution (monster.py), this because the object - special method - wasn't contemplated. We are like half way, so there is no risk something evolves.

kerizane (approved):

PR #1306, opened 2022-09-10
Support arbitrary dimensions in technique animations - clean

AndyMender:
Cleaned up version of: https://github.com/Tuxemon/Tuxemon/pull/1285

Split off effort from: https://github.com/Tuxemon/Tuxemon/issues/1267

The goal is to be able to download animations regardless of their size.

Here's how the animations look like in-game now (using a modified version of Agnite for testing)
- Breathe Fire - [Wiki entry](https://wiki.tuxemon.org/File:Breath_fire.gif)
breathe_fire
- Disintegrate - [Wiki entry](https://wiki.tuxemon.org/File:Disintegrate_83.gif)
disintegrate

### DIFF
--- a/scripts/download_technique_animations.py
+++ b/scripts/download_technique_animations.py
@@ -17,7 +17,6 @@ import requests
 from lxml import html
 from PIL import Image
 
-FRAME_SIZE = 64
 WIKI_URL = "https://wiki.tuxemon.org"
 
 TUXEMON_ROOT_DIR = pathlib.Path(__file__).resolve().parent.parent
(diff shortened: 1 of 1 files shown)


kerizane (approved):

kerizane:
> The goal is to be able to download animations regardless of their size (as long as height == width).

I can't see this constraint, but I think that's fine, we want t9 support any size.

AndyMender:
> > The goal is to be able to download animations regardless of their size (as long as height == width).
> 
> I can't see this constraint, but I think that's fine, we want to support any size.

Ah, thanks for the call-out! That was just a copy-paste from the original PR. I removed the constraint, because as you said we don't really want it.

kerizane (approved):

PR #1316, opened 2022-09-13
evolution and evolve item condition

JaskRendix:
~~I was creating some new item conditions, mostly because of evolutions. Everything works, but I noticed a _bug_ and I have no idea how to deal with it. I created the condition has_evolution, so you can use a specific object (boosters, etc) on a monster who evolves in another monster. It works, but it doesn't allow "accumulation". I'll explain.~~ Everything started with a "bug" now solved.

This PR includes:
1. evolution through objects (like boosters);
2. related to the 1, I added the following object conditions: _has_path_ (to get the object) and _level_ (to get the level);
3. related to the 1, I added the _evolve_ object action, it allows to address the evolution;
4. to fix the problem of consumable object, I added "booster", so this type of object if it's used on the wrong thing doesn't disappear (object-schema); I wanted to add an exception to "consumable", but 12 exceptions were a bit too much;
5. evolution through combat (traditional evolution after leveling up, like rockitten that at level 24 evolves in rockat);

What's missing?
1. evolution through stats (**katacoon**'s case);
2. evolution animation;
3. eventually we can delete the evolve_monsters (action and condition) as well as get_evolution

Inside the earth_booster.json object, something like this:
```
  "effects": [
    "evolve eruptibus"
  ],
  "conditions": [
    "level target,>=,0",
    "has_path target,earth_booster"
  ],
```

### DIFF
--- a/schemas/item-schema.json
+++ b/schemas/item-schema.json
@@ -112,7 +112,6 @@
       "title": "ItemType",
       "description": "An enumeration.",
       "enum": [
-        "Booster",
         "Consumable",
         "KeyItem"
       ],
--- a/schemas/item-schema.json
+++ b/schemas/item-schema.json
@@ -112,6 +112,7 @@
       "title": "ItemType",
       "description": "An enumeration.",
       "enum": [
+        "Booster",
         "Consumable",
         "KeyItem"
       ],
--- a/tuxemon/item/effects/evolve.py
+++ b/tuxemon/item/effects/evolve.py
@@ -1,0 +1,54 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Contributor(s):
+# Adam Chevalier <chevalieradam2@gmail.com>
+#
+
+from __future__ import annotations
+
+from typing import NamedTuple
+
+from tuxemon.item.itemeffect import ItemEffect, ItemEffectResult
+from tuxemon.monster import Monster
+
+
+class EvolveEffectResult(ItemEffectResult):
+    pass
+
+
+class EvolveEffectParameters(NamedTuple):
+    monster_evolve: str
+
+
+class EvolveEffect(ItemEffect[EvolveEffectParameters]):
+    """This effect evolves the target into the monster in the parameters."""
+
+    name = "evolve"
+    param_class = EvolveEffectParameters
+
+    def apply(self, target: Monster) -> EvolveEffectResult:
+        monster_evolve = self.parameters.monster_evolve
+
+        if any(d["monster_slug"] == monster_evolve for d in target.evolutions):
+            self.user.evolve_monster(target,monster_evolve)
+            return {"success": True}
+        else:
+            return {"success": False}
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -775,7 +775,7 @@ class NPC(Entity[NPCState]):
 
                 # TODO: implement an evolution animation
 
-              # Set new monster data
+                # Set new monster data
                 new_monster = Monster()
                 new_monster.load_from_db(evolution)
                 new_monster.set_level(old_level)
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -754,6 +754,38 @@ class NPC(Entity[NPCState]):
             self.monsters.remove(monster)
             self.set_party_status()
 
+    def evolve_monster(self, monster: Monster, evolution: str) -> None:
+        """
+        Evolve a monster from this npc's party.
+
+        Parameters:
+            monster: Monster to remove from the npc's party.
+            evolution: Monster to add to the npc's party.
+
+        """
+        if monster in self.monsters:
+            # Store old monster data
+            old_level = monster.level
+            old_current_hp = monster.current_hp
+            old_flairs = monster.flairs
+            self.monsters.remove(monster)
+
+            # TODO: implement an evolution animation
+
+            # Set new monster data
+            new_monster = Monster()
+            new_monster.load_from_db(evolution)
+            new_monster.set_level(old_level)
+            new_monster.current_hp = min(old_current_hp, new_monster.hp)
+            new_monster.slug = new_monster.slug
+            self.monsters.append(new_monster)
+
+            # If evolution has a flair matching, copy it
+            for new_flair in new_monster.flairs.values():
+                for old_flair in old_flairs.values():
+                    if new_flair.category == old_flair.category:
+                        new_monster.flairs[new_flair.category] = old_flair
+
     def remove_monster_from_storage(self, monster: Monster) -> None:
         """
         Removes the monster from the npc's storage.
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -763,28 +763,41 @@ class NPC(Entity[NPCState]):
             evolution: Monster to add to the npc's party.
 
         """
-        if monster in self.monsters:
-            # Store old monster data
-            old_level = monster.level
-            old_current_hp = monster.current_hp
-            old_flairs = monster.flairs
-            self.monsters.remove(monster)
+        if monster not in self.monsters:
+            return
+        else:
+            for slot, monster in enumerate(self.monsters):
+                # Store old monster data
+                old_level = monster.level
+                old_current_hp = monster.current_hp
+                old_flairs = monster.flairs
+                self.monsters.remove(monster)
 
-            # TODO: implement an evolution animation
+                # TODO: implement an evolution animation
 
-            # Set new monster data
-            new_monster = Monster()
-            new_monster.load_from_db(evolution)
-            new_monster.set_level(old_level)
-            new_monster.current_hp = min(old_current_hp, new_monster.hp)
-            new_monster.slug = new_monster.slug
-            self.monsters.append(new_monster)
+              # Set new monster data
+                new_monster = Monster()
+                new_monster.load_from_db(evolution)
+                new_monster.set_level(old_level)
+                new_monster.current_hp = min(old_current_hp, new_monster.hp)
+                new_monster.slug = new_monster.slug
+                self.monsters.append(new_monster)
 
-            # If evolution has a flair matching, copy it
-            for new_flair in new_monster.flairs.values():
-                for old_flair in old_flairs.values():
-                    if new_flair.category == old_flair.category:
-                        new_monster.flairs[new_flair.category] = old_flair
+                # If evolution has a flair matching, copy it
+                for new_flair in new_monster.flairs.values():
+                    for old_flair in old_flairs.values():
+                        if new_flair.category == old_flair.category:
+                            new_monster.flairs[new_flair.category] = old_flair
+
+                # Removing the old monster caused all monsters in front to move
+                # a slot back
+                # Bring our new monster from the end of the list to its
+                # previous position
+                for i in range(len(self.monsters) - 1, slot, -1):
+                    self.switch_monsters(i, i - 1)
+
+                # Only do one evolution per call
+                return
 
     def remove_monster_from_storage(self, monster: Monster) -> None:
         """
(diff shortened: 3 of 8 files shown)


kerizane:
Conditions is an "and" not an "or"

kerizane:
Maybe make has_evolution accept multiple targets.

Actually, can you add a condition that checks the data within the evolution path? I would rather not have to duplicate the evolution data.

JaskRendix:
That's right! It was simpler!

JaskRendix:
Another step has been done.

Now we can evolve monster using an item, I tested it.

the right monster -> it evolves -> item disappears
the wrong -> it fails -> item disappears

Do you have any advice on how to restore the item? (evolve.py) 
Maybe this is wrong **return {"success": False}**?

JaskRendix:
Another step, but now I'm stuck in the monster.py, here:
**self.owner.evolve_monster(self.name,evolution["monster_slug"])**
It gives me **AttributeError: 'NoneType' object has no attribute 'evolve_monster'**
I guess that it's because of this **self.owner: Optional[NPC] = None**

I don't know if **level_up** is the right place or maybe **end_combat** is better.

As soon as I come up with the solution, then I'll rename the PR and update the first post.

tuxemon/item/conditions/height.py
```diff
@@ -0,0 +1,69 @@
...
+
+    name = "height"
+    param_class = HeightConditionParameters
+
+    def test(self, target: Monster) -> bool:
+        lhs = target.height
+        op = cmp_dict[self.parameters.comparison]
+        rhs = self.parameters.value
+
+        return op(lhs, rhs)
```
kerizane: Lhs and RHS are a bit cryptic. Do we even need the variable?

kerizane: A few others with same issue.

kerizane:
Evolving after combat is less likely to cause bugs, imo.

kerizane:
FYI, there is a config item for making items be consumed even if they fail. You should make sure your item can't be used if it will fail.

tuxemon/item/conditions/height.py
JaskRendix: Yeah, I'm going to get rid of weight, height and monster, I'll keep only level and has_path

JaskRendix:
> Evolving after combat is less likely to cause bugs, imo.

You're right about it, especially if there will be the animation in the future. I'll move to end_combat (monster.py). Always that NPC thing.

JaskRendix:
I have added booster because we are talking about 12 items: water_booster, sweet_sand, metal_booster, sea_girdle, earth_booster, thunderstone, flintstone, tectonic_drill, stovepipe, miaow_milk, pyramidion and fire_booster.

Do you have any advice about the NPC in monster.py?

kerizane:
```
    def add_monster(self, monster: Monster) -> None:
        """
        Adds a monster to the npc's list of monsters.

        If the player's party is full, it will send the monster to
        PCState archive.

        Parameters:
            monster: The monster to add to the npc's party.

        """
        monster.owner = self
        if len(self.monsters) >= self.party_limit:
            self.monster_boxes[CONFIG.default_monster_storage_box].append(
                monster
            )
        else:
            self.monsters.append(monster)
            self.set_party_status()

```

```
    def set_state(self, session: Session, save_data: NPCState) -> None:
        """
        Recreates npc from saved data.

        Parameters:
            session: Game session.
            save_data: Data used to recreate the NPC.

        """
        self.facing = save_data.get("facing", "down")
        self.game_variables = save_data["game_variables"]
        self.battle_history = save_data["battle_history"]
        self.inventory = decode_inventory(
            session, self, save_data.get("inventory", {})
        )
        self.monsters = decode_monsters(save_data.get("monsters"))
        self.name = save_data["player_name"]
        for monsterkey, monstervalue in save_data["monster_boxes"].items():
            self.monster_boxes[monsterkey] = decode_monsters(monstervalue)
        for itemkey, itemvalue in save_data["item_boxes"].items():
            self.item_boxes[itemkey] = decode_inventory(
                session, self, itemvalue
            )
```

The owner was never set because monsters that are loaded from the save file don't go through the "add_monster" method.

JaskRendix:
Thanks for the information.

JaskRendix:
NPC problem solved, now I need to figure out why doesn't evolve the monster.

JaskRendix:
It works, now monsters evolve after the combat. I updated the first post and squashed everything into one commit.

schemas/item-schema.json
```diff
@@ -112,6 +112,7 @@
       "title": "ItemType",
       "description": "An enumeration.",
       "enum": [
+        "Booster",
```
kerizane: We don't really need this new item type. Can you make it work without it?

tuxemon/item/effects/evolve.py
```diff
@@ -0,0 +1,54 @@
...
+    """This effect evolves the target into the monster in the parameters."""
+
+    name = "evolve"
+    param_class = EvolveEffectParameters
+
+    def apply(self, target: Monster) -> EvolveEffectResult:
+        monster_evolve = self.parameters.monster_evolve
+
+        if any(d["monster_slug"] == monster_evolve for d in target.evolutions):
+            self.user.evolve_monster(target,monster_evolve)
```
kerizane, line 51: run black and isort on all files you modified

tuxemon/npc.py
```diff
@@ -754,6 +754,38 @@ def remove_monster(self, monster: Monster) -> None:
...
+            self.monsters.remove(monster)
+
+            # TODO: implement an evolution animation
+
+            # Set new monster data
+            new_monster = Monster()
+            new_monster.load_from_db(evolution)
+            new_monster.set_level(old_level)
+            new_monster.current_hp = min(old_current_hp, new_monster.hp)
+            new_monster.slug = new_monster.slug
```
kerizane: ?

tuxemon/npc.py
```diff
@@ -754,6 +754,38 @@ def remove_monster(self, monster: Monster) -> None:
...
+            self.monsters.append(new_monster)
```
kerizane: The order has changed :(

tuxemon/states/combat/combat.py
```diff
@@ -473,6 +473,9 @@ def transition_phase(self, phase: CombatPhase) -> None:
             self.players[0].set_party_status()
             self.end_combat()
 
+            for monster, player in product(self.active_monsters, self.active_players):
```
kerizane: Using product here is weird given that the next method has

`                    if player is self.players[0]:`

kerizane (changes requested):

tuxemon/states/combat/combat.py
```diff
@@ -1149,6 +1152,21 @@ def trigger_player_run(self, player: NPC) -> None:
...
+            if not evolution["path"]:
+                # check the level field.
+                if evolution["at_level"] <= monster.level:
+                    logger.info(
+                        "{} evolved into {}!".format(
+                            monster.name, evolution["monster_slug"]
+                        )
+                    )
+                    monster_evolve = str(evolution["monster_slug"])
+                    if player is self.players[0]:
```
kerizane: move this constraint into the calling code

kerizane (changes requested):

tuxemon/states/combat/combat.py
JaskRendix: I removed  **if player is self.players[0]:**, it wasn't necessary to the evolution.

tuxemon/states/combat/combat.py
JaskRendix: I rewrote that part.

tuxemon/npc.py
JaskRendix: Because it hasn't maintained the same slot position?

schemas/item-schema.json
JaskRendix: If you say so, I'll find another way.

tuxemon/npc.py
kerizane: Yes

tuxemon/item/effects/evolve.py
kerizane: There's a formatting script in the scripts directory, there will be a few changes as people haven't been running it.

Run this against the development branch and then raise a PR.

When that is merged in you can rebase this branch and run the script again against this branch.

tuxemon/states/combat/combat.py
```diff
@@ -1149,6 +1152,19 @@ def trigger_player_run(self, player: NPC) -> None:
         del self.monsters_in_play[player]
         self.players.remove(player)
 
+    def evolve(self, player: NPC, monster: Monster) -> None:
+        for evolution in monster.evolutions:
+            # check the path field.
```
kerizane: Don't add comments that are this obvious.

You could explain why we are skipping entries that have path set.

tuxemon/npc.py
```diff
@@ -754,6 +754,38 @@ def remove_monster(self, monster: Monster) -> None:
...
+    def evolve_monster(self, monster: Monster, evolution: str) -> None:
+        """
+        Evolve a monster from this npc's party.
+
+        Parameters:
+            monster: Monster to remove from the npc's party.
+            evolution: Monster to add to the npc's party.
+
+        """
+        if monster in self.monsters:
```
kerizane: I'd prefer a guard statement / early return

```
if monster not in self.monsters:
    return
```

tuxemon/states/combat/combat.py
kerizane: Can we use zip instead of product?

tuxemon/npc.py
JaskRendix: There is the slot now, but I need to figured out the right parameters.

tuxemon/npc.py
JaskRendix: Updated, slot included.

tuxemon/states/combat/combat.py
JaskRendix: Done

tuxemon/npc.py
JaskRendix: Because now evolution after battle is fine, now I need to fix the item based.

tuxemon/item/effects/evolve.py
JaskRendix: No

tuxemon/npc.py
```diff
@@ -754,6 +754,51 @@ def remove_monster(self, monster: Monster) -> None:
...
+
+        Parameters:
+            monster: Monster to remove from the npc's party.
+            evolution: Monster to add to the npc's party.
+
+        """
+        if monster not in self.monsters:
+            return
+        else:
+            for slot, monster in enumerate(self.monsters):
```
kerizane, line 769: This doesn't need to be in an else statement, as you have the early return.

kerizane, line 769: This for loop is too complicated. You can use self.monsters.index to get the index and then put the new monster where the original monster was.

tuxemon/states/combat/combat.py
```diff
@@ -473,6 +473,9 @@ def transition_phase(self, phase: CombatPhase) -> None:
             self.players[0].set_party_status()
             self.end_combat()
 
+            for monster, player in zip(self.active_monsters, self.active_players):
```
kerizane, line 476: Actually, I think we should just do this for the player, because modders might not want this for NPCs, and we don't save npc changes, so the evolution will disappear when you reload the game.

tuxemon/item/effects/evolve.py
kerizane: Ok, just run black and isort against the files you have added / modified. There's only 5 of them.

kerizane:
If it's too much work you can leave booster in, not a big deal.

tuxemon/npc.py
JaskRendix: Ok

tuxemon/states/combat/combat.py
JaskRendix: Ok

JaskRendix: If I remove monster (after for, leaving only player), it starts asking me:
```
    self.evolve(player, monster)
UnboundLocalError: local variable 'monster' referenced before assignment
```

JaskRendix: I'm going to try a thing, let's see if works

kerizane: Ideally you would track each monster in the players party who gained a level in the battle and try to evolve them.

If this is too difficult you could just try to evolve every monster in the party.

JaskRendix:
```
            for player in zip(self.active_monsters, self.active_players):
                self.evolve(player)
```
```
    def evolve(self, player: NPC) -> None:
        for monster in player.monsters: --> I added this because I needed something for monster
```
and the output 
```
    for monster in player.monsters:
AttributeError: 'tuple' object has no attribute 'monsters'
```
if I used this **for monster in self.active_monsters**
I got
```
    player.evolve_monster(monster,evolution["monster_slug"])
AttributeError: 'tuple' object has no attribute 'evolve_monster'
```

I'm sure it's something silly, but I cannot see now.

kerizane: It's the zip

JaskRendix:
Ok, I'm going to open another PR. Booster will leave out, as soon as we have the new name, we can insert easily

Issue #1314, opened 2022-09-12
Animation download script does not handle duplicate animations


AndyMender:
**Is your feature request related to a problem? Please describe.**
Currently, on the [Tuxemon wiki](https://wiki.tuxemon.org/index.php?title=Category:Technique_Animation) we have multiple copies of the same animation with different sizes. The problem is that the animation download script doesn't distinguish between them, because it strips the number suffix and from what I noticed locally it gives the impression as if the new animations have more frames.

**Describe the solution you'd like**
We have a couple of options
- Only download the first encountered animation
- Keep the suffix and treat animation variants of different dimensions as separate animations

**Describe alternatives you've considered**
Above

**Additional context**
N/A

kerizane:
There are a couple of options.

Including the number within the file would work, but would need you to update the json files again. 

64 x 64 is the preferred size, I think we should try to use this where possible. For a few of these I assume the other size was the original, and the 64 x 64 is an edited version. 

I think we should clean the wiki up a bit, as some of these animations aren't really used - options:
1. Mark the animation as unused in the wiki (Add a category?).
2. Have a list of skipped animations in the script.
3. Delete / move the larger versions within the wiki.
4. Only download a non-standard size if 64 x 64 isn't present.

AndyMender:
1. Sounds good. We can scrape that label via the animation download script.
2. Yup!
3. Delete/move as in from the git tree or in the wiki?
4. That should be easy to check. We cache the GIF locally in the anim download script so we can consider the 64x64 GIF to be the default (name of animation without any suffixes, etc.) and only use the differently sized GIF if no default is present

Who has access to the wiki to do the clean up by the way?

kerizane:
@Sanglorian

kerizane:
Up to you but it would go with option 2 or 4, as this doesn't depend on sanglorian. If we ever need to support multiple file sizes we can just change the script.

Option 2 is a bit manual, but it's the simplest, and gives you the most control. In this case you could detect that there are multiple sizes and require you to config all but one of them as being skipped? Could combine it with option 4 and only require config if there are multiple nonstandard sizes.

AndyMender:
I'll go for option 4. Option 2 can be implemented as a log line in the script's output :)

kerizane:
In option 2 I meant explicitly choose animations to skip as static config within the script.

Sanglorian:
[...]

PR #1318, opened 2022-09-13
Add handling of duplicate anims and non-64x64 anim variants to download script

AndyMender:
Addresses: https://github.com/Tuxemon/Tuxemon/issues/1314

Just a note that this still doesn't catch all corner cases like:
- https://wiki.tuxemon.org/File:Firelion_right.gif (the smaller sized anim is 64x61, not 64x64)
- https://wiki.tuxemon.org/File:Rockfall_193.gif (none of the anims are 64x64)

### DIFF
--- a/scripts/download_technique_animations.py
+++ b/scripts/download_technique_animations.py
@@ -15,7 +15,9 @@ from urllib.parse import urljoin
 
 import requests
 from lxml import html
-from PIL import Image
+from PIL import Image, ImageChops
+
+FRAME_SIZE = 64
 
 WIKI_URL = "https://wiki.tuxemon.org"
 
@@ -65,7 +67,7 @@ def download_animation_credits(gif_page_url: str) -> str:
     gifpage_source = requests.get(gif_page_url)
     gifpage_tree = html.fromstring(gifpage_source.content)
     credits_blocks = gifpage_tree.xpath("//div[@class='mw-content-ltr']/div/p")
-    
+
     credits_text = ""
     for credits_row in credits_blocks:
         credits_text += credits_row.text.strip() if credits_row.text else ""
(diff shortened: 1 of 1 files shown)


AndyMender:
@kerizane I think this PR covers what we discussed in the linked ticket. Are there any changes required? :)

kerizane:
Sorry, I know I suggested this implementation, but I think we should do it this way instead:

https://github.com/Tuxemon/Tuxemon/issues/1314#issuecomment-1249730021

If the animations are currently broken, then I will accept a PR that fixes the animations without updating this script.

scripts/download_technique_animations.py
```diff
@@ -87,24 +89,47 @@ def download_animation_credits(gif_page_url: str) -> str:
...
                 ANIMATION_DIR, f"{base_name}{frame:02}.png"
             )
+
+            gif_image.seek(frame)
+            # Check if animation is not a duplicate or a non-64x64 variant on an existing one
+            if os.path.isfile(frame_filename):
+                with Image.open(frame_filename) as frame_image:
+                    if not ImageChops.difference(
+                        gif_image, frame_image
+                    ).getbbox():
```
kerizane, line 118: What does this if statement check?

AndyMender: It compares 2 image objects. I should've put a comment above to make it clearer, apologies!

AndyMender:
> Sorry, I know I suggested this implementation, but I think we should do it this way instead:
> 
> [#1314 (comment)](https://github.com/Tuxemon/Tuxemon/issues/1314#issuecomment-1249730021)
> 
> If the animations are currently broken, then I will accept a PR that fixes the animations without updating this script.

I'll keep the branch just in case, but yeah, if the preferred way is having a strict category for used animations, I will need to implement that differently.

PR #1323, opened 2022-09-16
random monster

JaskRendix:
Really useful action for modders.
At the moment you can add_monster, but this requires that you know the monster names.
Now, since there are many monsters and since I remember only a few (common issue I guess), random_monster can help.

Simply **random_monster 5** where 5 is the level.

It can be useful in a tournament, where they store your monsters and they give you random 3 or 6, etc.

### DIFF
--- a/tuxemon/event/actions/random_monster.py
+++ b/tuxemon/event/actions/random_monster.py
@@ -1,0 +1,81 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+from __future__ import annotations
+
+import os
+import random as rd
+from typing import NamedTuple, Optional, Union, final
+
+from tuxemon import monster
+from tuxemon.db import db
+from tuxemon.event import get_npc
+from tuxemon.event.eventaction import EventAction
+from tuxemon.npc import NPC
+
+
+class RandomMonsterActionParameters(NamedTuple):
+    monster_level: int
+    trainer_slug: Union[str, None]
+
+
+@final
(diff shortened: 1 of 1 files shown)


tuxemon/event/actions/random_monster.py
```diff
@@ -0,0 +1,84 @@
...
+        else:
+            trainer = get_npc(self.session, trainer_slug)
+
+        assert trainer, "No Trainer found with slug '{}'".format(
+            trainer_slug or "player"
+        )
+        
+        path = "mods/tuxemon/db/monster/"
+        
+        monsters = os.listdir(path)
```
kerizane: Can you get the monster from the DB object instead?

kerizane (changes requested):

tuxemon/event/actions/random_monster.py
JaskRendix: I tried, but I was unable to get from the DB. Slug is necessary. Eventually we can insert temporary a list of monsters inside the file. Like this one:
_monsters = ["null", "rockitten", "rockat", "jemuar", "nut", "bolt", "arthrobolt", ..."]_

kerizane: Can't you get the slug from the DB?

JaskRendix: No, I can't, that's why I used os.listdir(path). I got the idea from locale.py, as you can see it utilizes fetch and listdir.
root = prepare.fetch("l18n")
for locale in os.listdir(root):

kerizane: from tuxemon.db import db

monster_slug = random.choice(db.database["monster"])

kerizane: I'd prefer this to having a hardcoded path.

JaskRendix: It was simpler than I thought, thanks, I'll update

tuxemon/event/actions/random_monster.py
```diff
@@ -0,0 +1,81 @@
...
+        if trainer_slug is None:
+            trainer = self.session.player
+        else:
+            trainer = get_npc(self.session, trainer_slug)
+
+        assert trainer, "No Trainer found with slug '{}'".format(
+            trainer_slug or "player"
+        )
+        
+        monster_slug = rd.choice(list(db.database["monster"]))
```
kerizane, line 75: Are you sure you need to convert to a list first?

JaskRendix: I think so, because it works. Without list I got:
```
  File "/home/robespierre/Tuxemon/tuxemon/event/actions/random_monster.py", line 74, in start
    monster_slug = rd.choice(db.database["monster"])
  File "/usr/lib/python3.10/random.py", line 378, in choice
    return seq[self._randbelow(len(seq))]
KeyError: 84 --> a different number each time
```

kerizane: Weird!

Ok, add a comment explaining this in case somebody is tempted to refactor.

`# list is required as choice expects a sequence`

kerizane (approved):

PR #1329, opened 2022-09-17
Don't strip dimension from animation names in download script

AndyMender:
Related to: https://github.com/Tuxemon/Tuxemon/issues/1314

Definitely up for discussion!


scripts/download_technique_animations.py
```diff
@@ -43,7 +43,7 @@ def download_bytes(url: str, filepath: str) -> None:
 def process_filename(filepath: str) -> str:
     """Extract base filename from an animation file path."""
     cleaned_filename = os.path.splitext(os.path.basename(filepath))[0]
-    cleaned_filename = cleaned_filename.strip("0123456789_")
```
kerizane, line 46: I think remove the strip and the split, but leave the "lower"

kerizane (changes requested):

kerizane:
Include the animations in the PR.

tuxemon/graphics.py
```diff
@@ -271,7 +271,7 @@ def animation_frame_files(
 
     """
     frames = list()
-    pattern = re.compile(rf"{name}\.?[0-9]+\.png")
+    pattern = re.compile(rf"{name}\.?_[0-9]+\.png")
```
kerizane: There is another folder than uses this, so it will need to be _?

scripts/download_technique_animations.py
kerizane: Will need to fix "12 hits for separation" in the JSON.

scripts/download_technique_animations.py
```diff
@@ -98,7 +98,7 @@ def gif_to_frames(filepath: str) -> None:
         animation_name = process_animation_name(base_name)
         for frame in range(0, image.n_frames):
             frame_filename = os.path.join(
-                ANIMATION_DIR, f"{base_name}{frame:02}.png"
+                ANIMATION_DIR, f"{base_name}_{frame:02}.png"
```
kerizane: The code already allows full stop as a delimiter, but I'm fine with adding _

scripts/download_technique_animations.py
kerizane: As we are keeping lower, it's still possible to have collisions. Track the names that you have seen in a set and skip / error if you see the same name twice.

kerizane:
You'll need to rename "capture 1" to "capture" on the wiki

scripts/download_technique_animations.py
AndyMender: I wasn't aware of the full stop. I think that would be okay, but on some platforms the whole part after a full stop could be treated as a file extension so perhaps `_` is safer?

tuxemon/graphics.py
AndyMender: Done!

scripts/download_technique_animations.py
AndyMender: > Will need to fix "12 hits for separation" in the JSON.

Do we actually use hits for separation anywhere? I tried searching in the repo, but didn't found any uses in technique JSONs.

> As we are keeping lower, it's still possible to have collisions. Track the names that you have seen in a set and skip / error if you see the same name twice.

We can also check if the gif file exists in `download_bytes`, the way it was kind of done in my other PR, but without checking the file content: https://github.com/Tuxemon/Tuxemon/pull/1318

Should we error out completely or just report the animation couldn't be downloaded?

One other issue I noticed is that the name of the animation is derived from the name of the gif and for some animations like "Explosion Total" the files have prefixes, for instance `120px-*` and `180px-*`. Not sure if these should be addressed on the Wiki or in the script.

AndyMender:
> You'll need to rename "capture 1" to "capture" on the wiki

@Sanglorian could you create an account on the Tuxemon wiki for me? :)

AndyMender:
> Include the animations in the PR.

As in, rename the existing animation frames or add the ones with the new name format in addition? I will probably have to rename the slugs in the technique JSONs as well then.

kerizane:
> > You'll need to rename "capture 1" to "capture" on the wiki
> 
> @Sanglorian could you create an account on the Tuxemon wiki for me? :)

You don't need an account.

kerizane:
I think we should use the name of the wiki page, not the gif file.

kerizane:
I would delete the existing animations, redownload, and fix the JSON slugs if necessary.

You can wait until the draft is complete, as that will make it easier to review.

scripts/download_technique_animations.py
kerizane: > Should we error out completely or just report the animation couldn't be downloaded?

Up to you.

Ideally either? Control this via a const or cli arg.

kerizane: > Do we actually use hits for separation anywhere? 

Not sure, I thought we did.

AndyMender:
> I think we should use the name of the wiki page, not the gif file.

You mean for the names of the animation frame files? I'll see how to incorporate that into the script.

> You don't need an account.

I'm getting a 404 from Parsoid if I try to edit a technique page via "Edit".

kerizane:
Ah, edit source lets you edit the category, but not change the name of the file...

kerizane:
You are able to change the category without an account. Whilst you're waiting for the account you could add the duplicate detection (by file name) in another branch and redownload the animations.

AndyMender:
> You are able to change the category without an account. If you like we could forget about the "don't strip numbers from name" logic, and just use the new category to avoid duplicates.
> 
> Can sort that out in the future.

If it's okay to keep the 12 in `12_hits_for_separation` and 1 in `capture_1` (probably fine, since we might have different Capture anims later) then we don't need to strip anything and just fix the technique JSONs.

kerizane:
Yes, we should allow both of the above.

Just check that "capture" isn't hard-coded anywhere (I was assuming it was, but it is probably just in the JSON).

AndyMender:
> Yes, we should allow both of the above.
> 
> Just check that "capture" isn't hard-coded anywhere (I was assuming it was, but it is probably just in the JSON).

Found the hardcoded location and added a TODO to later fix it. Fixed the rest and added new animations. Brief testing around Paper Town showed no immediate issues.

kerizane (approved):

kerizane (approved):

kerizane (approved):

kerizane:
Good work, thanks.

PR #1330, opened 2022-09-17
States should use __init__ method, not startup

### LINKED ISSUE #1167: Change `startup` method of states to `__init__`, and use instances instead of classes in `push_state`.

kerizane:
Fixes #1167

### DIFF
--- a/tuxemon/client.py
+++ b/tuxemon/client.py
@@ -580,14 +580,14 @@ class LocalPygameClient:
     @overload
     def push_state(
         self,
-        state_name: Type[StateType],
+        state_name: StateType,
         **kwargs: Any,
     ) -> StateType:
         pass
 
     def push_state(
         self,
-        state_name: Union[str, Type[StateType]],
+        state_name: Union[str, StateType],
         **kwargs: Any,
     ) -> State:
         """Push new state, by name"""
--- a/tuxemon/main.py
+++ b/tuxemon/main.py
@@ -76,21 +76,21 @@ def main(
     # since menus do not clean up dirty areas, the blank,
     # "Background state" will do that.  The alternative is creating
     # a system for states to clean up their dirty screen areas.
-    client.push_state(BackgroundState)
+    client.push_state(BackgroundState())
     if not config.skip_titlescreen:
-        client.push_state(StartState)
+        client.push_state(StartState())
 
     if load_slot:
-        client.push_state(LoadMenuState, load_slot=load_slot)
+        client.push_state(LoadMenuState(load_slot=load_slot))
         client.pop_state()
     elif config.splash:
-        client.push_state(SplashState)
-        client.push_state(FadeInTransition)
+        client.push_state(SplashState(parent=client.state_manager))
+        client.push_state(FadeInTransition())
 
     # TODO: rename this to "debug map" or something
     if config.skip_titlescreen:
         map_name = prepare.fetch("maps", prepare.CONFIG.starting_map)
-        client.push_state(WorldState, map_name=map_name)
+        client.push_state(WorldState(map_name=map_name))
 
     # block of code useful for testing
     if config.collision_map:
--- a/tuxemon/menu/input.py
+++ b/tuxemon/menu/input.py
@@ -34,7 +34,7 @@ class InputMenu(Menu[InputMenuObj]):
     background = None
     draw_borders = False
 
-    def startup(
+    def __init__(
         self,
         *items: Any,
         prompt: str = "",
@@ -54,7 +54,7 @@ class InputMenu(Menu[InputMenuObj]):
             initial: Optional string to pre-fill the input box with.
 
         """
-        super().startup(*items, **kwargs)
+        super().__init__(*items, **kwargs)
         self.input_string = initial
         self.chars = T.translate("menu_alphabet").replace(r"\0", "\0")
         self.n_columns = int(T.translate("menu_alphabet_n_columns"))
--- a/tuxemon/menu/input.py
+++ b/tuxemon/menu/input.py
@@ -36,7 +36,6 @@ class InputMenu(Menu[InputMenuObj]):
 
     def __init__(
         self,
-        *items: Any,
         prompt: str = "",
         callback: Optional[Callable[[str], None]] = None,
         initial: str = "",
@@ -54,7 +53,7 @@ class InputMenu(Menu[InputMenuObj]):
             initial: Optional string to pre-fill the input box with.
 
         """
-        super().__init__(*items, **kwargs)
+        super().__init__(**kwargs)
         self.input_string = initial
         self.chars = T.translate("menu_alphabet").replace(r"\0", "\0")
         self.n_columns = int(T.translate("menu_alphabet_n_columns"))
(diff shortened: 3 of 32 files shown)


kerizane:
You can pass the state in as a string or an object. I don't think there's any point in supporting the type any more.

If you pass in the type it will work, but you will get a deprecation warning. This is to make it so that any new states implemented whilst this is in review won't break the game when this is merged.

After this is in we can enforce the new interface during PRs, convert any remaining usages, and remove support for the deprecated interface.

tuxemon/state.py
```diff
@@ -227,23 +220,21 @@ def draw(self, surface: pygame.surface.Surface) -> None:
...
-        The state can prepare to be drawn during State.startup
-
         Parameters:
             surface: Surface to be rendered onto.
 
         """
 
     def startup(self, **kwargs: Any) -> None:
         """
+        DEPRECATED - Put use __init__ instead.
```
kerizane: Typo

tuxemon/state.py
```diff
@@ -359,7 +350,7 @@ def _instance(self, state_name: str) -> State:
             state = self._state_dict[state_name]
         except KeyError:
             raise RuntimeError(f"Cannot find state: {state_name}")
-        instance = state(self)
+        instance = state(parent=self, **kwargs)
```
kerizane: Wrong

tuxemon/states/items/__init__.py
```diff
@@ -88,17 +88,15 @@ class ItemMenuState(Menu[Item]):
...
+        super().__init__(**kwargs)
 
         # this sprite is used to display the item
         # its also animated to pop out of the backpack
         self.item_center = self.rect.width * 0.164, self.rect.height * 0.13
         self.item_sprite = Sprite()
         self.sprites.add(self.item_sprite)
 
-        # do not move this line
-        super().startup(**kwargs)
```
kerizane, line 106: This works.

tuxemon/states/items/__init__.py
```diff
@@ -253,16 +251,14 @@ class ShopMenuState(Menu[Item]):
...
-        self.state = "normal"
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
 
         # this sprite is used to display the item
         self.item_center = self.rect.width * 0.164, self.rect.height * 0.13
         self.item_sprite = Sprite()
         self.sprites.add(self.item_sprite)
 
-        # do not move this line
```
kerizane, line 269: This works.

tuxemon/client.py
```diff
@@ -580,14 +580,14 @@ def push_state(self, state_name: str, **kwargs: Any) -> State:
     @overload
     def push_state(
         self,
-        state_name: Type[StateType],
+        state_name: StateType,
```
vnmabus, line 583: We should use a `TypeVar` with `StateType` bound here to link the passed state and the returned one, as they are the same.

tuxemon/main.py
```diff
@@ -76,21 +76,21 @@ def main(
...
+        client.push_state(StartState())
 
     if load_slot:
-        client.push_state(LoadMenuState, load_slot=load_slot)
+        client.push_state(LoadMenuState(load_slot=load_slot))
         client.pop_state()
     elif config.splash:
-        client.push_state(SplashState)
-        client.push_state(FadeInTransition)
+        client.push_state(SplashState(parent=client.state_manager))
```
vnmabus, line 87: Why this parameter suddenly appeared?

tuxemon/menu/input.py
```diff
@@ -34,7 +34,7 @@ class InputMenu(Menu[InputMenuObj]):
     background = None
     draw_borders = False
 
-    def startup(
+    def __init__(
```
vnmabus, line 37: Remove the `kwargs` parameter from all `__init__` methods, as it was only necessary in `startup` for preventing a LSP violation.

tuxemon/state.py
```diff
@@ -85,24 +85,17 @@ class should be created. Update must be overloaded in the child class.
     transparent = False  # ignore all background/borders
     force_draw = False  # draw even if completely under another state
 
-    def __init__(self, parent: StateManager) -> None:
+    def __init__(self, **_: Any) -> None:
```
vnmabus: What happened with the parent parameter?

tuxemon/client.py
kerizane: StateType = TypeVar("StateType", bound=State)

tuxemon/state.py
kerizane: This attribute is only used by one class, so I changed made it a parameter of that class. Otherwise all states would require it to be passed in during pushstate (which would be annoying), or we'd have to do something hacky.

tuxemon/main.py
kerizane: See your other question.

kerizane:
@vnmabus Do you think the init parameters should be keyword only?

I will leave these as is. People are free to use args or kwargs for new classes.

We don't need to support args via push_state, as people can either pass in the object or pass the args in as kwargs if they are using the old interface.

tuxemon/menu/input.py
kerizane: We still need it for Menus.

Although, I guess the child classes don't need kwargs.

tuxemon/client.py
kerizane: Already in use.

tuxemon/menu/input.py
kerizane: I have removed it for the child classes, but am leaving it here.

kerizane:
@vnmabus could you take another look?

Would be nice to get this in, as I've had some merge conflicts, so please be clear as to what is essential and what can be left for a future ticket.

I will take a look at which parameters actually need defaults in another ticket.

Thanks!

tuxemon/client.py
vnmabus: Oops, my fault.

tuxemon/menu/input.py
vnmabus: Why not the explicit allowed parameters, with their actual types?

vnmabus (dismissed):
LGTM

tuxemon/menu/input.py
kerizane: Good idea.

There are 19 parameters, and they are used to override class attributes. Seemed slightly unusual so I didn't want to do it here.

Maybe in a future ticket.

kerizane:
> LGTM

Thanks!

xirsoi (dismissed):
It all looks good to me, but this is the kind of thing where my relatively shallow knowledge of Python makes me less comfortable in reviewing it.

So, i'm leaning on vnmabus' superior knowledge, and he has approved it.

bitcraft (approved):

PR #1332, opened 2022-09-18
Maybe_translate and replace

JaskRendix:
I discovered it after working on #1331 

1. The strings in _use_success_  as well as _use_failure_ weren't properly treated.  Eg {user} wasn't replaced, now it's replaced.

~~Getting rid of **maybe_translate** for **translate**, and since there are no other uses of maybe_translate, I have deleted it from **locale.py**.
I understand that the function was there because, I guess, the null value in use_success. About this we can utilize a similar function to use_tech. It was combat X on Y.~~ I'm not sure about what I wrote, because until now it works fine.

maybe_translate

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -980,9 +980,14 @@ class CombatState(CombatAnimations):
                     msg_type = (
                         "use_success" if result["success"] else "use_failure"
                     )
+                    context = {
+                        "user": getattr(user, "name", ""),
+                        "name": technique.name,
+                        "target": target.name,
+                    }
                     template = getattr(technique, msg_type)
                     if template:
-                        message += "\n" + T.translate(template)
+                        message += "\n" + T.format(template, context)
 
             self.alert(message)
             self.suppress_phase_change(action_time)


tuxemon/technique.py
```diff
@@ -149,9 +149,9 @@ def load(self, slug: str) -> None:
         # technique use notifications (translated!)
         # NOTE: should be `self.use_tech`, but Technique and Item have
         # overlapping checks
-        self.use_item = T.maybe_translate(results.get("use_tech"))
-        self.use_success = T.maybe_translate(results.get("use_success"))
-        self.use_failure = T.maybe_translate(results.get("use_failure"))
+        self.use_item = T.translate(results["use_tech"])
```
kerizane: Is there pydantic validation for these values?

kerizane: How quickly do we find out if they aren't set?

kerizane:
Leave maybe_translate, might be useful later

tuxemon/technique.py
JaskRendix: You're right. I thought it was necessary for the correct display during battles. But after a series of tests, it wasn't.

JaskRendix:
Done, minimal intervention, now we can see the use_success, use_failure, etc messages. It replaces correct {user}, etc.

kerizane (approved):

PR #1339, opened 2022-09-21
get player party

JaskRendix:
What is get_player_party? It's the upgraded version of get_player_monster.

**get_player_monster** helps you to store the instance_id in a variable of your choice, but get_player_monster opens a menu, you choose your monster, then your instance_id is associated to the variable (you need to come up with a name, but which name?)

**get_player_party** speeds things up for the modders. In one action you get all the instance_ids of the party. No need to choose your monster. No need to come up with a name for your variable.

You got 4 monsters, then you got this:
Screenshot from 2022-09-21 09-59-10

I wanted to implement this format because it's simple and it updates automatically the slot variables if you run again the action, so there is no risk of creating other (unused) variables.

(run isort and black)

### DIFF
--- a/tuxemon/event/actions/get_player_party.py
+++ b/tuxemon/event/actions/get_player_party.py
@@ -1,0 +1,60 @@
+#
+# Tuxemon
+# Copyright (c) 2020      William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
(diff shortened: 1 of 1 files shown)


kerizane:
Can you give me an example of how / when you would use this?

JaskRendix:
When you need to store all the monsters in a single action. Everything else you can use get_player_monster. At the moment I don't think there is something (plot) that requires get_player_party. I can always close and reopen later. No problem.

PR #1337, opened 2022-09-19
evolution and evolved by item

JaskRendix:
This PR includes:
1. evolution through objects (like boosters);
2. related to the 1, I added the following object conditions: _has_path_ (to get the object) and _level_ (to get the level);
3. related to the 1, I added the _evolve_ object action, it allows to address the evolution;
4. evolution through combat (traditional evolution after leveling up, like rockitten that at level 24 evolves in rockat);

Already run with isort and black

What's missing?
1. evolution through stats (**katacoon** and **vivipere** cases);
2. evolution animation;
3. new object type #1334 

Inside the earth_booster.json object, something like this:
```
  "effects": [
    "evolve eruptibus"
  ],
  "conditions": [
    "level target,>=,0",
    "has_path target,earth_booster"
  ],
```

PR #1345, opened 2022-09-22
remove evolve_monsters

JaskRendix:
Since we merged #1337 we can remove the old evolve_monsters (action and condition) as well as get_evolution (monster.py).


kerizane:
Could you raise a separate PR with the following:

Modify add_monster so that it takes an index (default behaviour should still be append to end)
Change your code from recent PR so that you are using remove_monster + add_monster rather than assigning directly to the list.
Use add_monster when building the party when loading a file.

JaskRendix:
>  Change your code from recent PR so that you are using remove_monster + add_monster rather than assigning directly to the list.

Do you realize it was with remove and add in the beginning? I'll do.

kerizane:
I think you were using list.append and list.remove, but if you're right then I'm sorry.

xirsoi (approved):
LG;TftPR

Issue #201, opened 2016-08-02
[Locale] New gender attribute for locale context


bitcraft:
I'll preface this by just saying I hope to avoid a long discussion about the differences of sex and gender in this thread.

In the future, it will be possible to give the translator hints about the gender of items, TXMN, and NPC's so that the machine-generated dialogs sound more natural (for example: using her/her/it when appropriate).  In gettext terms, this would be used in the same way as "context".

So, for this issue, it would be sufficient to go through all the the DB objects, edit the json to include a "gender" attribute.

For discussion purposes, we will just assume the following types for now: "male", "female", "neutral".  This issue is not directly related to fusion right now.

kerizane:
NPCs can have a gender, should be easy to figure that out.
Items sort of have a gender, but that gender is a language feature and doesn't need to be tracked within the code.

Individual monsters should have a gender. I guess we should just pick one when we generate the monster.

Most types of monster will be male/female with a 50/50 chance.
Some types monsters will always be male, always female, or always neutral.

PR #1351, opened 2022-09-24
gender monsters

JaskRendix:
Thank again @kerizane for the reply #201 
This was the draft I was talking about. Gender about individual monsters.
save_upgrader needs a fix, I hadn't figured out at the moment, because I was trying to get the right format.
I know there are shape that remain neutral (like robos), so it needs an additional if about the shape.

it needs a fix for add_monster, random_monster and evolve_monster, breeding I haven't looked into at the moment.


JaskRendix:
I need to update the development

Issue #466, opened 2018-10-07
Missing keys in en_US


airon90:
The following keys seems to be missing in en_US locale:

* ```npc_red_name```
* ```npc_professor_name```
* ```npc_xerogrunt1```
* ```npc_xerogrunt2```
* ```npc_xerogrunt3```
* ```npc_xerogrunt4```
* ```npc_xerogrunt5```
* ```npc_xerogrunt6```
* ```npc_xerogrunt7```
* ```npc_xerogrunt8```
* ```npc_xerogrunt9```
* ```npc_xerogrunt10```
* ```npc_misa```
* ```npc_allie```
* ```npc_liela```
* ```npc_cottonnurse```
* ```npc_aeble```

bitcraft:
Thanks for opening this.  Ive just created a new label called "Locale".  Please add it for new tickets with translation or localization errors.  I believe these keys are from @Qiangong2

airon90:
I will add labels next time :)

Qiangong2:
Will the names of the NPCs change between languages?

bitcraft:
Probably.  They should be in en_US anyway.

Qiangong2:
Well, at the very least, aeble should be changed to something else in different languages. It is meant to be pronounced like the word "able" as in "I am able to". Ideally, it would be a misspelled version of the word in that language. 

I don't know how you would translate the other names though. Honestly, I just came up with them when they were needed

airon90:
> Will the names of the NPCs change between languages?

In [Italian translation](https://github.com/Tuxemon/Tuxemon/blob/0e3f964658021d7bf6a092bdc404a4b7f58cb9b3/tuxemon/resources/db/locale/it_IT.json#L468) ```Cotton Town``` became ```Cotonopoli``` (Pokémonish. Literaly it means Cotton Town) while in [Esperanto translation](https://github.com/Tuxemon/Tuxemon/blob/0e3f964658021d7bf6a092bdc404a4b7f58cb9b3/tuxemon/resources/db/locale/eo.json#L184) ```Maple``` became [```Klara```](https://eo.wikipedia.org/wiki/Klara_Zamenhof). I think that each language community will decide if to localize the name or leave the original ones,

airon90:
In localized version of Pokémon games, many names changed. See [example](https://bulbapedia.bulbagarden.net/wiki/Pallet_Town#Name_origin)

Qiangong2:
[...]

PR #1355, opened 2022-09-24
missing keys

JaskRendix:
It started #466 with a dozen, now we are at more than 60 missing keys. 
This is everything missing, no exceptions.
I share as draft, so if there is any advice, etc. about the names, feel free.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -2336,6 +2336,12 @@ msgstr "Knight I"
 msgid "knight2"
 msgstr "Knight II"
 
+msgid "knight3"
+msgstr "Knight III"
+
+msgid "knight4"
+msgstr "Knight IV"
+
 msgid "liela"
 msgstr "Liela"
 
@@ -2378,6 +2384,9 @@ msgstr "Team Xero Grunt I"
 msgid "xerogrund2"
 msgstr "Team Xero Grunt II"
 
+msgid "xerogrunt1"
+msgstr "Team Xero Grunt I"
+
 msgid "xerogrunt2"
 msgstr "Team Xero Grunt II"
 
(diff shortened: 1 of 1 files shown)


mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -2405,6 +2405,198 @@ msgstr "Team Xero Grunt IX"
...
+msgid "taba_greeter"
+msgstr "Radio"
+
+msgid "tallon"
+msgstr "Tallon"
+
+msgid "tuxemart"
+msgstr "Tuxemart"
+
+msgid "xerogrunt1"
```
kerizane: There should already be a group of xerogrunts

JaskRendix: You gave me an idea

JaskRendix:
It was a mess. NPCs spread everywhere, now everything is below the right label and most important sorted in alphabetical order.

kerizane:
This diff is massive and difficult to read. What exactly did you do?

I would prefer the diff to be focused on adding the names.

JaskRendix:
NPCs where split in two sections and they weren't in alphabetical order. I just reordered.

As you wish. I'll update.

kerizane:
It might be good to order them in another ticket after this is merged.

JaskRendix:
Sure

kerizane:
If you could add your new names in alphabetical order, but  not reorder the existing ones, that would be good.

JaskRendix:
Sure, as it was in the beginning. I'll move only the knights and xerogrunt as you pointed out before.

JaskRendix:
Done, tell me if it's ok.

kerizane:
@Sanglorian @Qiangong2 what do you think?

I'm wondering if all of these are actually necessary. When do we use the names, is it just for battle?

Do we want to define names for everyone, or does this just lead to extra work for translators?

JaskRendix:
I don't think all these are necessary. Some are part of unused maps, but at the end the JSON exists. This shows the problem of recreating the same NPC again and again. Think about mom and homemaker, as well as postboy, we have already a spyder_postboy.

Qiangong2:
All of the spyder npcs are for Sanglorian's campaign, which is why they look like duplicates.

The NPC names are in there for localization purposes.

Sanglorian:
I believe names are shown when battle begins only. The NPCs without any name given are ones I use for NPCs on the overland only - so spyder_picnicker might stand in for multiple picnickers on different maps.

kerizane:
If they're only used for battle then I don't think we should be giving them names for no reason. @Sanglorian @Qiangong2 do you agree?

I'm planning to write some map validation logic at some point, we can check for missing names there.

Sorry for the wasted effort @JaskRendix

JaskRendix:
No problem, remember to close the #466

Issue #1121, opened 2022-02-10
Get the monster ID for each tuxemon from the wiki


ultidonki:
On the wiki, each tuxemon has a unique ID number. This isn't used yet, but the monster journal should probably show this number and sort all monsters by the ID number. 

The extractor - /scripts/tuxepedia/extractor.py and api.py should be updated to also get the ID number, and add it to the .json. 
Only the first 154 tuxemon have a monster ID, they're on the SECOND table of the Completed Monsters screen, see https://wiki.tuxemon.org/Completed_Tuxemon  

Also see issue #1120 for a similar wiki-scraping issue - but this one (getting the ID numbers) will be easier, as #1120 needs you to output everything in a .po format which the scripts don't currently do.

JaskRendix:
I was unable to fix api.py and extractor.py, but I imported the id, weight and height in the json (used Beatiful Soup). I noticed a series of things about the Tuxemon that we need to discuss. If you want to, then I can pull a PR, so we can discuss it and fix it.

Everything based on https://wiki.tuxemon.org/Completed_Tuxemon

ultidonki:
> I noticed a series of things about the Tuxemon that we need to discuss. If you want to, then I can pull a PR, so we can discuss it and fix it.

Yes, please open the PR, thanks a lot for that! I probably won't be able to answer every question but probably other people will be able to discuss if it you open the PR. Thanks! :smile:

JaskRendix:
I got curious about Journal/Tuxepedia. I wanted to ask you: how do you see the monster journal?
Do you see it as a dictionary? List? A JSON file (like encounters or economy or integrated like partymember inside NPC)?
I keep thinking that a dictionary or a list isn't enough.
It needs at least parameters like (monster.slug as connects everything, txmn_id, seen - boolean 0/1, and I don't know).

kerizane:
I think most of the info can be kept within the monster JSON file.

Some info will need to be stored in the save file: whether a monster has been seen and/or caught.

This could be a single dict of the slug to another dict/tuple, or it could be a set of seen monster slugs and a set of caught monsters slugs.

ultidonki:
@JaskRendix Thanks for implementing all these and letting me know all these issues can be closed! :smile:

PR #1358, opened 2022-09-25
seen or caught

JaskRendix:
It was a request from #1121 

Implementing a dictionary that allows to track the monsters seen or caught, and obviously unseen if there is no value. I opted for an int, because maybe in a future someone can expand the "levels". At the moment there is seen (1) and caught (2).

As usual the fix for add_monster and random_monster, included evolve_monster. Save_upgrader that automatically set caught (2) your monsters as well as in combat.py for a capture result. A couple of actions and one condition for modders.

### DIFF
--- a/tuxemon/event/actions/add_monster.py
+++ b/tuxemon/event/actions/add_monster.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 from typing import NamedTuple, Optional, Union, final
 
 from tuxemon import monster
+from tuxemon.db import SeenStatus
 from tuxemon.event import get_npc
 from tuxemon.event.eventaction import EventAction
 from tuxemon.npc import NPC
@@ -76,3 +77,4 @@ class AddMonsterAction(EventAction[AddMonsterActionParameters]):
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
--- a/tuxemon/event/actions/add_monster.py
+++ b/tuxemon/event/actions/add_monster.py
@@ -77,4 +77,4 @@ class AddMonsterAction(EventAction[AddMonsterActionParameters]):
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
-        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught
--- a/tuxemon/event/actions/random_monster.py
+++ b/tuxemon/event/actions/random_monster.py
@@ -21,12 +21,11 @@
 
 from __future__ import annotations
 
-import os
 import random as rd
 from typing import NamedTuple, Optional, Union, final
 
 from tuxemon import monster
-from tuxemon.db import db
+from tuxemon.db import db, SeenStatus
 from tuxemon.event import get_npc
 from tuxemon.event.eventaction import EventAction
 from tuxemon.npc import NPC
@@ -80,3 +79,4 @@ class RandomMonsterAction(EventAction[RandomMonsterActionParameters]):
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
--- a/tuxemon/event/actions/random_monster.py
+++ b/tuxemon/event/actions/random_monster.py
@@ -25,7 +25,7 @@ import random as rd
 from typing import NamedTuple, Optional, Union, final
 
 from tuxemon import monster
-from tuxemon.db import db, SeenStatus
+from tuxemon.db import SeenStatus, db
 from tuxemon.event import get_npc
 from tuxemon.event.eventaction import EventAction
 from tuxemon.npc import NPC
@@ -69,14 +69,14 @@ class RandomMonsterAction(EventAction[RandomMonsterActionParameters]):
         assert trainer, "No Trainer found with slug '{}'".format(
             trainer_slug or "player"
         )
-        
+
         # list is required as choice expects a sequence
         monster_slug = rd.choice(list(db.database["monster"]))
-      
+
         current_monster = monster.Monster()
         current_monster.load_from_db(monster_slug)
         current_monster.set_level(monster_level)
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
-        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught
--- a/tuxemon/event/conditions/tuxepedia_is.py
+++ b/tuxemon/event/conditions/tuxepedia_is.py
@@ -20,6 +20,7 @@
 #
 from __future__ import annotations
 
+from tuxemon.db import SeenStatus
 from tuxemon.event import MapCondition
 from tuxemon.event.eventcondition import EventCondition
 from tuxemon.session import Session
@@ -61,9 +62,11 @@ class TuxepediaIsCondition(EventCondition):
         monster_str = condition.parameters[1]
 
         if monster_key in player.tuxepedia:
-            if player.tuxepedia[monster_key] == monster_str:
-                return True
+            if monster_str == "caught":
+                if player.tuxepedia[monster_key] == SeenStatus.caught:
+                    return True
+            elif monster_str == "seen":
+                if player.tuxepedia[monster_key] == SeenStatus.seen:
+                    return True
             else:
                 return False
-        else:
-            return False
--- a/tuxemon/event/conditions/tuxepedia_is.py
+++ b/tuxemon/event/conditions/tuxepedia_is.py
@@ -68,5 +68,7 @@ class TuxepediaIsCondition(EventCondition):
             elif monster_str == "seen":
                 if player.tuxepedia[monster_key] == SeenStatus.seen:
                     return True
+                elif player.tuxepedia[monster_key] == SeenStatus.caught:
+                    return True
             else:
                 return False
--- a/tuxemon/event/conditions/tuxepedia_is.py
+++ b/tuxemon/event/conditions/tuxepedia_is.py
@@ -1,0 +1,69 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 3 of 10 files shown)


tuxemon/event/actions/add_monster.py
```diff
@@ -76,3 +76,4 @@ def start(self) -> None:
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
+        trainer.tuxepedia[monster_slug] = 2
```
kerizane: Can we use a string enum?

```
class SeenStatus(str, Enum):
    seen = "seen"
    caught = "caught"
```

kerizane: I am also considering counting the number of times you have seen / caught instead. There is a slight complication in that you might want to not count an NPC's monster multiple times, but we don't need to support this straight away.

What do you think?

tuxemon/event/conditions/tuxepedia_is.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Script usage:
+        .. code-block::
+
+            is tuxepedia_is <monster_slug>,<value>
+
+    Script parameters:
+        monster_slug: Monster slug name (e.g. "rockitten").
+        value: seen = 1 / caught = 2
+
```
kerizane, line 41: As I said earlier, I think a string is better than a magic number.

kerizane (changes requested):

tuxemon/event/actions/add_monster.py
JaskRendix: I'm open to your proposal. I think enum is a good concept.

kerizane: Let's go with counting then.

JaskRendix:
I wasn't sure if below NPC or Monsters (party), so I did only this just to be sure. Then I'll update the other parts.

kerizane:
It doesn't need to be on the npc or the monster. You're just using this in the tuxepedia instead of an int.

JaskRendix:
Multiple times seen or caught would result in another class?

Update:
I forgot value, I'm updating.

JaskRendix:
I tried to put SeenStatus.caught instead of caught, but I got
{'rockitten': <SeenStatus.caught: 'caught'>}

kerizane:
`{'rockitten': <SeenStatus.caught: 'caught'>}`

This is fine, that's just what an enum looks like.

If you're tracking the number of times then you don't really need to use an enum.

```
'rockitten': {
'seen': 10,
'caught': 0,
}
 ```

You can use a typeddict to define the structure.

**Comment by JaskRendix** [comment]
Thanks, I just added SeenStatus.*.value for simplifying, so it prints as "caught" or "seen"

I removed import os from random_monster because I forgot to delete it when I created the file. It's unused.

**Comment by JaskRendix** [comment]
We can integrate the counting separately by counting both.
Think about this following draft:
```
class GenericCount(TypedDict):
    creature: str
    seen_count: int
    caught_count: int
    evolved_count: int


    def get_count(
        self,
        monster: Monster,
        seen: int = None,
        caught: int = None,
        evolved: int = None
    ) -> GenericCount:
        return {
            "creature": monster,
            "seen_count": +seen,
            "caught_count": +caught,
            "evolved_count": +evolved,
        }
```
It can become a sort of statistic for the monster itself.

**Comment by kerizane** [comment]
Have you tested combat against a wild monster and against an NPC?

If you run `make validate` from the base of the repo are there any errors?

**Comment by JaskRendix** [comment]
I tested both combats.
Against wild monster registers the monster as seen, if I capture it, then it becomes caught. If I meet again the monster (already caught), then it remains caught as should.
Against trainers, it registers all the monsters as seen (I fought Billie in route2) and it registers correctly the evolution too (rockitten to rockat).

Before the battle with Billie
Screenshot from 2022-09-27 14-27-04
After the battle with Billie
Screenshot from 2022-09-27 14-28-38

make validate
```
PYTHONPATH=. python scripts/schema.py --validate
pygame 2.1.2 (SDL 2.0.16, Python 3.10.6)
Hello from the pygame community. https://www.pygame.org/contribute.html
```
and if I do "python scripts/schema.py --validate"
```
  File "/home/nik/Tuxemon/scripts/schema.py", line 5, in <module>
    from tuxemon.db import (
ModuleNotFoundError: No module named 'tuxemon'
```

**Comment by JaskRendix** [comment]
I need to check a couple of things with the actions and condition

**Comment by JaskRendix** [comment]
Checked actions and condition, it works.

**Comment by kerizane** [comment]
Some minor changes I want, will explain when I have time.

File: tuxemon/event/actions/add_monster.py [hunk 3]
```diff
@@ -76,3 +77,4 @@ def start(self) -> None:
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
```
kerizane:
```suggestion
        trainer.tuxepedia[monster_slug] = SeenStatus.caught
```

kerizane: I'd rather you didn't use value everywhere, because then the typehints that claim this is SeenStatus are wrong.

If you don't like the way this gets printed then you can modify the way SeenStatus prints things.

Add the following method to SeenStatus

```
def __str__(self):
    return self.value
```

Not tested, but I think it should work.

Just so you know, the behaviour for classes that inherit from str and enum is like so:

```
>SeenStatus.seen.value == "seen"
True
>SeenStatus.seen == "seen"
True
```

tuxemon/save_upgrader.py
```diff
@@ -83,6 +84,12 @@ def upgrade_save(save_data: Dict[str, Any]) -> SaveData:
 
     save_data["battle_history"] = save_data.get("battle_history", {})
     save_data["money"] = save_data.get("money", {})
+    save_data["tuxepedia"] = save_data.get("tuxepedia", {})
+
+    # set as captured the party monsters
+    if not save_data["tuxepedia"]:
+        for mons in save_data.get("monsters", []):
```
kerizane, line 91: Note we also have monster in storage.

tuxemon/states/combat/combat.py
```diff
@@ -379,6 +380,9 @@ def transition_phase(self, phase: CombatPhase) -> None:
                 var["battle_last_monster_type"] = monster_record.slug
                 var["battle_last_monster_category"] = monster_record.category
                 var["battle_last_monster_shape"] = monster_record.shape
+                # Avoid reset string to seen if monster has already been caught
+                if monster_record.slug not in self.players[0].tuxepedia.keys():
```
```suggestion
                if monster_record.slug not in self.players[0].tuxepedia:
```

tuxemon/states/combat/combat.py
```diff
@@ -977,6 +981,10 @@ def perform_action(
                     # TODO: Don't end combat right away; only works with SP,
                     # and 1 member parties end combat right here
                     if result["success"]:
+                        # Tuxepedia: set monster as caught (2)
+                        self.players[0].tuxepedia[
+                            self.monsters_in_play[self.players[1]][0].slug
```
kerizane: can we just use `target` rather than `self.monsters_in_play[self.players[1]][0].slug`  ?

kerizane (changes requested):

kerizane:
If you fix these I will merge. I don't want to overcomplicate it by counting the monsters.

Make sure you test your changes.

tuxemon/event/actions/add_monster.py
JaskRendix: Thanks, you're right about value. I appreciate the feedback.

tuxemon/states/combat/combat.py
JaskRendix: ok

tuxemon/save_upgrader.py
JaskRendix: You're right

tuxemon/event/actions/random_monster.py
```diff
@@ -80,3 +79,4 @@ def start(self) -> None:
         current_monster.current_hp = current_monster.hp
 
         trainer.add_monster(current_monster)
+        trainer.tuxepedia[monster_slug] = SeenStatus.caught.value
```
kerizane: `SeenStatus.caught`

please do this for all cases of SeenStatus

kerizane (changes requested):

tuxemon/event/actions/random_monster.py
JaskRendix: Exactly what I'm doing

JaskRendix:
Testing is ok, it lacks only monster_boxes in save_upgrader

kerizane (dismissed):

kerizane:
> Testing is ok, it lacks only monster_boxes in save_upgrader

Great, add that and I will merge.

tuxemon/event/conditions/tuxepedia_is.py
```diff
@@ -0,0 +1,72 @@
...
+        # Read the parameters
+        monster_key = condition.parameters[0]
+        monster_str = condition.parameters[1]
+
+        if monster_key in player.tuxepedia:
+            if monster_str == "caught":
+                if player.tuxepedia[monster_key] == SeenStatus.caught:
+                    return True
+            elif monster_str == "seen":
+                if player.tuxepedia[monster_key] == SeenStatus.seen:
```
kerizane, line 69: Shouldn't we allow caught monsters to be classed as seen? You could add "seen_only" as an option.

kerizane: Or do we leave it as is and add seen_or_caught?

Which is least surprising?

JaskRendix: It makes sense what you say.

I saw caught as the upgrade of seen. I did this for the Tuxedex, so it can calculate how many seen and how many caught of the total number of monsters as well as for modders that can use for trigger an event.

JaskRendix:
Kennel is the box, I guess, because with this:
```
        for monstervalue in save_data.get("monster_boxes"):
            save_data["tuxepedia"][monstervalue] = SeenStatus.caught
```
I got `Kennel': <SeenStatus.caught: 'caught'>`

kerizane:
Check the rest of the code, but monster_boxes is probably a dictionary of box names against lists of monsters.
```
        for monsters in save_data.get("monster_boxes", {}).values():             
```

JaskRendix:
Done, updated.

It could be useful, I printed the structure save_data["monsters"].items():
```
('Kennel',[{'current_hp': 85, 'level': 5, 'name': 'Cateye', 'slug': 'cateye', 'total_experience': 125, 'gender': 'female', 'instance_id': '69457433c83d49e8aab324ec2a34a791', 'moves': ['amnesia', 'eyebite', 'flow']},{another monster, etc..}])
```

tuxemon/event/conditions/tuxepedia_is.py
kerizane: So, don't we need to fix this?

If something has been caught then we should return true if we ask if is seen.

JaskRendix: True, I'll update.

kerizane (approved):

PR #1359, opened 2022-09-26
Support flipping of combat animations

AndyMender:
Addresses: https://github.com/Tuxemon/Tuxemon/issues/749
Continuation of: https://github.com/Tuxemon/Tuxemon/pull/788


AndyMender:
I ported the code from the linked PR. Some immediate TODOs:
- [x] https://github.com/Tuxemon/Tuxemon/issues/749#issuecomment-1258571203
- [x] Adjust the direction of breath_blue and breath_fire in the wiki (done by @Sanglorian )

tuxemon/states/combat/combat.py
```diff
@@ -152,6 +154,8 @@ def load_technique_animation(technique: Technique) -> Optional[Sprite]:
             image = graphics.load_and_scale(fn)
             images.append((image, frame_time))
         tech = SurfaceAnimation(images, False)
+        if fix_direction:
+            tech.flip(True)
```
kerizane: Note that some animations will need to be flipped in the y axis too.

AndyMender: No problem, I just need to handle both directions then.

AndyMender: To be honest, I'm not super happy with the current implementation, because it allows flipping only in X `or` Y. Should we support flipping in both directions?

Another question, I understand the various "front" animations should support the "up" mode, correct?

kerizane: > To be honest, I'm not super happy with the current implementation, because it allows flipping only in X or Y. Should we support flipping in both directions?

Yes

> Another question, I understand the various "front" animations should support the "up" mode, correct?

?

AndyMender: The "front" animations are facing down so if we would want to flip them in the Y-axis, that would be equivalent to the "up" animations.

Sanglorian:
I've flipped the two breath animations in the wiki. 

With e.g. Firelion front and Firelion back, or Snake front and Snake back, I don't think we'd want to flip them on the Y axis. If we did anything, we'd play Firelion front for a technique targeting the player's monster and Firelion back for one targeting the opponent's monster, but that might be getting too involved. 

I regret not setting up separate wiki pages for the animations, which would allow us to define their metadata separately (e.g. Source, whether to flip and on which axis or axes, etc) -- might be worth considering for a later time. Otherwise we can define this stuff at the Technique level, just means a bit of redundancy when the same animation is used by multiple techniques.

tuxemon/event/eventengine.py
```diff
@@ -148,6 +148,8 @@ def __init__(self, session: Session) -> None:
 
         # debug
         self.partial_events: List[Sequence[Tuple[bool, MapCondition]]] = list()
+        # TODO: What type are the list elements?
+        self.parameters: List = list()
```
kerizane: Is this needed?

tuxemon/states/combat/combat.py
```diff
@@ -1013,7 +1017,15 @@ def perform_action(
                     )
                 )
 
-        tech_sprite = self._technique_cache.get(technique)
+        # TODO: Flipping in X-axis and Y-axis should be allowed,
+        #       and supported in the technique or animation JSON
+        flip_axes = ""
+        for trainer in self.ai_players:
+            if user in self.monsters_in_play[trainer]:
+                flip_axes = "x"
```
kerizane: This should be a bool that defines if we should show the normal animation or the flipped animation. The info on how to flip it should be on the technique object.

kerizane (changes requested):

tuxemon/event/eventengine.py
AndyMender: Nope, removed. Not sure what the original intention here was.

tuxemon/states/combat/combat.py
AndyMender: Is the new implementation better?

AndyMender:
I still need to resolve the conflict in technique.py and add a dict with flip_axes to the technique download script.

tuxemon/states/combat/combat.py
```diff
@@ -114,11 +114,13 @@ class TechniqueAnimationCache:
...
         self._sprites: Dict[Technique, Optional[Sprite]] = {}
 
-    def get(self, technique: Technique) -> Optional[Sprite]:
+    def get(self, is_flipped: bool, technique: Technique) -> Optional[Sprite]:
         """
         Return a sprite usable as a technique animation.
 
         Parameters:
+            is_flipped: Flag to determine whether animation frames should be flipped.
+            flip_axes: Axes in which the animation should be flipped.
```
```suggestion
```

tuxemon/states/combat/combat.py
```diff
@@ -114,11 +114,13 @@ class TechniqueAnimationCache:
     def __init__(self) -> None:
         self._sprites: Dict[Technique, Optional[Sprite]] = {}
 
-    def get(self, technique: Technique) -> Optional[Sprite]:
+    def get(self, is_flipped: bool, technique: Technique) -> Optional[Sprite]:
```
kerizane: Fussy, but I would have these parameters the other way round. technique is the main parameter, is_flipped is secondary.

tuxemon/technique.py
```diff
@@ -196,6 +197,7 @@ def load(self, slug: str) -> None:
                 logger.error(
                     f"Cannot find animation frames for: {self.animation}",
                 )
+        self.flip_axes = results.flip_axes
```
kerizane: You need to set flip_axes on the pydantic model within tuxemon/db.py 

Try Literal["x", "y, "xy"] as the typehint.

kerizane (changes requested):

tuxemon/states/combat/combat.py
kerizane: Not sure about this.

tuxemon/states/combat/combat.py
kerizane: Yes

kerizane:
> If we did anything, we'd play Firelion front for a technique targeting the player's monster and Firelion back for one targeting the opponent's monster, but that might be getting too involved.

I think this would be pretty easy to implement. We could have a json entry called "enemy_animation". In the code we would use this if it's present, otherwise use flip_axes.

Andy, let's save this for another ticket. If you agree with this idea, make a note on the original issue, please. This PR should not flip the front/back animations.

kerizane:
Looks like this is nearly done. Please make sure you test all permutations of flip_axes before you submit this for review.

tuxemon/states/combat/combat.py
AndyMender: That would make sense, especially since it's an extra flag. Changed!

tuxemon/technique.py
AndyMender: Done.

AndyMender:
> > If we did anything, we'd play Firelion front for a technique targeting the player's monster and Firelion back for one targeting the opponent's monster, but that might be getting too involved.
> 
> I think this would be pretty easy to implement. We could have a json entry called "enemy_animation". In the code we would use this if it's present, otherwise use flip_axes.

That would be easy to implement, yes. However, it's adding extra complexity for the sake of 2 animations of which only firelion_front is currently used.

> Andy, let's save this for another ticket. If you agree with this idea, make a note on the original issue, please. This PR should not flip the front/back animations.

Yup, agreed. I will add a note about front/back animations to the issue.

tuxemon/db.py
```diff
@@ -335,6 +335,10 @@ class TechniqueModel(BaseModel):
     effects: Sequence[TechniqueEffect] = Field(
         ..., description="Effects this technique uses"
     )
+    filp_axes: Literal["x", "y", "xy"] = Field(
```
kerizane: Typo

AndyMender: Fixed

kerizane (approved):

kerizane:
Great work, thanks.

AndyMender:
Thanks to you and @Sanglorian as well! :)

PR #1354, opened 2022-09-24
JSONs gender

JaskRendix:
Genders for npcs and monsters.
This PR is for creating the data structure JSONs.

My question is: do we need to use the same field "gender" for both? Is it ok?

If not we could use "gender" for monsters and "identity" for NPCs as well as "genus".

If someone notices a wrong gender, then feel free to signal.

Some interesting numbers:
**create_npc** generates **213 NPCs with a JSON** and **12 NPCs without a JSON** (old maps) and there are _23 NPCs with a JSON that have never been created in a script_.

12: cfanatic1, cfanatic2, danteinstore, npc_aeble, npc_allie, npc_liela, npc_maples_mom, Rival_Dad, shopkeeperpapertown, Student1, Student2, Tuxemon_Collector

23: karrianna, penguin (!), professor-cardiling, professor-rockitten, professor-tweesher, red (npc_red), spyder_dragonscave_daeny, spyder_dryadsgrove_ferris, spyder_magician (!), spyder_nimrod_mace, spyder_professor (!), spyder_riverboatcaptainpapertown (!), spyder_wayfarer1_bismuth, spyder_wayfarer1_ratcher, xerogrund2, xerogrunt10, xerogrunt3, xerogrunt4, xerogrunt5, xerogrunt6, xerogrunt7, xerogrunt8, xerogrunt9
(!) no translation, only JSON

PR #1360, opened 2022-09-27
set_gender

JaskRendix:
Following #1354

**(1) integrate set_gender inside the add_monster function (the main one in npc.py) --> draft is based on this**

Files:
- addition of (M), (F) and (N) close the name in the menu (screenshot);
- addition of (M), (F) and (N) close the name in the combat menu (screenshot);
- gender among SIMPLE_PERSISTANCE_ATTRIBUTES, it saves correctly, there are no problems with old versions;
- inserted line in evolve, because if not there was the risk of changing gender after the evolution;
- inserted monster.possible_genders = npc_monster_details.gender so there is the correct display of the adversary gender (trainer battles);
- fixed Nudiflot male and female because of Nudiflot (M) (M) or (F) (F);
- make validate no problems

-> I tried to use ♂, etc, but I got **?** as result (symbol 1 character, (N) 3 characters)
chr(9794)  # ♂ unicodedata.lookup("MALE SIGN")
chr(9792)  # ♀unicodedata.lookup("FEMALE SIGN") 
chr(9906)  # ⚲ unicodedata.lookup("NEUTER") 
It looks like the font doesn't support the symbol.

Monster menu
Screenshot from 2022-09-28 21-28-27
Battle menu
Screenshot from 2022-09-28 10-56-36

### DIFF
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -63,6 +63,7 @@ SIMPLE_PERSISTANCE_ATTRIBUTES = (
     "status",
     "total_experience",
     "flairs",
+    "gender",
 )
 
 SHAPES = {
@@ -338,7 +339,7 @@ class Monster:
         self.txmn_id = results.txmn_id
         self.height = results.height
         self.weight = results.weight
-        self.possible_genders = results.possible_genders
+        self.gender = random.choice(list(results.possible_genders))
         self.catch_rate = (
             results.catch_rate
             or TuxemonConfig().default_monster_catch_rate
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -836,7 +836,7 @@ class NPC(Entity[NPCState]):
             )
             monster.set_level(monster.level)
             monster.current_hp = monster.hp
-            monster.possible_genders = npc_monster_details.gender
+            monster.gender = npc_monster_details.gender
 
             # Add our monster to the NPC's party
             self.add_monster(monster)
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -778,6 +778,7 @@ class NPC(Entity[NPCState]):
         new_monster.current_hp = min(old_monster.current_hp, new_monster.hp)
         new_monster.moves = old_monster.moves
         new_monster.instance_id = old_monster.instance_id
+        new_monster.gender = old_monster.gender
         self.remove_monster(old_monster)
         self.add_monster(new_monster, slot)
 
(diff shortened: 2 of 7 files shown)


tuxemon/monster.py
```diff
@@ -512,6 +513,25 @@ def set_level(self, level: int = 5) -> None:
         self.total_experience = self.experience_required()
         self.set_stats()
 
+    def set_gender(self, monster: Monster) -> None:
+        """
+        Set monster gender.
+
+        """
+        if monster.possible_genders == "male":
+            return
```
kerizane, line 522: This is weird.

tuxemon/npc.py
```diff
@@ -778,6 +779,7 @@ def evolve_monster(self, old_monster: Monster, evolution: str) -> None:
         new_monster.current_hp = min(old_monster.current_hp, new_monster.hp)
         new_monster.moves = old_monster.moves
         new_monster.instance_id = old_monster.instance_id
+        new_monster.possible_genders = old_monster.possible_genders
```
kerizane: Just calculate the gender here and set it to an attribute called gender.

JaskRendix: Do you mean set_gender instead of this?

kerizane: Didn't read this properly.

It doesn't make sense for possible_genders to sometimes be the possible genders, and to sometimes be the actual gender.

When we are loading a monster from db, do the following:
`monster.gender = random.choice(list(model.possible_genders))`    

\# note the object may not actually be called model

JaskRendix: I added
`self.gender = random.choice(list(results.possible_genders))`

JaskRendix: I didn't do other editing because I want to be sure about the following steps.

I guess I need to replace a bunch of things related to possible_genders

JaskRendix:
I know, I know, I said that I wanted to be sure. Is it correct? It wasn't necessary set_gender.

tuxemon/npc.py
```diff
@@ -835,6 +836,7 @@ def load_party(self) -> None:
             )
             monster.set_level(monster.level)
             monster.current_hp = monster.hp
+            monster.possible_genders = npc_monster_details.gender
```
kerizane:
```suggestion
            monster.gender = npc_monster_details.gender
```

kerizane (changes requested):

kerizane (approved):

PR #1363, opened 2022-09-29
new glyph font

JaskRendix:
@Sanglorian 

I added a few glyph to the font. Let me know if it's ok. So we can use: ♂ and ♀as neuter maybe ⚪

Some tests, in the original one there is a space, then I removed the space because it took precious space (characters)
Screenshot from 2022-09-29 13-25-28
Screenshot from 2022-09-30 12-45-21

Screenshot from 2022-09-29 13-26-15
Screenshot from 2022-09-30 12-42-15


kerizane:
This is great, well done!

We could just leave it blank for neuter.

Please add this to the credits.

kerizane:
Regarding the design, they might fit with the existing design better if they were slightly chunkier and/or less smooth.

The male symbol looks like it's too low in comparison with the normal text.

xirsoi:
Traditionally the symbol for neuter would be a circle with a line going down, like the female symbol but lacking the cross bar.
I like the additions! Thanks for the contribution!

JaskRendix:
> Traditionally the symbol for neuter would be a circle with a line going down, like the female symbol but lacking the cross bar.

Yeah, the U+26B2 ⚲, but at the moment I didn't find the right one, especially a font that shares the same unicode.

> Please add this to the credits.

No problem.

> Regarding the design, they might fit with the existing design better if they were slightly chunkier and/or less smooth.

I'll look into. Are you referring to all of them or only to male/female? Because some would be hard to convert.

> The male symbol looks like it's too low in comparison with the normal text.

I guess it's an impression, look at the measurement.

kerizane:
If you were to draw a horizontal line directly under the text I think both symbols would be partially under that line. For the male one it is more noticeable because it has more visual weight at the bottom.

kerizane:
Is it possible to add these symbols to a new file? This would make it easier to update the original fonts, and would also make it clearer that they are from a different source.

kerizane:
Did you make these or get them from somewhere? What is the program in the screenshot?

kerizane:
> I guess it's an impression, look at the measurement (image).

If you look at the existing characters the same way, do they actually fill the entire box?

JaskRendix:
FontForge is the software

I took the stuff from 
https://fonts.google.com/noto/specimen/Noto+Emoji/about

but maybe this font is more similar to the actual:
https://fonts.google.com/specimen/DotGothic16/glyphs?query=gothic&noto.query=gothic&noto.lang=got_Goth&noto.continent=Europe&noto.script=Goth

both are open source. About the symbol, you're right, I need to move the symbol up.

JaskRendix:
I got rid of everything excluding male and female, because it doesn't match the style. I'll update credits too. Now both symbols are on the correct line. Source DotGothic

JaskRendix:
Honestly I don't know, maybe a PNG (24x24 or 20x20) for each gender is better than updating the font.

kerizane:
Please add new game screenshots of the symbols so we can compare them.

JaskRendix:
Updated the first post.

kerizane:
The first one is better.

JaskRendix:
ok, I'll update

CREDITS.md
```diff
@@ -1268,6 +1268,8 @@ Fonts
 
 * ["Press Start 2P"](https://www.google.com/fonts/specimen/Press+Start+2P) by 
 [codeman38](mailto:cody@zone38.net) is licensed under [SIL Open Font License, 1.1](http://scripts.sil.org/OFL)
+adapted by JaskRendix, added symbols from DotGothic16 font (https://fonts.google.com/specimen/DotGothic16/)
+licensed under [SIL Open Font License, 1.1](http://scripts.sil.org/OFL)
```
kerizane, line 1272: @Sanglorian do you think this credit is ok?

JaskRendix: Link was wrong. I updated. Of course I'm ready to change the format if necessary.

xirsoi (approved):
Thanks for your contribution!

Issue #937, opened 2021-08-11
Refactor Technique class to support loadable modules


bitcraft:
Technique class currently defines status effect code inside of the Technique class.  This makes it impossible to load new status effects during runtime, and also means that a mod cannot defines its own unique statuses.

The Technique class should have the various status effects refactored into a new class.  This new class should follow the same system as Items and Events to allow the status effects to be loaded from a folder of python files.

PR #1365, opened 2022-10-01
Refactor Technique class

JaskRendix:
Following #1364 related to the refactor of the technique class (#937)

Issues:
~~(1) Now when a battle starts it shows me "pound" as techniques (self.name = "Pound") (screenshot)
Maybe we need to add a "for" in combat, but at the moment I don't know if it's a silly thing or not.
Or maybe getting rid of self.session in technique.py and fixing the parser, but at the moment I have no clue.~~

~~((2) If switching monster: crash -> see question "meta"~~(

A series of questions related to the refactoring:
1. self.category for techniques, it looks like has been created only for supporting the old system, I guess it can be deleted once the next system is applied. Is it correct?
2. self.conditions for techniques, it was planned in the beginning? If yes, then I have a PR ready to integrate the empty field inside the technique JSONs as well as in db.py.
3. what is "meta"? Do we need to create the meta.py? (as effect in status_faint and run)


tuxemon/technique/effects/status.py
```diff
@@ -0,0 +1,89 @@
...
+        already_applied = any(t for t in target.status if t.slug == slug)
+        success = (
+            not already_applied
+            and base.can_apply_status
+            and base.potency >= random.random(),
+        )
+        tech = None
+        if success:
+            if slug == "status_faint":
+                target.apply_status(base("status_faint"))
```
kerizane: Suspicious

JaskRendix: Agree, I'll update as lifeleech
```
tech = base(slug, carrier=target, link=self.user)
target.apply_status(tech)
```

JaskRendix: or maybe only target, but in that case we can remove the if about status_faint, what do you think?

kerizane: Surely it needs to be `Technique`, not `base`?

tuxemon/technique/effects/damage.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Returns:
+        Dict summarizing the result.
+
+    """
+
+    name = "damage"
+    param_class = DamageEffectParameters
+
+    def apply(self, target: Monster) -> DamageEffectResult:
```
kerizane: You need to pass the slug in somehow and load the actual technique.

kerizane: Same for all effects.

kerizane:
> (1) Now when a battle starts it shows me "pound" as techniques (self.name = "Pound") (screenshot)

This is because you're not loading the data for the technique that was used, you're loading the default data, which is Pound.

> (2) If switching monster: crash -> see question "meta"

meta is for techniques that aren't strictly techniques. I don't think we need meta.py, just need the code to not crash.




> * self.category for techniques, it looks like has been created only for supporting the old system, I guess it can be deleted once the next system is applied. Is it correct?

yeah, delete it


> * self.conditions for techniques, it was planned in the beginning? If yes, then I have a PR ready to integrate the empty field inside the technique JSONs as well as in db.py.

I think these will be useful later, but will need some refactoring. I will decide what to do with them later.

JaskRendix:
Moves are visible, no crash in swapping, status_faint works (I got teleported by losing against a wild monster).
I didn't touch category at the moment, I'll do in another PR.
I removed the two "meta" from run and status_faint.

kerizane:
Please test every status.

JaskRendix:
So, this is the situation:
1. **poison** and **faint** and **recover** work (but if the statuses are active -> set monster status), there is no way of getting poisoned (points 4/5).
2. **status_overfeed** doesn't exist as JSON, but there is **simple_overfeed** in formula.py, so I guess someone wanted to create it, but the job isn't complete (like hardshell). Do I need to create it?
3. I assigned the **status_lifeleech** to a monster and I went in a battle, it crashed:
```
Tuxemon/tuxemon/formula.py", line 195, in simple_lifeleech
    damage = min(target.hp // 2, target.current_hp, user.hp - user.current_hp)
AttributeError: 'NoneType' object has no attribute 'hp'
```
I think it's the user in lifeleech.py, but I haven't looked into at the moment.

4. There is no way of getting **status_recover**, **status_overfeed** or **status_lifeleech**. There is no technique with these effects as well as **recover**, **lifeleech** and **overfeed**.
5. As previously said (4), look at this:

Do I need to add **status status_poison**? Because right now (before too) there is/was no way of getting poisoned during the battle.

6. we need to move a bit to the right the symbol and reduce the space between these (screenshots).
```
    set_monster_status 0,status_poison
    set_monster_status 0,status_recover
    set_monster_status 0,status_lifeleech
```

7. "Faint" in status_faint? Inside the JSON status_faint there is "Faint" where normally there is wood, metal, etc. I guess it's a mistake. I didn't find trace about this type.

Other **status_XXX** don't have the field types, while **run** has **types": []**. Do I need to standardize everything with **types": []**?

9. There is a bool called **can_apply_status**, it appears only in 4 places (overfeed json - set on true -, technique.py - set on false -, damage.py and status.py - last two as success condition -); I would like to remove the field, I guess it was inserted only for overfeed, but by creating the status_overfeed JSON we can standardize and remove a condition that can impact other statuses (since it's in damage.py and status.py).

kerizane:
I am pretty sure poison used to work, I remember getting poisoned by cataspike.

kerizane:
> status_overfeed doesn't exist as JSON, but there is simple_overfeed in formula.py, so I guess someone wanted to create it, but the job isn't complete (like hardshell). Do I need to create it?

Confirm the behaviour of "overfeed" (technique owned by aardart) hasn't changed. If it hasn't then create a ticket to investigate overfeed.

> There is no way of getting status_recover, status_overfeed or status_lifeleech. There is no technique with these effects as well as recover, lifeleech and overfeed.

Create a ticket.

> we need to move a bit to the right the symbol and reduce the space between these (screenshots).
    set_monster_status 0,status_poison
    set_monster_status 0,status_recover
    set_monster_status 0,status_lifeleech

Create a ticket if this is new.

> "Faint" in status_faint? 

Remove.

> Do I need to standardize everything with types": []?

Just remove types from run.

>There is a bool called can_apply_status

This logic is used as part of setting poison. Remove from overfeed.json leave it in everywhere else.

mods/tuxemon/db/technique/status_poison.json
```diff
@@ -10,7 +10,8 @@
   "category": "poison",
   "sort": "meta",
   "effects": [
-    "status"
+    "poison",
+    "status status_poison"
```
kerizane: `"status status_poison"`
This shouldn't be here, it should be on the techniques "suck_poison" and "venemous_tentacle".

kerizane (changes requested):

mods/tuxemon/db/technique/status_poison.json
kerizane: Check all the techniques that should apply a status.

JaskRendix: Yeah, you know, that's what I wanted to say. Thanks for repeating it.

kerizane: Also, you should remove "poison" from those techniques.

JaskRendix:
Removed faint.
Removed types from run
Lifeleech still crashes, I need to look into.

The space between symbols isn't new, it happens on development too.
from
Screenshot from 2022-10-02 09-42-33
to
Screenshot from 2022-10-02 09-43-23

Removed can_apply_status from overfeed (screenshot where it remains)

kerizane:
Is lifeleech the only thing left?

JaskRendix:
Moved some parts in the #1372 since it happens the same thing in development.

kerizane:
Is lifeleech fixed now?

Remove the conditions stuff from this PR, but leave it in another branch in your repo in case we want it later.

JaskRendix:
No, it's not fixed lifeleech, if you have a clue about the user. It could be something silly, but I'm unable to see right now.

JaskRendix:
Purged conditions done. Now I have a branch with conditions (included the one for the technique JSONs).

kerizane:
Fixed

kerizane:
I think that's it. Want me to merge?

JaskRendix:
Sure, let's merge, then tell me if you want the draft about conditions

kerizane (approved):

PR #1369, opened 2022-10-02
fix moves

JaskRendix:
I was working on #1365 and I needed to test overfeed. So I used **add_monster aardorn,18**, because the level_learned is 11 for overfeed, but I noticed the following behavior. This PR is only for the actions, I don't know if we need to apply it somewhere else or maybe create a separate function.

Before. Why? because of the **load_from_db**, since the level is **random.randint(2, 5)**, it gives you the moves in this range, so 1 or 2 moves (it depends if you're lucky in the randint).
Screenshot from 2022-10-02 11-55-04

After the PR, since the level is 18:
Screenshot from 2022-10-02 11-43-17

### DIFF
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -512,6 +512,12 @@ class Monster:
         self.total_experience = self.experience_required()
         self.set_stats()
 
+        # Update moves
+        self.moves = []
+        for move in self.moveset:
+            if move not in self.moves and move.level_learned <= level:                
+                self.learn(Technique(move.technique))
+
     def experience_required(self, level_ofs: int = 0) -> int:
         """
         Gets the experience requirement for the given level.
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -513,7 +513,6 @@ class Monster:
         self.set_stats()
 
         # Update moves
-        self.moves = []
         for move in self.moveset:
             if move not in self.moves and move.level_learned <= level:                
                 self.learn(Technique(move.technique))


tuxemon/event/actions/add_monster.py
```diff
@@ -75,4 +76,11 @@ def start(self) -> None:
         current_monster.set_level(monster_level)
         current_monster.current_hp = current_monster.hp
 
+        # add moves related to the level
+        current_monster.moves = []
```
kerizane: Should be in set_level (I thought it already was...)

JaskRendix: No problem, I'll update

kerizane:
I'll need to double-check the existing code before merging this.

tuxemon/monster.py
```diff
@@ -512,6 +512,13 @@ def set_level(self, level: int = 5) -> None:
         self.total_experience = self.experience_required()
         self.set_stats()
 
+        # updates moves related to the level
```
kerizane: I'd prefer this
```
        for move in self.moveset:
            if move not in self.moves and move.level_learned <= level:                
```

kerizane (changes requested):

tuxemon/monster.py
JaskRendix: It sounds good

tuxemon/monster.py
```diff
@@ -512,6 +512,12 @@ def set_level(self, level: int = 5) -> None:
         self.total_experience = self.experience_required()
         self.set_stats()
 
+        # Update moves
+        self.moves = []
```
kerizane:
```suggestion
```

kerizane: This is called in a few other places. As there it is possible to learn new techniques via an action, I don't want to risk overwriting them.

kerizane (changes requested):

kerizane (approved):

PR #1373, opened 2022-10-04
draft maps folder

JaskRendix:
Following my first attempt Xero maps #1258
After the following discussion Remove unused maps #1257

This draft creates three subfolders (spyder, unused and xero), and moves the maps inside the respective folders.

maps
-------spyder
-------unused
-------xero

No problems with old savegames as well as with new ones.

- edited all the paths in the maps, so there are no problems / crashes with loading tilesets;
- all maps can be opened in Tiled;
- created maps folder and maps.json in db folder;
- maps.json lists all the spyder and xeno maps with the path (eg. spyder/spyder_bedroom.tmx);

Feel free to test, if it's not ok, then I close the PR, no problem


kerizane:
Can we store the folder we should be looking in somewhere, and only look in this folder?

JaskRendix:
We can consider to split maps.json into spyder.json and xero.json , so it can iterate through the maps of the scenario.

kerizane:
I'd rather not have the JSON file, as this is an extra thing that people will have to maintain.

If my idea is too difficult then you could build map.json as a dict internally (by iterating over the files in the folder). Raise an exception if there is a name clash. You don't need to output map.json as a file.

JaskRendix:
It makes sense, less files, less things to maintain. I think I'm going to close the PR for the moment. Just a question about generate the dict internally by iterating. Where is the right location to put the dictionary/function?

PR #1374, opened 2022-10-05
condition has_tech

JaskRendix:
I had this for a while among my branches.
I developed with the idea that it can help with swim (terrains), but recently I realized that maybe this isn't the right direction to solve the issue.
More, since we have already the has_item, it can be handy for modders.

It gives you true if a monster (in the party) has the technique.
If it's not useful, then no big deal.


Sanglorian:
I can't think of an immediate use I'd have for it, but it sounds useful for modders. I'm using items to get past obstacles but someone else might use techniques instead.

tuxemon/event/conditions/has_tech.py
```diff
@@ -0,0 +1,63 @@
...
+            condition: The map condition object.
+
+        Returns:
+            Whether the player has a technique in his party.
+
+        """
+        player = session.player
+        tech = condition.parameters[0]
+        for monster in player.monsters:
+            for moves in monster.moveset:
```
kerizane: Isn't moveset the techniques that you can learn?

JaskRendix: Nice spot, updated

xirsoi:
Please confirm that you have built a test scenario for this condition and tested it thoroughly.

JaskRendix:
The recent change to moves needs to adjust a thing.

JaskRendix:
Everything seems to work fine. Tested different times, for instance:

set **is has_tech stampede** to a condition of a NPC1 (stampede is lv8 move)
load the game, NPC1 doesn't appear
add **add_monster aardorn,5** by talking to NPC2
talk with NPC2, NPC1 doesn't appear.
add **add_monster aardorn,8** by talking to NPC2
talk with NPC2, NPC1 appears.

xirsoi (approved):
LG;TFPR

PR #1375, opened 2022-10-05
cannot escape wild monster

JaskRendix:
I had this for a while among my branches.
I developed with the idea of impeding players to run from wild monsters.
If the _level of the party_ is less than the _level of the wild monster_, then you can't run because "you're weak".

I'm not sure about this, because after starting working on techniques (refactoring) I bumped into the concept of conditions as well as this https://wiki.tuxemon.org/Category:Condition

Again, if it's not useful, then no big deal.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -123,6 +123,9 @@ msgstr "It missed..."
 msgid "combat_can't_run_from_trainer"
 msgstr "Can't run from trainer battles!"
 
+msgid "combat_cannot_escape"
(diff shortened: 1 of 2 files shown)


Sanglorian:
How does running away work at the moment? It makes sense for it to be harder to run away from some monsters, but it shouldn't be a binary: there should be a chance of running away regardless (so players can attempt to run over multiple turns).

Might be worth discussing in Discord so we can get group consensus on how it should work.

JaskRendix:
Right now you can run away always from wild monster, there is no restriction. Just for info:

Generation I and II:
This is an image
Generation III and IV:
This is an image
https://bulbapedia.bulbagarden.net/wiki/Escape

If in Discord do you agree about a formula, then let me know, so we can put it in formula.py

PR #1376, opened 2022-10-05
replace gender with symbols

JaskRendix:
Following #1363 
(M) -> ♂
(F) -> ♀
(N) ->⚲
~~In the neuter case, users will get the ?, but at the end since it's neuter, it makes sense.
If someone has another symbol (Greek letter, etc), then fell free to suggest.~~
Updated the font with the neuter symbol, similar to the female, but without the cross in lower part. So we can use the three symbols without problems. 

Screenshot from 2022-10-05 19-42-40


kerizane:
We don't necessarily need a neuter symbols, the Pokémon games just leave it blank in this case.

Could you provide screenshots of all three?

JaskRendix:
Sure, updated first post

kerizane:
The neuter one is quite similar to the female one, and I'm not sure how many people actually know what it means. Leave the glyph in the font pack, but don't use it.

kerizane:
M and f look good, thanks.

JaskRendix:
Done

kerizane (approved):

Issue #1014, opened 2021-09-28
Maybe remove hard-coded name selection screen when starting new game


ultidonki:
**Is your feature request related to a problem? Please describe.**
Currently the player name input screen is always the first thing that appears when starting any mod. 
If modders want a unique starting point, they can't remove this.

**Describe the solution you'd like**
The hard-coded player rename screen would be removed, and modders can add this themselves using rename_player if they really want to use it. 

**Describe alternatives you've considered**
Modders can use the rename_player action instead

**Additional context**
<s>Currently, the name input screen is badly behaved, and map events will immediately start playing 'underneath' the screen, eg. playing music or showing dialogue boxes (these actually appear *in front of* the name input screen and need to be continued before you can actually input a name). 
I've got around this in my mod by not playing music or showing dialog until the player moves. </s>
(This is now fixed)

bitcraft:
This is a good idea.  Perhaps we could add an action list to a mod's metadata that is run on a new game.

andrewhong04:
> Currently, the name input screen is badly behaved, and map events will immediately start playing 'underneath' the screen

I would like to point out that my recent pull request that got merged fixes the behavior of the start menu. Instead of map events playing underneath the screen, the player needs to input their name, and then the map events will start loading in.

ultidonki:
> I would like to point out that my recent pull request that got merged fixes the behavior of the start menu. Instead of map events playing underneath the screen, the player needs to input their name, and then the map events will start loading in.

That's awesome, thanks for letting me know! 

That'll make the start of my mod a bit nicer. I've edited the original issue to remove that point. Thanks again! :)

kerizane:
Hardcoded menu is here: https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/states/start/__init__.py

JaskRendix:
@ultidonki it has been removed, let me know if it's ok or you need something else.

PR #524, opened 2019-04-09
Added Don't Care button to generate random name for each locale

silvercam15:
As mentioned in Issue #473,  
> Adding default names in each locale would be a nice feature. In earthbound for snes, you could also pick a random name from a curated set( "Don't Care")....which might also be nice.

We added default names in each locale. Please see the screenshot below. 
imageedit_6_2399927701
Upon clicking the "Don't Care" button, a random name is generated (as seen in the screenshot above). The user can cycle through the list of random names by continuously pressing the "Don't Care" button until they land on one they like and can then select "END" to start the game. 
If you have any comments, suggestions, or concerns, please feel free to let us know! Thanks for letting us contribute!

PR #1377, opened 2022-10-05
remove hard-coded name selection

JaskRendix:
as requested in #1014 
- the PR removes the hard-coded input at the beginning of the game;
- the default name is Red ("player_npc", "npc_red") both Spyder and Xero;
- to help modders, now there is **assign_name <slug>** action, so they can assign the protagonist name in case Red isn't the right one, especially if there is a precise plot and story-line;

Now I don't know, maybe someone has the intention of inserting the classic discussion "Are you called Red?" then the "Yes or No" reply, then the **rename_player** action. I can implement if it's necessary.

This will help the ones - #524 and me - that don't care much about the name.
By they way, I tried a couple of times starting the game and it's so fast without the name selection.
In medias res starting point!

### DIFF
--- a/tuxemon/event/actions/assign_name.py
+++ b/tuxemon/event/actions/assign_name.py
@@ -1,0 +1,53 @@
+#
+# Tuxemon
+# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
+#                         Benjamin Bean <superman2k5@gmail.com>
+#
+# This file is part of Tuxemon
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


Sanglorian:
Eventually I would like the rename_player action for Spyder - ideally as part of the sequence described here: https://wiki.tuxemon.org/The_Spyder_in_the_Cathedral#Introduction

But in the short term I've got no objection to this approach.

JaskRendix:
Ah ok, I didn't know that Spyder beginning was like this.
We are in the old release now, so we need to update it to the new release by introducing the "introduction" in the long-term. So everything needs to be reworked.

Issue #1147, opened 2022-04-04
Allow teaching tuxemon new techniques from a Map Event


ultidonki:
**Is your feature request related to a problem? Please describe.**
You can currently teach tuxemon new techniques (if they're the right type of tuxemon, eg. water/fire/...) by using an item action.
It would be nice to be able to do it from a Map Event too. 

**Describe the solution you'd like**
A new map event along the lines of:
`teach_tuxemon_technique <tuxemon_id>,<technique_slug>`
Where tuxemon_id is not a tuxemon slug, but the tuxemon's unique ID (you can get it from the action `get_player_monster`)

**Describe alternatives you've considered**
Maybe the map event would just open the same screen where a 'Technique' item can be used? ie. you can see the current tuxemon you have, and can choose one just like when using an item with the 'learn' effect? 

**Additional context**
See the item action to learn a new technique: tuxemon/items/effects/learn.py
And a map event that uses the tuxemon ID to do an action on that monster: tuxemon/event/actions/store_monster.py
And the definition of an item that uses the Learn technique item action: mods/tuxemon/db/items/tm_avalanche.json

JaskRendix:
@ultidonki it has been implemented with #1664 you can close the issue.

PR #1378, opened 2022-10-05
teach technique

JaskRendix:
as requested in #1147
action that allows to teach techniques to monsters (utilizes instance_id)


tuxemon/event/actions/teach.py
```diff
@@ -0,0 +1,74 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TeachActionParameters(NamedTuple):
+    monster_id: str
+    technique: str
+
+
+# noinspection PyAttributeOutsideInit
```
kerizane: ?

JaskRendix: It was part of the file I utilized. I'm going to remove it.

This string there is also in:
clear_variable
copy_variable
get_player_monster
spawn_monster
store_monster

Do you want that I open another PR to remove it from all?

kerizane:
Seems a bit awkward to use.

Maybe we should reorder the args and make monster_id optional? If you don't provide it then it automatically triggers the selection menu for you.

JaskRendix:
Yeah, it's true, it's a bit awkward. You need a **get_player_monster variable**, so it selects the monster, then **teach variable,tech**.

What do you think now? There is no more instance_id, it opens the menu, so you can choose to who teach the technique.

xirsoi:
Please create an event that uses this code to demonstrate it works correctly.

JaskRendix:
And how am I supposed to do it? By screenshots?

JaskRendix:
As requested, added amnesia to rockitten
`<property name="act2" value="teach amnesia"/>`

it opened the menu with the monster in the party, then selected rockitten. First random encounter, as follow:

Screenshot from 2022-10-06 20-04-44

or
```
    <property name="act2" value="teach amnesia"/>
    <property name="act3" value="teach bullet"/>
    <property name="act4" value="teach flamethrower"/>
    <property name="act5" value="teach electrical_storm"/>
    <property name="act6" value="teach energy_field"/>
    <property name="act7" value="teach feint"/>
    <property name="act8" value="teach eyebite"/>
```
Screenshot from 2022-10-06 20-17-14


from rockitten JSON

```
    "slug": "rockitten",
    "category": "cute_boulder",
    "ai": "RandomAI",
    "moveset": [
        {
            "level_learned": 2,
            "technique": "ram"
        },
        {
            "level_learned": 2,
            "technique": "boulder"
        },
        {
            "level_learned": 2,
            "technique": "thunderball"
        }
    ],
```
* screenshots show (M) and (F) because it's an old version

The max number of techniques to show (without something outside the screen) is 11.
10 is the perfect number.

xirsoi:
I think the change to automatically pull up the monster select is worse than the original way, it removes control from the script and I want to give power to the modding community, not take it away.

kerizane:
What about if we made it Optional like I requested?

JaskRendix:
Thanks for the following consideration: _this it removes control from the script and I want to give power to the modding community, not take it away_

> What about if we made it Optional like I requested?

I'm considering it in the next version.

kerizane:
Actually, one issue with launching the menu is that there is no way to filter it (e.g. techniques taught by items can have conditions applied which stop you from selecting other tuxemon).

Not sure if `get_player_monster` supports this, but the way to go would probably be to implement it there, rather than adding extra complexity to this action.

Sorry for the confusion, but maybe it's best to stick with only supporting the UUID. @xirsoi ?

In the future it might be useful to have an action which triggers an item (either an actual item in your bag, or using the item as a template of the behaviour we want), but I think this is out of scope for this ticket and a different use case.

JaskRendix:
No worries, optional gave me the following idea. Instead of the menu, we are going to assign the move randomly to a monster in the party. So the modder can set up the action without thinking too much.

So, if the modder wants to make the player choose:
- get_player_monster variable
- teach variable,tech

_The possible scenario: the player wins a tournament, the organizer wants to gift him or her: "I'm going to teach WTR, ultra-rare technique to one of your monster. WTR is a spinning kick, where the kicker makes a complete spin before delivering a blow. Which one?" then you choose._ 

If the modders doesn't care about which monster:
- teach ,tech

_The possible scenario: the player wins a tournament, the organizer wants to gift him or her: "I have taught WTR  ultra-rare technique to your worthiest monster. WTR is a spinning kick, where the kicker makes a complete spin before delivering a blow."._ 

Possible variation:
**teach random,tech** instead of **teach ,tech**

JaskRendix:
I'm going to close the PR. I merged both.

It continues in #1382

PR #1385, opened 2022-10-07
Move EventAction parameter attributes into Action class

kerizane:
It was annoying me that we have the attributes nested within another attribute. The parameter namedtuple has been removed, the attributes are now on the class. We still have automatic type-casting.

Optional values now need to be explicitly defaulted to None. I can try to bring back automatic defaulting to None if necessary.

I tried using pydantic, rather than dataclasses, but it was a bit too awkward and opinionated.

Also implemented:
A script which performs basic validation on every action in the game.
A couple of minor script fixes discovered by the above.
Fix dialog + dialog chain: the action would fail crash if you had too many commas in the text (I know they're deprecated, but why not).
Support for "Literal" typehint.

### DIFF
--- a/scripts/test_actions.py
+++ b/scripts/test_actions.py
@@ -1,0 +1,28 @@
+"""
+Script to validate map actions
+"""
+from pathlib import Path
+from unittest.mock import MagicMock
+
+from tuxemon.constants import paths
+from tuxemon.db import db
+from tuxemon.event.eventengine import EventEngine
+from tuxemon.map_loader import TMXMapLoader
+from tuxemon.prepare import CONFIG
+from tuxemon.session import Session
+
+
+db.load("monster")
+engine = EventEngine(Session(None, None, None))
+loader = TMXMapLoader()
+loader.image_loader = MagicMock()
+
+for mod_name in CONFIG.mods:
+    for path in (
+        Path().joinpath(paths.mods_folder, mod_name, "maps").glob("*.tmx")
+    ):
+        txmn_map = loader.load(str(path))
+        for event in txmn_map.events:
+            for act in event.acts:
+                if not engine.get_action(act.type, act.parameters):
+                    print(f"{path} failed")
--- a/tuxemon/event/actions/battles_print.py
+++ b/tuxemon/event/actions/battles_print.py
@@ -22,19 +22,17 @@
 from __future__ import annotations
 
 import logging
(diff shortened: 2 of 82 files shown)


kerizane:
@vnmabus you might be interested in this one

tuxemon/event/actions/battles_print.py
```diff
@@ -52,12 +50,12 @@ class BattlesAction(EventAction[BattlesPrintActionParameters]):
     """
 
     name = "battles_print"
-    param_class = BattlesPrintActionParameters
+    variable: Optional[str]
```
kerizane: Fixme

kerizane:
I think it might be better if we added a factory classmethod, rather than having the logic in init.

JaskRendix:
I've tested it in the past week and it goes very well, no problems.
Can I try to apply the same method to items effects/conditions (then techniques)?

kerizane:
Go ahead

vnmabus:
This has even more magic that the old way XD. But it gets rid of that unnecessary repetition of the parameters class name, and it reduces the typing and code complexity, so I really like it.

kerizane:
I think it would be slightly less magic if I used a factory classmethod, rather than casting the values within the post-init. I don't think this is urgent, though.

kerizane:
Do you think I should preserve the automatic default of None for optional types? It's a bit magic, and I don't think it's that necessary.

scripts/test_actions.py
```diff
@@ -0,0 +1,28 @@
...
+
+for mod_name in CONFIG.mods:
+    for path in (
+        Path().joinpath(paths.mods_folder, mod_name, "maps").glob("*.tmx")
+    ):
+        txmn_map = loader.load(str(path))
+        for event in txmn_map.events:
+            for act in event.acts:
+                if not engine.get_action(act.type, act.parameters):
+                    print(f"{path} failed")
```
kerizane, line 28: Maybe this should be a real unittest?

kerizane: Not urgent though

kerizane:
@vnmabus Happy for this to be merged?

vnmabus:
I don't really have time now for a full review. I like the idea, and @JaskRendix tested it seemingly without problems. I think we can merge now, and fix the issues that arise later.

bitcraft (approved):

PR #1390, opened 2022-10-09
fix double move

JaskRendix:
I was doing some testing and I needed some monsters.
add_monster and random_monster keep trying to sneak the moves two times - screenshot.
The cause is the randint 2-5 of load_from_db (monster.py).

The PR addresses the issue, the fix is applied only in add_monster and random_monster.
I'm open to discuss other fixes if this is considered too much.

Screenshot from 2022-10-09 15-09-02

### DIFF
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -514,7 +514,10 @@ class Monster:
 
         # Update moves
         for move in self.moveset:
-            if move not in self.moves and move.level_learned <= level:
+            if (
+                move.technique not in (m.slug for m in self.moves)
+                and move.level_learned <= level
+            ):
                 self.learn(Technique(move.technique))
 
     def experience_required(self, level_ofs: int = 0) -> int:


kerizane:
Can you modify the method that teaches techniques so that it checks if you know it before teaching it again?

JaskRendix:
This one in set_level?
```
        # Update moves
        for move in self.moveset:
            if move not in self.moves and move.level_learned <= level:
```

kerizane:
Maybe this bit is failing? `move not in self.moves`

JaskRendix:
I think so, I tested without and it works without problems (still getting double move). I removed forget for the moment.

tuxemon/monster.py
```diff
@@ -528,7 +514,7 @@ def set_level(self, level: int = 5) -> None:
 
         # Update moves
         for move in self.moveset:
-            if move not in self.moves and move.level_learned <= level:
```
kerizane, line 517: My point was that `move not in self.moves` is meant to stop you from learning moves you already know. It's possible that python isn't realising that they are the same move, because we don't have object equality defined properly. So, you might need to do:

`if move.slug not in (m.slug for m in self.moves) and move.level_learned <= level`

JaskRendix: It was easier than I thought. I was thinking about removing duplicates or getting common elements inside the two lists. Thanks

JaskRendix:
Done, the only change, it wasn't move.slug, but move.technique

kerizane:
Did you run black?

JaskRendix:
Done

kerizane (approved):

Issue #1391, opened 2022-10-11
Aardorn bug


JaskRendix:
**Describe the bug**
Weird bug. You're obliged to quit the game because there are no more moves available.

**To Reproduce**
1. add aardorn (from lv 1 to lv7);
2. go into battle (trainer battle) with only aardorn or with the others fainted;
3. See screenshot;

![Screenshot from 2022-10-10 10-35-15](https://user-images.githubusercontent.com/64643719/195073466-c71ca003-e6a1-4361-baea-00aa64efc32a.png)

wild monster fight -> you can run
trainer fight -> you're stuck, you need to quit the game

**Expected behavior**
not being forced to quit the game

**Cause**
it's because of the recharge of the moves

**Desktop (please complete the following information):**
- OS: Ubuntu 22
- Python 3.10.6

Any ideas? 
we can add the object to restore moves;
we can add a move;
anything else?

kerizane:
I think if you are out of usable moves then you should just be able to skip your turn.

Add an additional technique for this, but only when you don't have any other moves. You should still be able to use an item or switch monster of you want to. Make sure that the enemy is unable to get into this same bad state.

Sanglorian:
Because moves refresh, it's actually okay if the enemy also gets into this state. But otherwise I agree with you. 

The only problem I see is a trainer battle where neither side can retreat, if it gets to a point where neither team has any monster that can do damage. You could get stuck in a loop, even with access to moves. But that's a separate issue - maybe trainer battles should have "Forfeit" where "Run" would appear in a wild battle.

kerizane:
Yeah, I meant make sure that the enemy can also skip their turn.

JaskRendix:
> Because moves refresh, it's actually okay if the enemy also gets into this state

The enemy doesn't get into this state. I tested by replacing Dollfin with Aardorn lv 5 (spyder_route2_billie), but nothing happens. Aardorn used 5 times in a row Tonguespear.

Do we need to address this?

> You should still be able to use an item or switch monster of you want to. 

Now you can access objects, you can use too, this happens in two cases:
- everyone else is fainted in the party (so it's not possible to switch)
- Aardorn is the only monster in the party (so it's not possible to switch)

So a Skip or "Forfeit" instead of "Run"?

JaskRendix:
I tested and NPCs aren't affected. An Aardorn was able to use Tonguespear 5 times in a row.
So the chance the NPCs retreat are equal to 0.

[...]

PR #1395, opened 2022-10-13
forfeit and escape

JaskRendix:
PR addresses the following points:
1. added forfeit (as suggested here #1391);
2. added escape formula (as discussed #1375);

**1** as suggested, it'is only a change of label; the idea is sending the player back to the last center (like status_faint), but at the moment the PR doesn't address this (thinking about it and how to implement it).

**2** since I was operating on forfeit, I took the chance to implement the escape formula;

Everything works.
I'm surprised about the variable run_attempts, from what I understood it keeps increasing without resetting, but it was requested.

A couple of screenshots:
Wild monster:
Screenshot from 2022-10-13 15-10-31
Trainer battle:
Screenshot from 2022-10-13 15-11-35


Sanglorian:
Nice work!

I think we should avoid using the same formula as Pokemon just for safety.  On Discord I suggested chance of run success could be = 

40% + (15% * (number of consecutive previous attempts + level of player's current monster - level of wild monster))

But I'm not fussy.

For forfeiting, I think it should take you back to the last healing centre and heal your team, exactly as if you'd been defeated in battle.

Thanks again!

JaskRendix:
Interesting formula, I'll update the one.
About forfeiting, I like your idea!

JaskRendix:
40% + (15% * (number of consecutive previous attempts + level of player's current monster - level of wild monster))
if (formula) >= 0: (yes, run)
```
att. lv us  lv tg               formula
1	5	7		0.25 -> run
data not relevant anymore
```
let me know if it looks good, with this method, if the difference of levels is more than 5 (player vs wild monster), then you can't run.

Sanglorian:
Thanks for that, just a couple of things: 

1. Escape should be the chance of escape, not automatic escape when greater than 0. 

Something like: 

    escape = 0.4 + (0.15 * (attempts + level_user - level_target))
    if ((random number between 0 and 1)) <= escape:
    return True

2. Attempts should increase by +1 _after_ success is checked, not before (i.e. the first attempt should use 0 rather than 1)

kerizane:
@Sanglorian please approve if you are happy with this

tuxemon/formula.py
```diff
@@ -202,3 +203,15 @@ def simple_overfeed(
 ) -> int:
     speed = target.speed // 2
     return speed
+
+
+def escape(level_user: int, level_target: int, run: int) -> int:
```
kerizane: This returns a book, not an int

tuxemon/formula.py
```diff
@@ -202,3 +203,15 @@ def simple_overfeed(
...
+    if run == 0:
```
kerizane: Could just be attempts = run, but why not just rename the "run" arg to "attempts"?

tuxemon/states/combat/combat.py
```diff
@@ -423,6 +425,7 @@ def transition_phase(self, phase: CombatPhase) -> None:
             self.players[0].set_party_status()
             var = self.players[0].game_variables
             var["battle_last_result"] = "ran"
+            var["run_attempts"] = +1
```
kerizane: Either += 1 or = 1

Not sure what you're trying to do here

tuxemon/states/combat/combat_menus.py
```diff
@@ -63,30 +63,70 @@ def startup(
...
         combat_state = self.client.get_state_by_name(CombatState)
-
-        if combat_state.is_trainer_battle:
+        player = combat_state.monsters_in_play[combat_state.players[0]][0]
+        target = combat_state.monsters_in_play[combat_state.players[1]][0]
+        var = combat_state.players[0].game_variables
+        var["run_attempts"] = 0
+        if (
+            formula.escape(player.level, target.level, var["run_attempts"])
+            is True
```
kerizane: Don't need the `is True` bit

tuxemon/formula.py
JaskRendix: You're right, it's a bool

tuxemon/formula.py
JaskRendix: Thanks, I didn't notice, renamed.

tuxemon/states/combat/combat.py
JaskRendix: += 1! Thanks

JaskRendix: > Attempts should increase by +1 after success is checked, not before (i.e. the first attempt should use 0 rather than 1)

Each successful attempt increases the number (+1), as requested.

tuxemon/states/combat/combat_menus.py
JaskRendix: Thanks again

kerizane (approved):

JaskRendix:
Do I need to close and reopen?

kerizane:
No, it's fine.

kerizane (approved):

Issue #1335, opened 2022-09-19
Should not be able to sell keyitems


kerizane:
Key items are presumably quest items that you shouldn't be able to lose. Don't show these in the sell items menu.

JaskRendix:
I noticed, we could add the exception for keyitem or we can update the items (shifting the sort from utility to quest); quest items aren't showed.

Sanglorian:
Should "keyitem" be its own variable? There might be items within a category that shouldn't be bought or sold even if others in the category should be. For example, under [Technique Teaching Items](https://wiki.tuxemon.org/Category:Item#Technique_Teaching_Items) the TMs perhaps shouldn't be sellable. 

I can set this up as a variable for items on the wiki if people think that's the best approach.

kerizane:
In that case we could probably track this in economy.

kerizane:
Yeah, set this up on the wiki please.

Sanglorian:
Done -- [items ](https://wiki.tuxemon.org/Category:Item) now have the "Key Item" property, which is either "yes" or blank.

JaskRendix:
fixed #1485 , eventually it can be closed

PR #1397, opened 2022-10-14
fix KeyItem

JaskRendix:
PR addresses #1335

### DIFF
--- a/tuxemon/states/items/__init__.py
+++ b/tuxemon/states/items/__init__.py
@@ -301,6 +301,7 @@ class ShopMenuState(Menu[Item]):
             item
             for item in self.seller.inventory.values()
             if not (self.seller.isplayer and item["item"].sort == "quest")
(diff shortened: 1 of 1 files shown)


kerizane:
Thanks, but we already have the logic implemented for quest, so is this change required? Probably some of the  items just need to be configured correctly as quest items.

Also, is this valid python? Two if statements look suspicious...

kerizane (changes requested):
Not convinced this is necessary or correct.

Sanglorian:
I think the wiki has confused things by having Story Items (a category) and key item (a binary variable). If "quest item" is a variable, then we can just use that instead of key item, and rename the field in the wiki.

But not all key items are Story Items, which is I think why @JaskRendix has set up this separate variable.

JaskRendix:
No problem in closing the PR, I got confused because the issue was still open. I tried to upgrade the original statement, but I wasn't able to do it. If it's ok already, then please close the issue.

PR #1398, opened 2022-10-15
money prize

JaskRendix:
I took inspiration from https://bulbapedia.bulbagarden.net/wiki/Prize_money

PR addresses the money prize after a battle. Utilizes the old experience_give_modifier (not utilized) and it converts it to money_modifier (set basic value10).

How it works? Formula: **monster_level * money_mod**
Example: battle against spyder_route2_billie
```
Dollfin 6 * 10 = 60
Eyenemy 6 * 10 = 60
Cardiling 3 * 10 = 30
total 150

150 if you make a perfect fight (monster beats all the enemy monsters);
more monsters participate, more money you can win;
so it works similar to experience, only against trainers (no wild encounters):
```
Advantage is that modders can set the modifier, so they can give more or less importance to a specific NPC.

Classes part moved to #1400


kerizane:
Haven't looked at the PR, but ways to get money are definitely needed!

Sanglorian:
Thanks - this sounds great! Doing it NPC by NPC sounds fine by me, no need to do it by class  

That said, harmonising the classes makes sense - do you want to open an issue for that and we can discuss it further there?

JaskRendix:
@Sanglorian I opened the issue.

JaskRendix:
I wasn't satisfied, because there wasn't a message. I did some edits and I created a new message, so we can distinguish between victory trainer and victory wild encounter. Why? I needed a new message to include the money prize (see below). Of course the message can be changed / edited if someone has a better string.

Screenshot from 2022-10-19 15-59-48

Sanglorian:
Looks good! I think you can use ${{currency}} to add the currency symbol -- it's currently just "$".

JaskRendix:
You gave me an idea to solve an annoying problem, but at the moment I can only add the "$" currency in the T.format. This will allow to change the position in the different languages.

tuxemon/npc.py
```diff
@@ -886,6 +886,9 @@ def find_tech(self, tech: str) -> Optional[Technique]:
 
         return None
 
+    def give_money(self, amount: int) -> None:
+        self.money["player"] = self.money.get("player") + amount
```
kerizane: Will player always be here?

`+= amount` if it is, otherwise do `.get("player", 0) + amount`

JaskRendix: Ok, thanks

kerizane (approved):

Issue #1222, opened 2022-08-06
Add flatpak instructions to README


kerizane:
We now have a flatpak image on flathub, as well as attached to the release on GitHub. Add instructions on how to use these.

awimoweh:
Hello if you don't mind, I'll try and write the instruction on how to use it.

kevin2kevin2:
It has to be closed (I have not the right), isn't it? (cf. https://github.com/Tuxemon/Tuxemon#flatpak)

PR #1403, opened 2022-10-18
Flatpak instruction

awimoweh:
From #1222

Here are the few line, on how to use Tuxemon with Flatpak

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import fnmatch
 import os
 
-from setuptools import setup
+from setuptools import setup, find_packages
 from setuptools.command.install import install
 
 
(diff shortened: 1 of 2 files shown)


kerizane:
Thanks for the contribution, but somebody else has already submitted a PR for this. I have merged theirs.

setup.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, *args, **kwargs):
     maintainer_email="info@tuxemon.org",
     url="https://www.tuxemon.org",
     include_package_data=True,
-    packages=modules,
+    packages=find_packages(),
```
kerizane, line 53: You may raise a new PR for this change, but delete the code rather than commenting it out.

setup.py
```diff
@@ -3,7 +3,7 @@
 import fnmatch
 import os
 
-from setuptools import setup
+from setuptools import setup, find_packages
```
kerizane, line 6: from setuptools import find_packages, setup

PR #1413, opened 2022-10-28
fix warning pc states init

JaskRendix:
PR addresses some minor things I noticed in pc/states/__init__ (undefined):
- two Strings in str;
- missing Optional;
- missing Animation;

Tested + black, no problems with drop and pick.

Screenshot from 2022-10-28 10-41-28

### DIFF
--- a/tuxemon/states/pc/__init__.py
+++ b/tuxemon/states/pc/__init__.py
@@ -38,6 +38,7 @@ from typing import (
     Any,
     Callable,
     Generator,
+    Optional,
     Sequence,
     Tuple,
     no_type_check,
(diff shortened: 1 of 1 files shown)


tuxemon/states/pc/__init__.py
```diff
@@ -46,6 +47,7 @@
 import pygame_menu
 
 from tuxemon import graphics, prepare
+from tuxemon.animation import Animation
```
kerizane: Move this one to within `if TYPE_CHECKING` (line 62)

JaskRendix: Ok

kerizane (approved):

PR #1415, opened 2022-10-28
technique: accuracy, power and potency

JaskRendix:
PR addresses the fields **power**, **potency** and **accuracy** (techniques).

I was working on the status_grabbed and status_stuck. These statuses have effects on power and potency. Unfortunately these fields are different than armour, etc (classical stats), so I developed this idea to simplify the system.

At the moment these values are defined in each JSON.
The idea is to centralize these values by range as you can see below (schema).

I admit my ignorance by saying that I don't know if there is a logic behind these values.
By running some tests, I discovered that all the techniques with **special** as range have:
- accuracy 1
- power 0
- potency 1

Advantages?
- centralized, we can get rid of the fields in the JSONs, as well as the ones in db.py;
- less values to maintain and update (in the case we add new techniques or statuses);
- simplify, we can implement variance in power, potency and accuracy more easily (thinking about status_grabbed and status_stuck, but it can be others too);
- using range, it exists, but it's not utilized much;

As usual, this is not definitive, just a draft. We can remove accuracy or potency or everything and trash everything, no problem.
As usual, @Sanglorian let us know what do you think. 

(tested in locale and works)

```
RANGES = {
    Range.melee: {
        "accuracy": 1,
        "potency": 1,
        "power": 1,
    },
    Range.ranged: {
        "accuracy": 2,
        "potency": 2,
        "power": 2,
    },
    Range.reach: {
        "accuracy": 3,
        "potency": 3,
        "power": 3,
    },
    Range.reliable: {
        "accuracy": 4,
        "potency": 4,
        "power": 4,
    },
    Range.special: {
        "accuracy": 5,
        "potency": 5,
        "power": 5,
    },
    Range.status: {
        "accuracy": 6,
        "potency": 6,
        "power": 6,
    },
    Range.touch: {
        "accuracy": 7,
        "potency": 7,
        "power": 7,
    },
}
```

### DIFF
--- a/mods/tuxemon/db/technique/run.json
+++ b/mods/tuxemon/db/technique/run.json
@@ -7,6 +7,7 @@
   "animation": null,
   "sfx": "sfx_blaster",
   "icon": "",
+  "range": "status",
   "sort": "meta",
   "effects": [],
   "flip_axes": "",
--- a/mods/tuxemon/db/technique/status_blinded.json
+++ b/mods/tuxemon/db/technique/status_blinded.json
@@ -6,6 +6,7 @@
   ],
   "flip_axes": "",
   "icon": "gfx/ui/icons/status/icon_blinded.png",
+  "range": "status",
   "sfx": "sfx_pulse",
   "slug": "status_blinded",
   "sort": "meta",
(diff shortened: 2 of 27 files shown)


Sanglorian:
Accuracy is the chance that a technique hits, potency is the chance that - if a technique hits - its condition is imposed, and power is the damage done by a technique. 

So they do all need to be set for each technique - not all melee attacks will do the same damage or have the same chance to hit, for example. 

Range should determine which of a monster's stats is used for an attack, and which stat the target uses to defend -- so it's quite important (but it might not be set up yet).

JaskRendix:
Ok, from what you say, it doesn't make sense.
I'm going to find another way for status_grabbed and status_stuck.
Thanks.

PR #1417, opened 2022-10-30
apply status

JaskRendix:
PR addresses:
added in db.py the category, repl_neg and repl_pos;
updated apply_status with status check;
updated apply_status with monster parameter;
removed status check from the status py files;

still not the definitive form, it'll be updated again for spyderbite and eliminated;

tested + black


tuxemon/event/actions/set_monster_status.py
```diff
@@ -65,7 +65,7 @@ def set_status(monster: Monster, value: Optional[str]) -> None:
             # TODO: own class for status effect
             # TODO: handle invalid statues
             status = Technique(value)
-            monster.apply_status(status)
+            monster.apply_status(monster, status)
```
kerizane: Why?

JaskRendix: Because it's the same function in monster.py, if not set_monster_status action wouldn't work.

kerizane: But it's a method on monster, why do you need to pass monster in again as an argument?

JaskRendix: You're right, it's a duplicate, I'm going to get rid of.

tuxemon/monster.py
```diff
@@ -442,29 +442,32 @@ def apply_status(self, status: Technique) -> None:
...
+                        self.status.insert(0, status)
+                    elif status.repl_pos == "remove":
+                        self.status.clear()
+                    else:
+                        # noddingoff, exhausted, festering, dozing
+                        return
+                elif self.status[0].category == "negative":
+                    if status.repl_neg == "replace":
+                        self.status.clear()
+                        self.status.insert(0, status)
```
kerizane: I think "append" is more straightforward.

kerizane: Should status be a single item rather than a list? It seems like we can only have a single status at a time, so would make the code simpler if we got rid of the the list.

kerizane (dismissed):

tuxemon/monster.py
JaskRendix: It was append in the beginning, I can reinsert it eventually. I put insert because I wanted to reserve the second slot to spyderbite or eliminated, but as you said it can complicate the code.

So, this is it: self.status: List[Technique] = []

We get rid of the list and we replace it with? Str.enum? Other ideas?

JaskRendix: We need two slots: first for all the status, second for spyderbite or eliminated

JaskRendix:
Replaced the insert with append. 

There is also the status_faint to keep in consideration. If there is an alternative with a couple of slot, why not.

kerizane (approved):

PR #1418, opened 2022-10-31
Move parameter attributes into ItemCondition and ItemEffect class

JaskRendix:
PR addresses the removal of namedtuple from item (effects and conditions), now the attributes are on the class.
Inspired by #1385 

Addition:
- added **hp** parameter to **revive** as well as 1 in the JSON

Everything works, there is the message when you can use an item or not, you can use the capture device and capture monsters, you can use potion on monster, etc.

### DIFF
--- a/tuxemon/item/conditions/shape.py
+++ b/tuxemon/item/conditions/shape.py
@@ -25,21 +25,15 @@
 
 from __future__ import annotations
 
-from typing import NamedTuple, Union
+from dataclasses import dataclass
+from typing import Union
 
 from tuxemon.item.itemcondition import ItemCondition
 from tuxemon.monster import Monster
 
 
-class ShapeConditionParameters(NamedTuple):
-    shape1: str
-    shape2: Union[str, None]
-    shape3: Union[str, None]
-    shape4: Union[str, None]
-    shape5: Union[str, None]
-
-
-class ShapeCondition(ItemCondition[ShapeConditionParameters]):
+@dataclass
+class ShapeCondition(ItemCondition):
     """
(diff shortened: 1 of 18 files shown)


tuxemon/item/conditions/shape.py
```diff
@@ -48,14 +42,18 @@ class ShapeCondition(ItemCondition[ShapeConditionParameters]):
     """
 
     name = "shape"
-    param_class = ShapeConditionParameters
+    shape1: str
+    shape2: Union[str, None]
+    shape3: Union[str, None]
```
kerizane: I think these need to be defaulted to None. Can you test?

JaskRendix: You're absolutely right

kerizane:
In my PR I have just added "test_actions" to validate all existing usages. Is this something you can do in this PR?

JaskRendix:
That's a pretty good script, I'm not so skilled in scripts, but you mean iterating all the files inside the folder /mods/db/item with ".json" inside the "effects or conditions" field?

tuxemon/item/conditions/__init__.py
```diff
@@ -1,26 +0,0 @@
-#
```
kerizane: Sorry, I would rather you leave these in for now, until I have confirmation that it's ok to remove them.

If you are creating a file from scratch then feel free to leave it out.

JaskRendix: No problem, I'm going to reverse.

kerizane (approved):

PR #1419, opened 2022-10-31
types in db.py

JaskRendix:
PR addresses:
- creation of the class ElementType(str, Enum) in db.py (see below);
- renaming of all Types (eg Water, Earth) to types (eg. water, earth) as technique types, so now everything is similar and centralized (there were some monster all lowercase -> pairaigrin?); if you want Earth instead of earth, no problem, we can switch.

I noticed it when I was working on type.py (the other PR). I hope it can be useful, if not I'll close it, no problem. If you want to add something, feel free to share.

```
class ElementType(str, Enum):
    aether = "aether"
    wood = "wood"
    fire = "fire"
    earth = "earth"
    metal = "metal"
    water = "water"
    normal = "normal"
    glitch = "glitch"
```

tested + black done

### DIFF
--- a/mods/tuxemon/db/item/tm_avalanche.json
+++ b/mods/tuxemon/db/item/tm_avalanche.json
@@ -1,6 +1,6 @@
 {
   "conditions": [
-    "type target,Earth,Water"
+    "type target,earth,water"
   ],
   "effects": [
     "learn avalanche"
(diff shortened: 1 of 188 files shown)


kerizane (approved):

kerizane:
Great! Stuff like this is always useful.

PR #1423, opened 2022-11-03
monster weight, height and capture

JaskRendix:
PR addresses weight, height and capture of the monster:
- added info (monster menu), before it wasn't implemented;
- added two msgid in the PO file related to info;
- added personalized weight for the monster in the party;
- added personalized height for the monster in the party;
- added personalized capture for the monster in the party;
- inserted metric/imperial field in config as well as in locale.py (now if you choose metric it'll show the data in kg, km and cm, while imperial in lb, mi and ft)

Updated screen:
Screenshot from 2022-11-04 15-53-18

Screenshot from 2022-11-03 14-02-50


Sanglorian:
Thanks for this. Would it be possible for pounds vs kgs to be based on locale, rather than both displayed? Average is fine for now but longer term I think we'll want a concise display with a lot of info on it, at which point I expect we'd drop it.

I think 10% is fine for the distribution. 

1000 days for date of birth is a bit funny because there might be some tuxemon that are unique or come from especially long-lived varieties. What about just recording date of capture (which would be birth for bred mons).

Sanglorian:
Forgot to mention, other interesting information longer term to display would be:

Type/element
Species name (if given a nickname)
Techniques
XP and XP till next level

JaskRendix:
You're right, some species, like ghosts or leviathans are hundred years old, etc. so it makes sense.
I'm going to:
- reverse age to capture, so the one generates would be the date of capture;
- drop the average, it occupies space, and trying to implement some of the infos you suggested;


What do you mean based on locale?

JaskRendix:
Number of monster per category + average Height (cm) and average Weight (kg).
It could help to complete the data. Always enjoyed Excel stuff.
```
Shape		Nr	Avg H	Avg W
Aquatic		7	150.7	83.6
Blob		30	48.5	26.3
Brute		20	116.6	402.9
Dragon		15	208.0	1024.9
Flier		28	20.4	2.8
Grub		17	5.1	0.6
Humanoid	21	68.8	64.4
Hunter		18	67.8	51.4
Landrace	7	176.4	1860.7
Leviathan	3	500.0	16733.3
Polliwog	11	13.3	2.5
Serpent		20	18.1	0.5
Sprite		12	24.6	6.7
Varmint		19	36.1	9.9
```
Number of monster per range of height and weight
```
H (CM)		NR	W (KG)		NR
0-10		0	0-10		34
11-20		2	11-20		8
21-30		4	21-30		3
31-40		6	31-40		2
41-50		4	41-50		1
51-60		6	51-60		8
61-70		8	61-70		1
71-80		4	71-80		1
81-90		5	81-90		2
91-100		3	91-100		1
101-150		15	101-150		8
151-200		18	151-200		11
201-250		10	201-250		3
251-500		4	251-500		5
501-1000	3	501-1000	1
1001-2500	2	1001-2500	1
2501-5000	0	2501-5000	3
5001		0	5001		3
```

kerizane:
@JaskRendix you need to fix the conflict.

JaskRendix:
Done

kerizane:
More conflicts

JaskRendix:
Done

tuxemon/formula.py
```diff
@@ -215,6 +216,99 @@ def escape(level_user: int, level_target: int, attempts: int) -> bool:
...
 
+def today_ordinal() -> int:
+    """
+    It gives today's proleptic Gregorian ordinal.
+    """
+    today = dt.date.today().toordinal()
+    return today
+
+
+def set_capture() -> int:
```
kerizane: Is this still used?

JaskRendix: Yes, it's used for all the existing monster, to give them a random capture date (one of the last 1000 days); eventually we can remove it and assign "today" date.

JaskRendix: Look, let's assign "today" date, I'll remove formula.doc too, at the moment it's not used, the day I'll find something useful I'll add again. It doesn't make sense the random backwards, it'll will start counting since the implementation.

kerizane (approved):

PR #1424, opened 2022-11-03
Move parameter attributes into TechEffect class

JaskRendix:
Technique turn, after #1418, following #1385

PR addresses the removal of namedtuple from tech (effects), now the attributes are on the class:
- edited effect files;
- added the param "amount int" to damage and fixed JSONs;
- added the param "all str" to poison and fixed JSONs;
- added the param "all str" to recover and fixed JSONs;
- added the param "all str" to statchange and fixed JSONs;
- added the param "all str" to lifeleech and fixed JSONs;
- added the param to swap, there is no JSON to fix;

The previous loading crash has been solved by the actions above (addition of params), now the game load, it can start a battle without a battle, but there is a crash when switching -> issue solved.

Testing all the status with statchange: no problem.
Testing poison: no problem.
Testing recover: no problem.
Testing lifeleech: no problem.

### DIFF
--- a/tuxemon/technique/techeffect.py
+++ b/tuxemon/technique/techeffect.py
@@ -1,27 +1,14 @@
 from __future__ import annotations
 
-import logging
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    ClassVar,
-    Generic,
-    Optional,
-    Sequence,
-    Type,
-    TypedDict,
-    TypeVar,
-)
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, ClassVar, Optional, TypedDict
 
-from tuxemon.tools import NamedTupleProtocol, cast_parameters_to_namedtuple
+from tuxemon.session import Session, local_session
+from tuxemon.tools import cast_dataclass_parameters
 
 if TYPE_CHECKING:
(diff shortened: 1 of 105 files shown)


tuxemon/technique/techeffect.py
```diff
@@ -73,44 +61,17 @@ class TechEffect(Generic[ParameterClass]):
...
-        self._done = False
+    name: ClassVar[str]
+    session: Session = field(init=False, repr=False)
+    _done: bool = field(default=False, init=False)
+
+    def __post_init__(self):
+        self.session = local_session
+        for tech in local_session.player.monsters:
+            self.user = tech
+            for moves in tech.moves:
```
kerizane: What's going on in this for loop?

kerizane: I don't understand what it's trying to do

JaskRendix: before there were Technique (self.move) and Monster (self.user), so yesterday I tried to come up with something (wrong).

JaskRendix: I think it's solved. I need to test some things.

kerizane (approved):

PR #1427, opened 2022-11-05
setcode action

JaskRendix:
PR addresses the creation of the set_code action (already structured for #1385) as well as adds three msgid PO.

(I have the other version with params, so it's not a problem to reverse)

Reason? Modders, modders can define a code (numeric or alphabetical or mixed) to access, open a door, receive or trigger something.

It's composed by three parameters:
- code (the actual code);
- characters (characters of the code; eg. code is "1234", then 4 characters;
- variable: the variable that if the code is correct will be set on "on", if not will be set on "off";

eg: **property name="act2" value="set_code lepeletier,10,revolution"**

The variable will help modders with the following actions (a scenario if it's correct, another if it's wrong); above the player has to guess "rockitten". It'll be created the variable prize, so if it's correct the value will be "on", otherwise "off". 

Question: is it ok case sensible? Too complicated? Do we need all upper or lowercase? Let me know. At the end the modders can play on the difficulty of the word too (or sequence of numbers or both).

The input:
Screenshot from 2022-11-06 08-14-34
The message, if it's correct, then it'll be correct.
Screenshot from 2022-11-06 08-14-47

### DIFF
--- a/tuxemon/event/actions/set_code.py
+++ b/tuxemon/event/actions/set_code.py
@@ -1,0 +1,61 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import final
+
+from tuxemon.event.eventaction import EventAction
+from tuxemon.locale import T
+from tuxemon.menu.input import InputMenu
+
+
+@final
+@dataclass
+class SetCodeAction(EventAction):
+    """
+    Set a code and checks if it's correct or not.
+    Case Sensitive:
+    ATTENTION and AtTenTION are two different words.
+
+    Script usage:
+        .. code-block::
+
+            set_code <question>,<answer>,<variable>
+
+    Script parameters:
+        question: The question the player needs to reply.
+                  (eg. "access_code")
+                  then you create the msgid "access_code"
+                  inside the English PO file, as follows:
+                  msgid "access_code"
+                  msgstr "Here the actual question?"
+        answer: The right answer to the question.
+        variable: Where the result (right/wrong) is saved.
+
+    """
+
+    name = "set_code"
+    question: str
+    answer: str
+    variable: str
+
+    def check_setcode(self, name: str) -> None:
+        player = self.session.player
+        if self.answer == name:
+            player.game_variables[self.variable] = "right"
+        else:
+            player.game_variables[self.variable] = "wrong"
+
+    def start(self) -> None:
+        self.session.client.push_state(
+            state_name=InputMenu,
+            prompt=T.translate(self.question),
+            callback=self.check_setcode,
+            escape_key_exits=False,
+            char_limit=len(self.answer),
+        )
+
+    def update(self) -> None:
+        try:
+            self.session.client.get_state_by_name(InputMenu)
+        except ValueError:
+            self.stop()


tuxemon/event/actions/set_code.py
```diff
@@ -0,0 +1,65 @@
...
+from tuxemon.menu.input import InputMenu
+from tuxemon.tools import open_dialog
+
+
+@final
+@dataclass
+class SetCodeAction(EventAction):
+    """
+    Set a code and checks if it's correct or not.
+    Case Sensible:
```
kerizane: Sensitive, not Sensible

tuxemon/event/actions/set_code.py
```diff
@@ -0,0 +1,65 @@
...
+    ATTENTION and AtTenTION are two different words.
+
+    Script usage:
+        .. code-block::
+
+            set_code <code>,<characters>,<variable>
+
+    Script parameters:
+        code: The actual code the player needs to know.
+        characters: The number of characters/digits.
```
kerizane: We could calculate this automatically by counting the number of characters in code. Or did you want to allow values longer than the actual code?

tuxemon/event/actions/set_code.py
```diff
@@ -0,0 +1,65 @@
...
+    variable: str
+
+    def check_setcode(self, name: str) -> None:
+        player = self.session.player
+        if self.code == name:
+            player.game_variables[self.variable] = "on"
+            open_dialog(
+                self.session,
+                [T.format("right_code", {"code": name})],
+            )
```
kerizane: I don't think we should have the dialogue hard-coded. Just let the modder trigger the dialogue manually using the output variable.

tuxemon/event/actions/set_code.py
JaskRendix: Automatically by counting the number of characters is perfect, so we can remove a parameter, that's perfect.

tuxemon/event/actions/set_code.py
JaskRendix: It makes sense, so in the following steps the modder can check the variable by existing actions/conditions.

tuxemon/event/actions/set_code.py
JaskRendix: Thanks

JaskRendix:
Removed right and wrong_code too. Modders will think about the right options;
Removed input_code hardcoded, now modders can set up;
Added instruction to add the input_code inside the PO file (translation);
Renamed parameters question and answer to be more intuitive;
Renamed on/off as right/wrong inside the variables;

tuxemon/event/actions/set_code.py
```diff
@@ -0,0 +1,54 @@
...
+        if self.code == name:
+            player.game_variables[self.variable] = "on"
+        else:
+            player.game_variables[self.variable] = "off"
+
+    def start(self) -> None:
+        self.session.client.push_state(
+            state_name=InputMenu,
+            prompt=T.translate("input_code"),
+            callback=self.check_setcode,
```
kerizane, line 52: We should probably make the prompt a parameter. I.e. don't hardcode "input_code" here.

JaskRendix: No problem, added some instruction too, so they can immediately define the msgid.

kerizane (approved):

PR #1428, opened 2022-11-05
fix crash escape

JaskRendix:
PR addresses the crash of escape (attempt of running from a wild encounter).

```
  File "/home/robespierre/Tuxemon/tuxemon/states/combat/combat.py", line 431, in transition_phase
    var["run_attempts"] += 1
KeyError: 'run_attempts'
```
var["run_attempts"] += 1 was in the wrong spot
Then I add the check if the variable exists.

### DIFF
--- a/tuxemon/states/combat/combat_menus.py
+++ b/tuxemon/states/combat/combat_menus.py
@@ -119,12 +119,15 @@ class MainCombatMenuState(PopUpMenu[MenuGameObj]):
         player = combat_state.monsters_in_play[combat_state.players[0]][0]
         target = combat_state.monsters_in_play[combat_state.players[1]][0]
         var = combat_state.players[0].game_variables
-        var["run_attempts"] = 0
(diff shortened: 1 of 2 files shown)


tuxemon/states/combat/combat_menus.py
```diff
@@ -119,12 +119,17 @@ def run(self) -> None:
         player = combat_state.monsters_in_play[combat_state.players[0]][0]
         target = combat_state.monsters_in_play[combat_state.players[1]][0]
         var = combat_state.players[0].game_variables
-        var["run_attempts"] = 0
+        # if the variable doesn't exist
+        try:
+            var["run_attempts"]
+        except:
```
kerizane:
```
if "run_attempts" not in var:
    var["run_attempts"] = 0
```

JaskRendix: Thanks

kerizane (approved):

PR #1429, opened 2022-11-06
output and update battle_history

JaskRendix:
PR addresses the following points:

1. added str enum in db.py
2. update of battle_history
3. moved the rounds and variables from combat.py to formula.py

- **added str enum in db.py**

```
class OutputBattle(str, Enum):
    won = "won"
    lost = "lost"
    draw = "draw"
    ran = "ran"
    forfeit = "forfeit"
```
I wasn't satisfied with the original "won", "lost", etc. I prefer to centralize like SeenStatus.

- **update of battle_history**

~~**battle_history: Dict[Tuple[str, OutputBattle], int]** -> npc_slug, (won/draw/lost), date~~
**Dict[str, Tuple[OutputBattle, int]]**
**changed, now the tuple is inside the value, not anymore in the key, I fixed all the other files. Why? Because when I tried to save the filegame, it crashed. Unable to save a tuple as key in JSON**

so if you are going to lose or win or draw against the same NPC, it updates the date, it gaves us the possibility to know xx days ago.

The missing variable was time (when the battle was fought?). Initially I wasn't able to include the variable, but now I can. The discover of ordinal, as well as working on other PRs, gave me the possibility to discover toordinal, the perfect value (all int) to indicate the date. So it's possible, easily, to understand when it was fought the last battle.

Now, for instance, battles_print gives (when you look for npc and output):
**won against spyder_route2_billie 0 days ago**

- **moved the rounds and variables from combat.py to formula.py**

The reason is order and cleanness, combat.py results a bit less chaotic and all the math ops are in formula.

As usual tested and black.

### DIFF
--- a/tuxemon/event/actions/battles_print.py
+++ b/tuxemon/event/actions/battles_print.py
@@ -59,16 +59,16 @@ class BattlesAction(EventAction[BattlesPrintActionParameters]):
 
     def start(self) -> None:
         player = self.session.player
+        character = self.parameters.character
+        result = self.parameters.result
         today = dt.date.today().toordinal()
 
-        value = (self.parameters.character, self.parameters.result)
-        if self.parameters.result:
-            if value[0] in player.battle_history:
-                battle_date = player.battle_history.get(value)
-                print(
-                    f"{value[1]} against {value[0]} {today - battle_date} days ago"
-                )
+        if character in player.battle_history:
+            output, battle_date = player.battle_history[character]
+            diff_date = today - battle_date
+            if result == output:
+                print(f"{result} against {character} {diff_date} days ago")
             else:
-                print(f"Never {value[1]} against {value[0]}")
+                print(f"Never {result} against {character}")
         else:
             print(player.battle_history)
--- a/tuxemon/event/actions/battles_print.py
+++ b/tuxemon/event/actions/battles_print.py
@@ -21,6 +21,7 @@
 
 from __future__ import annotations
 
+import datetime as dt
 import logging
 from typing import NamedTuple, Optional, final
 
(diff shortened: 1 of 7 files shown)


tuxemon/event/actions/battles_print.py
```diff
@@ -56,12 +59,16 @@ class BattlesAction(EventAction[BattlesPrintActionParameters]):
...
         player = self.session.player
-
-        variable = self.parameters.variable
-        if variable:
-            if variable in player.battle_history:
-                print(f"{variable}: {player.battle_history[variable]}")
+        today = dt.date.today().toordinal()
+
+        value = (self.parameters.character, self.parameters.result)
+        if value:
```
kerizane: This will always be true

kerizane:
```
if (None, None):
    print("This will print!")
```

JaskRendix: You're right, I change.

JaskRendix:
If I try to save the game:
```
  File "/usr/lib/python3.10/json/encoder.py", line 376, in _iterencode_dict
    raise TypeError(f'keys must be str, int, float, bool or None, '
TypeError: keys must be str, int, float, bool or None, not tuple
```

kerizane (dismissed):

tuxemon/states/combat/combat.py
```diff
@@ -477,31 +475,27 @@ def transition_phase(self, phase: CombatPhase) -> None:
...
-                    var["percent_lose"] = round(
-                        (var["battle_lost"] / var["battle_total"]) * 100
-                    )
                     # track battles against NPC
-                    self.players[0].battle_history[
-                        self.players[1].slug
-                    ] = "lost"
+                    self.players[0].battle_history[self.players[1].slug] = (
+                        OutputBattle.won,
+                        dt.date.today().toordinal(),
```
kerizane, line 497: Should be lost, not won?

JaskRendix: Yes, it was lost

kerizane (approved):

PR #1430, opened 2022-11-09
removal of levelrange

JaskRendix:
PR addresses:
- the removal of level_range from the JSON of encounters;
- the addition of the following code under random_encounters;

```
    # Set the monster's level based on the party level
    level = random.randint(
        player.game_variables["party_level_lowest"],
        player.game_variables["party_level_highest"],  --> we can use average instead of highest
    )
```
Reason? The random_encounters doesn't follow the development of the party. It's static, we have a range, but we are obliged to insert another field or JSON to mimic the party development.

The proposal is replace **level_range** with a random value that mimics the party. The range of the value is the **minimum** and the **maximum** level of the party. 

_For instance:
let's consider a party with 2 monsters, the weakest  lv 4, the strongest lv 12
the wild monster will be a random between 4 and 12
(if we use average, the random will be between 4 and 8)_

In this way we can remove a field from the JSON and centralizing the operation in the action file. For exceptions we can link the encounter_rate to the level, in some cases I saw 100, and the level range was 100, it takes nothing to add an exception (like if encounter_rate >= 50? only three cases or better == 100).

What do you think @Sanglorian ?

Pro? removing a field from the JSON, helping the player to level_up the new captured monsters
Cons? single monster party, very high level, like a single monster lv20, both value will be 20, but we can address it with counting the number of monsters in the party and setting the minimum to 1 (?).

### DIFF
--- a/mods/tuxemon/db/encounter/37707_town.json
+++ b/mods/tuxemon/db/encounter/37707_town.json
@@ -3,34 +3,18 @@
     "monsters":[
         {
             "encounter_rate": 0.5,
-            "level_range": [
-                1,
-                10
-            ],
             "monster": "fruitera"
         },
         {
             "encounter_rate": 0.5,
-            "level_range": [
-                1,
-                10
-            ],
             "monster": "dollfin"
         },
         {
             "encounter_rate": 0.5,
-            "level_range": [
-                1,
-                10
-            ],
             "monster": "legko"
         },
         {
             "encounter_rate": 0.5,
-            "level_range": [
-                1,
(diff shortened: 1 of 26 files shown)


Sanglorian:
I think this is a useful option to have, but most of the time we want a static range -- that lets players train in areas that have weaker monsters, or push ahead into more perilous ones to advance faster. Otherwise there's a feeling of being on a "treadmill" - however hard you go, you don't go any further or faster.

JaskRendix:
Understood

PR #1431, opened 2022-11-09
combatwin message

JaskRendix:
PR addresses the "combat_victory_trainer" in PO.

"combat_victory_trainer" is too long.
It gets cut when the NPC name is longer than usual.
Because of this it's not possible to see how much money the player won (the important part).

From: 
"You have won against {npc} and you get {currency}{prize}!"
To:
"You have won {currency}{prize}!"

Got rid of the NPC name, 20 characters saved (without considering the NPC name).

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -174,7 +174,7 @@ msgid "combat_victory"
 msgstr "You have won!"
 
 msgid "combat_victory_trainer"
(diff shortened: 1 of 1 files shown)


kerizane:
You could also split it over two pages. I don't mind.

@Sanglorian ?

JaskRendix:
Look, let's split it on two lines, I tested and it works. Not cut since there are two lines.

1st line "You have won against {npc}"
2nd line "You get {currency}{prize}!"

kerizane (approved):

PR #1433, opened 2022-11-10
katacoon and chromeye evolutions

JaskRendix:
PR addresses two missing evolutions: **katacoon** and **chromeye**.
katacoon as described in wiki
chromeye random

I'm unable to act on others due to missing monster. 

Deleted all the proposals, these weren't considered.

@Sanglorian

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1204,14 +1204,18 @@ class CombatState(CombatAnimations):
                 # check the path field, path field signals evolution item based
                 if not evolution.path:
                     if evolution.at_level <= monster.level:
-                        logger.info(
-                            "{} evolved into {}!".format(
-                                monster.name, evolution.monster_slug
-                            )
-                        )
                         self.players[0].evolve_monster(
                             monster, evolution.monster_slug
                         )
+                elif evolution.path == "gender":
+                    if evolution.at_level <= monster.level:
+                        if evolution.gender == monster.gender:
+                            self.players[0].evolve_monster(
+                                monster, evolution.monster_slug
+                            )
+                elif evolution.path == "exception":
+                    if evolution.at_level <= monster.level:
+                        self.players[0].evolve_exceptions(monster)
 
     def end_combat(self) -> None:
         """End the combat."""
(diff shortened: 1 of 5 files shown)


Sanglorian:
Thanks! 

So with Katacoon, the idea is that if its Melee stat is higher, it evolves into Bugnin, and otherwise into Sumchon. So not based on its techniques. 

But I'm not sure that we've implemented variable stats yet.

Re Vivipere's evolutions: Viviphyta, Vivitrans, Vivicinder and Vividactyl shouldn't have any requirement of being level 24. Just as soon as they level up while meeting that condition, they evolve.

Because Vivitron happens at level 25, it will only happen if someone rejected the level 24 evolution (Vivisource or Viviteel). So there's no need for neuter.

Good question re Botbot. My idea was that you could choose from any of them: a little menu would pop up and you could choose. Not sure if that's feasible. 

And as for Chromeye I haven't thought about it. Good question! Maybe we should throw it open to the Discord to see what they think.

Thanks heaps for progressing this!

JaskRendix:
No problem, let me know the result of the discussion about Chromeye, remember that we have other variables that we can exploit.

tuxemon/states/combat/combat.py
```diff
@@ -1204,14 +1204,18 @@ def evolve(self) -> None:
...
                             monster, evolution.monster_slug
                         )
+                elif evolution.path == "gender":
+                    if evolution.at_level <= monster.level:
+                        if evolution.gender == monster.gender:
+                            self.players[0].evolve_monster(
+                                monster, evolution.monster_slug
+                            )
+                elif evolution.path == "exception":
+                    if evolution.at_level <= monster.level:
```
kerizane, line 1217: For exception evolutions it looks like we will evolve them multiple times... Maybe we should return after running this method?

kerizane:
I will allow this, but it would be nice if you could do this properly. Raise a ticket to reimplement this if you don't want to do it now.

E.g. stat_melee > stat_armour

And random for the random methods. Ideally you would collect the random choices from the file rather than hardcoding them.

JaskRendix:
You're right about the hardcoding and you gave me an idea.

PR #1434, opened 2022-11-11
maps properties

JaskRendix:
PR addresses maps properties by expanding the use.

At the moment the only use is related to edges (clamped), but with this PR we expand the use of properties to other things.

- added scenario field: spyder, xero or none: so we can easily generate a list or a dict with all the maps of each scenario, without creating another file, etc.
- added the bool town: if town == true, if not None: my idea was a str field, but I wasn't sure about the options (town, route, dungeon?), so I stopped at bool, ideas are welcome.
- added the bool inside: if inside True, if not None: if the player is inside or outside, this parameter can help to define the background (battles) as well as dynamic such evolutions

since there are already the files tmx and the data is stored inside it, there is no need to create others JSONs, since we can access it; all these values can be accessed by the session.client and these can be useful to define new actions/conditions/effects

I need inside because of evolutions.

@Sanglorian are you interested in additional parameters to add?

### DIFF
--- a/tuxemon/map.py
+++ b/tuxemon/map.py
@@ -405,7 +405,7 @@ class TuxemonMap:
         collision_map: Mapping[Tuple[int, int], Optional[RegionProperties]],
         collisions_lines_map: Set[Tuple[Tuple[int, int], Direction]],
         tiled_map: TiledMap,
-        edges: str,
+        maps: dict,
         filename: str,
     ) -> None:
         """Constructor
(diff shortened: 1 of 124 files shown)


tuxemon/map.py
```diff
@@ -440,10 +440,19 @@ def __init__(
...
-        self.edges = edges
+        self.edges = maps.get("edges")
         self.data = tiled_map
         self.sprite_layer = 2
         self.filename = filename
+        self.maps = maps
+
+        # optional fields
+        # inside (true), outside (none)
+        self.inside = maps.get("inside")
```
kerizane: Can you convert inside and town to bools?

bool(maps.get("inside"))

kerizane:
Great idea!

tuxemon/map.py
JaskRendix: no problem

kerizane (approved):

PR #1436, opened 2022-11-12
monsters: Vivisource, Vividactil, B-ver.1

JaskRendix:
PR addresses:
https://wiki.tuxemon.org/Vivisource
https://wiki.tuxemon.org/Vividactil
https://wiki.tuxemon.org/B-ver.1

@Sanglorian the only thing missing are the techniques for each one.


Sanglorian:
Thanks! I think there's a default set of techniques for each type, for Set 2 monsters without their own technique list at this time. I'm not sure if it's been implemented yet, though -- and I don't remember what they are.

JaskRendix:
Pointless keep it open, the missing monsters will not be added (although are ready)

JaskRendix:
Reopened, let's see if the script will be back.
If someone has some ideas, then feel free to share, because the script isn't around since 2019.

JaskRendix:
@sanglorian

Trying to unstuck the situation. Let me know if you agree about this solution. 

Regarding **b_ver_1**:
- wall_of_steel
- glower
- battery_discharge

battery_discharge is the 3rd move of all the botbot evolutions.
wall_of_steel like picc and mrmorswitch.
glower like k9 and av8r.

Regarding **vivisource**:
- rust_bomb
- constrict
- perfect_cut

**Vivisource** is the male evolution of **vivipere**.
These are the exact moves of the female version **viviteel**. Why? I looked for example and I based my decision on **nudiflot male** and **female**, these monsters, different gender, share the same moves. Applied the same principle.

Regarding **vividactil**:
- headbutt
- wing_tip
- peck

Based on **rhinchus**. Why? Same flier/earth as **vividactil**.

Sanglorian:
The b.ver1 and Vividactil techniques look good! But I think Vivisource's should be water-based. How about Flood, Chill Mist and Wallow?

JaskRendix:
I would say perfect!

Update Vivisource, done.

kerizane (approved):

PR #1439, opened 2022-11-13
confirm evolution

JaskRendix:
PR addresses the introduction of the choice for evolution.

Right now the monsters evolve (the ones without path aka conditions) as soon as the battle is ended.

With this PR, it'll appear (after the last message of the trainer) the following menu:
Screenshot from 2022-11-13 15-53-09

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -733,6 +733,9 @@ msgstr "Are you sure you would like to release {name}?"
 msgid "tuxemon_released"
 msgstr "{name} has been released."
 
+msgid "evolution_confirmation"
+msgstr "Are you sure you would like to evolve {name}?"
+
 # Menu notifications
 msgid "log_off"
 msgstr "Log Off"
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1201,6 +1201,28 @@ class CombatState(CombatAnimations):
     def evolve(self) -> None:
         for monster in self.players[0].monsters:
             for evolution in monster.evolutions:
+                if evolution.at_level <= monster.level:
+                    self.client.pop_state()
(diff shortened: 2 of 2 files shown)


kerizane:
This won't work properly if multiple monsters try to evolve and you reject the first and accept the second.

tuxemon/states/combat/combat.py
```diff
@@ -1199,6 +1199,28 @@ def trigger_player_run(self, player: NPC) -> None:
...
+                            T.format(
+                                "evolution_confirmation",
+                                {"name": monster.name},
+                            )
+                        ],
+                    )
+                    tools.open_choice_dialog(
+                        local_session,
+                        menu=(
+                            ("yes", T.translate("yes"), self.positive_answer),
```
kerizane, line 1218: You need to define `positive_answer` within this method.
```
def positive_answer():
    self.players[0].evolve_monster(
        monster, evolution.monster_slug
    )
```
You can also use a library function: functools.partial, but this might be too complicated. Anyway, both of these ways are used frequently in the project.

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -733,6 +733,9 @@ msgstr "Are you sure you would like to release {name}?"
 msgid "tuxemon_released"
 msgstr "{name} has been released."
 
+msgid "evolution_confirmation"
```
kerizane, line 736: Needs more of a lead in, I think...

"{monster} is trying to evolve. Allow evolution?"

tuxemon/states/combat/combat.py
JaskRendix: Ok, I'll look into

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
JaskRendix: Yes, it makes sense.

PR #1443, opened 2022-11-15
strenum evolution

JaskRendix:
PR addresses the introduction of EvolutionType strenum in db.py
```
class EvolutionType(str, Enum):
    standard = "standard"
    item = "item"
    gender = "gender"
    mixed = "mixed"
```
I wasn't satisfied with the previous structure, it wasn't centralized and it was too risky to build on it.
I don't say now it's the ideal version, but it's a big improvement.

black done, isort done, testing done with both standard and items


tuxemon/item/conditions/has_path.py
```diff
@@ -44,6 +44,7 @@ class HasPathCondition(ItemCondition):
     expected: str
 
     def test(self, target: Monster) -> bool:
-        expect = self.expected
-
-        return any(d["path"] == expect for d in target.evolutions)
+        for evolution in target.evolutions:
+            if evolution.path.item:
+                if evolution.item == self.expected:
+                    return True
```
kerizane: Technically the signature is wrong because it returns true or none, not true or false.

Add return False after the for loop.

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +42,9 @@ class EvolveEffect(ItemEffect):
...
-            self.user.evolve_monster(target, monster_evolve)
-            return {"success": True}
-        else:
-            return {"success": False}
+        for evolution in target.evolutions:
+            if evolution.monster_slug == self.monster_evolve:
+                self.user.evolve_monster(target, self.monster_evolve)
+                return {"success": True}
+            else:
+                return {"success": False}
```
kerizane: This will only ever check the first evolution

JaskRendix: First and only one since it's an object. I'm considering to remove the "else", because at the end you select the monster in the previous passage, so there is no alternative.

tuxemon/item/conditions/has_path.py
JaskRendix: True

tuxemon/item/effects/evolve.py
JaskRendix: Yeah, the only exception is botbot

JaskRendix: Luckily for us booster_tech is the only way botbot evolves

JaskRendix:
To remember: booster_tech
```
  "conditions": [
    "level target,>=,0",
    "has_path target,booster_tech"
  ],
  "effects": [
    "evolve random"
  ],
```
Don't you find hilarious that we are setting up all the system before adding the items?
I have the JSONs in locale, that's so funny.

Update:
I checked and among the new (old) monsters (the ones in the other PR), no one has features like botbot.

tuxemon/item/conditions/has_path.py
kerizane: You added it within the for loop

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +43,15 @@ class EvolveEffect(ItemEffect):
...
+            self.user.evolve_monster(target, evolution)
             return {"success": True}
         else:
-            return {"success": False}
+            for evolution in target.evolutions:
+                if evolution.monster_slug == self.monster_evolve:
+                    self.user.evolve_monster(target, self.monster_evolve)
+                    return {"success": True}
+                else:
+                    return {"success": False}
```
kerizane: Evolution will only work if it's the first evolution option.

JaskRendix: Ok, I got it:
it doesn't work if monster A has:
evolution1 standard
evolution2 item
evolution3 item

Found and fixed another bug in has_path.py
Found and fixed another bug in evolve.py

I'm unsure if I addressed the correct one.

JaskRendix: Do I need to put the pop-up yes/no? In this way I can address through instance_id

kerizane:
Will need to amend https://github.com/Tuxemon/Tuxemon/pull/1450

tuxemon/item/effects/evolve.py
kerizane: You have a for loop but you always exit after the first entry.

You should move the success false return statement out of the for loop.

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +43,15 @@ class EvolveEffect(ItemEffect):
...
+            evolution = random.choice(choices).monster_slug
+            self.user.evolve_monster(target, evolution)
             return {"success": True}
+        elif self.monster_evolve != "random":
+            for evolution in target.evolutions:
+                if evolution.monster_slug == self.monster_evolve:
+                    self.user.evolve_monster(target, self.monster_evolve)
+                    return {"success": True}
         else:
             return {"success": False}
```
kerizane: Remove the final else block and deindent the final return statement. Otherwise it's theoretically possible for you to return None if line 53 is never true.

JaskRendix: Ok

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +43,14 @@ class EvolveEffect(ItemEffect):
...
-        if any(d["monster_slug"] == monster_evolve for d in target.evolutions):
-            self.user.evolve_monster(target, monster_evolve)
+        if self.monster_evolve == "random":
+            choices = [d for d in target.evolutions if d.path == d.path.item]
+            evolution = random.choice(choices).monster_slug
+            self.user.evolve_monster(target, evolution)
             return {"success": True}
-        else:
-            return {"success": False}
+        elif self.monster_evolve != "random":
```
kerizane: The Elif is pointless. It should just be an else clause.

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +43,14 @@ class EvolveEffect(ItemEffect):
...
+            for evolution in target.evolutions:
+                if evolution.monster_slug == self.monster_evolve:
+                    self.user.evolve_monster(target, self.monster_evolve)
+                    return {"success": True}
+        return {"success": False}
```
kerizane, line 56: This bit is right 👍

kerizane (changes requested):

tuxemon/item/effects/evolve.py
```diff
@@ -42,10 +43,14 @@ class EvolveEffect(ItemEffect):
     monster_evolve: str
 
     def apply(self, target: Monster) -> EvolveEffectResult:
-        monster_evolve = self.monster_evolve
-
-        if any(d["monster_slug"] == monster_evolve for d in target.evolutions):
-            self.user.evolve_monster(target, monster_evolve)
+        if self.monster_evolve == "random":
+            choices = [d for d in target.evolutions if d.path == d.path.item]
```
kerizane: This doesn't sound right...

`d.path == d.path.item`

kerizane (changes requested):

tuxemon/item/effects/evolve.py
JaskRendix: Originally it was "item", I wanted to centralize and remove the "item", I guess I need to put again "item"

JaskRendix: I hope it sounds better now.

kerizane (approved):

PR #1449, opened 2022-11-17
asiatic language (JP and CN)

JaskRendix:
Related to the issue #1445. This PR addresses the issue by adding a font for Chinese and another for Japanese (I noticed it on Weblate). We are talking about 10Mb of fonts, two files, a lot of glyph. Both are under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL).

I put zh_CN for Chinese, while ja for Japanese (as Weblate says)

Chinese: https://fonts.google.com/noto/specimen/Noto+Sans+TC
Japanese: https://fonts.google.com/noto/specimen/Noto+Sans+JP

This is the easy way to deal with the problem. 
There is a hard way, but it doesn't make sense.

Your choice.

Some screenshots of the font style (remember, it would be written in ideograms, not Latin characters):
chinese5
chinese4
chinese3
chinese2
chinese1

### DIFF
--- a/tuxemon/menu/menu.py
+++ b/tuxemon/menu/menu.py
@@ -192,11 +192,11 @@ class Menu(Generic[T], state.State):
     background_filename: Optional[str] = None
     menu_select_sound_filename = "sound_menu_select"
     if prepare.CONFIG.locale == "zh_CN":
-        font_filename = "NotoSansTC.ttf"
+        font_filename = prepare.FONT_CHINESE
     elif prepare.CONFIG.locale == "ja":
-        font_filename = "NotoSansJP.ttf"
+        font_filename = prepare.FONT_JAPANESE
     else:
-        font_filename = "PressStart2P.ttf"
+        font_filename = prepare.FONT_BASIC
     borders_filename = "gfx/dialog-borders01.png"
     cursor_filename = "gfx/arrow.png"
     cursor_move_duration = 0.20
--- a/tuxemon/menu/menu.py
+++ b/tuxemon/menu/menu.py
@@ -191,7 +191,12 @@ class Menu(Generic[T], state.State):
     # File to load for image background
     background_filename: Optional[str] = None
     menu_select_sound_filename = "sound_menu_select"
-    font_filename = "PressStart2P.ttf"
+    if prepare.CONFIG.locale == "zh_CN":
+        font_filename = "NotoSansTC.ttf"
+    elif prepare.CONFIG.locale == "ja":
+        font_filename = "NotoSansJP.ttf"
+    else:
+        font_filename = "PressStart2P.ttf"
     borders_filename = "gfx/dialog-borders01.png"
     cursor_filename = "gfx/arrow.png"
     cursor_move_duration = 0.20
--- a/tuxemon/prepare.py
+++ b/tuxemon/prepare.py
@@ -105,6 +105,11 @@ NATIVE_RESOLUTION = [240, 160]
 # Set the character limit for setting a player name.
 PLAYER_NAME_LIMIT = 30
 
+# Fonts
+FONT_BASIC = "PressStart2P.ttf"
+FONT_CHINESE = "NotoSansTC.ttf"
+FONT_JAPANESE = "NotoSansJP.ttf"
+
 # If scaling is enabled, scale the tiles based on the resolution
 if CONFIG.large_gui:
     SCALE = 2
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -1394,6 +1394,12 @@ Fonts
 adapted by JaskRendix, added symbols from Noto Emoji font (https://fonts.google.com/noto/specimen/Noto+Emoji)
 licensed under [SIL Open Font License, 1.1](http://scripts.sil.org/OFL)
 
+* ["Noto Sans Traditional Chinese"](https://fonts.google.com/noto/specimen/Noto+Sans+TC)
+licensed under [SIL Open Font License, 1.1](http://scripts.sil.org/OFL)
+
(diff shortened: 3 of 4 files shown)


kerizane:
Looks good! Can you show how they look with the actual glyphs.

The link mentioned styles, are we able to choose the style?

And what's the hard way?

kerizane:
The current size of the repo is 177MB, so an extra 10MB doesn't sound too bad.

tuxemon/menu/menu.py
```diff
@@ -191,7 +191,12 @@ class Menu(Generic[T], state.State):
     # File to load for image background
     background_filename: Optional[str] = None
     menu_select_sound_filename = "sound_menu_select"
-    font_filename = "PressStart2P.ttf"
+    if prepare.CONFIG.locale == "zh_CN":
```
kerizane, line 194: Can you make a function to get the filename?

kerizane: Or load it into a global const variable.

JaskRendix:
> And what's the hard way?

Inserting all the ideograms in our font.

> The link mentioned styles, are we able to choose the style?

No, I took only the boldest style possible (black).

> Looks good! Can you show how they look with the actual glyphs.

At the moment there is no much stuff in the Japanese PO, just fruits and few others.
Screenshot from 2022-11-18 08-46-43

DavidLiang1129:
> Related to the issue #1445. This PR addresses the issue by adding a font for Chinese and another for Japanese (I noticed it on Weblate). We are talking about 10Mb of fonts, two files, a lot of glyph. Both are under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL).
> 
> I put zh_CN for Chinese, while ja for Japanese (as Weblate says)
> 
> Chinese: https://fonts.google.com/noto/specimen/Noto+Sans+TC Japanese: https://fonts.google.com/noto/specimen/Noto+Sans+JP
> 
> This is the easy way to deal with the problem. There is a hard way, but it doesn't make sense.
> 
> Your choice.
> 
> Some screenshots of the font style (remember, it would be written in ideograms, not Latin characters): chinese5 chinese4 chinese3 chinese2 chinese1

Thanks for trying, I'm still happy because it proved that Tuxemon can at least display ideograms, can you keep it as a temporary solution for displaying ideograms?
——————————————————————
Chinese:
谢谢您的尝试，我仍然很高兴，因为它证明了Tuxemon至少可以显示表意文字，您可以保留它作为显示表意文字的临时解决方案吗?

tuxemon/menu/menu.py
kerizane: The issue was that you have the same code twice, here and in theme.

This change is a mistake because you call prepare.fetch again within the startup method.


I was asking for:
font_filename = get_font_filename(prepare.CONFIG.locale)

or 

font_filename = A_CONST_WE_IMPORTED_FROM_SOMEWHERE

kerizane (changes requested):

kerizane:
[DivedLiang0329](https://github.com/DivedLiang0329)

I will accept this, it just has a minor issue that needs fixing.

tuxemon/menu/menu.py
JaskRendix: ok

JaskRendix:
Is it ok the place for the constants?

tuxemon/prepare.py
```diff
@@ -105,6 +105,11 @@
 # Set the character limit for setting a player name.
 PLAYER_NAME_LIMIT = 30
 
+# Fonts
+FONT_BASIC = "PressStart2P.ttf"
```
kerizane, line 109:
```
    if prepare.CONFIG.locale == "zh_CN":
        FONT_FILENAME = prepare.fetch("font", "NotoSansTC.ttf")
    elif prepare.CONFIG.locale == "ja":
        FONT_FILENAME = prepare.fetch("font", "NotoSansJP.ttf")
    else:
        FONT_FILENAME = prepare.fetch("font", "PressStart2P.ttf")
```

JaskRendix: Yeah, you're right, I didn't understand

JaskRendix: Still here? Or I need to move somewhere else?

JaskRendix: I'm sorry, I'm confused, I have no idea what to do.

kerizane: No problem, let's just leave it how it was in your first commit.

kerizane: Actually... We may as well use these consts in theme.py as well.

JaskRendix: Like this?
```
    if prepare.CONFIG.locale == "zh_CN":
        font_filename = prepare.fetch("font", prepare.FONT_CHINESE)
    elif prepare.CONFIG.locale == "ja":
        font_filename = prepare.fetch("font", prepare.FONT_JAPANESE)
    else:
        font_filename = prepare.fetch("font", prepare.FONT_BASIC)
```

JaskRendix: Ok, for me no problem, just tell me how do you want theme.py

kerizane (approved):

DavidLiang1129:
> [DivedLiang0329](https://github.com/DivedLiang0329)
> 
> I will accept this, it just has a minor issue that needs fixing.

Thank you very much. May I ask what exactly is wrong? (I don't know programming yet, just curious)
————————————————————
Chinese:
非常感谢。我能问一下到底怎么了吗?(我还不懂编程，只是好奇)

kerizane:
He repeated the same logic in two different places. This means that if we change the logic we will have to do it in two places, and maybe we'll forget about one of them.

PR #1450, opened 2022-11-18
accept or refuse evolution

JaskRendix:
Following #1439 
implemented partial and instance_id to address the correct monster
tested with 6 monsters ready to evolve, I had a popup for each one (double checked with instance_id).

Screenshot from 2022-11-18 10-08-55

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1193,19 +1193,50 @@ class CombatState(CombatAnimations):
         self.players.remove(player)
 
     def evolve(self) -> None:
+        self.client.pop_state()
         for monster in self.players[0].monsters:
             for evolution in monster.evolutions:
-                # check the path field, path field signals evolution item based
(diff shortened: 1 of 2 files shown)


tuxemon/states/combat/combat.py
```diff
@@ -1193,19 +1194,53 @@ def trigger_player_run(self, player: NPC) -> None:
...
+                                ),
+                            ),
+                            ("no", T.translate("no"), self.negative_answer),
+                        ),
+                    )
+
+    def positive_answer(self, instance_id: uuid.UUID) -> None:
+        self.client.pop_state()
+        self.client.pop_state()
+        monster = self.players[0].find_monster_by_id(instance_id)
```
kerizane: On the right lines, but you could have just passed the monster in.

JaskRendix: No problem

kerizane (approved):

PR #1451, opened 2022-11-19
Dump locations

JaskRendix:
What to do with the dump locations? Work in progress.
I started working on these maps on September 2022.
I'm considering that the only way is a spin-off or a fork.

Dump of locations created or edited by me:
- Buddha Mountain
- Sphalian Town (Mountain town)
- Sphalian Town - Center
- Sphalian Town - Cafe
- Sphalian Town - House
- Route 7
- Peetlu'umo Town (Peetlu'umo is island in Yucatec Maya) - both the name of the city and the name of the island
- Peetlu'umo Center
- Peetlu'umo House 
- Peetlu'umo Inn
- Peetlu'umo Office
- Route 8

Still wondering about the name of the archipelago, probably I'll put south, so way way below Paper Town, so no-one can be afraid.


kerizane:
Good start! @Sanglorian ?

Sanglorian:
All looks good to me. I can't remember the original of Sphalian Town but Buddha Mountain was a test map I believe. I don't think they're needed in the Spyder campaign, but there's always room for more regions. (And at one point we were talking about optional maps, which you could choose from a list. Could be a good option here.)

JaskRendix:
Ah ok, so there is no way to connect these maps to the existing campaigns.

Pointless to keep it open, I wonder if there will be place for other campaigns in the future (2023 or 2024).
Especially since Xero is far from being complete and playable.

Sanglorian:
I might have confused things by saying they don't fit in the Spyder campaign -- there's room for any number of campaigns within the Tuxemon game. It wouldn't need a fork or spin off. If you have momentum to create a campaign (and from the looks of it you do) then I say go for it.

PR #1452, opened 2022-11-19
update Sea route

JaskRendix:
@Sanglorian 

I have updated the sea route, before there were 8 different tilesets, now only 3 and only CORE ones.
What do you think?

sea


kerizane (approved):

Sanglorian:
Looks good, thanks!

PR #1453, opened 2022-11-21
routed

JaskRendix:
@Sanglorian 

Still landlocked, no-one can access it.
Description about the waterfalls.
As much generic as possible with north/south/west/east, all depends if it'll be connected.
Also the new JSON with the encounters. Copy of route6, but still no idea what monster do you want to.

There is a teleport to the Dragon's Cave in the middle of the river, I didn't touch it, let me know if you want to remove it.

Are you interested in this version of routed?
If you're not, then no problem, I'm gonna trash it.

Spyder.world updated automatically, I guess it was stuck in an old version.

Update:
Waterfalls are now animated.

spyder_routed


kerizane (dismissed):

Sanglorian:
Cool - it's not needed for the Spyder campaign, but let's save it because I'm sure it'll come in handy in some campaign.

PR #1454, opened 2022-11-21
status grabbed and stuck + conditions

JaskRendix:
PR addresses some of the missing status and add conditions:
- created the values default_power and default_potency, so there no risk of getting half value each time a move is used;
- player with a grabbed or stuck monster is unable to switch monster as well as run (forfeit too) from wild monsters;
- added status.py and current_hp.py under tech conditions, the first one because of grabbed and stuck, the other mostly for testing, I left because it's useful for **diehard** status (conditions can simplify a lot the implementation of the missing statuses).

Big improvement since September, back then only 3 out of 22, now we are at 13 out of 22.
And this for only one campaign!

Tested + black.

Source: https://wiki.tuxemon.org/index.php?title=Category:Condition

Remember missing:
```
grabbed user: Bubble Trap
grabbed target: Bubble Trap, Clamp On, Constrict, Invictus, Overgrowth, Web
stuck user: Quicksand
stuck target: -


diehard user: Clamp On, Shuriken
diehard target: -
dozing user: -
dozing target: Petrify
noddingoff user: Hibernate, Sleep Bomb
noddingoff target: Sleep Bomb, Sleeping Powder
tired user: -
tired target: Feint, Starfall
charging user: Energy Field, Stampede, Surge
charging target: -
```

### DIFF
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -206,6 +206,26 @@ def simple_overfeed(
     return speed
 
 
+def simple_grabbed(monster: Monster) -> None:
+    for move in monster.moves:
+        if move.range.ranged:
+            move.potency = move.default_potency * 0.5
+            move.power = move.default_power * 0.5
+        elif move.range.reach:
+            move.potency = move.default_potency * 0.5
+            move.power = move.default_power * 0.5
+
+
+def simple_stuck(monster: Monster) -> None:
+    for move in monster.moves:
+        if move.range.melee:
+            move.potency = move.default_potency * 0.5
+            move.power = move.default_power * 0.5
+        elif move.range.touch:
+            move.potency = move.default_potency * 0.5
+            move.power = move.default_power * 0.5
+
+
 def escape(level_user: int, level_target: int, attempts: int) -> bool:
     escape = 0.4 + (0.15 * (attempts + level_user - level_target))
     if random.random() <= escape:
--- a/mods/tuxemon/db/technique/bubble_trap.json
+++ b/mods/tuxemon/db/technique/bubble_trap.json
@@ -3,7 +3,9 @@
   "accuracy": 0.75,
   "animation": "bomb_bright",
   "effects": [
-    "damage 100"
+    "damage 100",
+    "grabbed target",
+    "grabbed user"
   ],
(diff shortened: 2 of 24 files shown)


tuxemon/formula.py
```diff
@@ -206,6 +206,30 @@ def simple_overfeed(
     return speed
 
 
+def simple_grabbed(monster: Monster) -> None:
+    for move in monster.moves:
+        if move.range.ranged:
```
kerizane, line 211: This doesn't look right

tuxemon/formula.py
```diff
@@ -206,6 +206,30 @@ def simple_overfeed(
...
+            move.potency = move.potency * 0.5
+            move.power = move.power * 0.5
```
kerizane: This looks too permanent

tuxemon/formula.py
```diff
@@ -206,6 +206,30 @@ def simple_overfeed(
...
+        elif move.range.reach:
+            move.potency = move.potency * 0.5
+            move.power = move.power * 0.5
+        else:
+            return
```
kerizane: Don't think you need to return here

tuxemon/formula.py
```diff
@@ -206,6 +206,30 @@ def simple_overfeed(
...
+
+
+def simple_stuck(monster: Monster) -> None:
+    for move in monster.moves:
+        if move.range.melee:
```
kerizane, line 221: All comments in last section apply here

tuxemon/formula.py
JaskRendix: Yeah, it looks, because at the end we are dealing with parameters different from speed, etc. I tested, and after the battle, the values return as before. I was thinking about a set_stats for techniques where the focus was power and potency (like monster.py).

The tricky thing is that power and potency are written in the JSON of the technique, so I'm obliged to insert the slug of the technique. While monster stats are generated based on category and level. Initially I wanted to do it, but I didn't end well #1415

tuxemon/formula.py
JaskRendix: it addresses only the moves in these categories

kerizane (approved):

PR #1455, opened 2022-11-21
rematch

JaskRendix:
PR addresses the implementation of rematch: #1429 was a first step in this direction.

The situation now:
you fight against Trainer X (Rockitten lv5) in Route 2, it's done, you'll never meet Trainer X again.

This system obliges modders to create a ton of NPCs. The goal is making possible to fight again the NPC in another map later in the game (possibly with more powerful monster).

This situation tomorrow:
you fight against Trainer X (Rockitten lv5) in Route 2;
you fight against Trainer X (Rockitten lv32) - or the evolved version - in Route 6;

Why? Because creating new NPCs require a lot of work, especially choosing the monsters, if we can utilize again the "old" ones (that we meet at the beginning) later in the game, then that's a win win. We can easily double down the trainer battles.

This will be the string to put inside the maps to trigger the creation of the same NPC later (in other maps):
`    <property name="cond1" value="is battle_is NPC_SLUG,won"/>`
so if you don't fight it, it'll never create the NPC later.

Tested + black, it works.

Doubts about the parameters (values): average better? lowest? opponent monster level?
```
let's say you fight the same day TRAINER X, single monster, lv 10
diff_date = 0
high = low + 1

low = highest level of the party

let's say out of 3 monster (party), the strongest is 15
it means:
high = 15 + 1
high = 16

so when you fight again, the monsters will be in this range randint(15,16)
```

### DIFF
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -28,6 +28,7 @@
 
 from __future__ import annotations
 
+import datetime as dt
 import logging
 import random
 from typing import TYPE_CHECKING, NamedTuple, Optional, Sequence, Tuple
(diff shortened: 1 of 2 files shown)


tuxemon/formula.py
```diff
@@ -237,3 +239,40 @@ def battle_math(player: NPC, output: str) -> None:
...
+    if diff_date == 0:
+        high = low + 1
+    elif diff_date < 10:
+        high = low + 2
+    elif diff_date < 50:
+        high = low + 3
+    elif diff_date > 51:
+        high = low + 5
+    # different conditions related to the output
+    if output.won:
```
kerizane: This Elseblock is identical?

JaskRendix: I'm going to put an else, it's better.

tuxemon/formula.py
```diff
@@ -237,3 +239,40 @@ def battle_math(player: NPC, output: str) -> None:
...
+        high = low + 1
+    elif diff_date < 10:
+        high = low + 2
+    elif diff_date < 50:
+        high = low + 3
+    else:
+        high = low + 5
+    # different conditions related to the output
+    if output.won:
+        monster.level = random.randint(low, high)
```
kerizane: My point was that you have the exact same behaviour on lines 266, 268 and 270, so what's the point of the if/else?

JaskRendix: I thought about differentiate for output, but at the end is going to happen only after the output.won, so yeah, I'm going to remove it.

kerizane (approved):

PR #1456, opened 2022-11-22
simplify maps

JaskRendix:
Progressing the task of simplify maps (like the sea route).

I removed tons of unused tilesets. Some up to 20 tilesets, when they needed only 5 (I think with the centralization we can arrive to 2, maybe 3). By the way, the only two maps with no unused tilesets are bedroom and downstairs.
 
Xero: removed unused tilesets from towns and routes;
Spyder: removed unused tilesets and semplification of all the open space maps (towns and routes), everything is based on COREs tilesets now.

Moreover, I moved some pieces of inside the respective CORE_, so we can consider to delete these in the following step and mass replace the still existing PNGs.

Some examples:
- doors.png inside core_set pieces.png (because it's better to have the closed door in the same CORE_building)
- plants.png inside core_set pieces.png (the only one)
- kitchen.png inside core_set pieces.png (fridge, etc)
- electronics.png inside core_set pieces.png (same as above)
- interior tiles mike inside core_set pieces.png (some used tiles)
- buddha_statues inside core_outdoor.png (all the buddhas)
- Truck-red-all inside core_outdoor.png (centralization)
- city_tiles inside core_outdoor.png (statue, fountain, big bush, the only elements used)
- Car,_Black_by_Isaiah658 inside core_outdoor.png  (centralization)
etc.

There was a file with 2 dots (..tsx), I fixed and Tile automatically updated it.

Moreover, there were two Pacheesi Tiles (identical), one in maps folder (?), the other in tileset.
-> I moved the file from maps to tileset and replaced the one there.

Moved rubberduck_outdoor.png inside tilesets (it's a tileset);
tt_forest_route1.png deleted (maps), since it was an export (image) of the map;


kerizane (approved):

Sanglorian:
Wow, very useful consolidation, thank you.

PR #1465, opened 2022-11-30
Random AI, next_use and skip

JaskRendix:
PR addresses a bug I discovered a while ago #1391, fixes it and introduces a feature.

Before a random monster as well as a trainer's monster could use the same move like 7 times in a row (could be more, but I stopped testing after 7).
_(note: five times in a row if the random choice granted the same move)_
The point is that they had the advantage of using moves without recharging.
So the most powerful ones (the ones which take sometime 3 turns to recharge) could be used many times more.
Uneven advantage.

The PR inserts the recharging as well as the mechanism that impedes of being utilized in a row like before. Now before using a techniques that takes 3 turns to recharge, they need to wait 3 turns before this one is again available in the random list.

This pushed me to create a new technique called **skip**.
Skip will be used automatically by the adversary monster, while it'll appear in the player menu due certain conditions and it'll disappear once used.

This solves also the issue that if you're in a battle against a trainer, if your last monster is without moves, then you are forced to close abruptly the windows since you are stuck (no other monsters, no swap and back then no forfeit).

I'm not convinced about the IFs in the ai.py, especially the subtraction, at the moment I don't see a better alternative, maybe someone can see a way to beautify it.
Black + tested

Here the message that will appear when a monster uses skip, in this case the enemy after using his both moves, it was forced to use skip.
Screenshot from 2022-11-30 13-11-27

Where it'll appear skip in the monster menu:
Screenshot from 2022-11-30 14-03-14

### DIFF
--- a/tuxemon/ai.py
+++ b/tuxemon/ai.py
@@ -87,4 +87,12 @@ class RandomAI(AI):
         opponents: Sequence[Monster],
     ) -> Tuple[Technique, Monster]:
         # TODO: healing and whatnot
-        return choice(monster.moves), choice(opponents)
+        filter_moves = []
+        for mov in monster.moves:
+            if mov.next_use <= 0:
+                filter_moves.append(mov)
+        if not filter_moves:
+            skip = Technique("skip")
+            return skip, choice(opponents)
+        else:
+            return choice(filter_moves), choice(opponents)
--- a/mods/tuxemon/db/technique/skip.json
+++ b/mods/tuxemon/db/technique/skip.json
@@ -1,0 +1,25 @@
+{
+	"animation": null,
+	"effects": [],
+	"flip_axes": "",
+	"icon": "",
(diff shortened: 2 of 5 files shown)


tuxemon/ai.py
```diff
@@ -87,4 +87,12 @@ def make_decision(
         opponents: Sequence[Monster],
     ) -> Tuple[Technique, Monster]:
         # TODO: healing and whatnot
-        return choice(monster.moves), choice(opponents)
+        filter_moves = []
+        for mov in monster.moves:
+            if mov.next_use <= 0:
+                filter_moves.append(mov)
+        if len(monster.moves) - len(filter_moves) == len(monster.moves):
```
kerizane: `if not filter_moves:`

kerizane:
Looks good!

kerizane (approved):

PR #1466, opened 2022-12-01
add item NPC + random AI

JaskRendix:
PR addresses the addition of the slug parameter to add_item. Until now it was possible to add items only to player, now to NPCs too.

I got curious, so I added and I resorted some things in random AI:
- a couple of functions to individuate the strongest and weakest monster in the NPC's party;
- create **is_item_sort** in npc.py, a simple bool, so we can check if an item is in a certain category (sort);
- doubled **make_decision** in **make_decision_wild** and **make_decision_trainer**, so I can avoid to import combatstate inside ai (creating problem of circular importation -> Monster, test_monster and test_catch_monster); in this way we can define specific behavior for wild monsters (for which I have an idea); technically I could use random_dummy slug to define it, but it would be resulted as hardcoding, this way is more fluid.

I set the 15% as trigger for the potion, it can be changed.
At the end (if approved the item to NPC), modders can choose to spice up fights against NPCs by giving more items to them.
I tested a fight against a trainer with potions, and that's really another level, it can be handy for bosses, etc.. 

`<property name="act4" value="add_item ITEM,QUANTITY,NPC"/>`

Tested + black

Update:
added a fix about target/user, I noticed only now that when an NPC sent out a monster was:
MONSTER sent out NPC
now it's correct:
NPC sent out MONSTER

### DIFF
--- a/tuxemon/ai.py
+++ b/tuxemon/ai.py
@@ -102,15 +102,15 @@ class RandomAI(AI):
         opponents: Sequence[Monster],
     ) -> Tuple[Union[Monster, NPC], Union[Item, Technique], Monster]:
         """
-        Trainers battles.
+        Trainer battles.
         """
         if self.check_strongest(user, monster):
             if len(user.inventory) > 0:
                 inventory = list(user.inventory)
-                for i in inventory:
-                    if user.is_item_sort(i, "potion"):
-                        if self.check_hp_potion(monster):
-                            item = self.item_healing(user, i)
+                for item_slug in inventory:
+                    if user.is_item_sort(item_slug, "potion"):
+                        if self.need_potion(monster):
+                            item = self.item_healing(user, item_slug)
                             return user, item, monster
         technique, target = self.track_next_use(monster, opponents)
         # send data
@@ -178,7 +178,7 @@ class RandomAI(AI):
         else:
             return False
 
-    def check_hp_potion(self, monster: Monster) -> bool:
+    def need_potion(self, monster: Monster) -> bool:
         """
         It checks if the current_hp are less than the 15%.
         """
--- a/tuxemon/ai.py
+++ b/tuxemon/ai.py
@@ -29,17 +29,17 @@
 #
 from __future__ import annotations
 
-import logging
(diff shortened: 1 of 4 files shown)


tuxemon/ai.py
```diff
@@ -81,12 +95,49 @@ def make_decision(
...
+        user: NPC,
+        monster: Monster,
+        opponents: Sequence[Monster],
+    ) -> Tuple[Union[Monster, NPC], Union[Item, Technique], Monster]:
+        """
+        Trainers battles.
+        """
+        if self.check_strongest(user, monster):
+            if len(user.inventory) > 0:
+                inventory = list(user.inventory)
```
kerizane, line 109: what is the type of the inventory? Do we actually need to cast it into a list?

tuxemon/ai.py
```diff
@@ -81,12 +95,49 @@ def make_decision(
...
+                for i in inventory:
```
kerizane: `for item in inventory`
`for item_slug in inventory`

`i` is usually used for indexes

e.g.
`for i in range(10):`

tuxemon/ai.py
```diff
@@ -81,12 +95,49 @@ def make_decision(
...
+                    if user.is_item_sort(i, "potion"):
+                        if self.check_hp_potion(monster):
```
kerizane: I find `check_` to be a bit ambiguous as a name.

`needs_potion` ?

kerizane (changes requested):

tuxemon/ai.py
JaskRendix: Agree

tuxemon/ai.py
JaskRendix: item_slug sounds fine

tuxemon/ai.py
JaskRendix: inventory <class 'dict'>
Cast into a list because I thought it was more handy than loop in the dict.

kerizane (approved):

kerizane (approved):

PR #1471, opened 2022-12-09
Fix crash when loading plugins

NeonFlames:
When loading plugins if the path contained "tuxemon" before the actual game files the game would try to load a module that does not exist


kerizane (approved):

kerizane:
Well done for figuring this out.

PR #1473, opened 2022-12-11
fixing lifeleech status

### LINKED ISSUE #1472: Leech should not work after the opposing Tuxemon has fainted
Right now, when you fight a battle and the enemy Tuxemon uses a life leech move, it will leech HP even after the opposing Tuxemon has fainted.

I'm not necessarily opposed to keeping the move as it is, but where is your life being leeched to? If we do decide to keep it this way, maybe have the HP leech be decreased when the opposing Tuxemon faints?

JaskRendix:
Related to the recent  #1472 

Added check to the combat.py, so when the NPC monster faints, there is a check about the remaining monster of the PC (the one active in the battle);
if there is status_grabbed or status_lifeleech, then the status will be automatically cleaned up;

So we can solve 1 issue, 2 potential issues and cleaning up some unused parameters.

tested + black

update: moved most of the check in combat.py (no states) as check_status_connected, I want to avoid pointless ifs inside the other combat.py (states)

### DIFF
--- a/tuxemon/combat.py
+++ b/tuxemon/combat.py
@@ -49,6 +49,17 @@ def check_status(monster: Monster, status_name: str) -> bool:
     return any(t for t in monster.status if t.slug == status_name)
 
 
+def check_status_connected(monster: Monster) -> bool:
+    """
+    Statuses connected are the ones where an effect is present only
+    if both monsters are alive (lifeleech, grabbed).
+    """
+    if check_status(monster, "status_grabbed"):
+        return True
+    elif check_status(monster, "status_lifeleech"):
+        return True
+
+
 def fainted(monster: Monster) -> bool:
     return check_status(monster, "status_faint") or monster.current_hp <= 0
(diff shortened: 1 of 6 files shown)


tuxemon/combat.py
```diff
@@ -49,6 +49,17 @@ def check_status(monster: Monster, status_name: str) -> bool:
...
+def check_status_connected(monster: Monster) -> bool:
+    """
+    Statuses connected are the ones where an effect is present only
+    if both monsters are alive (lifeleech, grabbed).
+    """
+    if check_status(monster, "status_grabbed"):
+        return True
+    elif check_status(monster, "status_lifeleech"):
+        return True
+
```
kerizane, line 63: else return false

kerizane (approved):

kerizane (changes requested):
else return false

kerizane (approved):

PR #1476, opened 2022-12-12
Cut down on fresh start loading time

NeonFlames:
The locale cache was being built twice on launch of the game, once by prepare.py and once by the db module.


xirsoi (approved):
Looks good!

PR #1486, opened 2022-12-16
fix weight height monsters

JaskRendix:
Hi @sanglorian
Here the complete weight and height for all the 230 monsters, let me know what do you think.

I followed the same pattern as per the moves, I looked in a common value among the shape + type (eg earth + hunter), if the combination wasn't present, then I looked only the shape (eg hunter).

Where there was an evolutionary line (basic + stage1 or stage2, eg rockitten evolutionary line), I took into consideration the BMI (relationship between weight and double height); it was handy because considering the increment of the weight, I got a realistic height.

### DIFF
--- a/mods/tuxemon/db/monster/abesnaki.json
+++ b/mods/tuxemon/db/monster/abesnaki.json
@@ -24,8 +24,8 @@
     ],
     "possible_genders": ["neuter"],
     "txmn_id": 99,
-    "height": 0,
-    "weight": 0,
+    "height": 104,
+    "weight": 14,
     "catch_rate": 120.0,
     "lower_catch_resistance": 0.95,
     "upper_catch_resistance": 1.05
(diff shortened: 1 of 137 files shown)


Sanglorian:
Great idea, thank you. I have some tweaks to the numbers in green in the attached -- based on what the monster is based on, or to fit with my mental image of the monsters. 

Also, sorry I didn't think of it earlier, but Lunight and Solight are a matching pair. Would it be possible for them to be next to each other in the ID numbering? 

[weight-height.ods](https://github.com/Tuxemon/Tuxemon/files/10249902/weight-height.ods)

JaskRendix:
No problem, moved Spoilurm at 214, now Lunight 229 and Solight 230.

Everything is updated, thanks again.

Updated file with the new txmn_ids: [weight-height.ods](https://github.com/Tuxemon/Tuxemon/files/10251226/weight-height.ods)

kerizane (approved):

PR #1489, opened 2022-12-17
delete maps

JaskRendix:
@Sanglorian these maps are duplicate of already existing maps, I would like to know what do you think about deleting:

- cotton_town_spyder is the copy of cotton_town in Xero (unused);
- daycare_centre is the copy of daycare in Xero (unused);
- paper_scoop and paper_town don't exist in Xero, if someone wants to add those, it takes less effort to copy and past the Spyder version than building up based on unused maps;
- tuxe_mart.tmx is the old version of the new tuxe_mart_taba;

It can confuse newcomers too.
Let me know what do you think about.
Do you want to signal some maps that can be deleted too?


Sanglorian:
That's fine by me, just worth checking with @Qiangong2 too

Qiangong2:
The Xero campaign follows the exact same map as the Spyder campaign, so the maps should have the same base. Overall, I don't see any issue with the deletions.

Qiangong2 (dismissed):

kerizane:
Merge conflict

JaskRendix:
I was unable to solve the conflict for tuxe_mart, I'm going to open a separate PR.

kerizane (approved):

PR #1492, opened 2022-12-20
evolution gender and mixed

JaskRendix:
PR addresses the implementation of the evolution by gender and mixed:
- updated the **evolution_confirmation** (added into > the monster), so the user is aware of what's coming, he can refuse or accept it;
- updated the code of the message by putting the monster names **uppercase** (screenshot below);
- added the **evolved** = Monster, it was necessary for getting the evolution name;
- added the **evolved** monster into the **partial** of **positive_answer**, so it passes those info too;
- removed the **filters in positive_answer**, the job has been already done above (evolve());
- updated **find_tech** (now it's a **bool**);
- created **find_type** (bool) -> it was necessary;
- updated the **evolution types** inside **db.py** (added tech, location and type);
- added another **validator** (technique) because of tech;
- about **location** in **db.py**, it's a **bool**, true means inside, false means outside, it's written inside the description too;
- added **StatType** in db.py
- added **return_stat** in monster.py
- now there is **stat**, with **stat1** and **stat2**, so we can play with 2 statistics (like katacoon armour and melee), everything is automatized, the previous two (stattype and return_stat) where necessary to avoid hardcoding;
- getting rid of **mixed** parameter, replaced the last one (chromeye) with **season**; 

I decided to add other types so we can cover more easily different outputs, I have no idea if it'll appear another monster with evolution outside, or another techniques, or another element, etc.

Gender is working without issues;
Type is working without issues;
Location is working without issues;
Tech is working without issues;
Season is working without issues;
Stat is working without issues;

This closes the evolution, now we are covered on all the monsters in the db.

Screenshot from 2022-12-20 10-34-29

tested + black


JaskRendix:
@Sanglorian this closes the evolution issue, I implemented the various systems, for Chromeye I opted for seasonality, so people of the different side of the world (north/south hemisphere) can get opposite monster. Let me know what do you think or if you have other monsters with particular systems.

tuxemon/npc.py
```diff
@@ -871,23 +871,38 @@ def set_party_status(self) -> None:
...
         Parameters:
             tech: The slug name of the technique.
-
-        Returns:
-            Technique found.
-
         """
+        moveset = []
+        output = []
         for technique in self.monsters:
```
kerizane, line 881:
```
for technique in self.monsters:
    for move in technique.moves:
        if move.slug == tech:
            return True
return False 
```

tuxemon/npc.py
```diff
@@ -871,23 +871,38 @@ def set_party_status(self) -> None:
...
+                moveset.append(moves.slug)
+        output = [*set(moveset)]
+        if len(output) > 1:
+            if tech in output:
+                return True
+            else:
+                return False
 
-        return None
+    def find_type(self, element: ElementType) -> bool:
```
kerizane: has_type

tuxemon/npc.py
```diff
@@ -871,23 +871,38 @@ def set_party_status(self) -> None:
         self.game_variables["party_level_highest"] = level_highest
         self.game_variables["party_level_average"] = level_average
 
-    def find_tech(self, tech: str) -> Optional[Technique]:
+    def find_tech(self, tech: str) -> bool:
```
kerizane: has_tech

mods/tuxemon/db/monster/vivipere.json
```diff
@@ -36,10 +36,10 @@
             "gender": "female"
         },
         {
-            "path": "mixed",
+            "path": "location",
             "at_level": 0,
             "monster_slug": "vivitrans",
-            "mixed": "inside"
+            "location": true
```
kerizane: `"location": true`

This is a little confusing. Can we implement this some other way?

JaskRendix:
@Sanglorian I guess it's missing only your approval

bitcraft:
@kerizane does this look ok?

Sanglorian:
> @Sanglorian I guess it's missing only your approval

Looks great from a design point of view. Thanks!

mods/tuxemon/db/monster/vivipere.json
JaskRendix: Yeah, I'll use inside instead of location:
"inside" = true

tuxemon/db.py
```diff
@@ -250,9 +263,29 @@ class MonsterEvolutionItemModel(BaseModel):
...
+    stat2: Optional[StatType] = Field(
+        None, description="Stat parameter stat2 < stat1."
+    )
+    tech: Optional[str] = Field(None, description="Technique parameter.")
+
+    @validator("tech")
+    def technique_exists(cls, v):
+        if has.db_entry("technique", v):
+            return v
+        raise ValueError(f"no resource exists in db: {v}")
```
bitcraft: this error message is confusing

JaskRendix: I agree, blindly copy pasted the other one, I'm going to update the other as well.

JaskRendix: Changed a bunch of other resources too, let me know if it's ok for you

JaskRendix: Ah, I did black + isort as usual, and I included the small fix in prepare.py

tuxemon/states/combat/combat.py
```diff
@@ -1215,48 +1215,78 @@ def evolve(self) -> None:
...
-                            ),
-                        )
+                        self.question_evolution(monster, evolved)
+                elif evolution.path == "gender":
+                    if evolution.at_level <= monster.level:
+                        if evolution.gender == monster.gender:
+                            self.question_evolution(monster, evolved)
+                elif evolution.path == "element":
+                    if evolution.at_level <= monster.level:
+                        if self.players[0].has_type(evolution.element):
```
bitcraft, line 1229: just confirming how this works...  if a monster is ready to evolve along the "element" path, this check looks at all the monsters in the party?  so the requirement to evolve here is that *any* monster in the party will enable it, not the monster itself?  im a bit fuzzy here, just trying to understand.

bitcraft: same question for the "tech" evolution below.

JaskRendix: Exactly, @Sanglorian confirmed that it was related to any monster in the party as well as that there is no minimum level of evolution.
So if you have a Vivipere in your party (any level) and after a battle you have a monster wood type (element) or a monster with flamethrower (tech) or you're inside (location), Vivipere evolves.

bitcraft: ok.  so this is expected behavior?

JaskRendix: Yes

bitcraft (approved):
LGTM

PR #1496, opened 2022-12-21
add item category + items

JaskRendix:
PR addresses the addition of **category** (item) and **new items**:

**category** because we need more options to filter the items, sort and type aren't enough (it's also a mess to add new labels), we need a field free from other mechanism, so we can also easily implement other functions.

**categories**:
```
none
edible
badge
booster
fossil
morph
technique
revive
potion
fish
phone
capture
```

If someone has suggestions, feel free to share.

new items:
```
mm_earth
mm_fire
mm_metal
mm_water
mm_wood
+
tm_earth
tm_fire
tm_metal
tm_water
tm_wood
+
nu_phone
```

other **modifications**:
- deleted **red-book.png** in favor of **tm_fire.png** (identical png);
- deleted **purple-book.png** (unused) in favor of **tm_water.png** (identical png);
- updated image of **tm_avalanche** (earth) so **tm_earth.png**;
- updated image of **tm_blossom** (wood) so **tm_wood.png**;
- renamed the effect **learn** in **learn_tm**, because (following point);
- created the new effect **learn_mm** (it chooses randomly a technique from all the technique of the same type (eg. metal), then it removes the moves of the monster (to avoid duplicated), so it keeps giving new moves until it fails (when there are no more moves);
- added **ItemType** in item.py, it was missing, now it's correctly imported;
- added **Nu Phone**, the png was already present inside the folder, renamed the png and created the json;

As usual:
credits updated
tested + black
source: https://wiki.tuxemon.org/Category:Item

@sanglorian let me know if it's ok for you. It was tough the module in the beginning, but I'm happy I have figured out. What do you had in mind for the TMs? A single TM for each Type? Another randomization? Let me know.


tuxemon/item/effects/learn_mm.py
```diff
@@ -0,0 +1,45 @@
...
+        # monster moves
+        moves = []
+        for tech in target.moves:
+            moves.append(tech.slug)
+        # remove monster moves from type moves
+        set1 = set(filters)
+        set2 = set(moves)
+        res = list(set1 - set2)
+        # add a random move from what remains
+        if len(res) > 0:
```
kerizane: `        if res:`

kerizane: It is generally best to do 

`        if res:`
than
`        if len(res) > 0:`

Mainly because it's more concise.

kerizane (changes requested):

Sanglorian:
Thanks @JaskRendix . 

So, there should be one TM for each technique that's available to be learned that way, and one MM for each technique that's available to be learned that way. At the moment I don't think we've figured out which techniques should be learnable via TM or MM, so it might make sense to make a random selection or just all of them. 

In terms of learning the technique, it should fill an empty slot if there is one, and otherwise the player should get a choice of which existing technique to replace. 

Hope that helps! Let me know any questions.

JaskRendix:
Ok, that' an interesting thing to set up in another PR.
At the moment I added three things:
- created forgetting_pill, a new item;
- created a new condition called **moves** (so it's unable to add a technique if the monster has more than 4 moves);
- created a new effect called **forget_tm**, the effect makes forget a specific technique (by slug) or a random one (if using "random" instead of the tech slug);

I chose 4 as limit for move because the majority of the monster has three moves in the moveset; now I have no idea if it'll be 5 or 3, I started from this number, so we can set up some check and replacement.

JaskRendix:
Based on a reply in the other PR, I would say that probably the pill will not be added in the Spyder scenario, but it can come up handy in other ones (when, if, there will be other "playable" scenarios).

JaskRendix:
It was better to fix the problem about the replacement sooner, already two PRs, and already two reminders.

So:

- inserted a **check** inside **items/init.py** this because there will be object that don't use the use_success and use_failure pop_up;
- inside the check there is a connection to **check_max_moves** in **npc.py**, this one will open a **choice dialog** to decide which technique delete;
- inserted **MAX_MOVES** inside the **monster.py**, so we have a constant like **MAX_LEVEL**;
- chose this way because it wasn't possible to open a dialogue or a choice inside the effect file (eg learn_mm), the structure blocks everything until the **show_item_result_as_dialog** in **init.py**
- create the condition (items) **max_moves**, so there will be a check, the items with learn_mm and learn_tm can't be used if the monster has 5 moves, it's not the case, but preventive; I put **> max_moves** instead of >=max_moves, because in the second case there is the possibility to overwrite another move;

**Reason for MAX_MOVES = 4:**
Here: https://github.com/Tuxemon/Tuxemon/pull/1498#issuecomment-1363420627 first line.
I agree.

This is what's going to appear if there are already 4 moves. Once someone has dismissed the dialogue, then he can choose which one to overwrite.
Screenshot from 2022-12-23 15-50-43

bitcraft (dismissed):
LGTM

JaskRendix:
I made an update.

The **check_max_moves** in **npc.py** has been split.
Now there are **check_max_moves** and **overwrite_technique**.
The reason of this change is because I want to avoid a duplicate since I'm working on the overwriting method after leveling up and I need the choice dialogue separated from **check_max_moves**.

The PR will follow this one.

Tested, black and isort.

Update:
deleted forget_tm, lethe_pill.json, the lines inside the PO file and the credits line.
Eventually I'll open a separate PR to add a similar item.

Update x2:
added boilerplate (new files)

Update x3:
solved the conflict

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -99,6 +99,91 @@ msgstr "Rhincus Fossil"
...
+msgstr "Nu Phone"
+
+msgid "nu_phone_description"
+msgstr "You can do a lot with this, e.g. phone home."
+
+msgid "lethe_pill"
+msgstr "Lethe Pill"
+
+msgid "lethe_pill_description"
+msgstr "It resembles an eye and it makes forget about a random technique."
```
kerizane: Bad grammar

"It resembles an eye and it makes a monster forget about a random technique."

tuxemon/item/conditions/max_moves.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Accepts a single parameter and returns whether it is applied.
+
+    """
+
+    name = "max_moves"
+    max_moves: str
+
+    def test(self, target: Monster) -> bool:
+        if len(target.moves) > MAX_MOVES:
```
kerizane: "is max_moves" makes me expect the opposite logic.

tuxemon/item/conditions/max_moves.py
```diff
@@ -0,0 +1,25 @@
...
+class MaxMovesCondition(ItemCondition):
+    """
+    Checks max moves against the creature's moves.
+
+    Accepts a single parameter and returns whether it is applied.
+
+    """
+
+    name = "max_moves"
+    max_moves: str
```
kerizane: What does this param do?

JaskRendix: happy new year, yeah, cleaned up.

tuxemon/item/conditions/max_moves.py
JaskRendix: opposite logic updated

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
JaskRendix: right, done

kerizane (approved):

PR #1498, opened 2022-12-22
action "remove technique" + NPC

JaskRendix:
PR addresses and solves a problem related to #1496 

Still isn't clear of the number max of moves for monster, at the moment the rule is 3 moves per monster, the max can easily be 4, 5 or 6, but no more than 7, because if not the list inside the combat menu becomes a mess.

This requires a series of way to forget moves. In #1496 there is a pill that makes forget a random move, here there is a NPC called Mr Lethe (the guy of the pill) that you can easily find inside scoops.

There are two action:
- **replacing** where you choose the monster, only the monsters with more than 1 move can be selected;
- **removing** where it opens up a choice menu (screenshot below) and you choose what technique do you want to forget (same principle here, you are unable to forget the last technique);

The dialogue is the first thing that came up. If someone has suggestions, feel free to suggest.

@Sanglorian are you ok if I populate certain scoops with Mr Lethe and company? It's another way to forget moves.

### DIFF
--- a/tuxemon/event/actions/removing_move.py
+++ b/tuxemon/event/actions/removing_move.py
@@ -1,0 +1,93 @@
+from __future__ import annotations
+
+import logging
+import uuid
+from functools import partial
+from typing import NamedTuple, Optional, Union, final
+
+from tuxemon.event import get_npc
+from tuxemon.event.eventaction import EventAction
+from tuxemon.locale import T
+from tuxemon.npc import NPC
+from tuxemon.states.dialog import ChoiceState
+from tuxemon.tools import open_choice_dialog
+
+logger = logging.getLogger(__name__)
+
+
+class RemovingMoveActionParameters(NamedTuple):
+    npc_slug: Union[str, None]
+
+
+# noinspection PyAttributeOutsideInit
+@final
+class RemovingMoveAction(EventAction[RemovingMoveActionParameters]):
+    """
+    It removes a chosen technique from the monster moves.
+
+    Script usage:
+        .. code-block::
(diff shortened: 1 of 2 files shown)


Sanglorian:
Thanks, that looks good. The maximum number of techniques should be four. 

Ideally I'd also like it if, when gaining a new technique, you could overwrite an existing technique - but of course that could come later. 

I like the idea of Mr Lethe -- we probably only need him in one Scoop store to begin with.

JaskRendix:
> Ideally I'd also like it if, when gaining a new technique, you could overwrite an existing technique - but of course that could come later.

Yeah, you already wrote it in the other PR, I'm working on it.

JaskRendix:
> Ideally I'd also like it if, when gaining a new technique, you could overwrite an existing technique - but of course that could come later.

Yeah, you already wrote it in the other PR, I'm working on it.

I don't know honestly, I'm considering to close the PR, at the end who cares if it'll be implemented the automatic replacement?

Update: deleted the npc JSON + map actions, I'll leave only the two actions, at the moment I'm not going to populate the locations.

JaskRendix:
Look, I added the automatic delete choice after immediately after the use of the TM or MM item.

bitcraft:
Due to another PR being merged in, the actions now have a slightly more concise format.  Please take a look at them and make the changes required to fit.  If you have a problem with this, I can help.

JaskRendix:
Thank for merging the other PR. I'm going to close this PR because I'm not convinced about the system, I need to think about

PR #1501, opened 2022-12-24
tech menu

JaskRendix:
PR addresses the Technique screen + some small things:

As you can see, there is description and techniques;
- description because it was awful as second dialogue (after the infos);
- techniques because it was missing, there is no other place (excluding the battle screen, where you can see the monster techniques);

Screenshot from 2022-12-24 09-46-42
The techniques are sorted alphabetically.
Screenshot from 2022-12-24 09-42-41
Each technique screen shows only the techniques of the selected monster.

What else? Yeah, it was missing Aether, Glitch and Normal inside the PO file, so it can be translated automatically.

I left **usable_in** as in items, but I transformed in a simple bool. The bool is False as default. In a future we can integrate the possibility to use the technique on items, on monster, etc. but it wasn't the case right now. So if you try to use the techniques, this is the message: 
Screenshot from 2022-12-24 10-09-18

black + tested

@Sanglorian if you have ready a different image (instead of the bag), then feel free to drop a link, so I can update and replace it after.


Sanglorian:
Thanks this looks great! That bag image is the right one to use, thanks for asking.

kerizane:
Might be tricky, but I would use the image of the monster.

JaskRendix:
Yeah, thanks for the suggestion, no way to tell how much time to do it

JaskRendix:
@kerizane  Let me know if it's acceptable for you

Screenshot from 2022-12-25 08-17-09

kerizane:
Looks good! @Sanglorian ?

Sanglorian:
Thanks! Looks good. At some point we (the community) should write descriptions for each technique. Would be nice to display them here as well.

And sorry I got confused re the bag. Longer term we could remove the image and show nothing in particular there.

JaskRendix:
Perfect, updated and added the field description inside technique.py, so we have already done a step in the right direction!

bitcraft:
Looks good so far.  Please address the merge conflict and I will review it further.

JaskRendix:
Resolved the conflict.

Update again, now it works. A couple of () was missing as well as other small things related to #1330

tuxemon/db.py
```diff
@@ -431,6 +431,9 @@ class TechniqueModel(BaseModel):
     types: Sequence[ElementType] = Field(
         [], description="Type(s) of the technique"
     )
+    usable_in: bool = Field(
```
bitcraft: this could probably be clarified some

tuxemon/menu/interface.py
```diff
@@ -141,6 +141,48 @@ def in_focus(self, value: bool) -> None:
         self._in_focus = bool(value)
 
 
+class MenuTechnique(Generic[T], Sprite):
```
bitcraft: Is there no way that you can do this without duplicating the contents of another class?  last i knew, there was some work being done to remove the menus and replace them with another lib.  with that in mind, we should avoid adding obsolete code

JaskRendix: I'm going to remove it and stick to MenuItem without adding obsolete code.

tuxemon/db.py
JaskRendix: Updated the description: "Whether or not the technique can be used"
Do we need to add an additional # ?
Like # functionality still not activated?

tuxemon/menu/interface.py
JaskRendix: You were right, much better, less files involved.

tuxemon/db.py
bitcraft: > Whether or not the technique can be used _(where?, how?)_

i think that maybe that its not clear that the meaning is supposed to be "Whether or not the technique can be used _outside of combat_".  And "usable_in", is also not clear about this.  Does any of that make sense?

JaskRendix: Yesterday night it wasn't clear, today it makes sense. I'm going to address it.

JaskRendix: Ok, I'm going to change **usable_in** in **usable_on**, because it could usable outside the battle **on** something (monsters, items, techniques, world elements, etc.). I hope it makes sense.

JaskRendix:
I added the following changes inside **world_menu.py**.

**the first:**
from
`monster_menu = self.client.replace_state(MonsterMenuState)`
to
`monster_menu = self.client.push_state(MonsterMenuState())`

because the one above was triggering inside the terminal:
```
/home/robespierre/Tuxemon/tuxemon/state.py:577: DeprecationWarning: Calling push_state with Type[State] is deprecated, use an instantiated State instead
  warnings.warn(
```
every time someone clicked on Tuxemon to open the monster menu. Now it's fixed.

**the second:**
replacing the pushing towards **PygameMenuState** in the **def open_monster_submenu** with an open_choice_dialog. Identical output. No issues. Tested.

**the third:**
as above, but with **release_monster_from_party** where I removed the ChoiceState in favor of open_choice_dialog. Identical output. No issues. Tested.

It was an opportunity to catch since we are working on this file. If it's not ok, then I can reverse it.

update:
added boilerplate (new file)

kerizane (approved):

PR #1502, opened 2022-12-24
moves monster

JaskRendix:
PR addresses monsters and moves.
This change involves only the structure.

@Sanglorian 
Since the addition of the last monsters and the fix of weight and height, I was working on this PR.
I share this after the max_moves (as you said in another PR, max 4), a mechanism of overwriting moves that it's really intriguing.

Example: **Rockitten** > **Rockat** > **Jemuar**

Situation at the moment, 
**Rockitten** moves: ram, boulder and thunderball
**Rockat** moves: ~~ram~~, glover and ~~thunderball~~
**Jemuar** moves: ~~ram~~, ~~glover~~ and stampede
**Evolutions**: **Rockitten** > lv 24 > **Rockat** > lv 36 > **Jemuar**
`Total potential moves without duplicates: 5`

After the PR:
**Rockitten** moves: ram, boulder and thunderball
**Rockat** moves: gover, ice_claw, surge
**Jemuar** moves: stampede, mudslide, biting_winds
**Evolutions**: **Rockitten** > lv 24 > **Rockat** > lv 42 > **Jemuar** (changed because the monster learns 3 moves every 4 level, then evolves, it's explained inside the file attached)
`Total potential moves without duplicates: 9`

Confrontation **Jemuar** JSON:
before:
```
2 lv - ram
2 lv - glower
2 lv - stampede
```
after:
```
2 lv - ram
2 lv - boulder
12 lv - thunderball
28 lv - glower
32 lv - ice_claw
36 lv - surge
44 lv - stampede
48 lv - mudslide
52 lv - biting_winds
```
The mechanism:
- basic and standalone (no evo) monsters (eg. rockitten is basic) have 3 moves;
- 1 stage (rockat) have 6 moves;
- 2 stage (jemuar) have 9 moves;

I added also a new validator called multiple to check both **level_learned** for techniques and **at_level** for evolutions. Technique learning is multiple of 2, while evolution level is multiple of 3, the will is having two ways, so we can reduce drastically the probability that someone needs to overwrite the technique immediately after an evolution.

Using the script for the missing move (the other PR) I generated the missing techniques based on type, everything related the moves is here: [moves.ods](https://github.com/Tuxemon/Tuxemon/files/10298760/moves.ods)
Everything related to the change of "at_level" involving evolutions is here: [at_level evolution.ods](https://github.com/Tuxemon/Tuxemon/files/10298761/at_level.evolution.ods)
Both files contain all the changes and additions I made.

Do you want also standalone monster with more techniques? We can generate easily, it doesn't take much time.
Do you think we need to use the 9 moves list (jemuar) also for rockitten and rockat? At the end someone can also refuse the evolution and keeping developing rockitten until the max level.

Let me know.

I think in the second case, considering overwriting and the maximum number of moves (4), it can be more challenging.
I already tested some mechanism and these are really interesting.

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -4,7 +4,7 @@
     "ai": "RandomAI",
     "moveset": [
         {
-            "level_learned": 1,
+            "level_learned": 2,
             "technique": "tonguespear"
         },
         {
@@ -16,8 +16,16 @@
             "technique": "stampede"
         },
 	    {
-            "level_learned": 17,
+            "level_learned": 24,
             "technique": "feint"
+        },
+	    {
+            "level_learned": 28,
+            "technique": "sand_spray"
+        },
+	    {
+            "level_learned": 32,
+            "technique": "mudslide"
         }
     ],
     "evolutions": [],
--- a/mods/tuxemon/db/monster/aardorn.json
+++ b/mods/tuxemon/db/monster/aardorn.json
@@ -4,7 +4,7 @@
     "ai": "RandomAI",
     "moveset": [
         {
-            "level_learned": 1,
+            "level_learned": 2,
             "technique": "tonguespear"
         },
         {
@@ -14,16 +14,12 @@
         {
             "level_learned": 8,
             "technique": "stampede"
-        },
-	    {
-            "level_learned": 17,
-            "technique": "feint"
         }
     ],
     "evolutions": [
         {
             "path": "standard",
-            "at_level": 20,
+            "at_level": 18,
             "monster_slug": "aardart"
         }
     ],
--- a/mods/tuxemon/db/monster/abesnaki.json
+++ b/mods/tuxemon/db/monster/abesnaki.json
@@ -12,7 +12,7 @@
             "technique": "pseudopod"
         },
         {
-            "level_learned": 2,
+            "level_learned": 12,
             "technique": "sleep_bomb"
         }
     ],
--- a/mods/tuxemon/db/monster/agnidon.json
+++ b/mods/tuxemon/db/monster/agnidon.json
@@ -9,11 +9,23 @@
         },
         {
             "level_learned": 2,
+            "technique": "fire_ball"
+        },
+        {
+            "level_learned": 12,
+            "technique": "fume"
+        },
+        {
+            "level_learned": 28,
             "technique": "kindling_flame"
         },
         {
-            "level_learned": 2,
+            "level_learned": 32,
             "technique": "energy_claws"
+        },
+        {
+            "level_learned": 34,
+            "technique": "breathe_fire"
         }
     ],
     "evolutions": [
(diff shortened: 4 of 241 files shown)


Sanglorian:
This sounds like a great improvement. The 3 techniques all at the same time was just a place holder so I like your approach. The only thing I'd change is keeping the evolutions at level 36. Other evolution changes are fine if it's just a few levels difference.

For earlier evolutions, e.g. Rockitten, it'd be good if they got some techniques even after they would normally evolve, but not all of them.

JaskRendix:
No problem, I fixed the 4 cases you told me about (42 instead of 36), I changed also the level of the move, because I want to avoid a double trigger if I can.

I guess we can level up to 6 techniques for basic monsters (Rockitten). But we do this, then we can also give 6 techniques to standalone monsters (the ones with no evolutions)! Everything in another PR. @Sanglorian  what do you think?

Sanglorian:
Around 6 techniques for stand alone monsters makes sense -- but really I'd give them as many as makes sense. It's nice to have choice, so at least 5 makes sense but it could be 6 or 9 or 12 for an individual species.

JaskRendix:
Ok, updated the last moves, ready to merge. In another PR I'll deal with the monster without evolution.

tuxemon/db.py
```diff
@@ -233,6 +233,13 @@ def valid_level(cls, v):
             raise ValueError(f"invalid level learned: {v}")
         return v
 
+    @validator("level_learned")
+    def is_multiple(cls, v):
+        if v > 0:
+            if has.multiple(v, 2):
+                return v
+            raise ValueError(f"no resource exists in db: {v}")
```
bitcraft: this error message is misleading

tuxemon/db.py
```diff
@@ -254,6 +261,13 @@ class MonsterEvolutionItemModel(BaseModel):
     gender: Optional[GenderType] = Field(None, description="Gender parameter")
     mixed: Optional[str] = Field(None, description="Mixed parameter.")
 
+    @validator("at_level")
+    def is_multiple(cls, v):
+        if v > 0:
+            if has.multiple(v, 3):
+                return v
+            raise ValueError(f"no resource exists in db: {v}")
```
bitcraft: this error message is misleading

tuxemon/db.py
```diff
@@ -1019,6 +1033,21 @@ def db_entry(self, table: TableName, slug: str) -> bool:
             return True
         return False
 
+    def multiple(self, num: int, check_with: int) -> bool:
```
bitcraft: are you sure we need a function that just does `X % Y == 0`?

bitcraft (changes requested):

tuxemon/db.py
JaskRendix: I'm going to remove it, I used it for maintaining order

tuxemon/db.py
JaskRendix: I'm going to remove it, I used it for maintaining order

tuxemon/db.py
JaskRendix: I'm going to remove it, I used it for maintaining order

bitcraft (approved):

PR #1505, opened 2022-12-25
standardization center maps

JaskRendix:
PR addresses the standardization of the centers.

Right now the correct one was **spyder_candy_center** (13x11).

added use computer event to these locations (it was missing):
- healing_center
- cotton_cathedral
- sphalian_center (see below)

Standardized:
- the dimension of the map at: 13x11 (before 1 was 13x11, while the others 15x15, but these were full of pointless collision around the map as well as 2 tiles of empty space);
- the transition_teleport at: _center.tmx,6,10,0.3
- the use_computer at: is player_at 10,4
- the create_npc nurse at: 5,4
- the set_variable teleport_faint at: 6 10

updated the position of shady and happy guy inside the healing center (Xero), because of the resizing of the map.

deleted the cottonnurse.json, it was identical to tabanurse.json, so a duplicate;
the new message defines automatically the location (instead of the old cottonnurse_dialog, another duplicate);

Did you see above sphalian_center, yeah, I renamed in a more correct way the following maps:
- BuddhaMountain into  buddha_mountain
- healing_center_sphalian into sphalian_center

The reason is that I'm going to use these maps later on, so I wanted to fix some small stuff.

Last thing, I took the chance to update some transition_teleport (time?), some were still stuck at 0.5 and not updated at 0.3, like all the others transition teleport (37707 excluded).

In this way if someone wants to create a new center, everything is already set in the right standard. It helps if we want to mass replace something in the future.


Sanglorian:
Sounds good to me!

JaskRendix:
Solved the conflict

bitcraft (approved):
LGTM

PR #1508, opened 2022-12-25
taba + maple maps

JaskRendix:
PR addresses the following stuff:
- update maple_house (as usual, too big, I removed the empty tiles around as well as the pointless collisions);
- update taba_town (because of the different size, we need to update the teleport);
- connected maple_bedroom to maple_house (I left the old events, but I can remove) so with the stairs it's possible to access the bedroom (where I added the event to access the PC);
- updated because of the previous point the maple.json;
- renamed the old maple_downstairs (previous version of maple_house) as taba_town_house1 and I changed destination (by opening in taba_town another house);
- renamed and updated the old map1_apartment1 as taba_town_house2;
- renamed and updated the old map1_apartment2 as taba_town_house3;
- renamed and updated the old map1_apartment3 as taba_town_house4;

everything has been tested, the pathfinds work without issues.

@Qiangong2 what do you think? Do you want to integrate all these new locations inside Taba Town? In this way the player can access all the houses. These are still without NPCs, but if you agree with this, I'm going to populate it.
About maple_bedroom, do you want to remove the old events so you can set up something new?

Screenshots:
Taba Town House 4 -> it's a restaurant
taba_town_house4
Taba Town House 3 -> its an inn
taba_town_house3
Taba Town House 2 -> it's a library
taba_town_house2
Taba Town House 1 -> it's a normal house, I was thinking about a NPC called Mr Lethe
taba_town_house1
Maple House, now you can go up to the bedroom, but if you don't want to, I can remove it
maple_house
Maple bedroom with old events
maple_bedroom_events
Maple bedroom without old events
maple_bedroom


Qiangong2:
I think integrating them would be great! Previously, my plan was to just have a dialog saying "the door is locked", but this is a lot better :)

JaskRendix:
Thanks!

mods/tuxemon/maps/maple_bedroom.tmx
```diff
@@ -1,108 +1,72 @@
...
+  <object id="9" name="Go Downstairs" type="event" x="112" y="32" width="16" height="16">
    <properties>
-    <property name="act10" value="transition_teleport maple_house_downstairs.tmx,4,2,0.3"/>
+    <property name="act10" value="transition_teleport maple_house.tmx,11,2,0.3"/>
     <property name="act20" value="player_face down"/>
     <property name="cond10" value="is player_at"/>
     <property name="cond20" value="is player_facing left"/>
    </properties>
   </object>
-  <object id="19" name="Create Maple" type="event" x="0" y="32" width="16" height="16">
```
bitcraft, line 55: @Qiangong2 ok to remove this event?

Qiangong2: I don't believe I ever actually used the maple bedroom map.

I've used the player_bedroom map, but not maple.

JaskRendix: You're right, I forgot to remove it, I'm going to update.

bitcraft (approved):

bitcraft:
LGTM

PR #1513, opened 2022-12-27
Fix regression

NeonFlames:
Locale files would not be loaded properly if compiled by db.py
(this time it works)

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -57,7 +57,7 @@ from tuxemon.locale import T
 logger = logging.getLogger(__name__)
 
 # Load the default translator for data validation
-T.collect_languages(0)
+T.collect_languages(False)
 T.load_translator()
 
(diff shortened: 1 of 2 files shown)


tuxemon/db.py
```diff
@@ -57,7 +57,7 @@
 logger = logging.getLogger(__name__)
 
 # Load the default translator for data validation
-T.collect_languages(prepare.CONFIG.recompile_translations)
```
bitcraft, line 60: as a style guideline, we should not be using `0` in place of False, and instead should use `False` when our intention is for the boolean value.

bitcraft (changes requested):

tuxemon/db.py
NeonFlames: Alright

bitcraft (approved):

PR #1515, opened 2022-12-27
update transfer money

JaskRendix:
PR corrects and simplifies the action transfer money.

When I created it I made a mistake of adding the slug parameter. I recognized the error because I was using it for some maps and I found it very tricky, now it's more simple.

### DIFF
--- a/tuxemon/event/actions/transfer_money.py
+++ b/tuxemon/event/actions/transfer_money.py
@@ -23,7 +23,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from typing import final
+from typing import Union, final
 
 from tuxemon.event.eventaction import EventAction
 
@@ -39,7 +39,7 @@ class MoneyMathAction(EventAction):
     Script usage:
         .. code-block::
 
-            transfer_money <transaction>,<amount>,<slug>
+            transfer_money <transaction>,<amount>[,slug]
 
     Script parameters:
         transaction: Operator symbol.
@@ -47,13 +47,14 @@ class MoneyMathAction(EventAction):
         slug: Slug name (e.g. NPC, etc.)
         eg: +,100,mom (player gets 100 from mom)
         eg: -,100,mom (mom gets 100 from player)
+        eg: +,100 (player gets 100)
 
     """
 
     name = "transfer_money"
     transaction: str
     amount: int
-    slug: str
+    slug: Union[str, None] = None
 
     def start(self) -> None:
         player = self.session.player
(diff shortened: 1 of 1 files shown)


bitcraft:
im not familiar with the destination for where the funds are going.  have you found that we don't need shop keepers to have a wallet/money?

JaskRendix:
Great point, you're right about the wallet for shops, museums, etc. I have reintroduced the npc_slug
I wanted to simplify the action, before it was a bit confusing, now it's way more simple. 

**before**
`transfer_money -,100,spyder_leather_museum`
here it was already included inside the code that the player was going to lose 100 and the museum was getting 100
**now**
```
transfer_money spyder_leather_museum,+,100
transfer_money player,-,100
```
before it was too much hardcoded, modders need to have the choice if remove or add the money to the shop, museum or npcs, like if there is a gift, etc as well as the player

bitcraft:
im not really sure that the original was complicated at all.  did you ask other modders before making this change?  i suppose it would have been more easily "simplified" by just making the argument for a destination wallet optional.

JaskRendix:
Simplified the original, so the structure doesn't change. The npc is optional.

tuxemon/event/actions/transfer_money.py
```diff
@@ -39,58 +39,51 @@ class MoneyMathAction(EventAction):
...
-                value = 0
-                player.money[destination] = int(value) + amount
-            else:
-                player.money[destination] = int(wallet_slug) + amount
+            player.money["player"] = wallet_player - amount
+            if self.slug is not None:
+                wallet_npc = player.money.get(self.slug)
+                if wallet_npc is None:
+                    player.money[self.slug] = 0 + amount
+                    print(player.money[self.slug])
```
bitcraft: please do not include print statements

JaskRendix: I forgot to remove, it happens

tuxemon/event/actions/transfer_money.py
```diff
@@ -39,58 +39,50 @@ class MoneyMathAction(EventAction):
...
-            if wallet_slug is None:
-                value = 0
-                player.money[destination] = int(value) - amount
-            else:
-                player.money[destination] = int(wallet_slug) - amount
+            player.money["player"] = wallet_player + amount
+            if self.slug is not None:
+                wallet_npc = player.money.get(self.slug)
+                if wallet_npc is None:
+                    player.money[self.slug] = 0 + amount
```
bitcraft: `0 + amount`?  should this be `0 - amount`?

bitcraft: could also be `amount * -1`

tuxemon/event/actions/transfer_money.py
```diff
@@ -39,58 +39,50 @@ class MoneyMathAction(EventAction):
...
-            if wallet_slug is None:
-                value = 0
-                player.money[destination] = int(value) + amount
-            else:
-                player.money[destination] = int(wallet_slug) + amount
+            player.money["player"] = wallet_player - amount
+            if self.slug is not None:
+                wallet_npc = player.money.get(self.slug)
+                if wallet_npc is None:
+                    player.money[self.slug] = 0 + amount
```
bitcraft: why `0 + amount`?

tuxemon/event/actions/transfer_money.py
JaskRendix: true, it should be `0 - amount`, I'm going to put `amount * -1`

tuxemon/event/actions/transfer_money.py
JaskRendix: true, as above, I'm going to put `amount * 1`

bitcraft: why `amount * 1`?  you could just do `player.money[self.slug] = amount`?

kerizane (approved):

PR #1516, opened 2022-12-27
fix shop

JaskRendix:
PR addresses an issue with the shop.

The issue is related to the selling process.

Precisely, you can have some items inside your bag and these are going to pop-up inside the shop screen (until now only the items labelled as quest were excluded). Now, if the item isn't inside the economy file (in the Xero scenario was reg_papers and allie address) and you try to sell, well, it results in a crash.

Since we can't have all the shops with all the items listed with all the prices and costs (neither filter out by hardcoding every label), here the fix for the issue.

With this fix, you can sell in the shop only the items that are in the economy (so listed with price and cost), the others aren't going to appear anymore. 

before the fix (sell menu)
Screenshot from 2022-12-27 08-41-26
after the fix (sell menu)
Screenshot from 2022-12-27 08-58-22


bitcraft (approved):

bitcraft:
LGTM

Issue #1421, opened 2022-11-01
Can we remove licence boilerplate?


kerizane:
Do we really need the licence header at the start of every file? The dates are wrong for most of them, plus they will often have random names on because people just copy it from a nearby file.

It's annoying having half the screen taken up by this when you open a file (although I guess imports aren't that more interesting).

We could just have a single licence file within the repo.

JaskRendix:
Agree, let's take for instance this file: /tuxemon/event/actions/quit.py
with boilerplate: 1.4 kB (1,396 bytes)
without boilerplate: 583 bytes

kerizane:
There are already some files without the boilerplate, so I think we should encourage new files to not have it. I don't want to remove it from the existing ones without approval by @ShadowApex or @bitcraft

bitcraft:
I prefer to put them in the header.  We may find a way to shorten it, but the GPL faq states that it *should* be included in each file.  Though, its certainly a matter of preference, because it wouldn't be a violation.

https://www.gnu.org/licenses/gpl-faq.en.html#NoticeInSourceFile

JaskRendix:
No problem, as soon as we have the right version.
About the text still this? I guess we need to update the date from 2017 to 2023, right?
```
#
# Tuxemon
# Copyright (c) 2014-2017 William Edwards <shadowapex@gmail.com>,
#                         Benjamin Bean <superman2k5@gmail.com>
#
# This file is part of Tuxemon
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program. If not, see <http://www.gnu.org/licenses/>.
#
```
because some are still stuck in 2014, so we can do a mass replacement.

bitcraft:
yeah, the copyright notices need to be updated.  i haven't thought about this in a long time.  but poking around some other projects, notable the linux kernel, their source files start with a license ID and copyright. so, if we use the linux kernel sources as a guide, i think we would be fine to do just that.  perhaps
```
# SPDX-License-Identifier: GPL-3.0
[...]

PR #1518, opened 2022-12-27
update boilerplate

JaskRendix:
Related to #1421 
New boilerplate based on the proposal.
If something it's wrong or missing, then feel free to say it.


run_tuxemon.py
```diff
@@ -1,31 +1,5 @@
-#!/usr/bin/env python3
```
bitcraft, line 1: don't drop this line

bitcraft (commented):
no changes requested, but we may consider keeping the "contributor" lines.  not sure right now.

setup.py
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env python3
```
NeonFlames, line 1: Probably shouldn't drop this

run_tuxemon.py
JaskRendix: ok

setup.py
JaskRendix: ok

JaskRendix:
I can insert again the contributors, no problem, just tell me if you want to be inserted again (you're in 84 files).
I don't know if it's necessary to state it, but I'm not interested in having my name inside the file (created or edited/contributed).

bitcraft (approved):
looks good, thank you.  ill follow up with the other part of the issue, migrating the contributor lines and the ATTRIBUTIONS file

JaskRendix:
Ok, no problem, in that case this can be useful: [contributors.txt](https://github.com/Tuxemon/Tuxemon/files/10338876/contributors.txt) (it was my backup if I needed to insert again the contributors).

NeonFlames (approved):

kerizane (approved):

PR #1523, opened 2022-12-29
moves standalone monster

JaskRendix:
PR adds a more variegate set of moves for monster without evolution.

Reopened because of too many conflict of the previous one.

81 monsters with no evolution.

monster with earth in the types 19
monster with fire in the types 12 
monster with metal in the types 26 
monster with water in the types 17  
monster with wood in the types 18 

Moves:
```
level_learned: 2
level_learned: 2
level_learned: 6
level_learned: 8 -> additional move
level_learned: 12 -> additional move
level_learned: 16 -> additional move
level_learned: 20 -> additional move
level_learned: 24 -> additional move
level_learned: 28 -> additional move
level_learned: 32 -> additional move
level_learned: 36 -> additional move
level_learned: 40 -> additional move
```


JaskRendix:
@Sanglorian reopened the PR because of some issues, now solved.

I tagged you just to inform you that I populated the JSONs. You can see all the moves here: [standlone.ods](https://github.com/Tuxemon/Tuxemon/files/10319092/standlone.ods) or here in GitHub.

It shows all the 81 monsters with the types and the original 3 moves; the columns of the first three moves have grey background.
Then there are other 9 columns of moves. The method used is always the one related to the type of the move, the one we used in all the other PRs, the difference is that now we can "fish" in a bigger pond. 

Just some numbers, now we have (each category includes techniques with double type):
- 53 earth type moves
- 36 fire type moves
- 36 metal type moves
- 30 water type moves
- 42 wood type moves

It's a massive improvement.

JaskRendix:
Is it ok for you? @Sanglorian

Sanglorian:
Sorry for the delay, yeah that looks great - thank you!

kerizane (approved):

PR #1509, opened 2022-12-26
sprites

JaskRendix:
PR upload some of the sprites I edited.

Nothing new, changed hair color as well as clothes color.

The reason is that I'm creating some NPCs and I need more variety.

PR #1524, opened 2022-12-29
new sprites

JaskRendix:
PR addresses the sprites folder.

A couple days ago I added a bunch of new sprites, the sprites were all new, but based on existing ones (eg changed the hair color of the adventurer, changed the hair color of heroine, etc).

I noticed the names were confusing and it was difficult to understand on which original sprites the new ones are based. So I renamed the sprites. I added to the original sprite + the new feature (eg. guy12 is renamed knight_yellow because it was based on the knight sprites, but it has yellow features, etc.).

I left girl1, girl2, girl3, because these are all variations of girl1.

It's way more clear and reduce the possibility of confusion (especially considering the credits).

I added also new version of maniac (different clothes color), shop assist (different hair color), shopkeeper (different clothes and hair color) and swimmer (different color caps).


bitcraft (approved):
looks ok.  just a FYI for the future, mass renames need to also check the json and tmx files.  i don't think its a problem here since they are all new.  also, it would be best to open an issue and tag other modders because they would have potentially broken merges if they do not rebase.  for that reason, i tend to open an issue, or plan out changes on the discord so that they don't cause problems.

so generally, other modder would first commit and push their WIP changes, then they would wait until the mass rename is carried out, then they would pull in the file changes.  coordinating it with other will ensure that nobody has broken maps or sprites.

so ill approve this on the understanding that you are sure that none of the changed sprites are in use.

JaskRendix:
I'm sure, because it has been merged in #1509, two days ago.
Thanks for your FYI, I'm aware of what you wrote. Especially about checking JSON and tmx files.

PR #1537, opened 2023-01-04
Update build system

NeonFlames:
Switch from setup.py to pyproject.toml

### DIFF
--- a/tuxemon/states/pc/__init__.py
+++ b/tuxemon/states/pc/__init__.py
@@ -14,7 +14,6 @@ from typing import (
     Optional,
     Sequence,
     Tuple,
-    no_type_check,
 )
 
 import pygame_menu
@@ -28,7 +27,7 @@ from tuxemon.menu.menu import Menu, PopUpMenu, PygameMenuState
 from tuxemon.menu.theme import get_theme
 from tuxemon.session import local_session
 from tuxemon.states.monster import MonsterMenuState
-from tuxemon.tools import open_dialog, transform_resource_filename
+from tuxemon.tools import open_dialog
 
 logger = logging.getLogger(__name__)
 
@@ -74,7 +73,7 @@ class PCState(PopUpMenu[MenuGameObj]):
         else:
             storage_callback = change_state("MonsterBoxChooseStorageState")
 
-        if not local_session.player.monsters:
+        if len(local_session.player.monsters) <= 1:
             dropoff_callback = partial(
                 open_dialog,
                 local_session,


NeonFlames:
Looks like I have some things to fix up.

NeonFlames:
If anyone has an idea on how to fix the problem I'm all ears because I've looked through the setuptools documentation and found nothing (Although I may have missed it).

JaskRendix:
I don't know if this is the reason, I noticed that before there was this:
```
for root, dirnames, filenames in os.walk("tuxemon"):
    for filename in fnmatch.filter(filenames, "__init__.py"):
        matches.append(os.path.join(root, filename))
```
So technically:
```
[tool.setuptools.packages.find]
where = [""]
include = ["tuxemon*"]
namespaces = false
```
As advised here: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages below the section pyproject.toml (BETA), the note.

NeonFlames:
It still fails to build an editable install and a normal install with the namespaces key added as advised by the documentation.

JaskRendix:
Remember: https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior 
I was reading the raw logs of the failed action and I wonder if we need to add the wheel: requires = ["setuptools", "wheel"]
There are some deprecations in the raw logs pointing to it (neteria, etc).

NeonFlames:
> Remember: https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior I was reading the raw logs of the failed action and I wonder if we need to add the wheel: requires = ["setuptools", "wheel"] There are some deprecations in the raw logs pointing to it (neteria, etc).

I got help over at https://github.com/pypa/setuptools/discussions/3748 and they pointed out that the version should be statically defined or set via setup.py so I went for static definition in pyproject.toml but there are three packages we rely on that have deprecation warnings. (cbor, neteria, and another)

NeonFlames:
I noticed that one of the deprecation warnings comes from a dependencies dependency but I have no idea which one, however for now it's probably fine.

JaskRendix:
Exactly, you're right: pyperclip, neteria and cbor.

DEPRECATION: pyperclip is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

NeonFlames:
I'll add wheel to the build depends and that should probably silence the deprecation warning.

(also I have to redo the changes since I just accidentally nuked my branch)

PR #1540, opened 2023-01-05
PCState (kennel) + pygame_menu

JaskRendix:
PR addresses the PCState (kennel)

**we can still mass replace kennel with something else, I just need to insert a couple of lines in save_upgrader (so I can update automatically the old savegame where there is Kennel), please discuss it in Discord and let me know. There is no problem with translations, I checked and it was used storage, only a single string in German has the kennel translation**

Created a new folder below States called pc_kennel, inside there are all the kennel operation.
Fixed an old bug related to pick up the monster (discovered after setting up the move operation): shortly if you select nut, then the system picks up rockitten > solved by using uuid and instance_id.

Regarding states.pc.init.py:
- adjusted the pygame_menu **paths**;
- created def **repristinate** (init.py PC), so it can be used to repristinate the original theme (aligned left, no background, etc);
- created **HIDDEN** as standard hidden list (it can be used by anyone);
- created **HIDDEN_LIST**, it regroups all the hidden kennels;
- created **KENNEL** as constant;
- inserted function to create the kennel (new players) as soon as they access the PC (it replaces the config in npc.py);
- added submenu pick_up + move (https://github.com/Tuxemon/Tuxemon/pull/1540#issuecomment-1374835236)
- added submenu release (https://github.com/Tuxemon/Tuxemon/pull/1540#issuecomment-1374835236)

Regarding graphics:
- added new background (screenshot), battle background PNG, if someone has something better, then no problem;
- added **number of monster** inside the kennel (new Theme);

Regarding en_US PO:
- rewritten msgid **menu_dropoff_no_monsters**;
- added msgid **Kennel** and msgtr **shelter**, because **shelter** in my native language sounds more "neuter" and less "negative" than kennel. Too many never-ending scandals about kennels. I don't know in other languages. If there are other options, feel free to suggest;

Regarding actions:
- created **set_kennel**, brand new action to create a kennel with the option visible true/false (now the modder can create the kennel independently without forcing its creation in store_monster);
- updated **store_monster**, now it'll store only: (1) box is optional (2) checking if the kennel exist, if not raise valueerror (3) removed a copy pasted # line (4) removed import logging since it wasn't used (5) inserted the automatic creation of the HIDDEN kennel if the box is missing > this was necessary because possible cause of crash in the breeding process;

Regarding config.py:
- removed **CONFIG.default_monster_storage_box**;
- removed **CONFIG.default_item_storage_box** unused (we can set up the item_boxes in a similar way);

tested, black, isort as usual

(screenshots below)


JaskRendix:
@Qiangong2 how do you want to modify the following actions? File: daycare.tmx
From:
```
    <property name="act17" value="store_monster breeding_mother,cotton_breeder"/>
    <property name="act35" value="store_monster breeding_father,cotton_breeder"/>
```
To:
```
    <property name="act17" value="store_monster breeding_mother"/>
    <property name="act35" value="store_monster breeding_father"/>
```
Or (updated):
```
    <property name="act17" value="set_kennel cotton_breeder,false"/>
    <property name="act18" value="store_monster breeding_mother,cotton_breeder"/>
    <property name="act35" value="store_monster breeding_father,cotton_breeder"/>
```

Qiangong2:
I think the third option would be best as it gives more flexibility with updating it

JaskRendix:
Updated daycare.tmx with the third option.

JaskRendix:
Other actions:

removed monster from PC State (duplicate and useless) the player can access it in the main menu, where there are other functions like info, release, move (and techniques);
Screenshot from 2023-01-08 13-46-10
function pick_up if the player can access 1 kennel (no more clicking by mistake and immediate pick up)
Screenshot from 2023-01-08 12-27-25
function pick_up if the player can access 2 kennel (Kennel + 1 kennel called LMAO);
it'll give the possibility to move (all texts will be uppercase by the way)
Screenshot from 2023-01-08 12-09-26
function pick_up if the player can access 2+ kennel (Kennel + 4 kennels called L+M+A+O);
it'll give the possibility to move (all texts will be uppercase by the way)
Screenshot from 2023-01-08 11-26-34
if the player clicks move;
you can see all the options
Screenshot from 2023-01-08 11-28-14
____
I was doing some tests inside the kennels and it was frustrating pick up and then release, so added release option, if the player clicks, then there is the yes/no choice, no first choice (predefined), yes second choice (below)
Screenshot from 2023-01-09 13-54-44
____
Before
210948001-a3e07f8c-e458-4f11-8b55-0ad755820b16 - 211185758-064b868d-d2e8-4b02-a33b-359b0d049cd1

Before
211036297-b154d271-3728-46a7-ad81-0c08d9f1a4b6 - 211185762-cd8a154c-5ca1-4b86-b8d6-403137ff771a

JaskRendix:
Good news, the last pygame_menu update now offers banner (mix of button + image), now we can click directly on the image and open the submenu, this solves the issue about the monster images half cut, but we need to update the requirements.txt of pygame_menu from 4.2.8 to 4.3.6.

kerizane (approved):

PR #1553, opened 2023-01-11
leap year (true/false)

JaskRendix:
PR addresses the addition of leap year (game_variable["leap_year"])

until now we had day_of_year, but it was difficult to set precise anniversary or birthday (like Mom player birthday, city anniversaries, festivities in game, etc) if it was after February. Are you interested @Sanglorian and @Qiangong2 ? Let me know.

Examples for using it, let say Mom's birthday is on the 25th May (https://en.wikipedia.org/wiki/May_25), 145th day of the year (146th in leap years)
```
<property name="cond1" value="is variable_set day_of_year:146"/>
<property name="cond2" value="is variable_set leap_year:true"/>
or
<property name="cond1" value="is variable_set day_of_year:145"/>
<property name="cond2" value="is variable_set leap_year:false"/>

moreover day_of_year until 59 (28th February), it doesn't need (leap_year) cond:
<property name="cond1" value="is variable_set day_of_year:59"/>
```
_(removed the two ints for 2 reasons: (1) variable_set is optional[str] and (2) these are marked as ints in the following operations. Tested. Everything works without issues.)_

### DIFF
--- a/tuxemon/player.py
+++ b/tuxemon/player.py
@@ -56,8 +56,17 @@ class Player(NPC):
         %j - Day number of year 001-366
         """
         var = self.game_variables
-        var["hour"] = int(dt.datetime.now().strftime("%H"))
-        var["day_of_year"] = int(dt.datetime.now().strftime("%j"))
+        var["hour"] = dt.datetime.now().strftime("%H")
+        var["day_of_year"] = dt.datetime.now().strftime("%-j")
+        var["year"] = dt.datetime.now().strftime("%Y")
+
+        # Leap year
+        if (int(var["year"]) % 400 == 0) and (int(var["year"]) % 100 == 0):
+            var["leap_year"] = "true"
+        elif (int(var["year"]) % 4 == 0) and (int(var["year"]) % 100 != 0):
+            var["leap_year"] = "true"
+        else:
+            var["leap_year"] = "false"
 
         # Day and night basic cycle (12h cycle)
         if int(var["hour"]) < 6:


Sanglorian:
Good thinking!

kerizane (approved):

broccoder:
Great thinking, but Linux only. The problem lies in the `%-j`
To quote the [python docs](https://docs.python.org/3/library/time.html)
> Additional directives may be supported on certain platforms, but only the ones listed here have a meaning standardized by ANSI C. To see the full set of format codes supported on your platform, consult the [strftime(3)](https://manpages.debian.org/strftime(3)) documentation.

On some platforms, an optional field width and precision specification can immediately follow the initial '%' of a directive in the following order; this is also not portable. The field width is normally 2 except for %j where it is 3.

The [strftime(3)](https://manpages.debian.org/bullseye/manpages-dev/strftime.3.en.html#:~:text=string%20with%20spaces.-,%2D,(dash)%20Do%20not%20pad%20a%20numeric%20result%20string.,-0) says that - (dash) means "Do not pad a numeric result string."
However, Microsoft does not support this formatting on Windows. See [msdn docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?redirectedfrom=MSDN&view=msvc-170#:~:text=Remove%20leading%20zeros%20or%20spaces%20(if%20any).). They use `%#j` for removing leading zeros.
TL;DR: [This line](https://github.com/Tuxemon/Tuxemon/blob/03feef98d58ad88478ab0b758c2276d46bc25c13/tuxemon/player.py#L60) raises `ValueError: Invalid format string`, so please fix this @Sanglorian 

EDIT: There is a dead simple fix to this: just use `%j` as before
Python won't complain about the leading zeros and it'll convert it to int correctly in the leap year checks

JaskRendix:
@infibrocco thank you so much for the suggestion

PR #1556, opened 2023-01-12
fix tmx (class instead of type)

### LINKED ISSUE #1549: Bug type (files .tmx)
**Describe the bug**
The new version (version="1.9") will mass replace type with class.
All the maps are max 1.8.4.
So as soon as you enter in a map, well, no collisions, no NPCs and no events.
Opened issue on pytmx too

Eg.
```
<object id="107" type="collision" x="736" y="64" width="16" height="64"/>
<object id="107" class="collision" x="736" y="64" width="16" height="64"/>
```
or
```
<object id="22" name="Assistant Talk" type="event" x="112" y="112" width="16" height="16">
<object id="22" name="Assistant Talk" class="event" x="112" y="112" width="16" height="16">
```

**To Reproduce**
1. Obtain the newest Tiled version
2. Change a letter or whatever, then save.
3. It replaces all the type with class

or

1. Open a map, mass replace type with class.
2. Access the location in game.

**Expected behavior**
Respect collisions, events and NPCs.

**Screenshots**
Screenshot from 2023-01-10 08-18-53

JaskRendix:
PR addresses the change from type to class in the new version (>= 1.9).
Fix #1549 

Used __getattribute__ instead of obj.type, so I don't touch pytmx.

It works, I tested on some maps by mass replacing type with class (collisions works, as well as events, transitions, etc.), in this way we can maintain a hybrid of maps and we are not forced of mass converting maps to 1.9.

black + isort

### DIFF
--- a/tuxemon/map_loader.py
+++ b/tuxemon/map_loader.py
@@ -208,7 +208,11 @@ class TMXMapLoader:
                 colliders = gids_with_colliders.get(gid)
                 if colliders is not None:
                     for obj in colliders:
-                        if obj.type and obj.type.lower().startswith(
+                        if obj.type is None:
+                            obj_type = getattr(obj, "class")  # obj.class is invalid syntax
+                        else:
+                            obj_type = obj.type
+                        if obj_type and obj_type.lower().startswith(
                             "collision"
                         ):
                             if getattr(obj, "closed", True):
@@ -225,18 +229,22 @@ class TMXMapLoader:
                             collision_lines_map.add(line)
 
         for obj in data.objects:
-            if obj.type and obj.type.lower().startswith("collision"):
+            if obj.type is None:
+                obj_type = getattr(obj, "class")  # obj.class is invalid syntax
+            else:
+                obj_type = obj.type
+            if obj_type and obj_type.lower().startswith("collision"):
                 for tile_position, props in self.extract_tile_collisions(
                     obj, tile_size
                 ):
                     collision_map[tile_position] = props
                 for line in self.collision_lines_from_object(obj, tile_size):
                     collision_lines_map.add(line)
-            elif obj.type == "event":
+            elif obj_type == "event":
                 events.append(self.load_event(obj, tile_size))
-            elif obj.type == "init":
+            elif obj_type == "init":
                 inits.append(self.load_event(obj, tile_size))
-            elif obj.type == "interact":
+            elif obj_type == "interact":
                 interacts.append(self.load_event(obj, tile_size))
 
         return TuxemonMap(
@@ -389,7 +397,11 @@ class TMXMapLoader:
                     raise Exception
 
         # add a player_facing_tile condition automatically
-        if obj.type == "interact":
+        if obj.type is None:
+            obj_type = getattr(obj, "class")  # obj.class is invalid syntax
+        else:
+            obj_type = obj.type
+        if obj_type == "interact":
             cond_data = MapCondition(
                 "player_facing_tile", list(), x, y, w, h, "is", None
             )


kerizane:
Can you use getattr instead of __getattribute__?

kerizane:
Looks like you're repeating the code a bit. Can we do something like

`obj_type = obj.type if hasattr(obj, "type") else obj.class` ?

And then just use this variable.

JaskRendix:
You're right, I'm going to update it

JaskRendix:
Unfortunately I cannot use obj.class

JaskRendix:
Updated, I tested and it works, removed some duplicate code.

tuxemon/map_loader.py
```diff
@@ -208,7 +208,11 @@ def load(self, filename: str) -> TuxemonMap:
                 colliders = gids_with_colliders.get(gid)
                 if colliders is not None:
                     for obj in colliders:
-                        if obj.type and obj.type.lower().startswith(
+                        if obj.type is None:
+                            obj_type = obj.__getattribute__("class")
+                        else:
```
kerizane, line 213: Can you use `getattr(obj, "class")` ? It's more standard

JaskRendix: Done

tuxemon/map_loader.py
```diff
@@ -208,7 +208,11 @@ def load(self, filename: str) -> TuxemonMap:
                 colliders = gids_with_colliders.get(gid)
                 if colliders is not None:
                     for obj in colliders:
-                        if obj.type and obj.type.lower().startswith(
+                        if obj.type is None:
+                            obj_type = getattr(obj, "class")
```
kerizane: Will this definitely work with the newer version?

kerizane: I think one of the following is best. Doesn't need to be a oneliner. Please confirm it works.

obj_type = obj.type if obj.type is not None else obj.class
obj_type = obj.type if hasattr(obj, "type") else obj.class

kerizane: getattr and __getattribute__ are both pointless

JaskRendix: the problem with obj.class is that we need to modify pytmx, because type (obj.type) gets called from class TiledObject(TiledElement) (pytmx.py), here there is no self.class, so it'll never work here. 

```
class TiledObject(TiledElement):

    def __init__(self, parent, node):
        TiledElement.__init__(self)
        self.parent = parent

        # defaults from the specification
        self.id = 0
        self.name = None
        self.type = None <------
        self.x = 0
        self.y = 0
        self.width = 0
        self.height = 0
        self.rotation = 0
        self.gid = 0
        self.visible = 1
        self.closed = True
        self.template = None
```

JaskRendix: @Qiangong2 can you test this PR and see if it works?
I have seen you have reversed a commit because of this problem.

Qiangong2: Seems to work for me when editing with Tiled 1.9.2. I reverted my 1.9.2 commit and moved back to Tiled 1.8.6 for the time being, but of course it would be ideal to be on the latest Tiled version

kerizane: I don't think it's possible for your version to work and mine not.

Could you remind me what the original version was?

kerizane: Tested, the following works:
```
if obj.type is None:
    obj_type = getattr(obj, "class")  # obj.class is invalid syntax
else:
    obj_type = obj.type
```

kerizane (changes requested):

JaskRendix:
Updated

kerizane (approved):

PR #1562, opened 2023-01-14
update missing translation + mk01

JaskRendix:
PR addresses some missing translations and adds the missing proto, gamma, delta, omega.

There is also a small fix about the sorting inside EN PO, it wasn't really alphabetical.

@Sanglorian I didn't know there were also proto, etc. I updated the original alpha and beta (not anymore standalone, but stage1), so I added the new jsons as well as the icons. Tell me if it's ok, if not I can remove, no issues.


Sanglorian:
Sounds great, thanks!

kerizane (approved):

PR #1568, opened 2023-01-17
synch Professor PC

JaskRendix:
PR addresses the addition to the Professor PC (PC State).

@Sanglorian @Qiangong2 

Months ago we added SeenStatus, that saves the names of the monsters seen / caught, but it wasn't possible to access the data.

It gives back the status of monsters seen/caught related to the total. It gives the percentage of progression (eg. 7.8%), interesting stuff is that there will be a variable called tuxepedia_progress with the rounded value (if 7.8 > 8), this can be used to trigger new conversations with the professor or anyone else involved. Some examples:
```
<property name="cond1" value="is variable_is tuxepedia_progress,<=,10"/>
<property name="cond1" value="is variable_is tuxepedia_progress,>=,11"/>
<property name="cond1" value="is variable_is tuxepedia_progress,>=,21"/>
```
It's an archive because there is a choice:
- do you prefer the number for tuxepedia_progress? 
- do you prefer a string for tuxepedia_progress? (eg tuxepedia_progress = "novice, etc", which will be on the wiki too)

Let me know

Screenshot from 2023-01-17 09-38-08
Screenshot from 2023-01-17 09-38-24
Screenshot from 2023-01-17 09-38-59

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -1348,6 +1348,9 @@ msgstr "Load"
 msgid "menu_monster"
 msgstr "Tuxemon"
 
+msgid "menu_professor_pc"
+msgstr "Professor PC"
+
 msgid "menu_storage"
 msgstr "Pick Up Tuxemon"
 
@@ -1425,6 +1428,12 @@ msgstr "shelter"
 
 msgid "kennel_label"
 msgstr "{box} ({qty})"
+
+msgid "synchronize_pc1"
+msgstr "Do you want to synchronize your Tuxepedia with the Professor's PC?"
+
+msgid "synchronize_pc2"
+msgstr "Tuxemon: seen {seen} - caught {caught} - Total: {all} Tuxemon. Progress:{value}%!\n Professor: Keep me posted!"
 
 # Default mod names on selection menu
 msgid "xero_campaign"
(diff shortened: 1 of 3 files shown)


Qiangong2:
Out of curiosity, why the sync with professor's PC in the first place? Why not just have the Tuxepedia enabled by default?

I think it's great though!

JaskRendix:
There is no reason why I put in the first place, I was thinking of getting rid of Items (another duplicate) and put it there (above multiplayer). What do you mean with "enabled by default"? Do you mean accessing it through menu or item? Yeah, as soon as we are able to insert the item phone.

Sanglorian:
Great, thank you. I would rather go with the percentage/number than text like novice.

ultidonki:
It's a really nice feature, thanks for making it! I had an idea for a mod a few months ago, where you could only go to new maps by catching eg. 10, 20, or 30 tuxemon. So this will be useful!  :smiley: 

If the player needed to catch 10 tuxemon before something happened, like a door opened so they could go into the next room - would they need to go back to the PC and sync the data after catching 10 tuxemon before the tuxepedia_progress variable is updated?
It would be good if it could be done automatically, ideally, but not a big deal. :)

JaskRendix:
No worries, it's done automatically

JaskRendix:
The variable at the moment is update only when at the PC, but I would like to add the function to the phone too.

PR #1570, opened 2023-01-17
fix status (miss)

### LINKED ISSUE #1567: Tonguespear still leeches life if it misses
A simple issue really. The move Tonguespear should only leech life if it hits. Right now, even if it misses, it will still leech life after every turn

EDIT: aardorn uses the move

JaskRendix:
PR addresses the following:
- fixes #1567;
- creates the effect enhance (enhance.py) > the damage.py for special range techniques
- the point 2 allows to fix (1) the constant fail of special range techs (hibernate, boulder, etc), (2) shows the tackle and (3) shows the animation
- added the check if the target or user monster have more than 0 HP;
- ~~excluded items (before if the player used a potion and the target monster was under lifeleech, it leeched, not anymore);~~
- ~~excluded skip (when you skip move, the monster don't leech);~~

Schematically it's better to understand what we changed.

Situation before:
- there is no connection between damage.py (where we know if a technique misses or hits) and whatsoever.py (condition)
```
damage.py
tech.accuracy >= random.value (1)

poison.py
tech.potency >= random.value (2) -> no clue if the technique hits or misses (act independently)
```
Situation after:
- there is connection between damage.py and whatsoever.py (condition)
- random value 1 is saved in variable, so we can retrieve it inside whatsoever.py (condition)
```
combat.py
random.value (1) <----- moved here the random from damage.py (centralized + saved in variable)

damage.py
tech.accuracy >= saved(random.value (1))

poison.py
tech.accuracy >= saved(random.value (1)) <----- added so we can check if the technique hits or misses
tech.potency >= random.value (2)
-> now the condition considers if the technique hits or misses 
```

black + isort + tested

### DIFF
--- a/mods/tuxemon/db/technique/all_in.json
+++ b/mods/tuxemon/db/technique/all_in.json
@@ -12,7 +12,7 @@
   "is_area": false,
   "is_fast": false,
   "potency": 0.75,
-  "power": 3.0,
+  "power": 3,
   "range": "touch",
   "recharge": 3,
   "sfx": "sfx_blaster",
--- a/mods/tuxemon/db/technique/amnesia.json
+++ b/mods/tuxemon/db/technique/amnesia.json
@@ -3,6 +3,7 @@
   "accuracy": 0.75,
   "animation": "sparks_gold_alt",
   "effects": [
+    "damage 100",
     "exhausted target",
     "exhausted user"
   ],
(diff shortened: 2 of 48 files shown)


JaskRendix:
After editing, re-editing and re-editing, it looks like the behavior was correct.
Wasted time, but at least it ameliorates the statuses/conditions system.

kerizane (approved):

PR #1573, opened 2023-01-19
fix parse effects (techs and items)

JaskRendix:
It'll follow #1570 
PR allows to use effects (both techs and items) without an additional parameter. Simply damage instead of damage 100, swap instead of swap all, capture instead of capture 100, etc.

Before
```
  "effects": [
    "statchange all" --> "all" was necessary because of the None
  ],
```
After
```
  "effects": [
    "statchange"
  ],
```


JaskRendix:
As planned, it removes the 100 from damage and enhance, etc.
It ameliorates also the update of the hud, because there was a bug.
If the player's monster fainted, then it would update (the graphics) with the name of the winner.

eg rockitten vs aardorn,
rockitten faints, then the name would change temporarily in aardon
before changing again in the next one (the monster added)

kerizane (approved):

PR #1575, opened 2023-01-20
has status (items / techs conditions)

JaskRendix:
PR addresses the addition of has_status.

It was present status, but its duty is to confirm or not if the target has "lifeleech" or "poison" or whatever.

Has_status is a more generic condition check:
```
    "conditions": [
        "has_status target,true" --> target has a status / condition
        "has_status target,false" --> target is free from statuses / conditions
    ],
```
There was the possibility to add the parameter to the other status.py (tech and item), but it would be more complex and less clear to understand than a new set of condition checks.

### DIFF
--- a/tuxemon/item/conditions/has_status.py
+++ b/tuxemon/item/conditions/has_status.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from tuxemon.item.itemcondition import ItemCondition
+from tuxemon.monster import Monster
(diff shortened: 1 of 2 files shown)


tuxemon/item/conditions/has_status.py
```diff
@@ -0,0 +1,41 @@
...
+class HasStatusCondition(ItemCondition):
+    """
+    Checks if the creature has a status or not.
+
+    Parameters:
+    expected: true or false
+
+    Example:
+    "conditions": [
+        "has_status target,true"
```
kerizane: We can already do not has_status, so we don't need the second param

kerizane (changes requested):
comment

tuxemon/item/conditions/has_status.py
JaskRendix: True

kerizane (approved):

PR #1576, opened 2023-01-20
style create_npc (maps)

JaskRendix:
PR addresses a small style fix related to create_npc

it acts only on centers and fix the numeration of the postannouncement, so it's not resorted (every time) automatically in Tiled.

### DIFF
--- a/mods/tuxemon/maps/player_house_downstairs.tmx
+++ b/mods/tuxemon/maps/player_house_downstairs.tmx
@@ -144,14 +144,14 @@
   </object>
   <object id="35" name="postannouncement" type="event" x="16" y="80" width="16" height="16">
    <properties>
-    <property name="act1" value="wait 1"/>
-    <property name="act2" value="npc_face npc_mom,right"/>
-    <property name="act3" value="translated_dialog ohnoes"/>
-    <property name="act4" value="wait 0.7"/>
-    <property name="act5" value="translated_dialog ohnoes2"/>
-    <property name="act6" value="npc_face npc_mom,left"/>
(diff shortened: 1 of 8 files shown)


Qiangong2:
How does this change affect NPCs that are supposed to wander rather than stand?

JaskRendix:
It doesn't change anything, since the wander action is done by npc_wander.
The "wander" is related to the behavior, not the moving, but it's not used (excluding the only case in junkyard city mart above).

JaskRendix:
@Qiangong2  do you still need these NPCs files?

- professor-cardiling.json
- professor-rockitten.json
- professor-tweesher.json

Qiangong2:
Yes, I have to use three different NPCs to designate which Tuxemon the professor will use in battle (since it's based on whichever one you choose).

kerizane:
What is this fixing?

JaskRendix:
Simply style, removing the commas, so we can act on the action and leave only the behavior part.
(added create_npc.py - action, I was talking about this file)

Qiangong2:
I've personally never used `npc_wander`. I've always placed wander in the `create_npc` action, so I'm concerned it may break things

JaskRendix:
No problem

JaskRendix:
@Qiangong2 no problem, I removed entirely Xero.
By the way you used `npc_wander`:
```
    <property name="act1" value="create_npc bob,45,8,,stand"/>
    <property name="act2" value="npc_wander bob,2.0"/>
```

kerizane (approved):

Issue #1558, opened 2023-01-12
Wild Tuxemon stats and XP


Qiangong2:
This has been an issue for a while, but since effort is being made by @JaskRendix to update moves, I thought this would be good to open.

Right now, whenever you encounter a wild Tuxemon, its stats are far above whatever Tuxemon you have is. I don't know if this is a standardization issue, or something else, but stats should be set at a certain rate as well as the amount of xp gained. It should also scale based on level difference.

DavidLiang1129:
> This has been an issue for a while, but since effort is being made by @JaskRendix to update moves, I thought this would be good to open.
> 
> Right now, whenever you encounter a wild Tuxemon, its stats are far above whatever Tuxemon you have is. I don't know if this is a standardization issue, or something else, but stats should be set at a certain rate as well as the amount of xp gained. It should also scale based on level difference.

I noticed this too, and I was even more confused as to why the numbers that are normally displayed in other RPGS (HP, attack, defense, etc.) are not displayed in Tuxemon or are very "abstract" (only one HP slot, no numbers) and at the same time, It also doesn't show how much XP your Tuxemon got after the battle (sorry, I went off topic, maybe I should open a separate lssues)

JaskRendix:
Yeah, good point, I noticed too something unusual and it was on my radar.
I had something about on my pipeline, since I was going to tackle that part with techniques (as soon as the PR gets merged).

Some tests.

Stats order: armour, dodge, hp, melee, ranged, speed

```
Cataspike: grub
Rockitten: hunter

PC Rockitten lv 25 vs AI Cataspike lv 15 (wild encounter):
ai 154 110 154 88 176 110
pc 128 256 160 256 128 224

PC Rockitten lv 25 vs AI Cataspike lv 10 (wild encounter):
ai 119 85 119 68 136 85
pc 128 256 160 256 128 224

----> before PR
PC Rockitten lv 25 vs AI Cataspike lv 5 (wild encounter):
ai 119 85 119 68 136 85
pc 128 256 160 256 128 224

-----> after PR
PC Rockitten lv 25 vs AI Cataspike lv 5 (wild encounter):
ai 84 60 84 48 96 60
pc 128 256 160 256 128 224
```

JaskRendix:
Found the issue, addressed in a PR, shortly, there was an "if" for setting min stats at level 10. So of course all the wild monsters you met at lv5, well, they had stats at lv 10.

Qiangong2:
[...]

PR #1577, opened 2023-01-21
update set level()

JaskRendix:
PR addresses:
1. split set_level() into set_level() and set_moves()
2. removing the "learn" component from load_from_db (monster.py)
3. added set_moves close to each set_level -> only exclusion evolving process (only set_level)

Digging deeper I found another issue and the issue was connected to set_level:
`    def set_level(self, level: int = 5) -> None:`
it means that you can have a monster lv 3, but fighting like a lv 5 (related to the #1558 )

Why splitting set_level? Because of the evolution process and max_moves (4)
In this way we have two distinct processes: (1) for levels + stats and (2) for moves 
No interference, no mixing up together.

schematically before
```
-> load_from_db()
--> self.level = random.randint(2, 5) -> 5
--> learn moves (self.level(5)
---> set_level(50) -> learning all the moves not added from lv5 to lv50
----> add the monster
```
schematically after
```
-> load_from_db()
--> set_level(50) 
--> set_moves(50) 
---> add the monster
```
In this way, when we use add_monster or random_monster (eg. a monster lv 30 with 8 moves), it'll come out with the last 4 and not 8.


tuxemon/event/actions/set_monster_level.py
```diff
@@ -42,7 +52,11 @@ def start(self) -> None:
                 return
 
             monster = self.session.player.monsters[int(monster_slot)]
-            monster.set_level(monster.level + int(monster_level))
+            new_level = monster.level + int(monster_level)
+            monster.set_level(new_level)
+            update_move(monster, new_level)
```
kerizane, line 57: Can't we just use the existing monster.update_moves?

kerizane (changes requested):
See comment

tuxemon/event/actions/set_monster_level.py
JaskRendix: The existing monster.update_moves? Are you referring to the old piece of code that was inside set_level? I moved it inside set_monster_level
```
            if (
                move.technique not in (m.slug for m in self.moves)
                and move.level_learned <= level
            ):
```
Do you want that we create a new def update_moves inside monster.py?

kerizane: Can't we use monster.set_moves instead the local function update_move, which is pretty similar?

JaskRendix: Unfortunately we can't. 
Let's say someone uses set_monster_level, so it increases the level of rockitten from 10 to 20, 10 levels
Rockitten knows A, B, C techniques
if we use set_moves, then the result would be:
Rockitten knows A, A, B, B, C, C and D, E, F
(without considering there is the MAX_MOVES parameter in max_moves, so it would be reduced to the last 4, depending on the level)
alternatively we can remove the moves and leaving only the increase of the level, which it would make sense, and I would agree too

tuxemon/monster.py
```diff
@@ -544,13 +541,25 @@ def set_level(self, level: int = 5) -> None:
...
+        Set monster moves according to the level.
+
+        Parameters:
+            level: The level of the monster.
+
+        """
+        moves = []
         for move in self.moveset:
-            if (
-                move.technique not in (m.slug for m in self.moves)
```
kerizane, line 550: Why did you move this check into a new method?

kerizane (changes requested):

tuxemon/monster.py
JaskRendix: for the same reason I explained in the other comment:
- monster,py (it gives the monster all the techniques from 0 to LEVEL);
- set_monster_level.py (it gives the monster all the techniques from LEVEL to NEW_LEVEL)

JaskRendix: the ex lines 549 and 550 were necessary because there was this (ex lines 317, 318, 319)
```
                if move.level_learned <= self.level:
                    technique = Technique(move.technique)
                    self.learn(technique)
```
we were forced to add that method, now there is no need anymore

distinction:
- monster,py (it gives the monster all the techniques from 0 to LEVEL);
- set_monster_level.py (it gives the monster all the techniques from LEVEL to NEW_LEVEL)

kerizane (approved):

kerizane (approved):

Issue #1517, opened 2022-12-27
Unfinished Monsters


JaskRendix:
Here https://wiki.tuxemon.org/Creature_Progress_Tracker the complete list of all the unfinished monsters.

These monsters aren't in the game at the moment.

Here some of the monsters from the list above:

[Aerodin](https://wiki.tuxemon.org/Aerodin) evolves into [Aerostag](https://wiki.tuxemon.org/Aerostag) or [Aerodeer](https://wiki.tuxemon.org/Aerodeer)
**Axylightl** evolves into [Ampystoma](https://wiki.tuxemon.org/Ampystoma)
[Animatueur](https://wiki.tuxemon.org/Animatueur) no evolution
 [Appariron](https://wiki.tuxemon.org/Appariron) no evolution
**Banling**  evolves into [Bansaken](https://wiki.tuxemon.org/Bansaken) evolves into [Banvengeance](https://wiki.tuxemon.org/Banvengeance)
**Beenstalker** evolves into [Wendigger](https://wiki.tuxemon.org/Wendigger)
[Boxorox](https://wiki.tuxemon.org/Boxorox) no evolution
[Claymorior](https://wiki.tuxemon.org/Claymorior) evolves into **Regalance**
**Cubfort** evolves into [Ambearance](https://wiki.tuxemon.org/Ambearance)
[Delfeco](https://wiki.tuxemon.org/Delfeco) is one of **Dollfin** evolutions (miss back and face sprites)
[Eskipup](https://wiki.tuxemon.org/Eskipup) evolves into [Houndice](https://wiki.tuxemon.org/Houndice)
[Flaminggo](https://wiki.tuxemon.org/Flaminggo) no evolution
[Flounce](https://wiki.tuxemon.org/Flounce) evolves into [Knindling](https://wiki.tuxemon.org/Knindling)
[Forighness](https://wiki.tuxemon.org/Forighness) no evolution
[Gladiatorbug](https://wiki.tuxemon.org/Gladiatorbug) is one of **Katacoon** evolutions
[Helipi](https://wiki.tuxemon.org/Helipi) evolves into [Coppi](https://wiki.tuxemon.org/Coppi) evolves into [Parappi](https://wiki.tuxemon.org/Parappi)
[Jeluna](https://wiki.tuxemon.org/Jeluna) no evolution
[Konuit](https://wiki.tuxemon.org/Konuit) no evolution
[...]

PR #1580, opened 2023-01-22
fusion1 and fusion2 fields - JSONs monsters

JaskRendix:
PR originally was the addition of the fusion1 and fusion2 fields (last thing missing from the wiki)
Missing fusion 1 and 2: [fusion.ods](https://github.com/Tuxemon/Tuxemon/files/10460792/fusion.ods) 

Updated as per https://github.com/Tuxemon/Tuxemon/issues/1517#issuecomment-1408117994

BASIC = children, all the others are adults and can breed, so:
- updated the action breeding
- updated the condition has_breeder
- fixed a bunch of monster that needs to be neuter


JaskRendix:
monster
```
{
    "slug": "rockitten",
    "category": "egg",
    "ai": "RandomAI",
    "fusion1": "",
    "fusion2": "",
    "moveset": [],
    "evolutions": [
        {
            "path": "standard",
            "at_level": 5,
            "monster_slug": "random monster VARMINT"
        }
    ],
    "shape": "varmint",
    "stage": "stage0 or baby",
    "types": [
        "earth"
    ],
    "possible_genders": ["neuter"],
    "txmn_id": 0,
    "height": 0.30, ostrich egg
    "weight": 1.4, ostrich egg
    "catch_rate": 170.0,
    "lower_catch_resistance": 0.95,
    "upper_catch_resistance": 1.25
}
```
or item
```
{
  "conditions": [
  1 week, 2 weeks, 1 month?
  ],
  "effects": [
    "evolve random VARMINT"
  ],
  "slug": "varmint_egg",
  "sort": "utility",
  "sprite": "gfx/items/varmint_egg.png",
  "target": {
    "enemy monster": 0,
    "enemy team": 0,
    "enemy trainer": 0,
    "own monster": 2,
    "own team": 0,
    "own trainer": 0,
    "item": 0
  },
  "type": "Consumable", (because it's going to evolve)
  "usable_in": [
    "WorldState"
  ],
  "use_item": "combat_used_x",
  "use_success": "generic_success"
}
```

Sanglorian:
Thanks heaps, this is exciting! I've flagged in the Discord to see if there are any thoughts, to make sure there's full feedback.

JaskRendix:
Updated as per https://github.com/Tuxemon/Tuxemon/issues/1517#issuecomment-1408117994
updated also the action breeding + the condition has_breeder

mods/tuxemon/db/monster/aardart.json
```diff
@@ -2,6 +2,8 @@
     "slug": "aardart",
     "category": "anteater",
     "ai": "RandomAI",
+    "fusion1": "aard",
```
kerizane: Good work, but you probably should have done this in the po file.

tuxemon/event/conditions/has_party_breeder.py
```diff
@@ -32,8 +33,9 @@ def test(self, session: Session, condition: MapCondition) -> bool:
 
         """
         player = session.player
-        if any(t for t in player.monsters if t.gender == "male"):
-            if any(t for t in player.monsters if t.gender == "female"):
-                return True
+        if any(t for t in player.monsters if t.stage != "basic"):
```
kerizane: What happens if you have an adult male and a basic female?

kerizane (changes requested):

tuxemon/event/conditions/has_party_breeder.py
kerizane: Plus the typehint is wrong, because it's possible for you to return `None`

I suggest this:
```
        if any(t for t in player.monsters if t.stage != "basic" and t.gender == "male"):
            if any(t for t in player.monsters if t.stage != "basic" and t.gender == "female"):
                    return True

        return False
```

JaskRendix: I agree with your suggestion

mods/tuxemon/db/monster/aardart.json
JaskRendix: Why?... Do we want to be translatable?
Could you be more specific what I should have done in the PO file?
Do you mean in locale.py? Splitting in monster.py?

JaskRendix: Do we need to remove all the fusion1/2 from here and from the wiki?

JaskRendix:
removed neuter fix and split in another PR

I have a list with all the msgid for the particles fusion1/2 that we can add in PO as requested.
410 msgid (total 820 lines) by the way  [fusion.txt](https://github.com/Tuxemon/Tuxemon/files/10579350/fusion.txt)


we throw the data in the JSONs or inside the PO, your choice @kerizane 
alternatively we split in the middle the name based on the length, so we have no data at all, no JSONs or PO, but then we need to clean up or update the wiki.

kerizane:
I think the fusion1, fusion2 fields should be in the PO file, as we will want them to be translatable.  @Sanglorian what do you think?

JaskRendix:
Honestly I'm not so sure.
There were discussion against translating the monster name. I agree and I'm not native English.

Sanglorian:
> I think the fusion1, fusion2 fields should be in the PO file, as we will want them to be translatable. @Sanglorian what do you think?

Yeah, that makes sense to me.

JaskRendix:
Okk, as you wish, added all the 400+ msgid to the PO file.

msgid: "fusion_XXX" so we can avoid potential duplicate, added validator in db.py too.

probably I'll add the fusion1 and fusion2 to Tuxepedia, but not in this PR.

kerizane:
Hmmm, I was expecting something like:
fusion_aardorn_p1 = aar
fusion_aardorn_p2 = dorn

I'm worried that the way you've done it might cause issues in the future. What if the translator wants to use different prefixes for Aardart and Aardorn? I think this should be allowed. There are a lot of different languages, and I don't know how things will sound in various languages, so I don't want to enforce a strict mapping between the prefixes and suffixes for each name.

My way will require duplicate PO entries, which is annoying, but if we do hit an issue later on it will be a pain to sort out if we have translations implemented in multiple languages and we have to refactor it. I would rather go for maximum configurability from the start.

As an example, lets look at the translations for the pokemon bulbasaur.

English:
Bulbasaur, Ivysaur, Venusaur
German:
Bisasam, Bisaknosp, Bisaflor

There isn't a one-one mapping here.

kerizane (changes requested):
see comment

JaskRendix:
It makes sense, now I understand why you were talking about the PO file.
I'm going to fix it.

JaskRendix:
Done as you asked, it's better, it works like description, no pointless field inside the JSON, everything in the PO

vnmabus:
For me that looks like a nightmare to translate and maintain... I would rather prefer to have a heuristic that works in the 90% of the cases, and a way to put a special mark inside the Tuxemon name for splitting in the cases that fail (such as a zero-width space, or some character that does not appear in names and can be removed safely (e.g. a vertical var `|`).

JaskRendix:
I don't know what to say. I proposed something similar two weeks ago, but I agree with @kerizane

kerizane:
Hmm, the idea of a separator is quite good actually... A single entry in the po file "Aar|dorn" would be easier than having three entries that need to be kept in sync.

Apologies, @JaskRendix this was probably a pain to do... 

@JaskRendix, I'm aware this has been hanging around for a while. If you want to get this feature merged quickly then we could go with a simple heuristic. Keep your changes in another branch, in case we decide to go with my proposal. We can double-check with other contributors what they think. Don't update the wiki yet.

I don't mind what heuristic you go with, we can always change it later. Split down the middle? We need to make a choice about how to handle names with an odd number of characters. You could validate your heuristic against the manually created data and see how well it performs. This step is optional, I don't want to create too much work for you.

I don't think you should implement the override "|" yet until we have agreement from other people, as I don't want people using it, and triggering more work that might need to be undone.

Also, I appreciate the work you've done on this ticket and others. Sorry if I take a while to review things, but I'm very busy at the moment.

JaskRendix:
No, I have no rush, that's ok.
It was the right choice to get rid of the JSONs fields, 5 files involved are better than 200+
I'm going to close the PR so I can think about the splitting. Eventually we can always reopen it.

kerizane:
One heuristic would be that we take into account the pattern in name in the evolutionary tree. So it would notice that aardorn and aardart both start with "Aar" and make that the prefix. And if there isn't a tree or a pattern then we just split down the middle.

PR #1581, opened 2023-01-23
level up combat.py + fixes

JaskRendix:
The PR:
- removes **Learn New Moves** from monster.py;
- creates two defs related to **Learn New Moves** inside combat.py;
- adds the variables **self._level_before** and **self._level_after** in combat.py, so we can know if there is a level_up (included if the monster levels up multiple times);
- adds the **tuxemon_new_tech** to the PO file as well as the self.alert (the old logger string) -> see screenshot;
- adds **build_hud** refresh if leveling up,;

As you can see (testing)
Screenshot from 2023-01-23 10-35-40

Fixed:
- before there was no update in combat (if the monster leveled up) and the only way to know it: switch out and switch in or at the end checking the monster menu -> now the hub will refresh, the number lv XX will update;
- before if the monster was leveling up multiple times (eg. from lv 5 to lv 9), then there was only the checking at lv9. In case there was a technique set at lv8, well, nothing -> now it checks lv6, lv7, lv8 and lv9;

This isn't the definitive version, because it lacks of the overwriting step (if the monster has more than 4 moves), but it helps to simplify the next steps. It remains to know if we want the overwriting method (techniques) to happen between turns (after the adversary monster faints) or at the end of the fight like the evolution.

black + isort + tested

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -1312,6 +1312,9 @@ msgstr "Are you sure you would like to release {name}?"
 msgid "tuxemon_released"
 msgstr "{name} has been released."
 
+msgid "tuxemon_new_tech"
+msgstr "{name} learned technique {tech}!"
+
 msgid "evolution_confirmation"
 msgstr "{name} is trying to evolve into {evolve}. Allow evolution?"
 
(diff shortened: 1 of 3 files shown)


kerizane:
I would do it between turns. @Sanglorian ?

kerizane:
I will wait until your have implemented this fully before I merge.

kerizane (approved):

kerizane:
nvm, it doesn't break anything so I merged it

PR #1583, opened 2023-01-25
nu phone: functions

JaskRendix:
PR addresses the implementation of the phone.
the item has been moved to #1600 while the other functions will appear here.

4 columns since at the end there will be 4 apps, but we can expand later eventually
Screenshot from 2023-02-07 17-02-20
banking
Screenshot from 2023-02-07 15-40-37
it opens a choice menu with the following options: **100, 200, 500, 1k, 2k, 5k, 10k, 20k, 50k and 100k**
depending on how much the player has, you cannot deposit more than you have, etc.
why not using the input? because it needs to be refactored (pygame_menu)

----

contacts, default mom and professor
contacts is saved (npc.py), a simple dict[str, str]

the number is max **3 digits** (5 before, but it was too many digits)
Screenshot from 2023-02-07 15-40-48
when you click on MOM, it'll appear Call MOM and if it doesn't exist the msgid inside the PO, then it'll appear: **Phone keeps ringing but no answer!**

```
            # new string map + phone
            new_label = map_name + phone
```
eg spyder_cotton_scoop + 692 (phone number Professor)
`spyder_cotton_scoop692`

Some modders can set different phone number and create new dialogues.
Some modders can totally ignore this implementation, no issue, the phone will keep ringing at each try from the player.

Modder's choice.

new actions:
```
    <property name="act1" value="add_contacts slug,123"/> --> number 3 digits
    <property name="act1" value="remove_contacts slug"/>
```

if there are other suggestions. I would like to autogenerate comprehensible msgids, so that the modders can fill the PO with different conversation depending on the map. In this way the only PO hardcoded is the one about the phone ringing.

----
map
Screenshot from 2023-02-08 08-34-11

prototype, more functions will follow
it needs to be updated: the river, inserting Taba Town, eventually other location of the Fondent region.


Sanglorian:
Thanks @JaskRendix 

Rather than have the phone behave as an item, I was imagining it as its own entry in the start menu. Clicking it would reveal a sub-menu with each app (as you've done with E-BANKING). 

In the Spyder campaign, you get the phone when you first enter the cafe in Cotton Town. You get the Map App/Map at the same time, and the Tuxepedia soon after (Event 6). You get the Renaming App/Rename in the Mansion. Contacts App/Contacts in Timber Town. 

I forgot the Banking App/E-Banking -- I think you should get that at the start too. 

The Banking App/E-Banking should only be available in towns and cities. But the other apps should be available anywhere.

JaskRendix:
Ah ok, it wasn't clear.

JaskRendix:
@Sanglorian 
added the phone at the beginning (with e-banking included), as you wanted it's unable to access the e-banking outside town (excluded center and shops).

As you wished it's doesn't behave like an item. Look at the screenshot above and tell if it's fine.
Tuxepedia app and other apps will come later.

Sanglorian:
Thanks, that looks good to me! Longer term, I'd like the player to get the Nu Phone in Cotton Town rather than at the start.

What I imagined for E-Banking is that you can deposit money in an account, so you don't lose it even if you lose a trainer battle. I think just displaying the player's money should be done either under BAG or PLAYER.

JaskRendix:
@Sanglorian it makes me smile, I didn't know about the bank account. Same about the player losing money if loses a trainer battle (how much the player is going to lose? half of the potential win? because it's pretty generic information and I would like to avoid more surprises). 

"longer term", random questions popped:
- What are the other events where the player loses money (excluding buying)? Here the modders can create multiple events, but I'm referring to coded mechanism, because loses during the battle is coded.
- Are you planning to integrate loans or other functions?
- Will the payment in the shop through the bank account or pocket money or both?
- Will the big amount of money at the start inside be in the bank account (that the player will be able to access only in Cotton Town)? If yes, what percentage will be in the bank account and what in the pocket?
- Are you planning multiple currency and currency conversion?
- Has the community agreed about the new name of the actual currency (symbol included)?
- Without the phone, from where the player is supposed to access the bank account? PC?
- Do we need to integrate at least a location (map) bank inside at least one town?
- Is the app renaming truly necessary or is it a Spyder feature? Renaming what? Player? Monsters? Both? I'm asking because in Xero there are two NPCs inside the Center that do this. I know it's another scenario, just wondering.
- How it works the app contact? There will be inside all the NPCs that the player fought? Is it choice of the modder to add specific people (it would require a new set of actions)?
- Is the phone supposed to do calls? If yes, is the phone able to call without the app contacts? If yes, which are the numbers included in the default version? Mom? Professor? Modder choice?
- Is this only related to Spyder or it'll be a common mechanics for other scenarios too? I'm asking about the discussion in Discord, because I'm unaware about the forum, when I arrived the forum was already gone since months, maybe a year.

@Qiangong2 please tell me if you want the phone at the start, or at some random point, or not at all.

Qiangong2:
I hadn't planned on a phone feature, but I could definitely write it into the story. Make some cheesy omnichannel quip about it. "Why's it called the NU phone? Because it's Never Unneeded!" Or something to that regard.

I'd personally like it to only be enabled after you defeat the first battle area since that would be easier to integrate into the campaign.

JaskRendix:
@Qiangong2 no problem, I adjusted the PR, so you can insert later on in Xero.

Adjusted the add_item, now inside the Cafe, updated the save_upgrader too.

Sanglorian:
Sorry for the delay, really appreciate your work on this. 

> @Sanglorian it makes me smile, I didn't know about the bank account. Same about the player losing money if loses a trainer battle (how much the player is going to lose? half of the potential win? because it's pretty generic information and I would like to avoid more surprises).

I think the player should lose half of their money. That's why the bank account is so important, to limit the amount of money lost.

> "longer term", random questions popped:
> 
>     * What are the other events where the player loses money (excluding buying)? Here the modders can create multiple events, but I'm referring to coded mechanism, because loses during the battle is coded.

Losing a battle is the only one I can think of. 

> 
>     * Are you planning to integrate loans or other functions?
> 

No.

>     * Will the payment in the shop through the bank account or pocket money or both?
> 

Through pocket money. 

>     * Will the big amount of money at the start inside be in the bank account (that the player will be able to access only in Cotton Town)? If yes, what percentage will be in the bank account and what in the pocket?

None in the bank account to start.

> 
>     * Are you planning multiple currency and currency conversion?

No, I think that would be very confusing. 

>     * Has the community agreed about the new name of the actual currency (symbol included)?
> 

I think if we use {{currency}}, different symbols could be used. Some options previously discussed are 飞T$ㄞ¤§. I think $ keeps it simple for now. 

>     * Without the phone, from where the player is supposed to access the bank account? PC?
> 

Yeah, maybe a PC in cafes. 

>     * Do we need to integrate at least a location (map) bank inside at least one town?
> 

No, I don't think so. 

>     * Is the app renaming truly necessary or is it a Spyder feature? Renaming what? Player? Monsters? Both? I'm asking because in Xero there are two NPCs inside the Center that do this. I know it's another scenario, just wondering.
> 

Renaming App is for monsters. If it's easier to do it via NPCs, I don't mind that either. 

>     * How it works the app contact? There will be inside all the NPCs that the player fought? Is it choice of the modder to add specific people (it would require a new set of actions)?

Not developed yet. Certain NPCs would add their names when you talk to them. Then calling them would initiate a conversation. 

Happy to avoid it for now. 

> 
>     * Is the phone supposed to do calls? If yes, is the phone able to call without the app contacts? If yes, which are the numbers included in the default version? Mom? Professor? Modder choice?
> 

Eventually yes, via the Contacts App -- but we can omit for now. 

>     * Is this only related to Spyder or it'll be a common mechanics for other scenarios too? I'm asking about the discussion in Discord, because I'm unaware about the forum, when I arrived the forum was already gone since months, maybe a year.
> 
> 

It might just be for Spyder, but some of the functions will be useful for other campaigns too I think. 

> @Qiangong2 please tell me if you want the phone at the start, or at some random point, or not at all.

JaskRendix:
@Sanglorian some shit has been done, look at the screenshots.

Who knows, maybe there is around some unknown mockup.
By the way, did you notice that Tuxepedia is based on your mockups?

Qiangong2:
The phone should also only be available after the first battle area

JaskRendix:
@Sanglorian now the player will receive phone + tuxepedia (other PR) + bank + map in cotton town cafe;
```
  <object id="20" name="First Visit to Cotton Cafe" type="event" x="80" y="96" width="112" height="16">
   <properties>
   [...]
    <property name="act55" value="set_variable visitedcottoncafe:yes"/>
    <property name="act56" value="add_item nu_phone"/>
    <property name="act57" value="add_item app_banking"/>
    <property name="act58" value="add_item app_map"/>
   [...]
   </properties>
  </object>
```
while app contacts, you said Timber Town, but I have no idea where or when, I looked all the PO file and there is no sign of "contacts". So, do you want to create a new dialogue? Do you already know the precise event? Do you want that I come up with an event in Timber Town where someone installs the app?

Here in this event (Timber Town): 
`  <object id="329" name="Stop! - Scoop" type="event" x="464" y="368" width="16" height="16">`
where Dante says (**spyder_timber_dante2**)
`msgstr "Hey buddy! \n Are you planning to break into these HQ? Cool! \n Make sure you're well prepared before you do, because it's difficult to get out once you're inside!"`
we can add
`You'll need to call your mom to let her know you're safe. Use this app.` or something else

Sanglorian:
That's perfect, thank you - including the Timber Town new event. Very nice way of doing it. 

And those screenshots look great!

kerizane (approved):

PR #1585, opened 2023-01-26
player menu

JaskRendix:
PR addresses the implementation of the player menu.
- created the variable "**date_start_game**", it'll record the date when the player start the game, so it'll show the number of days (screenshot), other fields if necessary can be added in a second time;
- updated the **PLAYER_NAME_LIMIT** of characters (player) from 30 to 15 (screenshot);

font_size RED = 30
font_size everything else = 15
the position is obtained through percentages, not px

Screenshot from 2023-01-30 19-09-40
the S of seen and C of caught are uppercase.


Sanglorian:
Looks great! 

Another other thing we could add: "Date your adventure began" (i.e. date the game was started). And anything else that comes up along the way.

tuxemon/formula.py
```diff
@@ -333,3 +333,9 @@ def rematch(
     # restore hp evolved monsters
     for mon in opponent.monsters:
         mon.current_hp = mon.hp
+
+
+def synch(player: NPC, value: int, total: int) -> float:
```
kerizane: I'd prefer "sync".

https://www.merriam-webster.com/dictionary/sync

tuxemon/formula.py
```diff
@@ -333,3 +333,9 @@ def rematch(
...
+    percent = round((value / total) * 100, 1)
+    player.game_variables["tuxepedia_progress"] = str(round(percent))
```
kerizane: Why are we rounding it twice?

kerizane (changes requested):

tuxemon/formula.py
JaskRendix: Thanks for the dictionary, I'll update

tuxemon/formula.py
JaskRendix: Forgot, update

kerizane (approved):

DavidLiang1129:
It seems inappropriate for me to suggest this (since the changes have been merged), but is the second half a bit empty? For the second half, my suggestion is to add a "introduction to the Player" field that can be seen by other players in multiplayer mode, what do you think? (This is not a problem to be solved in a hurry, and I'm sorry if it bothers you.)

PR #1586, opened 2023-01-27
filter items during combat

JaskRendix:
During combat, after opening items

before
Screenshot from 2023-01-27 09-45-13
after
Screenshot from 2023-01-27 09-45-41

it doesn't make sense to show items that cannot be used during the fight (without MainCombatMenuState in usable_in)


Sanglorian:
Such a good idea!

kerizane (approved):

PR #1587, opened 2023-01-27
set gender sprite

JaskRendix:
until now it wasn't possible to switch between male and female easily, this PR addresses this issue
it creates two new actions **set_player_gender** and **set_player_name**
the first to switch gender (sprite too)
`<property name="act1" value="set_player_gender female"/>`
the second to change the name without opening the input (useful for modders to set a default name for their adventure)
`<property name="act1" value="set_player_name donald"/>`

Steps:
- removed **player_npc** from Tuxemon.cfg, left only NPC_RED inside the **worldstate.py**;
- added the field **playable** (NPC) in db.py > **bool** true/false > set default as **False**, the only **True** is NPC_RED;
- create two fields inside npc.py, both are saved;

-> (1) player_gender -> unsure about the name "player_gender", copied player_name, if there is some suggestion, is welcome
-> (2) player_sprite -> unsure about the name "player_sprite", copied player_name, if there is some suggestion, is welcome

- updated **save_upgrader** by inserting the new field with **male** and **adventurer** as default;

now when the game loads, it checks if the NPC is playable, if it's playable, then based on the saved gender, it switches sprite; the same happens with the set_player_gender, it switches sprite immediately.

basic option: adventurer is associated to male vs heroine is associated to female
fixed combat_back too

tested, isort, black


Sanglorian:
Very nice, thank you.

kerizane:
I like the idea, but maybe we should be changing the player NPC? Potentially we would like to support more NPC sprites.

So rather than set_player_gender we would have, set_player_npc and set it to npc_heroine (or whatever)

JaskRendix:
Removed the **player_gender**, left only **player_sprite**, edited **set_player_gender** in **set_player_sprite**

I would like to avoid to save the NPC slug, I think sprite is more flexible, I'm going to upload a generic combat_back to cover all the sprites different from adventurer and heroine (the only ones who have _back)

**PCSprites class in db.py lists the complete ones: sprite_name, front and back (temporary solution)**

JaskRendix:
@Sanglorian now we can change to any sprites, but there is the issue about the missing _back for many sprites. Do you agree about putting the following as default combat_back sprite if it's missing? It could drive people to create new ones. It's the same you used for mk01
default_back

Sanglorian:
Even better, thanks!

JaskRendix:
Now the the sprite will change in the player menu too.

kerizane:
If the back sprite is missing could we just skip the combat animation that requires it? It's not as important as the overworld sprites.

kerizane:
@vnmabus could you take a look at this?

JaskRendix:
> If the back sprite is missing could we just skip the combat animation that requires it? It's not as important as the overworld sprites.

It could be an option, but I need to understand how it works the animation.

kerizane:
Ok, not a blocker. Could you check if the animation looks better with a ? or with don't set anything at all?

kerizane:
The ideal solution would probably be to have an action to set the pronouns that dialogue will use (do we need to think about the dialogue now? are we assuming the player is male?) and set_sprite to give us more flexibility about what sprites we can use.

But this solution is fine.

kerizane:
I will merge this in a day or two, assuming nobody objects. Sorry for taking so long.

JaskRendix:
That's ok, because the more I look at this, the less I'm convinced.
Sprites are connected with template (class) and we need to simplify NPC and company, because if not it becomes something similar to the old item/inventory (really messy)
I'm going to close it. I need to do some tests.

PR #1588, opened 2023-01-28
route7 + update sphalian and buddha mountain

JaskRendix:
Updated maps + new route7
still work in progress, but at least it's a good first step, a sort of open air dungeon, still missing NPCs in the maps (trainer battles), will appear later
buddha mountain -> monsters are range level 55-65 day 65-75 night
route 7 -> monsters are range level  40-50 day 50-60 night
changed the base of cliff background because grass is more good looking than granite

the maps:
sphalian_town
route7
buddha_mountain

some background wild encounters:
mountain
Screenshot from 2023-01-28 10-26-36
route7
Screenshot from 2023-01-28 10-25-45


JaskRendix:
@Sanglorian it needs your approval because it's Fondent region, although there is no connection with Spyder.

I wonder if we need to create a map with all the Fondent region, technically where you would put route7? East Paper Town? East of Route3? North of Route3? Maybe the best option is another exit for Spyder_tunnel, let me know.

Sanglorian:
If there's no connection to the Spyder campaign, then my approval not needed :) The maps look neat!

kerizane (approved):

PR #1589, opened 2023-01-29
template field NPC json

JaskRendix:
Right now combat_back (npc.json) is a field with only two options: adventurer_back and heroine_back

since we are working on implementing a change of sprites #1587, this situation is going to change, the field combat_back will be "pointless", since the combat_back image will change automatically based on the saved sprite.

**_combat_back and combat_front are only used during trainer battles, precisely during the loading of the battle.
This PR doesn't touch any sprite_name / sprite_**

Now, what I propose is to convert the combat_back fields in template. Template is the NPC class: https://wiki.tuxemon.org/index.php?title=Category:Trainer as resumed on wiki by @Sanglorian 

It consists in:
- strenum with all the 84 classes (templates);
- **combat_back** will be changed in **template**;
- **combat_front** becomes optional;
- **combat_front** will be load only if the NPC template class is **noclass**
example: spyder_nimrod_xeon
```
{
  "slug": "spyder_nimrod_xeon",
  "gender": "neuter",
  "sprite_name": "robot",
  "monsters": [...],
  "combat_front": "xeon", --> so that during the battle will load xeon.png (gfx/sprites/battle/xeon.png)
  "template": "noclass"  ---> noclass
}
```
- if there is a **template** (one of the 84), then it'll be load the respective **template**
example:
```
{
  "sprite_name": "florist", 
  "monsters": [...],
  "template": "florist" --> since it exists the florist.png, then it's not necessary anymore combat_front
}
```
if someone uses a class without image: 
```
beachcomber, childactor, creepydoll, granny, homemaker, maniac, master,
postbox, riverboatcaptain, robot, rocker, shiningsun, shopkeeper and xerogrunt
hell knight too
```
it will load the question mark:
default_back


JaskRendix:
@Qiangong2 since Spyder isn't in development as your scenario, I wanted to ask you some questions. I don't want to break your version.

About xerogrund/xerogrunt, I need to know what do you desire, so I can update the JSONs.
```
  "slug": "xerogrund1",
  "gender": "male",
  "sprite_name": "xerogrunt",
  "combat_front": "adventurer",
  "template": "xerogrunt" 
```
it would be 

```
  "combat_front": "knight", -> or something else (anything inside gfx/sprites/battle)
  "template": "noclass" 
```
or 

```
  "template": "xerogrunt" 
```
--> xerogrunt.png doesn't exist, there will be the question mark image (see above)

I've seen you used:
```
    "combat_front": "disciple.png",
    "combat_back": "adventurer_back.png",
```    
it would be
```
    "template": "disciple",
```
I've seen you edited also omnigrunt.json, so from
```
  "combat_front": "enforcer_rookie.png",
  "combat_back": "adventurer_back.png",
```
to
`  "template": "enforcer_rookie",`

Qiangong2:
Not every NPC needs to battle, so I was just duplicating NPCs and changing the names. That's why some have "adventurer" and some have "enforcer." None of the NPCs will ever be viewable from the back, so I don't know if the template format is needed(?)

Regardless, `noclass` would probably be the best for NPCs without a back sprite.

JaskRendix:
> None of the NPCs will ever be viewable from the back, so I don't know if the template format is needed(?)

I could say the same thing about the old combat_back. Template is a category, we needed to categorize and import that important piece of data from the wiki (useful to filter NPCs and allow eventually new implementations).

PR #1596, opened 2023-01-30
ai validate tech

JaskRendix:
PR implements the condition checking inside ai.py
Until not AI didn't check the conditions (techniques), example:
```
    "conditions": [
        "current_hp target,<,1.0",
    ],
```

### DIFF
--- a/tuxemon/ai.py
+++ b/tuxemon/ai.py
@@ -110,17 +110,19 @@ class RandomAI(AI):
         """
         Tracks next_use and recharge, if both unusable, skip.
         """
-        # tracks next_use NPC ai
-        filter_moves = []
-        # it choses among the last 4 moves
+        actions = []
+        # it chooses among the last 4 moves
         for mov in monster.moves[-monster.max_moves :]:
             if mov.next_use <= 0:
-                filter_moves.append(mov)
-        if not filter_moves:
+                for opponent in opponents:
+                    # it checks technique conditions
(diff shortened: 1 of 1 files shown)


tuxemon/ai.py
```diff
@@ -112,10 +119,12 @@ def track_next_use(
         """
         # tracks next_use NPC ai
         filter_moves = []
-        # it choses among the last 4 moves
+        # it chooses among the last 4 moves
         for mov in monster.moves[-monster.max_moves :]:
             if mov.next_use <= 0:
-                filter_moves.append(mov)
+                # checks technique conditions
+                if mov.validate(self.opponent(opponents)):
```
kerizane: You use always check the first opponent here, but could end up returning a different one.
I would do something like this:

```
actions = []

for mov in monster.moves[-monster.max_moves :]:
    if mov.next_use <= 0:
    	for opponent in opponents:
    		if mov.validate(opponent):
	    	    actions.append((mov, opponent))
if not filter_moves:
    skip = Technique("skip")
    return skip, choice(opponents)
else:
    return choice(actions)
```
Please test it.

kerizane (changes requested):

tuxemon/ai.py
JaskRendix: It works without issues, I don't know why I didn't think about inserting the for before instead of creating a new def.

kerizane (approved):

PR #1598, opened 2023-01-31
technique overfeed + status_overfeed

JaskRendix:
PR fixes overfeed technique and creates the status_overfeed.
Until now overfeed has not been used because of crashes.

Status overfeed is a negative status, it can be replaced (both positive and negative) and it reduces by 50% (as it was written in the original formula) the enemy monster speed. If you agree I'm going to sync the wiki too.

@Sanglorian is it ok? 

tested, isort + black

reduced recharge from 8 to 5 because on wiki max value is 5


Sanglorian:
Sounds good!

kerizane (approved):

PR #1600, opened 2023-02-01
implementation journal menu

JaskRendix:
PR addresses the implementation of journal.
- adds shapes, stats (speed, armour, etc) and short stats in the PO file
- allows to access the new info monster from the monster menu (independent from Tuxepedia)
- allows to access the new info monster from the PC (pick_up) monster (independent from Tuxepedia)
- it distinguishes between seen, captured and "unseen" monsters by colors / underline

---

Considerations:
- types icons above, words below, the reason is that the icons are fixed, so we need to put them in a places without dynamic element (the name of the type is a dynamic element);
- unable to insert icon type in the monster screen (the nr4 with exp, etc). too much stuff;
- in the nr4 I was able to increase the font only of description/evolution, I can increase the font, but we are going to lose the difference with the monster name, if for your it's ok, then tell me.
- unable to switch by clicking left/right;

regarding left/right, I tried to trigger it, but it doesn't work:
```
        for event in self.client.key_events:
            if event.pressed and event.button == intentions.LEFT:
                change_monster
            if event.pressed and event.button == intentions.RIGHT:
                change_monster
```

I hope this is ok. Please, if still something is wrong, then address precisely the picture by the number.
Because generic indications like "bigger font" don't help, by the way I increased the font where it was possible.

1
Screenshot from 2023-02-09 13-40-11
2
Screenshot from 2023-02-09 13-37-31
3
Screenshot from 2023-02-09 13-46-43
4
Screenshot from 2023-02-09 14-26-32

Next step: #1583
Then we can close: #1093


JaskRendix:
Since I was implementing the thing, I'm going to get rid of the popups here:
Screenshot from 2023-02-02 13-22-41
info will bring here, slightly different from tuxepedia because it needs more info (levels, exp, etc)
Screenshot from 2023-02-02 13-40-37

JaskRendix:
@Sanglorian what do you think? Is it ok?

Sanglorian:
Sorry for the delay. Thanks so much for this, it's looking good. This is what we've been calling the Tuxepedia App, but it probably is better for it to come up as its own menu item ("TUXEPEDIA"), not be buried under PHONE. 

I mocked up some ideas for how it could look, on the wiki. Obviously the functionality is more important though. 

1. By pages, 10 or 20 per page, is fine. 
2. Only tuxemon that have been seen or captured should have information provided. The way I've done it in the mock up is that captured = face sprite, seen = blue underlined text, not seen or captured = red. 
3. Once it works this way, perhaps there could be a command that makes other monsters count as "captured" for Tuxepedia purposes. 

image

Here is an idea for how each monster's page could look: 

image

In the Spyder campaign, Tuxepedia would ideally become available after [Event 6](https://wiki.tuxemon.org/Cotton_Town) in Cotton Town.

JaskRendix:
Do you realized that I used your mock-up by removing the info?

JaskRendix:
@Sanglorian 
Do you realize that I used your mock-up by removing the info?
Are you showing me the mock-up for...?
Adding the icon of the type?
Moving on the left the name of the monster?
Changing font?
Increasing the font?
Moving the txmn_id on the right and allow to change monster based on this?

I'll trying to add the mini icon in front of the capture monster as well as changing the color based on seen/captured

I don't know what to think honestly. I wonder what's the difference between the journal and Tuxepedia, because I based my judgement on this #1093 and I thought it was similar with different graphics/names.

JaskRendix:
Screenshot from 2023-02-05 11-51-23
now it allows to click only on monsters blue (seen) or red (captured), black isn't accessible

JaskRendix:
@Qiangong2 do you want to unlock Tuxepedia at the beginning or do you want to mimic Spyder and make it appear only when the player receives the phone? (as you said in the other PR after the first battle area) let me know, if not we can thing about something else.

Qiangong2:
Ideally, after the first battle area

Sanglorian:
Thanks! So, for the mock up I was thinking of: 

* Adding the icon of the type. 
* Increasing the font size
* Moving the tmxn_id on the right and allowing the player to change monster by clicking the left and right buttons

I understand what you say about the red text being distracting. I was going off Wikipedia, where if you link to a webpage that doesn't exist, it shows in red. But maybe it would look bad.

JaskRendix:
It feels like you're not read what I resumed in the first post. 
Moreover you didn't precise if you were referring to the first or second screen (the with or without exp).
So I'm confused, I'll try to interpret.

kerizane (approved):

kerizane:
Looks good. Everyone happy for me to merge?

ultidonki:
@JaskRendix - I've just tested the new tuxepedia journal, it's fantastic! Thanks so much! :smiley:

PR #1604, opened 2023-02-03
update tech menu

JaskRendix:
PR addresses a minor update in the tech menu.
While working on #1600  I noticed a small bug, if a tech has two types (eg. Wood/Earth), it appeared without space **WoodEarth** instead of **Wood Earth**)
Since I was fixing that I added some changes.
The data (monster) is sent by kwargs and not anymore through variable, less code, less pointless circles.

Before
Screenshot from 2023-02-03 10-24-44
After (finally sorted by tech_id, reason why I added the ID in the main menu), more space also for the description or other useful info
Screenshot from 2023-02-03 10-19-45

black + isort + tested
this is only a minor change, because I'm working on refactoring the tech menu and getting rid of pygame/menu for pygame_menu

### DIFF
--- a/tuxemon/states/techniques/__init__.py
+++ b/tuxemon/states/techniques/__init__.py
@@ -2,7 +2,6 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-import uuid
 from typing import Generator
 
 import pygame
@@ -25,7 +24,9 @@ class TechniqueMenuState(Menu[Technique]):
     background_filename = "gfx/ui/item/item_menu_bg.png"
     draw_borders = False
 
-    def __init__(self) -> None:
+    def __init__(self, monster: Monster) -> None:
+        self.mon = monster
+
         super().__init__()
 
         self.item_center = self.rect.width * 0.164, self.rect.height * 0.13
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks! Would be good to have power, potency etc added to the description as well. 

I'd prefer ID in the description not in front of the technique name, but I'm not fussy either way.

JaskRendix:
@Sanglorian done
Screenshot from 2023-02-05 10-50-19

tuxemon/states/world/world_menus.py
```diff
@@ -228,10 +228,9 @@ def open_monster_techs() -> None:
             """Show techniques."""
             self.client.pop_state()
             monster = monster_menu.get_selected_item().game_object
-            self.client.push_state(TechniqueMenuState())
-            local_session.player.game_variables[
-                "open_monster_techs"
-            ] = monster.instance_id.hex
+            self.client.push_state(
+                TechniqueMenuState(kwargs={"slug": monster})
```
kerizane: This isn't how you use kwargs. You do it like this:

`TechniqueMenuState(slug=monster)`
`slug = kwargs["slug"]`

And it's not a slug, it's a monster, right?

tuxemon/states/techniques/__init__.py
```diff
@@ -25,7 +24,11 @@ class TechniqueMenuState(Menu[Technique]):
     background_filename = "gfx/ui/item/item_menu_bg.png"
     draw_borders = False
 
-    def __init__(self) -> None:
+    def __init__(self, **kwargs) -> None:
```
kerizane: Don't really need to pass it via kwargs. Just do:

`def __init__(self, monster: Optional[Monster] = None) -> None:`

kerizane (changes requested):
comments

tuxemon/states/techniques/__init__.py
JaskRendix: Ok

tuxemon/states/world/world_menus.py
JaskRendix: exactly, it's a monster: Monster

JaskRendix: Fixed

tuxemon/states/techniques/__init__.py
```diff
@@ -25,7 +24,9 @@ class TechniqueMenuState(Menu[Technique]):
     background_filename = "gfx/ui/item/item_menu_bg.png"
     draw_borders = False
 
-    def __init__(self) -> None:
+    def __init__(self, monster: Optional[Monster] = None) -> None:
```
kerizane: Sorry, do we expect TechniqueMenuState to always have a monster? In that case it should be 
`def __init__(self, monster: Monster) -> None:`

tuxemon/states/techniques/__init__.py
```diff
@@ -126,42 +127,42 @@ def open_choice_menu() -> None:
...
+        for moves in self.mon.moves:
+            moveset.append(moves)
+
+        output = sorted(moveset, key=lambda x: x.tech_id)
 
-        output = sorted(moveset)
         for tech in output:
-            obj = Technique(tech)
-            type2 = ""
+            obj = Technique(tech.slug)
```
kerizane: Wasn't tech already a Technique? Why are we recreating it?

kerizane (changes requested):

tuxemon/states/techniques/__init__.py
JaskRendix: Yeah, always. No problem, I'm going to address it

tuxemon/states/techniques/__init__.py
JaskRendix: Well done, I almost missed. Yeah, it's a repetition, I'm going to remove it

tuxemon/states/techniques/__init__.py
```diff
@@ -2,8 +2,7 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-import uuid
-from typing import Generator
+from typing import Generator, Optional
```
kerizane: Optional isn't used any more.

kerizane (changes requested):

tuxemon/states/techniques/__init__.py
JaskRendix: right, fixed

kerizane (approved):

PR #1606, opened 2023-02-04
action "add_collision"

JaskRendix:
PR addresses the addition of the action "add_collision", a way to solve the walking on water issue.

The action adds a collision zone based on the event area.

`  <object id="452" name="sea" type="event" x="32" y="240" width="32" height="16">`

it takes the coordinates (32,240) and generates the area based on width 32 and height 16, making the area not accessible (it adds the tuples to the collision_area). Example:
```
  <object id="452" name="sea" type="event" x="32" y="240" width="32" height="16">
   <properties>
    <property name="act1" value="add_collision 452"/> <---- same id number of the event
   </properties>
  </object>
```
This will allow to set specific condition to walk on the sea. I tested it with "ram", because having a rockitten in the party.
`<property name="cond1" value="not has_tech ram"/>`
In this way I was able to walk on the water.
`<property name="cond1" value="is has_tech ram"/>`
In this way not anymore.

I'm aware, it sounds confusing, but because we are adding an inaccessible zone, so there is double negative.

Added the zones on the coast line, moved the captain discussion on the bottom. More ordered. 
spyder_paper_town

Potentially we can do it with objects, specific monsters, etc. Eventually I'm ready to replace on other maps too.


Sanglorian:
This sounds like a major step forward, thank you!

kerizane:
Ready for review? Only problem is this is slightly hard to spell...

What if we want to make it accessible again?

Could call it add_collision or activate_collision and del_collision or remove_collision

JaskRendix:
if we want to make accessible again? yeah, we are obliged to edit the tmx
add_collision sounds better

but you gave me an idea with the remove collision

PR #1607, opened 2023-02-04
encounters other maps

JaskRendix:
Following #1579 
it updates the other wild encounters files.
No big variation, as usual half day, half night, while night slightly stronger than the day ones.

@Sanglorian a question related to spyder_landrace, concini, etc. now there is one for the day, and one for the night, do you want to do the night one stronger? Like 10 levels more, example below. I'm asking because there is only one monster.
ex volcoli
```
      "level_range": [
        50, --> 60
        51 --> 71 
      ]
```
I deleted the file **JSON-o-matic for Encounters.ods** inside the folder.
The reason is: do you want to move the info inside the wiki?
If you want to leave it there, then I can restore, no issues.


Sanglorian:
Thanks, please keep JSON-o-matic for Encounters.ods

No need for different level ranges at night for landrace, volcoli, etc.

JaskRendix:
ok

kerizane (approved):

PR #1610, opened 2023-02-06
fix cafe map

JaskRendix:
PR fixes the spyder_cotton_cafe map, there was an issue in the texture.

Unable to see the wood counter (pieces of the carpet?) as well as the rest of the potions.
spyder_cotton_cafe1
Now it's fixed, since I was there I also replaced the tilesets with CORE_
Fixed CORE_outdoor too because a series of tileset was translated of 1px up


JaskRendix:
@Sanglorian I guess the carpet over the counter was a mistake... right?

Sanglorian:
Thanks - yes, carpet was definitely a mistake. 

BTW, I will get to your journal comments but not tonight sorry -- really appreciate you working on it though.

JaskRendix:
No problem!

PR #1612, opened 2023-02-06
fix tilesets interiors maps

JaskRendix:
Following #1613 
PR removes the following tilesets and replaces these by COREs.

it contains also a fix (teleport) in cotton_town Xero

/stairs.png ----------------> integrated in CORE_stairs
/floorsandwalls.png ----> already inside COREs
/furniture.png ------------> already inside COREs
/electronics.png ---------> already inside COREs
/kitchen.png --------------> already inside COREs
/plants.png ---------------> already inside COREs

It addresses only interior maps (inside).
I removed the above tilesets also from unused maps, just because I wanted to remove every trace.
I have other questions for addressing other tilesets, but I'll open specific PRs.

----

I add for info the unused tilesets (inside the folder, but not used):
/stairs_down.png
/prototyping_outdoor.png
/signs.png
/shade.png
/shade.gif
/shade.bmp
/Medieval_Tiles_by_Chickenshowman.png
/couch.png
/doors.png
/gothicvania.png

already in COREs and replaced, but still inside tileset folder:
/Warmachine_black_and_yellow.png
/Warmachine_yellow_demilitarised.png
/Truck-red-all.png
/Car,_Black_by_Isaiah658.png


JaskRendix:
request @Sanglorian approval

the above 20 files are not used, we can move in another folder (or subfolder) if you want to, so there is no confusion for future modders (always the risk to see again plants.png, etc)

Sanglorian:
Great idea, thank you. Moving to a separate folder seems safest. Most but not all will be on the wiki as a back up.

JaskRendix:
That's good. We need to figure out where we can place the folder and how we can name it, but we can do it in a separate PR.
At least the unused are all listed here.

kerizane (approved):

PR #1613, opened 2023-02-08
fix floor texture

JaskRendix:
PR addresses:
- replacement in Cotton Cafe (both scenarios) - identical maps
- replacement in all the scoops (both scenarios) - identical maps (2 versions)
- replacement in lab professor
- integrated floorsandwalls inside CORE (both walls and floors)

there are still a lot of maps with floorsandwalls.png, but slowly we are going to fix it

@Sanglorian involves the floor texture

I don't know if you created this, but a lot of maps have this tileset:
floorsandwalls
the majority of the elements are present in other COREs, but there is the floor and the blue wall without similar texture.
Example:
scoop now
spyder_paper_scoop
scoop after with alternative wall 1 (only COREs)
spyder_paper_scoop1
scoop after with alternative wall 2 (only COREs)
spyder_paper_scoop1
scoop after with alternative wall 3 (only COREs)
spyder_paper_scoop2

What do you want to do?
Do you want to change to CORE (as in the second picture)?
Do you want to import the 4 texture inside the CORE_floors and respectively CORE_walls?

There are tons of options in CORE_floors, the one above is the most similar for colors, same for CORE_walls.

Let me know


Sanglorian:
Good catch, and thanks for laying it out so clearly. I'd like the missing tiles to be imported into CORE if possible.

JaskRendix:
No problem

JaskRendix:
No problem
I'm going to integrate #1610 here

kerizane (approved):

PR #1616, opened 2023-02-09
fix interiors

JaskRendix:
It follows #1612 and addresses other interiors.

This PR concludes the interior maps.

UPDATED list unused, see below.
there will be a separate PR at the end to move the files in another folder.
---

tilesets "unused"
```
unused tilesets after this PR:

/Objects_by_ArMM1998.png <---- new entry
/Interior_Floors_by_George.png <---- new entry
/Interiors by Redshrike.png <---- new entry
/Interior_Tiles_by_Mike_Bramson.png <---- new entry
/Pacheesi Tiles.png <---- new entry
/Interior_Walls_by_George.png <---- new entry
/Terrain_by_KelvinShadewing.png <---- new entry


unused tilesets:
/stairs.png
/floorsandwalls.png
/furniture.png
/electronics.png
/kitchen.png
/plants.png
/Warmachine_black_and_yellow.png
/Warmachine_yellow_demilitarised.png
/Truck-red-all.png
/Car,_Black_by_Isaiah658.png
/stairs_down.png
/prototyping_outdoor.png
/signs.png
/shade.png
/shade.gif
/shade.bmp
/Medieval_Tiles_by_Chickenshowman.png
/couch.png
/doors.png
/gothicvania.png
```

---

spyder_omnichannel2 was without collisions, I added and:
spyder_omnichannel2 - spyder_omnichannel3
the color is different because events and collisions aren't selected.

Is it ok if I added some office stuff? I like the office stuff texture. If not I can remove it, no problem.

PR #1619, opened 2023-02-10
fix exteriors

JaskRendix:
following #1616 so we can distinguish between interiors and exteriors:
- flowers animated everywhere;
- extended the Tuxemart animation (M moving to almost all the cities);
- minor fixes (often collisions zone outside the map);
- added a new building, red hospital (before there was only the green one);
- created the new core_city_and_country.png, added the source inside the PNG and deleted the old one (identical);
- fixed a bunch of signs and added some;
- fixed some east/west inside the maps;

Spyder and Xero maps exteriors are done.


JaskRendix:
@Sanglorian since the tileset **Outdoor_Tiles_-_City_and_Country_-_by_ArMM1998** is used a lot in cities and routes (less in interiors), I guess we can promote it to CORE_, because honestly I don't know if it makes sense import all of it inside other COREs.

If you agree, then what about the name? Is **core_city_and_country** good? (we'll update the credits too)

JaskRendix:
@Qiangong2 same question about the trees, for Xero do you want to keep the actual trees or do you want one of the two new options? It could be a way to distinguish one from the other. Let me know.

Qiangong2:
I don't necessarily have a preference either way. If I had to choose, I personally prefer the look of tree1 over tree2.

Sanglorian:
Thanks heaps! Promoting city and country to core under that name sounds good to me.

I'm honestly not fussed which tree design we use - up to which you prefer, I'd say.

JaskRendix:
No changes in trees, kept the same, I noticed that we need more signs, but I'm going to complete it in another PR.

Next PRs: signs and move unused files.

kerizane (approved):

PR #1625, opened 2023-02-15
refactor inventory

JaskRendix:
PR involves:
- refactor inventory (massive reduction and simplification, especially in npc.py)
- addition PC State Items (majority of the lines)
- addition items support to JSON (nothing much, it can be used or not by modders)

The functions (actions and conditions) are unchanged, same inputs and outputs.
For the player nothing will change.

Some weeks ago I started to look how to implement the Locker (PC for Items) and it confirmed my opinion that the inventory system needed an upgrade. This PR goes in this direction. Everything can be resumed with:

from **inventory: Mapping[str, Optional[int]]** to **items: Sequence[Mapping[str, Any]]**

inside the PR there is a function to transform all the items saved below "inventory" under "items". So **Item** becomes like a **Monster**, now each item has finally a **uuid** and the quantity is managed as a monster level or weight. It simplifies a lot, it reduces the number of messy functions (especially the ones related to buy/sell).

---

@Sanglorian I need a number, how many items the bag can fit before these are sent automatically in the PC?

**Waiting for max number of items inside the bag (30 at the moment) as well as shop items: do you want that while buying the store remains without items or do you want infinite items for each store?**

Before there was a bool for infinite, I guess you were obliged because of the inventory structure; but now we can set the number and keep it hidden, so the buying menu shows only the items without the quantities. This will allow a consequent simplification of the inventory jsons too (in another PR) - maybe also a total removal since we can add the items to the NPC.

By the way this PR will allow to have inside the NPC json:
```
{
  "slug": "nurse",
  "gender": "female",
  "sprite_name": "nurse",
  "items": [ <----
    {
      "slug": "potion",
      "quantity": 10,
    }
  ],
  "monsters": [...],
  "combat_front": "nurse.png",
  "combat_back": "adventurer_back.png"
}
```

---

tested + black + isort (all selling / buying fixed + using the item eg potion or capture device during the battle + drop off & pick up items inside the pc)


JaskRendix:
Some images PC Items:

similar screen to monster pc (as per @Sanglorian  suggestion, so we can understand types vs items)
Screenshot from 2023-02-17 13-46-19
options when clicking on an item
Screenshot from 2023-02-15 19-44-08
when deposit the player can choose how many
Screenshot from 2023-02-15 15-56-46
PC screen updated
Screenshot from 2023-02-15 15-15-18
Standard message when items got disbanded
Screenshot from 2023-02-15 19-46-45

Sanglorian:
Cool! Thank you. 

How about allowing for 99 types of item per backpack as a default (any number of an item allowed, so e.g. you could have 50 capture devices and it would only count as one type of item)? 

There's a proposed style of gameplay that would only allow 10 items (not 10 types of item, just 10 items total), so it'd be good to be able to set different amounts if possible. 

In the Spyder campaign, shops should usually have infinite items. However, on the rare occasions where they have limited numbers of items, I'd like the number of items remaining to be displayed.

kerizane (dismissed):

JaskRendix:
Ok, I fixed some points, but I'm going to define the distinction between types and items in another PR.
Limit set temporarily at 99 types

to make it simple to understand:
5 capture devices = 1 type (capture device) and 5 items (quantity)

kerizane (approved):

PR #1564, opened 2023-01-15
Change bg and intro spyder

JaskRendix:
PR addresses discussions happened here #1381 and #1377

Originally the intention was to remove the hardcode, but it was missing an alternative.

Alternative that finally I developed. It's a new action called change_bg and it allows modders to change background. To test it, I recreated the intro spyder that @Sanglorian wanted to implement (which originated this video: https://youtu.be/GyQjCrsVJX4)

I guess it can be useful for @Qiangong2 too, like the announcement in television, as well as finding the right spot where they can ask the name in game.

I removed the hardcoding, so when a player click newgame > spyder, it appears this:
by the way, don't give much importance to the dialogues, I was filling the void.
if you want to watch the intro:
Screenshot from 2023-01-15 14-10-39
question no/yes
Screenshot from 2023-01-15 13-41-10
if the player says yes, no intro, he'll appear inside spyder_bedroom exactly like now.
if the player says no, it continues (exactly like in the video)
Screenshot from 2023-01-14 23-11-19
ended the intro, the player will be teleported inside the old spyder_paper_mart, it's a duplicate of spyder_paper_scoop, identical, but unused and not accessible in any way. Perfect location for setting up an event like this, the player will never know.
By the way, not worry about the face "down" of the player, it has been fixed, now it looks right.
Screenshot from 2023-01-15 13-42-06
assistant distribute the papers to fill in
Screenshot from 2023-01-15 13-42-27
first question of the papers
Screenshot from 2023-01-15 13-44-09
it appears the input
Screenshot from 2023-01-15 13-44-24
second questions (originally @Sanglorian planned for three question, but at the moment we can ask only 2)
Screenshot from 2023-01-15 13-44-50
choice among the monsters
Screenshot from 2023-01-15 13-37-36
trainers need to give back the papers (the position of the face has been fixed), it's animated by pathfinding, exactly like the distribution, but here it's not the shop assistant to move, but each trainer, last one our hero.
Screenshot from 2023-01-15 13-37-53
last one
Screenshot from 2023-01-15 13-38-24
here the player gets teleported inside the bedroom, exactly like if it didn't choose the intro.

tested + isort + black

PR #1629, opened 2023-02-18
template and change_sprite

JaskRendix:
@kerizane before I wasn't satisfied, but this version is more complete and acceptable than the previous.

PR:
- created **TemplateModel** (single JSON file with listed all the categories https://wiki.tuxemon.org/Category:Trainer) -> we needed this list inside the db, I kept the list in a single JSONs because it's more manageable;
- creation of **self.template** (exactly like self.monsters and self.items);
- **self.template** of the player is saved (**slug**, **sprite_name** and **combat_front**)
- edited the **JSONs** (since there is a different structure)
- removed **Gender** from the NPCs JSONs, we added back then, but we never used; the gender needs to be associated to the template / category, we can expand it in a second time (eg template nurse is female, template adventurer male, etc.);
- combat_back is replaced with slug, so there is the connection with TemplateModel, this will allow to import the specific data of the template (expansion gender, etc);
- **save_data** ("template") + **save_upgrader** with the basic data (as usual for the active players);
- create the action **set_player_template**, it'll allow to change sprite and combat_front (if you choose heroine as combat_front, it'll automatically update the image inside the player menu);
- create the condition **player_sprite**, that it'll check the player sprite

---

regarding the condition:
`<property name="cond1" value="not player_sprite heroine"/>`
or
`<property name="cond1" value="is player_sprite heroine"/>`

so the modder can trigger different "quests" regarding the type of sprite (eg a mission where you need to be a miner for entering inside a mine, etc)

---

so at the beginning of the game, in the gender choice, we can set up:
`<property name="act1" value="set_player_template adventurer,adventurer"/>`
or
`<property name="act1" value="set_player_template heroine,heroine"/>`

this will allow to change the sprite to swimmer when entering the water too, connected to #1606 

---

regarding the pronouns, at this point, we can add a game variable to save the first choice
I can do in #1564
then it'll follow another PR with the update of **locale.py**, where we are going to define which language will have the "gender choice": French, Italian, Spanish, Portuguese, etc and which not.

tested + isort + black


JaskRendix:
Some screenshots, we have 102 sprites
maniac sprite
Screenshot from 2023-02-19 07-08-09
heroine sprite
Screenshot from 2023-02-19 07-07-43
rockitten sprite
Screenshot from 2023-02-19 07-17-18

kerizane (approved):

PR #1633, opened 2023-02-20
updated spyder map

JaskRendix:
@Sanglorian for your approval.

updated spyder.world by adding the new locations, I added the missing route B, a simple vertical stripe of trees as well as trees inside route D (it was only grass before), nothing else.

Graphically:
https://user-images.githubusercontent.com/64643719/220191380-51776112-a040-43f9-a0af-7a6a73e695b6.png

If it's ok for you, if not I can remove that vertical stripe of trees. It wasn't really cool the hole between the two maps.

Unrelated question: do you have any plans for these maps (eg cities or dungeon) or are you open to proposals?

### DIFF
--- a/mods/tuxemon/maps/spyder.world
+++ b/mods/tuxemon/maps/spyder.world
@@ -2,315 +2,481 @@
     "maps": [
         {
             "fileName": "spyder_routea.tmx",
+            "height": 0,
+            "width": 0,
             "x": 960,
             "y": 0
         },
         {
             "fileName": "spyder_route5.tmx",
+            "height": 0,
+            "width": 0,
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks! Looks good to me.

No plans for the other parts, but since the Spyder plot is basically complete I would suggest anything you add should be for the post game. 

The academy at the top right of Timber Town is also unclaimed, except for the concept: https://wiki.tuxemon.org/Walled_Garden

JaskRendix:
Ok, no problem, thanks for the info.

kerizane (approved):

PR #1638, opened 2023-02-24
tuxeball +  tuxeball types

JaskRendix:
PR addresses:
- the change of name of capture device in **Tuxeball** (Xero already utilized in the PO file the term tuxeball, it's stated here too: https://wiki.tuxemon.org/index.php?title=Capturing#Names);
- the change of name includes a **save_upgrader** for all the items already present inside the bag;
- addition of a **new tuxeball types** (see resume below);
- fixed **status_modifier** (I'm not sure it was working before);
- removed **item_power** from capture.py because it was set to 1 (no influence) and there is no field "power" in Items();
- removed hardcoded **tuxeball.png** (old capture_device.png) from **animate_capture_monster** (**combat_animations**), now the tuxeball will change automatically based of what the player uses (the slug of each used item will be saved in a variable, so we are able to retrieve after);
- removed hardcoded **tuxeball.png** (old capture_device.png) from **animate_monster_release** (now the type of tuxeball that will be used for capturing a monster will be saved in the monster.capture_device field). Since we already save the data of capture, I thought it would be great to save the tuxeball type too, more personalization;
- related to the previous point (self.capture_device), I have added it to the Journal (the one you can access through the monster menu), see below screenshot;
- added game variable in combat.py, so we can save the instance_id of the monster on the ground, this will simplify the action to retrieve it (for effects or conditions battle related);

I'm going to add everything to Tuxepedia as soon as my temporary block expires.

from
`   (3 * target.hp - 2 * target.current_hp) * target.catch_rate * item_power * status_modifier / (3 * target.hp)`
to
`   (3 * target.hp - 2 * target.current_hp) * target.catch_rate * status_modifier * tuxeball_modifier / (3 * target.hp)`

using the standard tuxeball, the **tuxeball_modifier** will be 1 (no influence)

In details:
```
            if self.tuxeball == "fire":
                if target.type1 != "fire":
                    tuxeball_modifier = 0.2
                else:
                    tuxeball_modifier = 1.5
```
if the player uses a tuxeball **fire** for catching a fire monster, then it'll be a **bonus** (x1.5); while if the player uses the same ball for catching a metal monster, then it'll be a **malus** (x0.2);

resume:
```
tuxeball fire -> obviously for a bonus with fire (malus everything else)
tuxeball water -> same as above
tuxeball earth -> same as above
tuxeball metal -> same as above
tuxeball wood -> same as above

tuxeball male -> obviously for a bonus with male (malus everything else)
tuxeball female -> same as above
tuxeball neuter -> same as above

tuxeball candy -> no malus/bonus, but it increases the level of the wild monster (+1) upon capture.
tuxeball hardened -> no malus/bonus, but it doesn't break when it fails
tuxeball lavish -> bonus 1.5
```
added some lines to show the different Tuxeball during the capture:
Screenshot from 2023-02-24 15-40-09
inside the journal (the one you can access through monster menu), the tuxeball will change if you use a different one to catch the monster (automatically)
Screenshot from 2023-02-24 19-50-11


JaskRendix:
@Sanglorian are ok the values for you? Did you have some suggestion?

You wrote (Tuxepedia): 
_The only change I would make is limiting the status_modifier to negative conditions._ 
That's fixed (1.2 bonus as it was before), do you want to introduce a malus (eg 0.8) regarding positive status too?

By the way, I saw the other types listed (Flavoured, etc.) and I'm considering to implement these too sooner or later.

JaskRendix:
@Sanglorian are ok the values for you? Did you have some suggestion?

You wrote (Tuxepedia): 
_The only change I would make is limiting the status_modifier to negative conditions._ 
**That's fixed (1.2 bonus as it was before), do you want to introduce a malus (eg 0.8?) regarding positive status too?**

By the way, I saw the other types listed on Tuxepedia:
```
Flavoured Tuxeball: The tuxemon's positive taste changes to the one specified by the flask
comment: positive taste? if you change parameter, then we can do it, I like the name

Smooth Tuxeball: The tuxemon likes you better after capture.
comment: likes... ? if you change parameter, then we can do it, I like the name
```
Or if you have other ideas, etc.

@Qiangong2 are you interested in some particular Tuxeball or do you have some ideas like the ones above? Let me know.

I don't know, like the Omnichannel Tuxeball, Able Tuxeball, Cotton Tuxeball, etc.

Qiangong2:
Maybe something llke the following:

`Ancient Tuxeball: This Tuxeball was made a long time ago and is capable of catching any Tuxemon without fail` (basically, our version of a master ball)

`Crusher Tuxeball: This Tuxeball breaks through defenses and increases the catch rate the higher the opposing Tuxemon's defense is. It will fail to catch if a protection move is used` (The higher a Tuxemon's defense, the higher a chance it will be catchable, up to 40%. It would increase by 0.01 points for every 5 defense points)

`Xero Tuxeball: This odd Tuxeball increases the chance of succeeding if the opposing Tuxemon is a different type than yours, but has a decreased chance if they are the same type` (If your tuxemon is wood type for example, and you're trying to catch a water type, it would have a 1.4 times chance to catch, compared to a wood type which would have 0.3)

`Omni Tuxeball: This odd Tuxeball increases the chance of succeeding if the opposing Tuxemon is the same type as yours, but has a decreased chance if they are different types` (Same as above, except the stats are flipped)

I hope this gives some good ideas! I named the Omni and Xero Tuxeballs as I did because Omnichannel is all about everything and everyone falling in line and being the same, whereas Xero thrives on having individuality and differences.

JaskRendix:
Thanks! I appreciate a lot!

Sanglorian:
Love it, thanks!

No malus for positive conditions, thanks. Lavish tuxeball, I would say 1.5.

Don't worry about Flavourful or Smooth for now - they would only be needed if we implemented Tastes (Flavourful) or some kind of happiness or friendship score to track how much your tuxemon like you (Smooth).

More details on Tastes here: https://wiki.tuxemon.org/Stats#Tastes

JaskRendix:
Thanks @Qiangong2 for your proposals, all is implemented.
Thanks @Sanglorian for the info.

kerizane (dismissed):

kerizane (dismissed):

kerizane (approved):

PR #1641, opened 2023-02-26
fix typehints

JaskRendix:
continuing #1211 and separate from #1642
from
Found 452 errors in 90 files (checked 288 source files)
to
Found 336 errors in 69 files (checked 288 source files)

there are others, but this is the first package with the less tricky.

there are a bunch in the journal, etc, but the majority is always related to the translate component (41 in total), it's necessary for moving the widget, I have already asked on pygame_menu about.

---

regarding **tuxemon/states/monster/__init__.py**
before it was 6.5, but I tracked the value and it was always interpreted as 6.
Reason why I put 6 and we got rid of 2 typehints in one shot.


tuxemon/event/actions/open_shop.py
```diff
@@ -89,4 +88,4 @@ def sell_menu() -> None:
         elif menu == "sell":
             push_sell_menu()
         else:
-            assert_never(menu)
```
kerizane, line 92: Why did you remove this

tuxemon/formula.py
```diff
@@ -54,13 +55,15 @@ def simple_damage_multiplier(
...
         for target_type in target_types:
-            body = TYPES.get(target_type, TYPES["aether"])
-            if body.extra_damage is None:
-                continue
-            if attack_type == body.extra_damage:
-                m *= 2
-            elif attack_type == body.resist_damage:
-                m /= 2.0
+            if target_type:
+                assert target_type
```
kerizane: this assert shouldn't be needed

tuxemon/menu/quantity.py
```diff
@@ -160,9 +162,12 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
         count_len = 3
 
         cost = self.cost if self.quantity == 0 else self.quantity * self.cost
+        cost_tag: Union[str, int]
         if int(cost) == 0:
-            cost = T.translate("shop_buy_free")
+            cost_tag = T.translate("shop_buy_free")
+        else:
+            cost_tag = cost
```
kerizane: Can we just do 

```
        if int(cost) == 0:
            cost_tag = T.translate("shop_buy_free")
        else:
            cost_tag = str(cost)
```

tuxemon/monster.py
```diff
@@ -322,12 +322,16 @@ def load_from_db(self, slug: str) -> None:
...
 
         # Look up the monster's sprite image paths
-        self.front_battle_sprite = self.get_sprite_path(
-            results.sprites.battle1
-        )
-        self.back_battle_sprite = self.get_sprite_path(results.sprites.battle2)
-        self.menu_sprite_1 = self.get_sprite_path(results.sprites.menu1)
-        self.menu_sprite_2 = self.get_sprite_path(results.sprites.menu2)
+        if results.sprites:
+            assert results.sprites
```
kerizane: Don't need this assert

tuxemon/states/pc_locker/__init__.py
```diff
@@ -441,21 +444,21 @@ def find_monster_box(itm: Item, box: list) -> Optional[Item]:
...
-                        player.remove_item(itm)
-                    else:
-                        stored = player.find_item_in_storage(
-                            find_monster_box(itm, box).instance_id
-                        )
-                        stored.quantity += quantity
-                        itm.quantity = diff
+                retrieve = find_monster_box(itm, box)
+                if retrieve is not None:
+                    assert retrieve
```
kerizane: Not going to point this out any more. Please check all your asserts.

kerizane: The if check covers this, so the assert is pointless

tuxemon/states/phone/__init__.py
```diff
@@ -474,8 +473,9 @@ def add_menu_items(
             fix_width(menu._width, -0.20), fix_height(menu._height, -0.03)
         )
 
+        widgets
```
kerizane: ?

tuxemon/states/world/world_menus.py
```diff
@@ -103,6 +106,7 @@ def monster_menu_hook() -> None:
                 # get the newly selected item.  it will be set to previous
                 # position
                 original_monster = monster_menu.get_selected_item().game_object
+                assert original_monster
```
kerizane, line 108: This one is fine :)

kerizane (changes requested):
Yay typehints :)

tuxemon/states/world/world_menus.py
JaskRendix: ok

tuxemon/states/phone/__init__.py
JaskRendix: forgot to remove

tuxemon/states/pc_locker/__init__.py
JaskRendix: you're right, I went in goblin mode with asserts, I'm going to fix it

tuxemon/formula.py
JaskRendix: ok

tuxemon/menu/quantity.py
JaskRendix: done

tuxemon/event/actions/open_shop.py
JaskRendix: because of:
`tuxemon/event/actions/open_shop.py:92: error: Argument 1 to "assert_never" has incompatible type "str"; expected "NoReturn"`

JaskRendix: if you have another idea, then feel free

kerizane: Ok, let's raise an exception.

JaskRendix: ok

kerizane: Please provide a useful error msg

kerizane (changes requested):

tuxemon/event/actions/open_shop.py
JaskRendix: `"The parameter {self.menu} can be only 'both', 'buy' or 'sell'."`

kerizane (approved):

PR #1648, opened 2023-03-01
update start_battle (action) + new random_battle (action) + update trainer_battle (client)

JaskRendix:
PR addresses the files:

- **start_battle**.py (action): getting rid of **get_entity**() in favor of **get_npc**()

- **random_battle**.py (new action): it works like start_battle, but it doesn't need the npc_slug, since it chooses a random npc without monsters

`random_battle nr_txmns,min_level,max_level`

nr_txmns = number of txmns of the opponent (from 1 to 6)
min and max level = it'll be the range of each monster

eg nr_txmn 6 with min: 1 and max 99, it can generate:
first monster lv 11, second monster level 42 and third monster level 69

- **trainer_battle**.py (client): connected to **random_battle**, so it's possible to start the random battle also directly from client (without action); the previous structure was broken and source of typehints.

black +  isort + tested + typehints free

_Right now we need to create a JSON, then add the monsters inside the JSON, but now we can have everything immediately through an action (money and exp modifier are equal to the level of the monster). So the strongest, the more exp/money you get. It'll simplify a lot the battles as well as we can have a lot more trainer battles than now._

### DIFF
--- a/tuxemon/event/actions/random_battle.py
+++ b/tuxemon/event/actions/random_battle.py
@@ -1,0 +1,114 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import final
+
+from tuxemon import formula
+from tuxemon.combat import check_battle_legal
+from tuxemon.db import db
+from tuxemon.event.eventaction import EventAction
+from tuxemon.monster import Monster
+from tuxemon.npc import NPC
+from tuxemon.states.combat.combat import CombatState
+from tuxemon.states.world.worldstate import WorldState
+
+logger = logging.getLogger(__name__)
+
+
+@final
+@dataclass
+class RandomBattleAction(EventAction):
+    """
+    Start random battle with a random npc with a determined
+    number of monster in a certain range of levels.
+
+    Script usage:
+        .. code-block::
+
+            random_battle nr_txmns,min_level,max_level
+
+    Script parameters:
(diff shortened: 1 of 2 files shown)


tuxemon/cli/commands/trainer_battle.py
```diff
@@ -35,12 +36,7 @@ def invoke(self, ctx: InvokeContext, line: str) -> None:
...
-            try:
-                action = ctx.session.client.event_engine.execute_action
-                action("create_npc", (trainer, 7, 6))
-                action("start_battle", npc_slug=trainer)
-                action("remove_npc", (trainer,))
-            except:
-                traceback.print_exc()
+        else:
+            ctx.session.client.event_engine.execute_action(
+                "random_battle", [args[0], args[1], args[2]]
```
kerizane: Random battles are fine, but you should leave the original feature in.

kerizane (changes requested):

tuxemon/cli/commands/trainer_battle.py
```diff
@@ -35,12 +36,7 @@ def invoke(self, ctx: InvokeContext, line: str) -> None:
...
-            trainer = args[0]
-            try:
-                action = ctx.session.client.event_engine.execute_action
-                action("create_npc", (trainer, 7, 6))
-                action("start_battle", npc_slug=trainer)
-                action("remove_npc", (trainer,))
-            except:
-                traceback.print_exc()
+        else:
+            ctx.session.client.event_engine.execute_action(
```
kerizane: Maybe leave this command as is, and add a new one for random battle?

JaskRendix: I'm going to remove and leave the trainer as it was before, at the end they can access through action

kerizane (approved):

Issue #1649, opened 2023-03-01
Calculations crash when passing float values to number_or_variable


I2SAM7:
**Describe the bug**
This bug occurs when doing calculations between different types.

**To Reproduce**
Steps to reproduce the bug:
1. Launch the game with the debugger console enabled
2. > action set_variable money:10
3. > action variable_math money,+,10.5,money
4. ERROR:
Traceback (most recent call last):
  File "/home/iss/Bureau/Tuxemon/tuxemon/tools.py", line 247, in number_or_variable
    return float(player.game_variables[value])
KeyError: '10.5'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/iss/Bureau/Tuxemon/tuxemon/cli/commands/action.py", line 76, in invoke
    ctx.session.client.event_engine.execute_action(name, args)
  File "/home/iss/Bureau/Tuxemon/tuxemon/event/eventengine.py", line 287, in execute_action
    return action.execute()
  File "/home/iss/Bureau/Tuxemon/tuxemon/event/eventaction.py", line 134, in execute
    with self:
  File "/home/iss/Bureau/Tuxemon/tuxemon/event/eventaction.py", line 98, in __enter__
    self.start()
  File "/home/iss/Bureau/Tuxemon/tuxemon/event/actions/variable_math.py", line 55, in start
    operand2 = number_or_variable(self.session, self.var2)
  File "/home/iss/Bureau/Tuxemon/tuxemon/tools.py", line 249, in number_or_variable
    raise ValueError(f"invalid number or game variable {value}")
ValueError: invalid number or game variable 10.5
Cannot execute action. Check the input and try again.

**Expected behavior**
I expected the calculation to happen without errors. 
When doing 10+10.5, I expect to see 20.5 and not an error.

**Desktop (please complete the following information):**
 - OS: Debian GNU/Linux 11 (bullseye)
 - Python 3.9.2

PR #1650, opened 2023-03-01
Number_or_variable: fix crash when passing float values (#1649)

I2SAM7:
An error occurred when doing calculations with two different types.

Example:
> action set_variable money:10
> action variable_math money,+,10.5,money
> ERROR

The function number_or_variable receives a string, then checks whether this string is a number or a variable name. This way of doing this, crashes when the function receives a float value.
Indeed, this float value is seen as a variable name because it was not recognised as an integer.
Therefore, I upgraded the test so that it checks whether the value passed is an integer, a float or a variable name.

One test has been added, a new test class in tests/tuxemon/test_tools.py called "TestVariableMoney" with a method called "test_var". This method should try all kinds of calculations possible.

### DIFF
--- a/tests/tuxemon/test_tools.py
+++ b/tests/tuxemon/test_tools.py
@@ -2,8 +2,18 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 import unittest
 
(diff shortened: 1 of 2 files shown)


JaskRendix:
Thank you, now I realized you were talking about game variables!

kerizane (approved):

PR #1656, opened 2023-03-03
Format all files that needs to be formatted

I2SAM7:
I ran the command "tox -efmt" in order to format the files that needed to be formatted.


JaskRendix:
I ran the same command periodically and I never got these.
Which version of python do you use?

kerizane (approved):

I2SAM7:
I use Python 3.9.2, well I guess it does not matter anymore since it has been merged :) !!
I decided to run that command after realizing that some pull requests didn't pass the style check, but now it's all green so it's good !!

PR #1657, opened 2023-03-03
transparent bg tuxemon sprites

JaskRendix:
Few tuxemon sprites had white bg instead of transparent bg


kerizane (approved):

Sanglorian:
Good spotting @JaskRendix , and thanks for updating in the wiki too.

PR #1662, opened 2023-03-06
random moves (moveset)

JaskRendix:
@Sanglorian at the moment we need to list each technique inside the JSON eg:
```
    "moveset": [
        {
            "level_learned": 2,
            "technique": "ram"
        }
    ],
```
with this implementation
```
    "moveset": [
        {
            "level_learned": 2,
            "element": "water" --> or whatever element
        }
    ],
```
we can choose a random technique of the same monster type (eg Earth -> all earth moves).
The system takes in consideration the monster moves already present (saved), the monster moves in the moveset (JSON), so we can avoid duplicates.
This will help in having more techniques learned, it happens sometimes that there are holes (like no techs from lv 6 to lv 12)

Is it ok for you?

### DIFF
--- a/mods/tuxemon/db/item/tm_avalanche.json
+++ b/mods/tuxemon/db/item/tm_avalanche.json
@@ -2,7 +2,7 @@
   "conditions": [
     "is type earth,water",
     "is max_moves",
-    "is lack_tech avalanche"
+    "not has_tech avalanche"
   ],
   "effects": [
     "learn_tm avalanche"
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks -- it's a good option to have. Longer term, rather than a random technique of the monster's type, perhaps we could specify the type? So - random-water or random-earth?

JaskRendix:
@Sanglorian integrated
```
            "level_learned": 2,
            "element": "water"
```
just these two

Sanglorian:
Perfect, thanks!

JaskRendix:
technically it'll follow #1670

Issue #1665, opened 2023-03-07
The screen background before the game starts


DavidLiang1129:
I thought the current black background was a bit drab, so I made a new one. What do you think? Your comments and suggestions are welcome!![ewm_1678184118178.jpg](https://user-images.githubusercontent.com/118408976/223394352-c06939e7-adb8-4559-8699-92e7df191222.jpg)

DavidLiang1129:
@JaskRendix Thank you for all your hard work

PR #1666, opened 2023-03-07
start background (state)

JaskRendix:
as suggested in #1665 

At the moment isn't possible to change bg, but this PR will allow to change the bg more easily.
(when we have the right background - or multiple among which we can choose)
For this I used the standard empty screen.
Font size is 30.

Note:
- load will appear only if there is a savegame;
- options and load will open on the same background;
- removed ModChooserMenuState(), since it was a duplicate. The choice of the campaign happens in the debug.tmx;

Screenshot from 2023-03-07 13-25-37


kerizane (approved):

DavidLiang1129:
The new background is great! 😃

PR #1667, opened 2023-03-09
dontcare button input (rename_player)

JaskRendix:
completing #524 with the addition of the button "Don't Care" close to the "End" (input).
The button will be hidden if it's not specified.

button "Don't care" visible:
`    <property name="act081" value="rename_player random"/>`
button "Don't care" hidden:
`    <property name="act081" value="rename_player"/>`

Mostly because the input is used also for other operations. By the way I kept the original names in a random "list" for no gender variable (someone can create a campaign without distinction) as well as a list with male and female names.

random_names_male: Cromwell - Edwardson - Lothair - Andrea - Brady - Demarcus - Brady - Farant - Birch - Justin - Colbey - Efrem - Garnier - Florentus - Alvyn - Hadden - Julianus - Hugh

random_names_female: Robine - Magnolia - Svana - Simone - Eila - Krimhild - Charlot - Yara - Gracie - Anke - Theresia - Diana - Kathrin - Anne - Madita - Slainie - Nastasia - Nelly

random_names: Beep-beep - Toast - Chair - Lizard - Chris - Carter - Kyndi - Wednesday - Tapioca - Pom - Indiana - Kansas - Dont Care

I don't care much about the names listed, I generated those randomly, I'm open to change or add new ones, feel free to suggest here or to open an issue if the PR gets merged.

Screenshot from 2023-03-09 13-22-45

update: I put 99 in char_limit, because --strict flagged:
`tuxemon/menu/input.py:46: error: Incompatible default for argument "char_limit" (default has type "None", argument has type "int")`


freshreplicant:
Nice feature! The names are a bit odd/archaic sounding. 

I don't really object to any of them however, except for 'Cromwell'. That name has very [negative historical connotations in Ireland](https://en.wikipedia.org/wiki/Cromwellian_conquest_of_Ireland) and in the Irish diaspora around the world. It may sound a bit pedantic, but really, ask pretty much any Irish person if they know who Oliver Cromwell is and you will hear nothing good about him. He has a very different and more positive historical legacy in the UK, but it's probably a good idea to avoid controversy.

JaskRendix:
Yeah, I know the story.
No problem, replaced it with Jarrod.

kerizane (approved):

DavidLiang1129:
Does the same apply to other languages?

JaskRendix:
@DavidLiang1129 feel free to use other names in your language, at the end these are the random names, there is no need to translate those literally.

DavidLiang1129:
> @DavidLiang1129 feel free to use other names in your language, at the end these are the random names, there is no need to translate those literally.

OK,Thanks!

PR #1670, opened 2023-03-11
refactor tech

JaskRendix:
Refactor of techniques, upgrade of the savegame too.
Mostly because on the long term can be a bottleneck.

Each technique will have an **instance_id** (each one is unique) and the parameters saved will be **slug**, **power**, **potency** and **accuracy** (+ **counter** and **counter_success**)

These are the main numerical parameters, this will allow to "personalize" the technique for each monster (eg increase or reduce the power if something happens, etc. with less issues than before).

if someone wants to add other parameters, feel free to suggest

update:
the following parameter, at the moment with no use
```
        self._combat_counter = 0
        self._life_counter = 0

        def reset_combat_counter()
```
will be changed in:
```
        self.counter = 0 --> generic counter 
        self.counter_success = 0 --> if it succeeds

        def reset_combat_counter() -> deleted, it doesn't appear anywhere else
```
both will be saved. These can help to develop a mechanism to personalize the technique for each monster.
Screenshot from 2023-03-12 08-52-29


JaskRendix:
Solved conflicts

kerizane (approved):

PR #1672, opened 2023-03-13
fix battles + various

JaskRendix:
PR addresses:
- fix **combat.py**, the missing **.hex** was causing a crash when trying to save the game after a combat (wild encounter or trainer);
- resorted the **spyder_route2_billie**, so it locks the controls and unlocks, the reason is that it was possible move the player (now both turn faces);
- some fixes in **battle.py** and the action **set_battle** (it changes nothing), mostly because I wasn't satisfied about the structure;
- added to **evolve_monster** in **npc.py** the missing parameters **capture_device** (the specific tuxeball that captured the monster) and the new **taste_cold** and **taste_warm**;
- fixed **battle_last_monster_type** in **combat.py**, it was saving the wrong parameter;
- fixed **set_monster_status** and **set_monster_health**, since both parameters are Optional, this action can be used without the comma at the end;
- replaced **center** with **clinic**, because **mypy --strict** flagged me **center** in **db.py**;
- fixed player variable `var["day_of_year"] = dt.datetime.now().strftime("%j")` as suggested here: https://github.com/Tuxemon/Tuxemon/pull/1553#issuecomment-1399851365

related to the center in db.py issue:
`tuxemon/db.py:143: error: Incompatible types in assignment (expression has type "str", base class "str" defined the type as "Callable[[str, SupportsIndex, str], str]")`
it was in conflict because of the word "center", so another typehint is gone.

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -27,8 +27,9 @@ from typing import (
 import pygame
 from pygame.rect import Rect
 
-from tuxemon import audio, battle, graphics, state, tools
+from tuxemon import audio, graphics, state, tools
 from tuxemon.animation import Task
+from tuxemon.battle import Battle
 from tuxemon.combat import (
     check_status,
     check_status_connected,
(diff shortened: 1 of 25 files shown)


tuxemon/states/combat/combat.py
```diff
@@ -713,7 +715,7 @@ def add_monster_into_play(
             # save iid monster fighting
             self.players[0].game_variables[
                 "iid_fighting_monster"
-            ] = monster.instance_id
+            ] = monster.instance_id.hex
```
JaskRendix, line 718: this was the cause of the crash

kerizane (approved):

PR #1655, opened 2023-03-03
upgrade random_monster and add_monster

JaskRendix:
This PR will allow to set the monster for the NPCs directly on the tmx file, no more modifications in the JSONs.
It's more intuitive for modders than listing all the monsters inside the JSON.

eg: spyder_route2_billie
technically we can delete this part in the JSON
```
  "monsters": [
    {
      "name": "Dollfin",
      "level": 6,
      "slug": "dollfin",
      "money_mod": 10,
      "exp_req_mod": 27,
      "gender": "male"
    },
    {
      "name": "Eyenemy",
      "level": 6,
      "slug": "eyenemy",
      "money_mod": 10,
      "exp_req_mod": 27,
      "gender": "male"
    },
    {
      "name": "Cardiling",
      "level": 3,
      "slug": "cardiling",
      "money_mod": 10,
      "exp_req_mod": 27,
      "gender": "female"
    }
  ]
```
and updating the TMX
```
    <property name="act2" value="create_npc spyder_route2_billie,3,15"/>
    [...]
    <property name="act4" value="translated_dialog spyder_route2_billie1"/>
    <property name="act7" value="start_battle spyder_route2_billie"/>
    <property name="act71" value="translated_dialog spyder_route2_billie2"/>
    [...]
    <property name="cond1" value="not variable_set route2billie:yes"/>
```
into
```
    <property name="act2" value="create_npc spyder_route2_billie,3,15"/>
    [...]
    <property name="act4" value="translated_dialog spyder_route2_billie1"/>
    <property name="act5" value="add_monster dollfin,6,spyder_route2_billie,27,10"/>
    <property name="act6" value="add_monster eyenemy,6,spyder_route2_billie,27,10"/>
    <property name="act7" value="add_monster cardiling,3,spyder_route2_billie,27,10"/>
    <property name="act8" value="start_battle spyder_route2_billie"/>
    <property name="act71" value="translated_dialog spyder_route2_billie2"/>
    [...]
    <property name="cond1" value="not variable_set route2billie:yes"/>
```
eventually you can use also random_monster (it doesn't need the monster slug):
`    <property name="act5" value="random_monster 3,spyder_route2_billie,27,10"/>`

PR #1677, opened 2023-03-14
update professor_lab.tmx, remove unused NPCs Xero and fix "loser" bug

JaskRendix:
PR addresses:
- updates the add_monster in the professor_lab;
- updates the starting "money" in Xero, the player will start with 500;
- the player will get 120 coins, no matter if wins or loses against the professor;
- introduces a different response if the player wins or loses;
- resizes the map, removes pointless collisions and updates the movements;
- fix "loser" bug by getting rid of `var["battle_lost_faint"] = "true"` in combat.py + Taba Town tmx (the "loser" bug appears if the player loses against the professor);
- it removes Karrianna JSON (unused) and it cleans up the PO file with a couple of dialogues;
- it removes professor-rockitten, etc. JSONs (unused) and it cleans up the PO file;

---

@Qiangong2 a couple of weeks ago #1655 has been merged. It upgrades the add_monster action, it equals the add_monster to the list of monsters inside the JSON (since there are two additional - optional - parameters).

I noticed in the professor lab you assign the monster to the professor:
`    <property name="act42" value="add_monster cardiling,3,professor"/>`
I was wondering if you want to add the exp multiplier or the money multiplier or both
`    <property name="act42" value="add_monster cardiling,3,professor,27,10"/>`

Because at the moment the battle doesn't give exp (almost nothing) or money. 

I tried with 27 (exp) and 10 (money), random values, and this will allow the monster to grow from lv 5 to lv 7 as well as to gain 30 coins. Feel free to give me new values, if you're not interested, then no problem, I'm going to close the PR. Let me know.

Unrelated question:
at the moment the [tuxemart](https://wiki.tuxemon.org/Category:Item) in Taba Town sells only potion, revive and tuxeball, let me know if you want to add other items (eg types of tuxeball, feel free to give me the prices / costs - cost is how much the player will get for reselling it to the store.) I'll post below the actual situation, feel free to change prices or costs.
```
potion price 20 cost 5
revive price 100 cost 20
tuxeball price 50 cost 10
```


Qiangong2:
I think the xp multiplier is a good idea. It makes a lot more sense than having to set it in the json file. 

I think 100 coins would be better, that way you can purchase a couple potions or tuxeballs if you need to. 

As far as other items, I imagined that the new tuxeball types would be sort of "restricted substances" where you could only purchase them from a specific tuxemart, maybe in Cotton Town.

JaskRendix:
@Qiangong2 excellent, so the modifier is at 40 (money), so the player will get 120 (lv 3 * 40) in case of victory.

1. someone put this at the beginning of Xero (debug.tmx)
`<property name="act5" value="set_money player,10000"/>`
this means that the player will start with 10'000, do you agree with this amount?
If not feel free to tell me how much money do you want to give to the player.

2. I have inserted the add_monster also for liela and omnigrunt, so it was possible to remove the monster from the JSONs and having it directly in the map. More handy for your to check.
```
    <property name="act12" value="add_monster aardorn,2,omnigrunt,27,10"/>
    <property name="act20" value="start_battle omnigrunt"/>
```

Qiangong2:
Probably 500 is a better starting point than 10,000. 500 gives you a couple revives, and you won't be able to purchase anything anyways until after you encounter the Professor, so you'll have 620 by the time you can actually purchase items.

JaskRendix:
@Qiangong2 I was digging deeply in Xero and I found a bug, I have already addressed it. Shortly, in the case the player loses against the professor (remote possibility), then as soon as the player is outside the lab, he or she is unable to access any building (teleport doesn't work).

The cause was this `var["battle_lost_faint"] = "true"` inside the combat.py. I removed it as well as I adjust the automatic teleport (if someone loses) inside Taba. There wasn't anywhere else **battle_lost_faint**.

---

Updated the professor_lab.tmx too, got rid of the empty tiles, at the end the player doesn't risk to fall of the map, so I removed the collision zone and resized the map. I updated movement too, so it's like before, already tested and no issues.

---

I tagged you because I want to ask you if you desire to trigger a different dialogue with the professor if the player loses or wins.

It would change from:
```
  <object id="114" name="start first battle" type="event" x="240" y="48" width="16" height="16">
   <properties>
    <property name="act4" value="start_battle professor"/>
```
to
```
  <object id="114" name="start first battle" type="event" x="240" y="48" width="16" height="16">
   <properties>
    [...]
   </properties>
  </object>
  <object id="115" name="first battle won" type="event" x="240" y="64" width="16" height="16">
   <properties>
    [...]
   </properties>
  </object>
  <object id="116" name="first battle lost" type="event" x="240" y="80" width="16" height="16">
   <properties>
    [...]
   </properties>
  </object>
  <object id="117" name="shesarrived" type="event" x="240" y="192" width="16" height="16">
   <properties>
    [...]
```
if you agree, then tell me how do you want to change the dialogue.
As well as if you want to give something else, because by losing the player will not receive the 120 coins, I don't know like a couple of tuxeball, a revive, etc.
```
msgid "wewbadfight" ------------------------> if the player loses
msgstr "Wow, that was a good fight!\n"
"Let me heal your Tuxemon real quick...\n"
"And... There! All set!\n"
"With a little more practice and maybe a few more Tuxemon, you'll definitely be able to take on the Trial.\n"

msgid "wewnicefight" ------------------------> if the player wins (unchanged)
msgstr "Wow, that was a good fight!\n"
"Let me heal your Tuxemon real quick...\n"
"And... There! All set!\n"
"With a little more practice and maybe a few more Tuxemon, you'll definitely be able to take on the Trial.\n"
```
If you don't want it, no issue, I'm going to revert it.

---

Last question. Since we have introduced the add_monster, then we can delete the following unused JSONs files if you want to:
- professor-rockitten
- professor-cardiling
- professor-tweesher
- karrianna

I'm asking because I don't know if you're using these in your expansion of Xero.

Let me know

tested + black  + isort

x64x2 (dismissed):

Qiangong2:
I personally think that since it's just a test battle, giving 120 money to the player is fine if you lose.

As for the different NPCs, karrianna can be deleted, she was just used for testing a couple years ago. How would the different Tuxemon the professor could possibly use be implemented then? Through variables and conditions? Because the Tuxemon the professor uses changes depending on which one you choose.

JaskRendix:
>> I personally think that since it's just a test battle, giving 120 money to the player is fine if you lose.

No problem, do you want to modify what the professor is going to tell to the player if loses?

>> How would the different Tuxemon the professor could possibly use be implemented then? Through variables and conditions? Because the Tuxemon the professor uses changes depending on which one you choose.

Shortly, yes, through variables and conditions. Each action adds a different monster.

More specifically, those files have never been used. If you look at the original action:
`    <property name="act42" value="add_monster tweesher,3,professor"/>`
you add the monster to **professor**, not **professor-tweesher**

you can delete **professor-tweesher** (and the other two files), and nothing happens, because these are not used.

Screenshot from 2023-03-15 18-48-09
as you can see these files don't pop up in any maps.

JaskRendix:
@Qiangong2 

>> I personally think that since it's just a test battle, giving 120 money to the player is fine if you lose.

No problem, added the action to give 120 coins.
Do you want to modify the text if the player loses/wins?
Because at the moment both have the same text:
```
"Wow, that was a good fight!\n"
"Let me heal your Tuxemon real quick...\n"
"And... There! All set!\n"
"With a little more practice and maybe a few more Tuxemon, you'll definitely be able to take on the Trial.\n"
```

>> How would the different Tuxemon the professor could possibly use be implemented then? Through variables and conditions? Because the Tuxemon the professor uses changes depending on which one you choose.

Shortly, yes, through variables and conditions. Each action adds a different monster.

More specifically, the files have never been used. If you look at the original action:
`    <property name="act42" value="add_monster tweesher,3,professor"/>`
you add the monster to **professor**, not **professor-tweesher**

you can delete **professor-tweesher** (as well as the other two files), and nothing happens, because these are not used.

Screenshot from 2023-03-15 18-48-09
as you can see these files don't pop up in any maps.

Qiangong2:
I think the dialog fits for both honestly. Removing the duplicated professor jsons should be fine then.

kerizane (approved):

PR #1678, opened 2023-03-15
add monster (spyder maps) and remove monsters from JSONs

JaskRendix:
@Sanglorian this PR:
- it'll remove all the info regarding the monster from the JSONs;
- it'll add the info inside the files tmx (maps);
- it'll remove the piece of code (hardcoded) inside start_battle (formula.py too), because by adding monster to the specific battle, the modders can decide if the monster is going to be more powerful or not;

Reasons:
- unique place to check (just open the map with Tiled) and update everything without being forced to open multiple files;
- it'll make simpler to introduce the rematch (done by modders without hardcoding);
- it'll make simpler to add new trainer battles (since the JSON npc doesn't contain info regarding monster, then you can utilize again the same file without worrying about conflicts - just playing with conditions);

since everything is added, I removed also the rematch part, because now modders can set new match by using different actions.
Another will follow.

If you want to maintain the actual situation, then no problem, I'm going to close the PR.
Let me know.

```
  "monsters": [
    {
      "name": "Dollfin",
      [...]
    },
    {
      "name": "Eyenemy",
      [...]
    },
    {
      "name": "Cardiling",
      [...]
    }
  ]
```
into
```
    <property name="act2" value="create_npc spyder_route2_billie,3,15"/>
    [...]
    <property name="act4" value="translated_dialog spyder_route2_billie1"/>
    <property name="act5" value="add_monster dollfin,6,spyder_route2_billie,27,10"/>
    <property name="act6" value="add_monster eyenemy,6,spyder_route2_billie,27,10"/>
    <property name="act7" value="add_monster cardiling,3,spyder_route2_billie,27,10"/>
    <property name="act8" value="start_battle spyder_route2_billie"/>
    <property name="act71" value="translated_dialog spyder_route2_billie2"/>
    [...]
    <property name="cond1" value="not variable_set route2billie:yes"/>
```


Sanglorian:
Sounds good, thanks @JaskRendix

JaskRendix:
remember to change "ready for review" #1691 and #1692 as soon as it gets merged

JaskRendix:
remember to change "ready for review" #1691 and #1692 as soon as it gets merged
remember to add #1701 to the maps after this / if gets merged.

kerizane (approved):

PR #1684, opened 2023-03-16
add statuses (diehard + festering)

### LINKED ISSUE #731: Boulder move doesn't hit
Boulder seems to have a 0% hit rate no matter what.  I've tested it in multiple battles and it seems to be a dead move.  Rockkitten is the specific Tuxemon using the move boulder.

### LINKED ISSUE #1170: Techniques with range set to special crash game
**Describe the bug**
Monsters attempting to use a technique with a "special" range, such as boulder, eyebite, energy_field will crash the game.  The file formula.py does not have a handler for this setting and crashes.

**To Reproduce**
1. Pick Lambert
2. Use Boulder

**Screenshots**
unknown

**Additional context**
In order to fix this, we will need to determine what modifiers or settings are required for the special range and then add the required code to formula.py

### LINKED ISSUE #1673: HP bar will sometimes over-extend the available space
As you can see here:
image

During random battles, the HP bar will sometimes over-extend. It doesn't always happen, but it seems to happen pretty regularly after loading into a save. I am running Tuxemon on Windows on Python 3.10

JaskRendix:
PR addresses the addition of diehard. As planned by @Sanglorian (tuxepedia).
```
When a monster with the diehard condition is reduced to <1 HP
Set their HP to 1.
They lose the diehard condition.
```
diehard effect on the user, techniques with diehard: **shuriken** and **clamp on**

isort, black, tested, no typehints
technically it'll follow #1670 

fix #1673 added mechanism to remove the status if the current_hp >= hp as well as it fixes a bug related to recover (crash)
fix #731 issues with boulder solved
fix #1170 no crash anymore


JaskRendix:
@Sanglorian I have a question about the status/conditions.

Take for instance https://wiki.tuxemon.org/Sudden_Glow
```
User Condition 	Recovering
Target Condition 	Focused
```
It means that the condition focused appears only if the monster is affected the recovering condition?
Let me know

Sanglorian:
Hi @JaskRendix 

No mistake! Techniques can give the target positive/neutral conditions or the user negative conditions. Focused and enraged are good in some circumstances and bad in others, so you might want to make your opponent focused or enraged.

It's also okay to go straight to tired - it is less of a negative than having exhausted first.

JaskRendix:
ok, thanks

JaskRendix:
@Sanglorian added status_festering, when the status/condition is present, it wouldn't be possible to use any sort of potion.

Last question:
no technique at the moment has festering, which one do you want to give this negative effect?

JaskRendix:
@Sanglorian added status_festering, when the status/condition is present, it wouldn't be possible to use any sort of potion.

**1° question:**
no technique at the moment has festering, which one do you want to give this negative effect?

**2° question:** regarding charging and exhausted.
_For “Charging” and “Exhausted”, the point is that the monster only loses the condition when they start the technique with that condition. They shouldn’t gain and lose the condition as part of the same technique use. I’m not sure how to arrange the logic to make sure that this happens._
example
```
rockitten has three moves: A, B, C (C gives exhausted to itself)
rockitten gets exhausted.
rockitten uses C again: it becomes tired?
rockitten uses A or B: it becomes tired?

or
rockitten uses C again: loses exhausted?
rockitten uses A or B: becomes tired?
```

let me know, because this will be the same also for charging > charged up, nodding off > dozing.

Sanglorian:
Thanks @JaskRendix , I'll think up some techniques that could use festering. 

For no. 2, it's 

```
rockitten has three moves: A, B, C (C gives exhausted to itself)
rockitten gets exhausted.
rockitten uses C again: it becomes exhausted
rockitten uses A or B: it becomes tired
```

JaskRendix:
@kerizane this PR is ready and fixes a lot of issues with statuses.
We are ready to add almost all the statuses (excluded spyder_bite and confusion - working on it).
But I'm going to do it in the next PR (if not it gets too big)


reminder:
```
Hibernate (Nodding Off - None)
Sleep Bombo (Nodding Off - Nodding Off)
Feint (Focused - Tired)
Sleeping Powder (Focused - Nodding Off)
Starfall (Focused - Tired)
Energy Field (charging - exhausted)
Stampede (charging - None)
Surge (charging - None)
Petrify (None, dozing)
```

kerizane (approved):

Sanglorian:
Have just taken a look through the techniques. What do you think about adding the festering condition to:

Fester (replacing poison), Meltdown, Pseudopod and Slime?

Thanks again @JaskRendix , it's great to have these conditions available.

JaskRendix:
@Sanglorian agree, well done

PR #1691, opened 2023-03-19
fix maps (3) useless spaces

JaskRendix:
PR removes useless spaces in **spyder_scoop4.tmx**, **professor_lab.tmx** and **test_pathfinding.tmx**


kerizane:
@Sanglorian ?

Sanglorian:
I know the way I set it up is awkward, but I needed a way for it to count as a wild battle - so for example the monster is catchable and the player can run away without forfeiting.

JaskRendix:
@Sanglorian ok
@kerizane now the PR removes only useless spaces

kerizane (approved):

PR #1696, opened 2023-03-20
history (monster JSON)

JaskRendix:
it adds the history field, it's like the evolution, but it shows all the evolution path, this will help to retrieve all the stages from a single monster, without being forced to reload the JSON db again.

it fixes also some wrong stages (tuxepedia)

it fixes breeding too, before
a S2 and a S3 -> could have a baby S2 or S3
now
a S2 and a S3 -> S1

before: Rockat and Jemuar -> baby Rockat or Jemuar
after: Rockat and Jemuar -> baby Rockitten
it happens the same also with mixed tuxemon, it retrieves the stage1

S1 = Basic, while the others are Stage 2 and Stage 3

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -29,6 +29,12 @@
         }
     ],
     "evolutions": [],
+    "history": [
+        {
+            "mon_slug": "aardorn",
+            "evo_stage": "basic"
+        }
+    ],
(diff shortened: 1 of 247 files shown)


JaskRendix:
@Sanglorian is it ok? I don't know if I need to ask you or someone else.

Sanglorian:
Sounds great, thank you!

kerizane (approved):

PR #1699, opened 2023-03-21
flavoured tuxeball (items) + fix bug capture

JaskRendix:
Following #1638
PR adds 5 items (tuxeball for each taste, upon capture the taste warm will be the one of the tuxeball)
Source: [Capturing Tuxepedia](https://wiki.tuxemon.org/Capturing#Item_additions)

Tuxeball zesty will change the taste warm of the captured monster in zesty.

@Sanglorian there is some malus? Some indication about the taste warm or is it ok like this? Let me know.

Unrelated: if someone has some ideas for a new Tuxeball type, then feel free to share.


Sanglorian:
That looks good as is I think, no need for a malus. 

Thanks!

NeonFlames:
Maybe sweet and savory could be added.

kerizane (approved):

PR #1702, opened 2023-03-23
bonding (monsters)

JaskRendix:
PR addresses the refactoring of steps.

The counting of **steps** (player) is recorded inside a **game variable** called **steps**.
This PR will remove this game variable and it'll create a field called **steps** inside **npc.py** that will be saved (like player_name, etc.).

the PR includes:
- creation of the field **steps** in **npc.py** (saved);
- creation of the field **stored** in **monster.py** (saved);
- creation of the field **steps** in **monster.py** (saved);
- new **condition** called **stored**, this will help us to understand how much time (steps) the monster has been stored, it'll be used for breeding too since the change;
- save_upgrader fix for old savegames (after the implementation, all the monster in the party will have the same amount of steps of the player);
- various fixes (old "steps" into new "steps");

until now there was no clear distinction (or advantage) to have monsters inside the party (excluding for fighting), but it lacks of the bonding component.

isort, black, tested, no new typehints
tested with 6 monsters in the party and there are no variation in speed, etc.

---

I would like to implement the steps value inside the info of the monster, but not like kilometers or miles, it would be a bit lame (bad copy of the player menu) and no one would care about it.

@Sanglorian do you have some sort of ranking for bonding? Or this parameter would be more close to friendship/bonding? For sure being inside the party has its advantages as bonding. Or we can rename the step monster into something else and make it grow like by 1 each 100 steps, etc. Let me know.

By the way, for giving you all a measurement, if you go from the Tuxecenter of Cotton to the upper  west (where the statue block the path), it's like 100 steps (by following the path).

If someone has some ideas, then feel free to share.


JaskRendix:
@Sanglorian

JaskRendix:
@Sanglorian ?

Sanglorian:
Steps is a good way to measure it instead of km. So far I haven't given much thought to bonding/friendship. I'll post on the Discord and see if there's any ideas.

Sanglorian:
> Steps is a good way to measure it instead of km. So far I haven't given much thought to bonding/friendship. I'll post on the Discord and see if there's any ideas.

Hi @JaskRendix , I got some suggestions on the Discord -- I've put them [up on the wiki.](https://wiki.tuxemon.org/Bonding) But it's just brainstorming at this point -- I don't think there's any rush to implement them.

JaskRendix:
@Sanglorian that's good, I'm going to close because I was on the wrong path. I'll think about it.

PR #1706, opened 2023-03-25
fix typehints

JaskRendix:
progress #1211 (mypy --strict)

from
Found 380 errors in 63 files (checked 291 source files)
to
Found 253 errors in 53 files (checked 291 source files)

some files are obsolete, but at least the count is going down.

### DIFF
--- a/tuxemon/clock.py
+++ b/tuxemon/clock.py
@@ -3,7 +3,7 @@
 import collections
 import time
 from heapq import heapify, heappop, heappush, heappushpop
-from typing import Any, Deque, List, Optional, Union
+from typing import Any, Callable, Deque, List, Optional, Union
 
 __all__ = ("ScheduledItem", "Scheduler", "Clock")
 
@@ -37,9 +37,8 @@ class ScheduledItem:
 class Scheduler:
     """Class for scheduling functions."""
 
-    def __init__(self, time_function: float = time.perf_counter()) -> None:
+    def __init__(self, time_function: Callable[[], float] = time.perf_counter):
         """Initialise a Clock, with optional custom time function.
-
         :Parameters:
             `time_function` : function
                 Function to return the elapsed time of the application,
@@ -59,7 +58,7 @@ class Scheduler:
         probably want to be scheduled together.
         """
         last_ts = self._last_ts
-        ts = self._time
+        ts = self._time()
         if ts - last_ts > 0.2:
             last_ts = ts
         return last_ts
@@ -186,7 +185,7 @@ class Scheduler:
             was the first tick.
 
         """
-        delta_t = self.set_time(self._time)
+        delta_t = self.set_time(self._time())
         self._times.append(int(delta_t))
         self.call_scheduled_functions(delta_t)
         return delta_t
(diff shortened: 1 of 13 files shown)


tuxemon/clock.py
```diff
@@ -35,7 +37,7 @@ def __lt__(self, other):
 class Scheduler:
     """Class for scheduling functions."""
 
-    def __init__(self, time_function=time.perf_counter):
+    def __init__(self, time_function: float = time.perf_counter()) -> None:
```
kerizane: You sure about this?

tuxemon/mod_manager/__init__.py
```diff
@@ -81,14 +84,14 @@ def update_all(self):
...
-        name,
-        release=None,
-        repo=None,
-        dont_extract=False,
-        install_deps=True,
-        installed=None,
-    ):
+        author: str,
+        name: str,
+        release: Optional[float],
```
kerizane, line 89: Behaviour change for release.

tuxemon/mod_manager/__init__.py
```diff
@@ -132,10 +135,10 @@ def download_package(
         )
 
         filename = os.path.join(
-            paths.CACHE_DIR, f"downloaded_packages/{name}.{release}.zip"
+            paths.CACHE_DIR, f"downloaded_packages/{name}.{str(release)}.zip"
```
kerizane: Pointless, but we can leave it in.

tuxemon/networking.py
```diff
@@ -81,7 +82,7 @@ def __init__(self, game: LocalPygameClient, server_name=None):
             server_name=self.server_name,
         )
 
-    def update(self):
+    def update(self) -> Any:
```
kerizane: Optional[bool]

kerizane (changes requested):
See comment

tuxemon/mod_manager/__init__.py
JaskRendix: Fixed all

tuxemon/networking.py
JaskRendix: Thanks

tuxemon/clock.py
JaskRendix: I think this is better:
```
    def __init__(self) -> None:
        """Initialise a Clock, with optional custom time function.
        """
        super().__init__()
        self._time: float = time.perf_counter()
```
same results, no new typehints. 

The issue is that if we go back to `def __init__(self, time_function=time.perf_counter):` then it creates a bunch of typehints related to `self._time`. Eg. before it was called in other parts with `ts = self._time()` or here `return max(next_ts - self._time(), 0.0)`

tuxemon/clock.py
```diff
@@ -337,11 +333,11 @@ def get_idle_time(self) -> Optional[float]:
 
         try:
             next_ts = self._scheduled_items[0].next_ts
-            return max(next_ts - self._time(), 0.0)
+            return max(next_ts - self._time, 0.0)
```
kerizane: This was a function because it needs to return the current time. You can't just call it once and then reuse it. If you want to typehint it you need to use typing.Callable

kerizane (changes requested):
Comment

tuxemon/clock.py
JaskRendix: Thanks, you're right

kerizane (approved):

kerizane:
Great, thanks.

PR #1708, opened 2023-03-27
fix xero

JaskRendix:
some small fixes related to Xero:
- fixed three `npc_face taba_greeter,player`, since `npc_face` accept only **up**, **down**, **left** and **right**;
- removed to_talk in favor of behavior talk (it's the same function);
- updated the omnigrunt line based on your last PR and removed the monster from the JSON (if not the NPC will have 2 monsters);
from
`<property name="act12" value="add_monster aardorn,2,omnigrunt,27,10"/>`
to
`<property name="act12" value="add_monster aardorn,2,omnigrunt,80,10"/>`
- same as above, I did the same for the acolyte, remove JSON monster and added in map:
`<property name="act30" value="add_monster chloragon,3,acolyte,80,100"/>`

@Qiangong2 let me know if you're ok

tested, no issues

### DIFF
--- a/mods/tuxemon/db/npc/omnigrunt.json
+++ b/mods/tuxemon/db/npc/omnigrunt.json
@@ -6,14 +6,5 @@
       "combat_front": "knight",
       "slug": "knight"
     }
-  ],
-  "monsters": [
-    {
-      "level": 2,
-      "slug": "aardorn",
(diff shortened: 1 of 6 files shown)


Qiangong2:
No complaints from me

kerizane (approved):

PR #1711, opened 2023-03-29
fix typehints

JaskRendix:
progressing #1211 updated mypy, always --strict
from
```
Found 260 errors in 48 files (checked 291 source files)
-- strict Found 305 errors in 57 files (checked 291 source files)
```
to
```
Found 214 errors in 46 files (checked 291 source files)
--strict Found 248 errors in 53 files (checked 291 source files)
```

### DIFF
--- a/tuxemon/map.py
+++ b/tuxemon/map.py
@@ -464,5 +464,6 @@ class TuxemonMap:
             image_loader=scaled_image_loader,
             pixelalpha=True,
         )
+        assert self.renderer
         self.renderer.data.tmx.images = data.images
         self.renderer.redraw_tiles(self.renderer._buffer)
--- a/tuxemon/map_loader.py
+++ b/tuxemon/map_loader.py
@@ -372,6 +372,8 @@ class TMXMapLoader:
         # Conditions & actions are stored as Tiled properties.
         # We need to sort them by name, so that "act1" comes before "act2" and so on..
         for key in keys:
+            if not isinstance(key, str):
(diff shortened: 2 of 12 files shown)


tuxemon/map.py
```diff
@@ -464,5 +464,6 @@ def reload_tiles(self) -> None:
             image_loader=scaled_image_loader,
             pixelalpha=True,
         )
+        assert self.renderer
```
kerizane, line 467: Indentation looks wrong

tuxemon/map_loader.py
```diff
@@ -373,30 +373,37 @@ def load_event(
...
-                condition = MapCondition(
-                    cond_type, args, x, y, w, h, operator, key
-                )
-                conds.append(condition)
-            elif key.startswith("act"):
-                act_type, args = parse_action_string(value)
-                action = MapAction(act_type, args, key)
-                acts.append(action)
+            assert key
+            if isinstance(key, str):
```
kerizane: Let's do

```
if not is instance(key, str):
    continue
```

kerizane: To avoid increasing the indentation

kerizane (changes requested):
Comment

tuxemon/map_loader.py
JaskRendix: it makes sense

kerizane (approved):

PR #1715, opened 2023-04-01
fix typehints

JaskRendix:
progressing #1211 mypy (1.1.1) --strict, now below 200
mostly because I was able to fix the typehint with pygame_menu, it was triggered by the list;
the other are imports;
forgot itertools in tools.py from yesterday PR;

Found 247 errors in 53 files (checked 291 source files)
Found 184 errors in 45 files (checked 291 source files)

tested, black and isort

### DIFF
--- a/tuxemon/networking.py
+++ b/tuxemon/networking.py
@@ -494,8 +494,8 @@ class TuxemonClient:
                 del self.client.event_notifies[euuid]
 
             if event_data["type"] == "NOTIFY_CLIENT_INTERACTION":
-                world = self.game.get_state_by_name(world.WorldState)
-                world.handle_interaction(event_data, self.client.registry)
+                _world = self.game.get_state_by_name(world.WorldState)
+                _world.handle_interaction(event_data, self.client.registry)
                 del self.client.event_notifies[euuid]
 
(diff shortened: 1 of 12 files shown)


kerizane (approved):

tuxemon/networking.py
```diff
@@ -24,7 +24,7 @@
 from tuxemon.middleware import Controller, Multiplayer
 from tuxemon.npc import NPC
 from tuxemon.session import local_session
-from tuxemon.states import world
+from tuxemon.states.world import WorldState
```
kerizane: Do we need this change? I think this is causing the python3.10 tests error.

kerizane (changes requested):
Fix tests

JaskRendix:
No, I'm not going to fix the tests right now.
Found 186 errors in 45 files (checked 291 source files)

tuxemon/networking.py
JaskRendix: No, we don't need it at the moment.

kerizane (approved):

kerizane:
I was asking for you to fix the tests in the python3.10 build, which you have done. All good.

PR #1716, opened 2023-04-02
fix status icon combat + fix types (item condition)

JaskRendix:
PR fixes the bug related to the status icon during the combat.

Before:
- a monster could get status poison, then switch monster, and the new one had the poison icon (although not the status).
- a monster could get status poison, the another one, and it would have been resulted in an icon over another (bad effect).

The PR fixes this behavior.
It simply adds the icon to the already present (but unused) **self._status_icons** list.
I tracked the list and it resulted empty all the time.

Someone opened an issue about this, I tried to find it, but I wasn't successful.

---

added a small fix for type condition too

---

removed **itm_slug = self.players[0].game_variables["save_item_slug"]** from combat.py, because the variable pass already through technique (both item/technique). It wasn't necessary. It'll follow a more precise division in #1661 

### DIFF
--- a/tuxemon/item/conditions/type.py
+++ b/tuxemon/item/conditions/type.py
@@ -39,7 +39,7 @@ class TypeCondition(ItemCondition):
                 )
                 if p is not None
             )
-        if target.types[1] is not None:
+        if len(target.types) > 1:
             ret = ret or any(
                 target.types[1] == p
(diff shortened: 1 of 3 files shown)


JaskRendix:
added a small fix for type condition too

kerizane (approved):

PR #1717, opened 2023-04-02
lack_tech (condition - item)

### LINKED ISSUE #691: Items can teach the same technique multiple times
I've encountered this in two ways so far.

1) Leveling a Tuxemon from 1. All Tuxemon currently seem to have all three of their techniques known regardless of level requirements. They *also* seem to learn techniques on leveling based on level requirements. **Moved to its own issue.**

2) A TM will happily teach a Tuxemon a technique it already knows. This can be resolved in a couple ways;
 a. The most sustainable approach would be to have the learn effect fail if the Tuxemon already knows that technique. It would be really crappy to have the TM be consumed in this case though, so something will have to be done about that.
b. Create a "lacks_technique" condition and require that the monster doesn't have that technique. But this would require that every TM include this condition, which feels like a bad thing.

JaskRendix:
PR adds the condition lack_tech, so that it wouldn't be possible to teach a technique if the monster already knows it.
eg blossom, if the monster knows blossom, then it wouldn't be possible to teach it.

fix #691 

waiting for the merging of #1703


kerizane (approved):

kerizane:
Wait, couldn't we have used `not has_tech`?

JaskRendix:
You're right, I'm going to update it

JaskRendix:
I was wrong, has tech is an event condition

PR #1718, opened 2023-04-02
improvement kennel (pygame_menu)

JaskRendix:
PR
- improves the **graphics**;
- changes formula for **columns/rows**;
- removes both results **db.lookup**, the parameter was already in the monster;

Before it was painful to find the right monster, now with the level and the HP bar is more easy. It fixes dimensions and columns too. The font is "higher" because it's our font, other fonts display correctly, more "central" (eg the 100%). So it's not possible to "lower" the font.

Before
Screenshot from 2023-04-02 11-02-14

After: it looks smaller because everything is distributed on three columns and there is the progression bar (HP), that adds an element to the counter (of all widgets).
Screenshot from 2023-04-02 10-53-33

it applies the same mechanism to locker items too.

unrelated - remember
```
opening box with 15 monsters takes 0.13 seconds
opening box with 33 monsters takes 0.74 seconds
opening box with 42 monsters takes 1.38 seconds
opening box with 51 monsters takes 2.25 seconds
opening box with 69 monsters takes 5.11 seconds
opening box with 87 monsters takes 9.59 seconds
opening box with 303 monsters takes 327.41 seconds (5m 27s)

do we need to set a nr max of monsters per box?
```


kerizane:
@Sanglorian ?

JaskRendix:
I have already an idea about how to implement it, but we need to set the max number.

kerizane:
30 seems reasonable.

JaskRendix:
@kerizane ok, updated the labels too with 30, see below.
I'm going to open another PR to set the limit mechanism (if @Sanglorian proposes another number, then we can change it there).
If you can review #1703 it would be great.

Screenshot from 2023-04-04 08-27-20
Screenshot from 2023-04-04 08-27-04

JaskRendix:
@kerizane ok, updated the labels too with 30, see below.
Opened the PR #1722 to set the limit (if @Sanglorian proposes another number, then we can change it there).
If you can review #1703, then it would be great.

Screenshot from 2023-04-04 08-27-20
Screenshot from 2023-04-04 08-27-04

JaskRendix:
let's generate new ones

JaskRendix:
done @kerizane 
items and monsters inside the boxes are sorted by slug

```
            if len(self.monster_boxes[KENNEL]) >= MAX_BOX:
                i = sum(
                    1
                    for ele, mon in self.monster_boxes.items()
                    if ele.startswith(KENNEL) and len(mon) >= MAX_BOX
                )
                self.monster_boxes[f"{KENNEL}{i}"] = self.monster_boxes[KENNEL]
                self.monster_boxes[KENNEL] = []
```

Sanglorian:
Looks good! Thanks for this. Might be worth trying to fit three rows on the screen if it doesn't look too cramped.

JaskRendix:
Done, now it's 2 columns, not 3.

Sanglorian:
Oh sorry, I actually meant 3 rows x 3 columns, although this looks good too

JaskRendix:
Done

kerizane (approved):

PR #1722, opened 2023-04-04
set limit box (monsters)

JaskRendix:
PR continues #1718 and implements the limit (30 or whatever).  Possibly after #1703 

It consists in a **condition** called **space** for the **tuxeball** (category capture) items. The condition checks if the player has party and box (Kennel - standard one) full. If it's true, then it appear the following messages when trying to use the item.
Screenshot from 2023-04-04 10-27-45
this will avoid all the possible issues with capture.

on the other side, for modder actions (eg. **store_monster**, **add_monster**, etc.), there will be a new **condition** called **space** too. The mechanism is identical to the item condition. This will avoid problems, because they can choose to trigger the event if the condition space is true.

I don't like the name **space**, if someone has alternatives, then feel free to share.

tests are failing because it's supposed to work after #1718

### DIFF
--- a/tuxemon/map.py
+++ b/tuxemon/map.py
@@ -24,7 +24,7 @@ from pytmx import pytmx
 from pytmx.pytmx import TiledMap
 
 from tuxemon import prepare
-from tuxemon.compat import ReadOnlyRect
+from tuxemon.compat.rect import ReadOnlyRect
 from tuxemon.event import EventObject
 from tuxemon.graphics import scaled_image_loader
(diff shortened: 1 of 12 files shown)


kerizane:
I'd rather not block capturing if the kennel is full, it's annoying for the players. Can we have the game automatically create a new kennel / switch to a new kennel?

JaskRendix:
I don't know, I'm disappointed.
ok, let's generate a new one

PR #1724, opened 2023-04-05
pathfind_to_player (Spyder)

JaskRendix:
Following #1678 

PR adds pathfind_to_player to all the trainer battles in Spyder.

added the action **pathfind_to_player** to all the **start_battle** where there is the condition **player_at**, in this way the NPC comes closer to the player, talk and then the battle starts. Here the example: #1701 (screenshot)

@Sanglorian let me know if it's ok for you


Sanglorian:
Sounds great, thanks!

kerizane (approved):

PR #1725, opened 2023-04-06
add npc (searoute) + stop (paper town)

JaskRendix:
PR adds NPC swimmer in Sea Route and the stop event in Paper Town.

**Stop event**, it's exactly the same event to block the player towards spyder_route1, but towards **sea route**.
In this way we avoid players going around without monsters.

to avoid ruining the experience, added also a NPC swimmer here:
Screenshot from 2023-04-06 14-11-58
```
    <property name="act2" value="add_monster picc,26,spyder_searoutec_more,27,10"/>
    <property name="act3" value="add_monster incandesfin,25,spyder_searoutec_more,27,10"/>
    <property name="act4" value="add_monster nudimind,24,spyder_searoutec_more,27,10"/>
    <property name="act5" value="add_monster nudikill,23,spyder_searoutec_more,27,10"/>
    <property name="act6" value="add_monster bigfin,19,spyder_searoutec_more,27,10"/>
```

@Sanglorian tell me if it's ok.


Sanglorian:
Hi @JaskRendix , sounds good as a precaution. In practice I don't think players will get the chance to swim until they already have some monsters.

JaskRendix:
Sure, but right now someone can wander without limits because of this issue.

kerizane (approved):

PR #1726, opened 2023-04-06
upgrade party_size condition

JaskRendix:
PR upgrades the condition in the title:

before:
operator: One of "equals", "less_than" or "greater_than".

after:
operator: Numeric comparison operator. Accepted values are "less_than", "less_or_equal", "greater_than", "greater_or_equal", "equals" and "not_equals".

updated: added the NE (not_equals) too, it can be really useful.

### DIFF
--- a/tuxemon/event/conditions/party_size.py
+++ b/tuxemon/event/conditions/party_size.py
@@ -2,13 +2,11 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-import logging
+from operator import eq, ge, gt, le, lt, ne
 
 from tuxemon.event import MapCondition
 from tuxemon.event.eventcondition import EventCondition
 from tuxemon.session import Session
-
-logger = logging.getLogger(__name__)
 
 
 class PartySizeCondition(EventCondition):
@@ -21,7 +19,9 @@ class PartySizeCondition(EventCondition):
             is party_size <operator>,<value>
 
     Script parameters:
-        operator: One of "equals", "less_than" or "greater_than".
+        operator: Numeric comparison operator. Accepted values are "less_than",
+            "less_or_equal", "greater_than", "greater_or_equal", "equals"
+            and "not_equals".
         value: The value to compare the party size with.
 
     """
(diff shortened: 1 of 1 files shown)


tuxemon/event/conditions/party_size.py
```diff
@@ -21,7 +22,8 @@ class PartySizeCondition(EventCondition):
             is party_size <operator>,<value>
 
     Script parameters:
-        operator: One of "equals", "less_than" or "greater_than".
+        operator: Numeric comparison operator. Accepted values are "less_than",
```
kerizane, line 22: Can we change this to use normal operators? (> < = etc). We use normal operators in most other conditions, right?

Keep it backwards compatible to avoid breaking any in progress work.

kerizane (changes requested):

tuxemon/event/conditions/party_size.py
JaskRendix: Done, backwards compatible, but I noticed that there aren't many greater_than, etc.
It was affected also has_item condition (action)

JaskRendix:
We cannot use < or >, because it's going to crash. Tested right now.
```
  File "/home/nik/.local/lib/python3.10/site-packages/pytmx/pytmx.py", line 501, in __init__
    self.parse_xml(ElementTree.parse(self.filename).getroot())
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 1222, in parse
    tree.parse(source, parser)
  File "/usr/lib/python3.10/xml/etree/ElementTree.py", line 580, in parse
    self._root = parser._parse_whole(source)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 99, column 48
```
in some parts someone put `&lt;` instead of < and `&gt; `instead of >, but it's not really easy / accessible / understandable.

ok, done, **variable_is** is updated too.

kerizane (approved):

tuxemon/event/conditions/variable_is.py
```diff
@@ -49,17 +49,17 @@ def test(self, session: Session, condition: MapCondition) -> bool:
         operand2 = number_or_variable(session, condition.parameters[2])
 
         # Check if the condition is true
-        if operation == "==":
```
kerizane: I don't want to remove these, as it can be annoying for people who have work in progress stuff that is using it. There have been a couple of complaints about us breaking things before.

You can add support for the new comparisons, and change the existing comparisons to use the word versions, but let's leave these in for now.

kerizane (changes requested):
.

kerizane:
Do these (\&lt;) look right if you view the files with Tiled?

\&lt; looks better in tiled (probably), but less_than looks better when editing the XML files directly. Tricky...

JaskRendix:
Sure, I don't want to trigger complaints where I can't read them. It was only about party_size at the end.

JaskRendix:
> Do these (&lt;) look right if you view the files with Tiled?

By the way: 
Screenshot from 2023-04-21 08-53-20
so technically we can replace the **greater_than** with `>`, but on the tmx we need to do a mass replacement with `&gt; `, easy task, but the best approach is waiting for someone opening an issue requesting it (because for sure this change will trigger complaints, there are some people accustomed to greater_than)

kerizane (approved):

tuxemon/event/conditions/variable_is.py
JaskRendix: @kerizane by the way this:

> There have been a couple of complaints about us breaking things before.

Kinda annoying.
No one replying to any PR telling: "this could break my progress".
I don't know what to think honestly. Disappointed.

kerizane: Please don't be put off by this, you have done so much great work. I'm only aware of two "complaints", and that's over-stating it, really.

6 - 12 months ago, somebody (maybe @Qiangong2 ?) updated their repo, and had a few issues due to some changes you'd made. A month ago they posted a warning to other users saying that "dialog" crashes the game.

That's it.

PR #1732, opened 2023-04-08
remove collision (walk on water)

JaskRendix:
It follows #1606 (closed because it was the wrong approach) @kerizane 

added the property **water** in **map_loader**, so there will be the possibility to define the collision
```
  <object id="313" type="collision" x="192" y="272" width="16" height="128">
   <properties>
    <property name="water" value="1"/>
   </properties>
  </object>
```
exactly below custom properties, how it happens for the "jumper" with continue, etc.
then there is the new action **remove_collision** that will make disappear **water**
```
  <object id="343" name="Remove obstacle 1" type="event" x="304" y="448" width="16" height="16">
   <properties>
    <property name="act1" value="remove_collision water"/>
    <property name="cond1" value="is has_tech surf"/>
   </properties>
  </object>
  <object id="343" name="Remove obstacle 2" type="event" x="304" y="448" width="16" height="16">
   <properties>
    <property name="act1" value="remove_collision water"/>
    <property name="cond1" value="is has_item ???"/>
   </properties>
  </object>
```
as you can see we can create multiple actions, because there can be multiple possibility to "walk on water".
Everything can be done easily from Tiled.

@Sanglorian do you want to distinguish between the river and the sea? Maybe **surf** on sea and **rafting** on rivers?

What are all the possible way to swim?
- technique: surf (type water) ----> planned #1734 technique + item)
- monster shapes: all the aquatic species? ----> planned #1733 to check if there is species in the party
- item: ??

water properties to all the water collision zones (river included), done (Spyder)
transition_teleport to rivers, done (Spyder)


Sanglorian:
Thanks! This is a big, important improvement to the game. Surfing for all types of water is fine. 

How to overcome water will depend on the scenario, but for Spyder it should be via an item, the Surfboard.

JaskRendix:
@Sanglorian I guess you're referring to this? https://wiki.tuxemon.org/Hoverboard
I don't know what you planned originally, but we can change the name into Surfboard (wiki). Let me know.
Added the item to #1734 (as surfboard)

Sanglorian:
Thanks @JaskRendix , that sounds good. I've fixed it up on the wiki. Originally I intended the Surfboard to be an upgrade to the Hoverboard, but at the moment the Hoverboard isn't needed.

kerizane:
This is good, but I think it can be better!

I would like it if it was more dynamic, so the modder can easily define types of dynamic collision without changing the code. E.g. river, lava, acid, invisible wall.

So rather than water = 1, maybe we should do key = water?

Should probably implement "add_collision" as well, so we can turn the collisions on again, if we want to.

JaskRendix:
@kerizane replaced water with **key** as you suggested as well as the 1.

if the conditions change, then the collision reappear.
I tested it and if the player doesn't have the surfboard, then he cannot walk on water anymore.
We can play it with all the conditions (event) we have.
We can change the name in **collision** instead of **remove_collision**

Or do you mean to give the modder the function to insert the coordinates?

JaskRendix:
added **add_collision** action, it adds again the collision by recalling the same label.

moreover I added the possibility to block a single tile by indicating the coordinates (x,y)
`add_collision whatever,6,8`
(without acting in Tiled - tmx, not accessible tile 6,8, you can remove it with `remove_collision whatever`)
`add_collision whatever` (requires that there is the label inside the tmx.

it works, but we got 3 new typehints:
```
tuxemon/event/actions/add_collision.py:39: error: Unsupported target for indexed assignment ("Mapping[Tuple[int, int], Optional[RegionProperties]]")  [index]
tuxemon/event/actions/add_collision.py:45: error: Unsupported target for indexed assignment ("Mapping[Tuple[int, int], Optional[RegionProperties]]")  [index]
tuxemon/event/actions/remove_collision.py:39: error: "Mapping[Tuple[int, int], Optional[RegionProperties]]" has no attribute "__delitem__"; maybe "__getitem__"?  [attr-defined]
```
if you have any idea on how to get rid of those typehints, then feel free to share
regarding **RegionProperties**
```
RegionPropertiesOptional = TypedDict(
    "RegionPropertiesOptional",
    {
        "continue": Direction,
        "key": str,
    },
    total=False,
)
class RegionProperties(RegionPropertiesOptional):
    enter: Sequence[Direction]
    exit: Sequence[Direction]
```

tuxemon/states/world/worldstate.py
```diff
@@ -620,6 +620,25 @@ def get_all_entities(self) -> Sequence[NPC]:
...
+    ) -> Optional[Tuple[int, int]]:
+        """
+        Returns coords (tuple) of specific collision zones.
+
+        Returns:
+            The coordinates.
+
+        """
+        for coords, props in map.items():
+            if isinstance(props, Dict):
```
kerizane: This works, but strictly speaking should be dict, not Dict.

JaskRendix: True

kerizane (approved):

PR #1733, opened 2023-04-08
upgrade monster_property

JaskRendix:
after #1732 I was going to create a condition to get the shape (eg if there was an aquatic monster in the party), but I bumped into **monster_property**.

With my surprise **monster_property** didn't pop up anywhere, but I understood as soon as I looked inside. Slot conditions are tough, because at the end who cares which monster the player has in the first position, or in the second, how you can forecast... ?
I upgraded the condition by putting a loop for all the monsters in the party and new valid values. 

Now the condition is more powerful, but I realized that it can replace the existing **has_tech** (still not used in all the maps) and **has_monster** used 19 times (easily replaceable, **spyder_flower_shop** 7 and **spyder_flower_city** 12).

I'm referring to event conditions.

Let me know if you want to get rid of these two.

### DIFF
--- a/tuxemon/event/conditions/monster_property.py
+++ b/tuxemon/event/conditions/monster_property.py
@@ -9,23 +9,27 @@ from tuxemon.session import Session
 
 class MonsterPropertyCondition(EventCondition):
     """
-    Check to see if a monster property or condition is as asked.
+    Check to see if a monster in the party has one of
+    the following property.
 
     Script usage:
         .. code-block::
 
-            is monster_property <slot>,<property>,<value>
+            is monster_property <property>,<value>
 
     Script parameters:
-        slot: Position of the monster in the player monster list.
-        property: Property of the monster to check (e.g. "level"). Valid values
-            are:
-                - name
-                - level
-                - level_reached
-                - type
-                - category
-                - shape
+        property: Property of the monster to check (e.g. "level").
+        Valid values are:
+                - slug (slug, rockitten)
+                - level (level, 8)
+                - level_reached (level_reached, 8)
+                - stage (stage, standalone)
+                - shape (shape, aquatic)
+                - taste_cold (taste_cold, mild)
+                - taste_warm (taste_warm, peppy)
+                - type (type, fire)
+                - gender (gender, female)
+                - tech (tech, ram)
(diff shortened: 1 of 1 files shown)


tuxemon/event/conditions/monster_property.py
```diff
@@ -43,25 +48,46 @@ def test(self, session: Session, condition: MapCondition) -> bool:
...
+            elif prop == "stage":
+                if mon.stage == val:
+                    return True
+            # monster shape (eg. aquatic, brute, etc.)
+            elif prop == "shape":
+                if mon.shape == val:
+                    return True
+            # monster taste_cold (eg. mild, etc.)
+            elif prop == "taste_cold":
+                if mon.shape == val:
```
kerizane: This

tuxemon/event/conditions/monster_property.py
```diff
@@ -43,25 +48,46 @@ def test(self, session: Session, condition: MapCondition) -> bool:
...
+                    return True
+            # monster taste_warm (eg peppy, etc.)
+            elif prop == "taste_warm":
+                if mon.shape == val:
```
kerizane: This

tuxemon/event/conditions/monster_property.py
```diff
@@ -9,31 +9,36 @@
...
-            is monster_property <slot>,<property>,<value>
+            is monster_property <property>,<value>
 
     Script parameters:
-        slot: Position of the monster in the player monster list.
-        property: Property of the monster to check (e.g. "level"). Valid values
-            are:
-                - name
-                - level
-                - level_reached
```
kerizane, line 25: Why did you remove level_reached?

JaskRendix: I thought it was too similar to level, but I have added it again.

tuxemon/event/conditions/monster_property.py
JaskRendix: Thanks

tuxemon/event/conditions/monster_property.py
JaskRendix: Thanks

kerizane (approved):

PR #1734, opened 2023-04-08
surf (technique) + tm surf (item) +  surfboard (item)

JaskRendix:
PR related to #1732 
@Sanglorian I don't know if you want to call it surf or something else (technique for swimming), but I have prepared the file, as well as the item (like the TM Blossom).

technique surf, more important info
```
  "tech_id": 174,
  "accuracy": 0.9,
  "animation": "waterfall",
  "effects": [
    "damage"
  ],
  "is_area": true,
  "is_fast": false,
  "potency": 0.8,
  "power": 2.6,
  "range": "ranged",
  "recharge": 3,
  "slug": "surf",
  "sort": "damage",
```
let me know if it's ok or if you planned other moves similar to surf.


Sanglorian:
Thanks, it looks good but it makes me think that there are some techniques that shouldn't be learned randomly because players shouldn't have access to water too early in the game.

Or maybe it could require an item AND surf.

JaskRendix:
@Sanglorian just tell me which other technique we need to exclude from the learning randomly, we can implement it.

JaskRendix:
@Sanglorian just tell me which other technique we need to exclude from the learning randomly, we can implement it.

update: added surfboard item too.

regarding: 
> Thanks, it looks good but it makes me think that there are some techniques that shouldn't be learned randomly because players shouldn't have access to water too early in the game.

I opened #1735, feel free to list here all the techniques you don't want to teach randomly.

JaskRendix:
regarding: 
> Thanks, it looks good but it makes me think that there are some techniques that shouldn't be learned randomly because players shouldn't have access to water too early in the game.

I opened #1735, feel free to list here all the techniques you don't want to teach randomly.

@Sanglorian I added the event based on Tuxepedia since it was missing (https://wiki.tuxemon.org/Timber_Town)
```
msgid "spyder_timber_mom1"
msgstr "Hey bojo, your BOARD doesn't work on water. Unless you've got power."
msgid "spyder_timber_mom2"
msgstr "What I mean to say is, here's a Surfboard! Ride it onto the river and see what happens!"
```
and 
```
    <property name="act01" value="player_stop"/>
    <property name="act02" value="lock_controls"/>
    <property name="act03" value="create_npc spyder_mompapertown,26,23"/>
    <property name="act04" value="pathfind spyder_mompapertown,17,21"/>
    <property name="act05" value="unlock_controls"/>
    <property name="act06" value="npc_face player,down"/>
    <property name="act07" value="npc_face spyder_mompapertown,up"/>
    <property name="act08" value="translated_dialog spyder_timber_mom1"/>
    <property name="act09" value="add_item surfboard"/>
    <property name="act10" value="translated_dialog spyder_timber_mom2"/>
    <property name="act11" value="pathfind spyder_mompapertown,26,23"/>
    <property name="act12" value="remove_npc spyder_mompapertown"/>
    <property name="act13" value="set_variable timbermom:yes"/>
    <property name="cond1" value="not variable_set timbermom:yes"/>
    <property name="cond2" value="is variable_set scooparachne:yes"/>
```
exactly where the player is going to appear after exiting scoop4 (after defeating arachne)
Screenshot from 2023-04-09 12-04-14

let me know if it's ok

Sanglorian:
Looks good, thanks!

JaskRendix:
Ok, so this one is ready to go.

kerizane (approved):

PR #1738, opened 2023-04-09
trading (event action)

JaskRendix:
PR insert a new event action called **trading**. I noticed it was necessary after bumping into this event:

```
    <property name="act1" value="translated_dialog spyder_flower_cady_willtrade"/>
    <property name="act2" value="remove_monster grintot"/>
    <property name="act3" value="add_monster hampotamos,10"/>
```
This one crashed, it wasn't updated. This is the reason why I created the event action trading.
```
    <property name="act1" value="translated_dialog spyder_flower_cady_willtrade"/>
    <property name="act2" value="trading grintot,hampotamos"/>
```
Switch in, switch out. Same level: grintot lv 25 -> hampotamos lv 25

@sanglorian is it ok or do you want to decide the level of the traded monster (eg grintot lv 25 -> hampotamos lv 10)?
Let me know, so I can fix it.

### DIFF
--- a/tuxemon/event/actions/trading.py
+++ b/tuxemon/event/actions/trading.py
@@ -1,0 +1,83 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import uuid
+from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thanks! This is another important feature for finishing off the Spyder campaign. 

The traded monster having the same level as the monster that the player traded is perfect.

kerizane (approved):

PR #1735, opened 2023-04-09
randomly (technique json)

JaskRendix:
PR introduce the field **randomly** in the **technique** json. The field is a simple bool, set on True. Maybe there are some techniques people are not interested in being taught randomly.

Issue pointed out by @Sanglorian  here: https://github.com/Tuxemon/Tuxemon/pull/1734#issuecomment-1501042280

At the moment only the future **surf** technique doesn't need to be taught randomly, but if there are others, then feel free to tell me. 

Set **healing_power** to **0** in db.py, deleted from all the JSONs where it was set 0.

tested, isort, black, no new typehints

learn_mm updated too, I'm going to tackle #1662 as soon as it gets merged in another PR, probably #1734

PR #1746, opened 2023-04-11
healing (effect) 

JaskRendix:
PR adds the effect healing based on healing_power (following #1735) as suggested by @Sanglorian 

Techniques with healing_power:
```
sudden_glow 2
wallow 4
refresh 4
battery_discharge 4
hibernate 8
```


JaskRendix:
@Sanglorian I noticed that [Battery Discharge](https://wiki.tuxemon.org/Battery_Discharge) wasn't in the wiki, so I added it. 
I have a question because from the JSON seems like that Battery Discharge has both:
```
  "effects": [
    "damage",
    "healing"
  ],
```
Is it correct? If I look in the wiki and sort by healing power, then I notice the majority with healing power are Aether techniques.

If it's correct, can healing fail? I guess if the monster has already 100% HP, but specifically, must healing be connected to the damage? 

if battery fails (damage) because `tech.accuracy >= random_value` healing must fail too?

eg considering **battery** has an accuracy of 80%, what happens to the healing component in the remaining 20%? It works? It fails?

Let me know.

updated wiki: https://wiki.tuxemon.org/Category:Technique

Sanglorian:
Thanks for spotting that! Yes, I think accuracy should apply to healing as well - I think of accuracy as the odds that the power works at all.

JaskRendix:
@Sanglorian ready to go.

Unrelated: I'm not satisfied with [Battery Discharge](https://wiki.tuxemon.org/Battery_Discharge), the mix "damage" and "healing" isn't the best. Eventually we can consider to create a third effect with both elements (damage and healing) together, but with another name because mixing both in a single one isn't the best choice, it makes me uncomfortable. Think about it, and let me know (and the new name too).
Like implosion (damage, healing), explosion (damage, and fainted monster), etc.

Sanglorian:
Thanks @JaskRendix -- 

Re Battery Discharge, what is it that makes you uncomfortable about mixing healing and damage? And just checking -- Battery Discharge should heal the user and damage the opponent. I might have been quite confusing about that, sorry if so!

JaskRendix:
No worries @Sanglorian at the end I solved.

JaskRendix:
@Sanglorian I was checking the wiki and among the new techniques I saw:
`Contrary Healing: This technique heals the enemy, not the user. (How much it heals is determined by healing_power)`
I implemented it inside healing.
```
"healing user" traditional healing
"healing target" contrary healing
```

Sanglorian:
Thanks @JaskRendix ! I've created a technique to make use of the new feature: https://wiki.tuxemon.org/Life_Surge

JaskRendix:
@Sanglorian I see, do you want to insert it now? If yes, we need potency (I put 100% as default) and animation (I put null). It would be the 175th. Let me know. You can see the JSON among the files here (PR).

By the way, also these technique lack animation: glower, hibernate, levitate, overfeed, overgrowth and spray

Sanglorian:
> @Sanglorian I see, do you want to insert it now? If yes, we need potency (I put 100% as default) and animation (I put null). It would be the 175th. Let me know. You can see the JSON among the files here (PR).
> 
> By the way, also these technique lack animation: glower, hibernate, levitate, overfeed, overgrowth and spray

Thanks! Potency 100% is good, and I've added animations for Life Surge, Glower and Overgrowth. I think the others are okay without for now.

JaskRendix:
@sanglorian Perfect, everything is updated. Ready to go.

kerizane (approved):

Issue #666, opened 2020-02-15
Map objects that can be added and deleted, as NPCs are


Sanglorian:
There's a few map objects that my story needs, that I need to be able to delete once certain variables are met. 

For example, there's a barrier in front of a jetty that needs to be removed once the jetty is available for use. 

This would also allow us to set up bushes/rocks that block paths, and that can be destroyed by a player with the right items. 

I guess a workaround is to just create all these items as static NPCs. 

Related to #193

bitcraft:
The problem that we were having is how do you reference dynamically created objects?  Right now, we only have slugs to reference objects in scripts.  What I can think of is something like:

This would allow totally free-form object (and npc) creation:
```
create_object npc_red my_unique_id
delete_object my_unique_id
```

What about making an object without adding the unique id?  What if we don't care? 
```
create_object npc_red
set_variable last_created_object_id my_unique_id
delete_object my_unique_id
```

We could also make barriers appear on conditionals
```
# event 1
act: enable my_object
con: is variable_set show_my_barrier:true
# event 2
act: disable my_object
con: is variable_set show_my_barrier:false
```

kerizane:
@JaskRendix can this be closed?

JaskRendix:
@kerizane Yes, it can be closed

PR #1763, opened 2023-04-17
painting events (Spyder)

JaskRendix:
PR adds painting events.

money_is part of #1726 

Painting events between Cotton and Candy are done. The player can buy: starry or monsters or trepidation or bundle (all 3).

the player can buy the bundle (all three) or each one by clicking on the painting. When bought it'll replace the text with.
Screenshot from 2023-04-18 07-58-48
added this here
Screenshot from 2023-04-18 07-57-39
if the player buys all three paintings
Screenshot from 2023-04-18 08-03-59

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -5856,6 +5856,24 @@ msgstr "Welcome back. Shall I chuck your tuxemon in the healing unit?"
 msgid "spyder_cottonart_barmaid"
 msgstr "Do you like my set? \n They capture my early days as a professional tuxemon trainer, before I retired on my winnings. \n  I call them: 'Monsters' Eyes Meet', 'Starry, Starry, Starry Night' and 'Trepidation'. They're for sale! \n Get them while they're hot! Just 1,000 each."
 
+msgid "spyder_cottonart_barmaid_bundle"
+msgstr "Do you want to buy all three paintings for ${{currency}}3,000?"
+
+msgid "spyder_cottonart_barmaid_single"
+msgstr "Do you want to buy this painting for ${{currency}}1,000?"
+
+msgid "spyder_cottonart_barmaid_painting_sold"
+msgstr "The painting has been bought by ${{name}}! This is a copy!"
+
+msgid "spyder_cottonart_barmaid_painting_sold_billie"
+msgstr "The painting has been bought and donated back to the gallery by Billie!"
+
+msgid "spyder_cottonart_barmaid_bundle_sold"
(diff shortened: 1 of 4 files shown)


JaskRendix:
@Sanglorian tell me if it's ok

proposals:
- do you like the idea of adding a meeting (no fight) with Billie while she is going out of the Art Shop? (the dialogue could be around the painting she bought / new capture / something else).
- do you agree that if the player buys a painting he/she doesn't have to pay anymore the fee (50)?

Sanglorian:
> @Sanglorian tell me if it's ok
> 
> proposals:
> 
>     * do you like the idea of adding a meeting (no fight) with Billie while she is going out of the Art Shop? (the dialogue could be around the painting she bought / new capture / something else).
> 
>     * do you agree that if the player buys a painting he/she doesn't have to pay anymore the fee (50)?

Yeah, those both sound good!

Longer term it could be good to have the paintings disappear when bought (although I like the idea of Billie having one, maybe it could have been bought by her but donated back to the gallery? Then you could annoy the player by having an NPC talk about how generous Billie is). 

Obviously removing the paintings from display would require #666 to be implemented, so I think your solution of saying it's a copy is a great one in the meantime.

JaskRendix:
> Obviously removing the paintings from display would require https://github.com/Tuxemon/Tuxemon/issues/666 to be implemented, so I think your solution of saying it's a copy is a great one in the meantime.

Yeah, I'm aware of this, and I was thinking about too, but the solution doesn't go far than your NPC workaround.
Probably with the only difference that there will be an action called **create_item** (like create_npc)

> I like the idea of Billie having one, maybe it could have been bought by her but donated back to the gallery

That's good

> Yeah, those both sound good!

Since you're Billie's voice, then I need some dialogues!

JaskRendix:
@Sanglorian 

> Obviously removing the paintings from display would require https://github.com/Tuxemon/Tuxemon/issues/666 to be implemented, so I think your solution of saying it's a copy is a great one in the meantime.

Yeah, I'm aware of this, and I was thinking about too, but the solution doesn't go far than your NPC workaround.
Probably with the only difference that there will be an action called **create_item** (like create_npc)

> I like the idea of Billie having one, maybe it could have been bought by her but donated back to the gallery

That's good.
- added the donated back to the gallery
- added the exclusion to pay if buy a painting.

> Yeah, those both sound good!

Since you're Billie's voice, then I need some dialogues!

Sanglorian:
> Since you're Billie's voice, then I need some dialogues!

Oh, yes! How about:

"I see you're trying to broaden your horizons. That's cute. Just try not to touch anything."

JaskRendix:
That's on point! I like it!

JaskRendix:
Everything is done. Ready to go.

kerizane (approved):

PR #1764, opened 2023-04-18
log_action, combat and techs

JaskRendix:
PR:
- introduces two techniques (**scope** and **panjandrum**) - testing purpose;
- renames **self._round** in **self._turn** (combat.py);
- renames **check_status** in **has_status** (combat.py) -> reason verbosity;
- renames **check_status_connected** in **has_status_bond** (combat.py) -> reason verbosity;
- adds **scope** in combat.py (not the state) to avoid a T.format breaking the "code" -> style reason;
- adds **self._log_action** in combat.py (see below);
- adds **local_damage** technique effect (in addition to the default **damage**);

regarding **_log_action**: `self._log_action: List[Tuple[int, EnqueuedAction]] = []`
the **int** is nothing more than the **turn**, the list will be appended during the fight, then emptied at the end.

I noticed @Sanglorian created (is going to create) techniques that require to know if the player used 1 or more times (or in sequence or exactly before) certain techniques.
At the moment this isn't possible, but with this component we can do something about.

@Sanglorian I decided for those techniques for testing purpose, we can get rid of eventually, no problem at all. If you think we can keep those, then we can fill the missing details (accuracy, power, animation, etc.).


Sanglorian:
> PR:
> 
>     * introduces three techniques (**scope** and **panjandrum**) - testing purpose;
> 
>     * renames **self._round** in **self._turn** (combat.py);
> 
>     * renames **check_status** in **has_status** (combat.py) -> reason verbosity;
> 
>     * renames **check_status_connected** in **has_status_bond** (combat.py) -> reason verbosity;
> 
>     * adds **scope** in combat.py (not the state) to avoid a T.format breaking the "code" -> style reason;
> 
>     * adds **self._log_action** in combat.py (see below);
> 
>     * adds **local_damage** technique effect (in addition to the default **damage**);
> 
> 
> regarding **_log_action**: `self._log_action: List[Tuple[int, EnqueuedAction]] = []` the **int** is nothing more than the **turn**, the list will be appended during the fight, then emptied at the end.
> 
> I noticed @Sanglorian created (is going to create) techniques that requires to know if the player used 1 or more times (or in sequence or exactly before) certain techniques. At the moment this isn't possible, but with this component we can do something about.
> 
> @Sanglorian I decided for those techniques for testing purpose, we can get rid of eventually, no problem at all. If you think we can keep those, then we can fill the missing details (accuracy, power, animation, etc.).

That's great, thanks! I've added the missing detail on their wiki pages: https://wiki.tuxemon.org/Scope and https://wiki.tuxemon.org/Panjandrum

JaskRendix:
Thank you so much, updated.

kerizane (approved):

PR #1749, opened 2023-04-12
plague (Spyder Bite)

JaskRendix:
PR introduce the mechanics for plague (Spyder Bite).
- creates a strenum inside db.py (PlagueType: sickless, inoculated and infected);
- creates a field plague in npc.py (PlagueType.sickless) -> as default;
- creates a field plague in monster.py (PlagueType.sickless) -> as default -> the field will be saved;
- creates new condition "party_infected" to check how many infected monsters inside the party;
- creates new action "npc_plague" to set the NPC as infected;
- creates new action "quarantine" to move all the infected monsters inside the box (in) or outside the box (out) - (event Spyder campaign - Candy Town);
- creates new action "set_monster_plague" to heal all the infected monster (or infect all, or inoculate all);

tested, isort, black, no new typehints

@Sanglorian  

- Do not display the icon indicating that a monster has Spyder Bite -> done
- It is not healed by a “full heal”. -> done
- A monster can have other conditions as well as Spyder Bite. -> done
- Even being eliminated does not cause a monster to lose Spyder Bite. -> done
- Some enemy trainers have the property “infected”, which means that their monsters have Spyder Bite. -> done
- Some enemy trainers have the property “inoculated”, which means that their monsters cannot be afflicted by Spyder Bite. -> done
- There will be a special location in the game that cures monsters that have Spyder Bite -> where?
- combat mechanism -> done

Questions:
- if a monster evolves? If rockitten is infected, rockat is infected?
- all the trainers in the wayfarer inn are infected? other places?

PR #1766, opened 2023-04-18
Billie (Spyder) meeting route4 and route6

JaskRendix:
Split from #1749 
PR introduces two new encounters with Billie (route4 and route6) as planned by @Sanglorian


JaskRendix:
@Sanglorian the first monster, the starter, is it evolved in route6?

Sanglorian:
Good question, yes I think it should be evolved once by Route 6 (and for Hospital), and evolved twice for Omnichannel HQ.

JaskRendix:
> Good question, yes I think it should be evolved once by Route 6 (and for Hospital), and evolved twice for Omnichannel HQ.

Done, the Omnichannel we can fix it in the other PR (Cotton).

kerizane (approved):

PR #1768, opened 2023-04-19
Chsjhe

LeoKids:


kerizane:
I appreciate your effort, but we don't want to support Python2. Are you sure you can't install Python3?

kerizane:
Also, I really doubt this works with python2, as we are using Python3 features in the codebase.

PR #1772, opened 2023-04-22
template (interactive object)

JaskRendix:
PR adds **interactive_obj** to the **template.json**. it follows an edit inside **npc.py** to avoid looking for front, back, left, right.
That's because an interactive obj doesn't move. It creates the folder **sprites_obj** to separate obj sprites from npc sprites.

files changed 100, but:
- 57 are the sprites of log, boulder and statue (deleted - 19*3);
- 15 are the sprites of objects (eg tuxeball_red, etc);
- 21 are the JSONs edited or created in the db/npc folder;
- the actual code is in 2 files: **npc.py** and **db.py**

There are already some interactive objects inside the game (eg rock, log, boulder, statue, etc.), and the enormous bottleneck about it, it's that we need 20 identical PNGs for an object that doesn't move.

It changes something? For the player nothing, for the modders, less copy paste 20 times for an object that doesn't move.

moreover I added some interactive object (5 different type of tuxeball)
eg this ball will give the player 5 potions (just a test) and the it'll disappear.
Screenshot from 2023-04-22 11-05-02

```
event1
    <property name="act1" value="create_npc tuxeball,11,8"/>
    <property name="cond1" value="not npc_exists tuxeball"/>
    <property name="cond2" value="not variable_set tuxeballcotton1:yes"/>
event2
    <property name="act2" value="translated_dialog xxxxx"/>
    <property name="act3" value="add_item potion,5"/> <---- whatever you want, monster, etc.
    <property name="act4" value="set_variable tuxeballcotton1:yes"/> 
    <property name="act5" value="remove_npc tuxeball"/>
    <property name="cond1" value="is player_facing_tile"/>
    <property name="cond2" value="is button_pressed K_RETURN"/> <--- without this it doesn't disappear immediately, you need to exit and reenter
    <property name="cond3" value="not variable_set tuxeballcotton1:yes"/>
```
Screenshot from 2023-04-22 15-48-21

tested, black, isort, no new typehints

PR #1774, opened 2023-04-22
template (interactive object): choice

JaskRendix:
connected to #1772 

same method, it makes disappear the tuxeball once chosen. 

Screenshot from 2023-04-22 19-03-28
Screenshot from 2023-04-22 19-01-54

work in progress, black, isort, no new typehints


JaskRendix:
@Qiangong2 with the last PR we can make disappear the object (at the moment only a few), I have already done some tests and as you can see from the screenshots above, both tuxeballs are going to disappear if chosen (one player, one professor).

Are you interested? If yes, do you have a specific set of colors?
Alternatively you can put three identical tuxeballs without distinction (all red, all blue, etc.)

let me know!

(right now rockitten red, cardiling yellow and tweesher green, but we can change those without problems)
colors available: red, yellow, green, violet and blue (blue is the default one).

Qiangong2:
That looks great! Thanks for implementing it!

As for the tuxeball colors, I think Rockitten being gray, Cardiling being red, and Tweesher being yellow is the best combination. Gray, red, and yellow fit the Tuxemon better in my opinion.

JaskRendix:
That's great, fixed color order.
Fixed also the environment (background battle), before it was giving **grass** background during the fight vs Professor, now it gives the **interior** background.
You can set different backgrounds (options here: https://wiki.tuxemon.org/Battle_Terrain)

kerizane (approved):

PR #1775, opened 2023-04-22
core (elements)

JaskRendix:
PR adds some new elements to the COREs (old stuff, recolored).

new colors for vending machines
Screenshot from 2023-04-22 20-35-35
single piece for enlarging table, sofa back and sofa front (until now it was fixed on 2 tiles, now it can be expanded)
Screenshot from 2023-04-22 20-35-04
it can help to use this furniture instead of the old one (with the old one is possible a table 3 tiles)


kerizane (approved):

Sanglorian:
Good ideas!

PR #1776, opened 2023-04-22
wild encounter (event action)

JaskRendix:
PR introduces wild_encounter (event action)

indirect request by @Sanglorian  here: https://github.com/Tuxemon/Tuxemon/pull/1691#issuecomment-1498424793

right now if a modder wants to trigger a fight with a wild monster (without trainer), it needs to create a JSON file inside db/encounter and setting `"encounter_rate": 100`, all this for 1 single monster, sometimes in 1 single event.

Too many files, with this PR it can be done with:
`<property name="act1" value="wild_encounter rockitten,5"/>`
complete with optionals
`<property name="act1" value="wild_encounter monster_slug,level,exp_mod,money_mod,environment"/>`
default environment is grass, but it can be replaced with any of these: https://wiki.tuxemon.org/Battle_Terrain (created this page the other day, eg Night Desert -> night_desert, Bridge -> bridge, etc.)

tested,black,isort, no new typehints

### DIFF
--- a/tuxemon/event/actions/wild_encounter.py
+++ b/tuxemon/event/actions/wild_encounter.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Union, final
+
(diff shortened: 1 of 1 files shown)


Sanglorian:
That seems so much easier, thank you!

JaskRendix:
Ok, ready to go, it'll follow another PR to fix the three events with wild_encounter.

kerizane (approved):

PR #1745, opened 2023-04-11
Cotton Town + Omnichannel (event Spyder)

JaskRendix:
PR fixes some events in the Spyder campaign (Cotton Town)

This one: https://wiki.tuxemon.org/Cotton_Town#Event_5
Screenshot from 2023-04-11 16-39-30
in this way it gives correctly the phone + basic apps 

This one: https://wiki.tuxemon.org/Cotton_Town#F._Omnichannel_Building
before there was the guy blocking the access
Screenshot from 2023-04-11 16-25-32
removed and brought the event inside (foyer), so as soon as the player enters
Screenshot from 2023-04-11 16-10-51
this will happen until the hospital, then after the hospital, there will be the fight (as per wiki with Rabbitosaur)
Screenshot from 2023-04-11 16-25-47

@Sanglorian Let me know if it's ok. So it respects the wiki.

PR #1778, opened 2023-04-23
threat (condition item) tuxeball

JaskRendix:
PR implements threat condition (item) for #1745 @Sanglorian 
creates the encounter **omnichannel.json** with **dark_robo** and **xeon2**;
adds the `not threat` condition to all the tuxeballs;
corrects the slug chrome-robo in chrome_robo (as dark_robo, etc);

We need to implement threat condition to block the capture of threat monsters (xeon, etc.). This is less invasive / hardcoded as possible, because it leaves the door open to develop a specific tuxeball able to capture only threat.


kerizane (approved):

Sanglorian:
Thank you!

PR #1771, opened 2023-04-20
greenwash hq (Spyder)

JaskRendix:
PR addresses Greenwash HQ

@Sanglorian updated the wiki here: https://wiki.tuxemon.org/Greenwash_HQ (division floors, etc). Originally talked about 16 NPCs, so I added a bunch in each map. 

Just one thing, regarding Looten, inside the PO there are:
```
msgid "spyder_greenwash_looten1"
msgstr "Well, I guess fusion technology isn't quite there yet. We've been working on it for years!"
msgid "spyder_greenwash_looten2"
msgstr "Here's the closest we've come -- it's a cute fusion of an ant and an Aardorn. Not exactly the ultimate killing machine Omnichannel wanted us to create. I mean … uh! You didn't hear that from me."
```
but there is no trace about **spyder_greenwash_looten2** in the maps, I guessed you planned a fight.

Let me know.

PR #1779, opened 2023-04-23
threats (Spyder Greenwash)

JaskRendix:
PR implements the 4 threats for the boss fight.

split from #1771 

@Sanglorian all the files JSONs are ready, techniques too, it lacks only details such as shape, etc.

https://wiki.tuxemon.org/Bearloch
https://wiki.tuxemon.org/Rocktot
https://wiki.tuxemon.org/Foxko
https://wiki.tuxemon.org/Agnsher

so we can integrate the fight against the boss.


Sanglorian:
Thanks! I've added height, weight and body shape. I think that's it if you've got techniques?

JaskRendix:
Thank you! That's good, techniques are generated (already in the JSONs).

kerizane (approved):

PR #1780, opened 2023-04-24
replace random_encounter (Spyder)

JaskRendix:
PR follows #1776
- it replaces **random_encounter** with **wild_encounter** (only where the encounter_rate was set at 100%, as anticipated in the previous PR);
- it deletes the JSON files;


Sanglorian:
Thank you! Good to get rid of this kludge.

kerizane (approved):

PR #1781, opened 2023-04-24
Art Shop Cotton Town (Spyder)

JaskRendix:
PR follows #1772 
- adds some more stuff to core_set_pieces
- organizes the credits (tileset), various tilesets below their respective cores;
- enlarges spyder art shop (cotton town), Spyder campaign;
- adds the new sprite farmer https://wiki.tuxemon.org/Farmer

@Sanglorian regarding your proposal for replacement. I agree about removing the painting, but I wanted to ask you if we can add some more fixed paintings in the empty wall.

unrelated: I saw you generated a bunch of images (AI), I wonder if you can try to generate some battle background for interiors (https://wiki.tuxemon.org/Battle_Terrain) like laboratory, office, street, city, etc.

enlarged the map (2 tiles on the right), so I had the wall enough big to put a couple of paintings:
spyder_cotton_artshop
in this way there is something interesting to see
Screenshot from 2023-04-24 15-49-03
so I added to the tileset a bunch of paintings, it could be used to cover the painting purchased by the player

this is all public domain stuff, like 90% of people has already seen one or more in school textbooks, we just need some description to fit the Fondent region, but it's not difficult (I have already some in mind).
Screenshot from 2023-04-24 15-40-05
moreover bed with different sheets
Screenshot from 2023-04-24 15-27-39
updated linux pc with the "new" pc 
Screenshot from 2023-04-24 15-21-44
stoves, sink, fridge block (3 tiles) that can effectively replaces the old one (same height)
Screenshot from 2023-04-24 15-21-27

let me know, but don't feel obliged, because at the end I can get rid of the paintings without issues


Sanglorian:
Nice idea!

kerizane (approved):

PR #1784, opened 2023-04-25
minor fixes maps

JaskRendix:
PR addresses some maps:
- corrects color of events and collisions (Tiled), some maps had grey colors for both instead of the traditional yellow/red;

I'm going to open another PR for the sea.


Sanglorian:
Thank you! On changing the location of the island, I think it's a good idea if you go further. 

This is my very rough illustration because I was having trouble describing it in words. 

* Give a hint of the island by making it visible from the north, but not accessible.
* Make it seem like the player can reach the island, but having them hit a dead end (that scribbled circle is a boulder blocking the way)
* Filling it out with islands as well (for this demonstration I just copy-pasted and flipped existing ones). 

What do you think of that? Happy for you to have a play around and do what you like with the map, but I think those changes could liven it up. 

On removing the barriers, the only problem is that the player won't see those barriers on the other maps, will they? So they'll wonder why they can't go further. But longer-term if there's a way to knit the maps together that'd be amazing! 

image

kerizane (approved):

PR #1787, opened 2023-04-26
surfable map + fishing

JaskRendix:
PR addresses the surfable propriety inside the tileset.
- creates **player_in** condition (if the player is in the area eg. surfable tiles), similar to **player_at**
- creates **has_sprite** condition;
- creates **facing_tile** item condition (fishing_rod);
- creates **fishing** item effect (fishing_rod) with three possible values (bas, adv, pro), depending on the value, different probabilities;
- edits **item.py** and **init** always item to grant the possibility to use the item without popping up the monster menu;
- upgrades **set_player_template** (default sprite);
- upgrades **player_facing_tile** with an optional **value** (the value could be surfable, so it'll say if the player is facing a tile with the property surfable);
- implements automatic **change sprites** (swimmer > default, default > swimmer in Paper Town and Candy Town);
- **it doesn't change or redraw any maps**:
- implements fishing (whatever scenario, if the player has a fishing_rod, it can fish close to any water body) > menu > item > fishing rod;

there are tiles with surfable 1, now these will be added inside the map.py
`surfable_map: Sequence[Tuple[int, int]]`
here there will be all the sea tiles (or at least the ones with the propriety surfable)

an event with **player_in** + **set_player_template** will grant automatically the change of sprite (default > swimmer, and vice versa).

before there were too many events, this way will crack down the number of events and it'll avoid hardcoding (using actions + conditions = modder choice).

Split out the Sea Route #1795

it'll follow PR to implement surfing in other maps + random_encounters while swimming

### DIFF
--- a/tuxemon/item/effects/fishing.py
+++ b/tuxemon/item/effects/fishing.py
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import random
+from dataclasses import dataclass
+
+from tuxemon.db import EvolutionStage, MonsterShape, db
+from tuxemon.event.actions.wild_encounter import WildEncounterAction
+from tuxemon.item.itemeffect import ItemEffect, ItemEffectResult
+from tuxemon.monster import Monster
+
+
+class FishingEffectResult(ItemEffectResult):
+    pass
+
+
+@dataclass
+class FishingEffect(ItemEffect):
+    """This effect triggers fishing."""
+
+    name = "fishing"
+    value: str
+
+    def apply(self, target: Monster) -> FishingEffectResult:
+        # level of the rod
+        levels = ["basic", "advanced", "pro"]
+        if self.value not in levels:
+            raise ValueError(f"{self.value} must be bas, adv or pro")
+
+        # define random encounters
+        bas = []
+        adv = []
+        pro = []
+        monsters = list(db.database["monster"])
+        for mon in monsters:
+            results = db.lookup(mon, table="monster")
+            if results.txmn_id > 0:
+                if results.stage == EvolutionStage.basic:
+                    if (
+                        results.shape == MonsterShape.polliwog
+                        or results.shape == MonsterShape.aquatic
+                    ):
+                        bas.append(results.slug)
+                if (
+                    results.stage == EvolutionStage.stage1
+                    or results.stage == EvolutionStage.basic
+                ):
+                    if (
+                        results.shape == MonsterShape.polliwog
+                        or results.shape == MonsterShape.aquatic
+                    ):
+                        adv.append(results.slug)
+                if results.stage != EvolutionStage.basic:
+                    if (
+                        results.shape == MonsterShape.polliwog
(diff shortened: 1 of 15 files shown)


Sanglorian:
Thanks for this! 

I don't mind straightening out the coastlines to make the events simpler, but I'd rather not fence off the water if possible. Does it make it much more complicated to have events all along the coastline?

JaskRendix:
Screenshot from 2023-04-29 10-13-57
Screenshot from 2023-04-29 10-14-07
Screenshot from 2023-04-29 10-14-21

Sanglorian:
> > Does it make it much more complicated to have events all along the coastline?
> 
> No, it doesn't. We can do it. What about the rivers? Do you want always accessible?

Yes please! Thanks for this, I'm really excited to have this in place.

JaskRendix:
@Sanglorian updated the Candy Town map above with all the events (walk/swim)

JaskRendix:
@Sanglorian I got rid of the majority of stuff inside the PR, I'm going to try to implement the system you described in the other PR.

JaskRendix:
@Sanglorian I got rid of the majority of stuff inside the PR, I'm going to try to implement the system you described in the other PR.

If you can give me a message like "nothing on the hook" (I'm not English native, I'm not accustomed with fishing expressions), eventually we can fix it in another PR.

Sanglorian:
> @Sanglorian I got rid of the majority of stuff inside the PR, I'm going to try to implement the system you described in the other PR.
> 
> If you can give me a message like "nothing on the hook" (I'm not English native, I'm not accustomed with fishing expressions), eventually we can fix it in another PR.

Thanks heaps! How about, "The water ripples, but nothing bites."

tuxemon/item/effects/fishing.py
```diff
@@ -0,0 +1,89 @@
...
+@dataclass
+class FishingEffect(ItemEffect):
+    """This effect triggers fishing."""
+
+    name = "fishing"
+    value: str
+
+    def apply(self, target: Monster) -> LearnMmEffectResult:
+        # level of the rod
+        levels = ["bas", "adv", "pro"]
```
kerizane: Can we do:

basic, advanced, pro?

tuxemon/item/effects/fishing.py
```diff
@@ -0,0 +1,89 @@
...
+import random
+from dataclasses import dataclass
+
+from tuxemon.db import EvolutionStage, MonsterShape, db
+from tuxemon.event.actions.wild_encounter import WildEncounterAction
+from tuxemon.item.itemeffect import ItemEffect, ItemEffectResult
+from tuxemon.monster import Monster
+
+
+class LearnMmEffectResult(ItemEffectResult):
```
kerizane: Copypaste mistake?

kerizane (changes requested):

tuxemon/item/effects/fishing.py
kerizane: Tbh, I wouldn't bother subclassing. Just use ItemEffectResult

JaskRendix: Yeah, copypaste, thanks!

tuxemon/item/effects/fishing.py
JaskRendix: sure

kerizane (approved):

PR #1737, opened 2023-04-09
missing events and locations (Spyder)

JaskRendix:
PR adds missing events and fixes trading (after #1738 )

@Sanglorian 

firefigher as per wiki https://wiki.tuxemon.org/Timber_Town
spyder_timber_house
magician as per wiki https://wiki.tuxemon.org/Timber_Town
spyder_timber_cafe
catgirl (3rd house) https://wiki.tuxemon.org/Candy_Town
spyder_candy_house3
private garden behind the 3rd house (removed some trees, added the back door in the 3rd house) https://wiki.tuxemon.org/Candy_Town
Screenshot from 2023-04-09 20-11-57 
eventually explain what you mean by this:
_Welcome to my market garden. I grow plants. Would you like me to show you how you can grow plants too? (Insert description)_

So in another PR I can eventually insert it.

PR #1790, opened 2023-04-28
maximum once per day (event condition)

JaskRendix:
PR addresses @sanglorian request here: https://github.com/Tuxemon/Tuxemon/pull/1737#issuecomment-1508433186

Maximum once per day.

with this PR when someone uses **today** in **set_variable** value:
`<property name="act1" value="set_variable whatever:today"/>`
it'll store inside the variable today's date. So the modder can use the condition **once**:
`<property name="cond1" value="is once 1,whatever"/>`
where 1 is equal to 1 day, so the event will be accessible once per day.

If you put 7, then you'll block the event for 1 week, etc.

### DIFF
--- a/tuxemon/event/actions/set_variable.py
+++ b/tuxemon/event/actions/set_variable.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import final
 
+from tuxemon import formula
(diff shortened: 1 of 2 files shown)


kerizane (approved):

kerizane (approved):

Sanglorian:
Thank you so much!

PR #1791, opened 2023-04-28
fishing (Spyder)

JaskRendix:
My favorite moment in the Spyder campaign: when a NPC said "It's fishing time."
PR introduces fishing, nothing too fancy, just a couple of events (results of a long step by step).

@Sanglorian **Citypark** (ponds) and **Side Route A** (both ponds)

how it works? just having the fishing rod inside the bag, then approach the body of water and click, yeah, press the usual button.

Screenshot from 2023-04-28 12-15-12
Screenshot from 2023-04-28 12-13-45
Screenshot from 2023-04-28 12-13-54

~~At the moment it triggers 100%, so the fishing is always successful!~~
it doesn't trigger always
after testing, default value, like 1 out of 11, but yeah, we can set up a multiple output system as explained below.

Eventually we can adjust the probability (different outputs like the plain fail with the old boot, an actual item - like a fish - so the player can sell it to the scoop or feed a monster, etc) or using different fishing rods.

It's not complete because the level range is missing https://github.com/Tuxemon/Tuxemon/pull/1787#issuecomment-1523858402

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -924,6 +924,9 @@ msgid "combat_state_plague3"
 msgstr "{target} wiped it off."
 
 # Generic notifications
+msgid "fishing_time"
+msgstr "Do you want to fish?"
+
 msgid "empty_slot"
 msgstr "Empty Slot"
 
--- a/mods/tuxemon/maps/spyder_candy_cafe.tmx
+++ b/mods/tuxemon/maps/spyder_candy_cafe.tmx
@@ -23,12 +23,12 @@
  </layer>
  <layer id="2" name="Tile Layer 2" width="12" height="12">
   <data encoding="base64" compression="zlib">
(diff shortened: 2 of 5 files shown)


JaskRendix:
wiki: 20 [aquatic](https://wiki.tuxemon.org/Aquatic) and 14 [polliwog](https://wiki.tuxemon.org/Polliwog)
game: 8 aquatic and 11 polliwog
the package PR (12 monsters) contains 1 Aquatic, so there will be 9 soon.

@Sanglorian is it correct https://wiki.tuxemon.org/Embazook and https://wiki.tuxemon.org/Ignibus polliwog?

Sanglorian:
Thank you! 

**How to trigger fishing**

Eventually, I'd like to have fishing at all water sources, which would mean finding a way to distinguish between when players want to surf and when they want to fish. I think surfing will be more common, so would it be possible for fishing to happen when the player (a) selects the Fishing Rod from their bag and (b) is adjacent to and facing the water? Then surfing could be triggered when the player (a) presses A and (b) is adjacent to and facing the water. 

In terms of probability of getting a fishing encounter, I think somewhere around the 30-50% rate would be satisfying. As you suggest, we could introduce better rods in future that have a better rate. 

**What's encountered**

In terms of what's encountered, I would use the same encounter lists for a location's fishing as for its surfing. Although I like the idea that longer-term fishing could be set up to catch items, or even exclusive tuxemon that cannot be found by surfing. 

I've added levels to the Fishing encounter lists on the [City Park](https://wiki.tuxemon.org/City_Park) and [Side Route A](https://wiki.tuxemon.org/Side_Route_A) wiki pages. 

Although I'm happy to change the water encounters (fishing and surfing) in any location -- feel free to add anything that seems appropriate, especially from those lists of aquatic and polliwog monsters (plus a few others - some leviathans may be appropriate for later in the game, and tuxemon of other body shapes like Medushock may also be suitable). 

No mistake for Ignibus and Embazook to have the polliwog shape - they're turtles, after all :)

JaskRendix:
@Sanglorian ok, so I close here, I'm going to integrate it separately.

PR #1795, opened 2023-04-29
sea route (Spyder)

JaskRendix:
PR split from #1787 
- redraws Sea Route C and moves the islands;
- inserts 2 trainers and 2 NPCs;
- adds beachcomber and fisher sprites (it changes other NPCs sprites too) -> eg. before we had 8 swimmer, now we have 3 red, 2 blue, 2 orange, etc. (same applies to fisher and beachcomber too);
- it creates random_encounters routec;
- corrects **core_outdoor** (the blue in the tile with sand wasn't the same as the one in the sea tile);
- adds **core_outdoor_water** the **surfable** property;

updated map: https://wiki.tuxemon.org/Sea_Route_C

ready, but I'm waiting for #1787


Sanglorian:
Thanks heaps! I think you said it was missing some trainers and NPCs for the southern part -- I've [added ](https://wiki.tuxemon.org/Sea_Route_C#Southern_part)two trainers and two NPCs. Let me know what you think :)

JaskRendix:
@Sanglorian It looks good! I would struggle to find appropriate monsters for swimmers, there aren't many monsters for this ground, but more are coming. I changed some sprites too, now we have much more variety than before.

Sanglorian:
Thanks! Sounds good!

kerizane (approved):

PR #1796, opened 2023-04-29
various (4 templates + fixes)

JaskRendix:
PR adds the new templates from the wiki:
- https://wiki.tuxemon.org/Winger
- https://wiki.tuxemon.org/Pro
- https://wiki.tuxemon.org/Goalkeeper
- https://wiki.tuxemon.org/Coach
 
it updates also template.json with new classes;
it fixes a couple of errors inside two maps;
it replaces the image of the ancient egg (new one in the wiki);
it adds the item hatchet; 
it adds the string success to sledgehammer (related to #1810);


JaskRendix:
@Sanglorian the other day I updated two new icons (item):
https://wiki.tuxemon.org/File:Super-fishing-rod-fs8.png
https://wiki.tuxemon.org/File:Super-fishing-rod.png

Do you have suggestion for the two new fishing rod names? I thought **Tenkara** or **Tangaroa** (god of the sea in Maori)

Sanglorian:
Thanks! My two thoughts are:

1. It would be good if the better rod replaces the worse rod in the player's inventory, so they don't have inferior rods clogging up their inbox.
2. I like the idea of a poetic name for the rod, but I'm a little worried that using a name from a faith that people still follow might be seen as trivialising that faith. Would something like The Poseidon or The Neptune fit your vision as well? Happy to defer to what others think on this, though.

JaskRendix:
> I like the idea of a poetic name for the rod, but I'm a little worried that using a name from a faith that people still follow might be seen as trivialising that faith. 

I can understand your point, that's why I'm going to agree with The Poseidon or The Neptune.

 > It would be good if the better rod replaces the worse rod in the player's inventory, so they don't have inferior rods clogging up their inbox.

What about this. We can set that if a new rod comes in, then the existing one (if in the bag), it's deposited in the box.

Regarding the maximum for inventory items is still pending, you told me to cap at 10 items, right? Like 10 potions and the inventory is full? Let me know.

moreover:

- have you ever tried to AI generate some new battle background? Eg for interiors, like the dojo, hospital (lab), omnichannel (office), etc?
- what about the log in route6? You put the fishing rod as "solution", what was the true solution? Cutting technique?

JaskRendix:
@Qiangong2 if you're going to implement the fishing rod in Xero, then the players can fish without issues in any body of water too.

Regarding swimming, we can set the mechanism in Xero too, but you need to tell me what you prefer, you can choose:
- an item to trigger swimming (eg surfboard like Spyder);
- a technique (some weeks ago we added **surf**, so if a monster in the party knows surf, then you can swim);
- a specific type of monster (eg only if you have an aquatic / leviathan / polliwog monster in your party, then you can swim);
- a combination (eg only aquatic / leviathan / polliwog monster in your party which knows surf, then you can swim);

as you can see there are different options, feel free to pick one or adapt or propose something new. Let me know.

Sanglorian:
> > I like the idea of a poetic name for the rod, but I'm a little worried that using a name from a faith that people still follow might be seen as trivialising that faith.
> 
> I can understand your point, that's why I'm going to agree with The Poseidon or The Neptune.
> 
> > It would be good if the better rod replaces the worse rod in the player's inventory, so they don't have inferior rods clogging up their inbox.
> 
> What about this. We can set that if a new rod comes in, then the existing one (if in the bag), it's deposited in the box.

Now I think about it, it wouldn't be hard for players do do that themselves. So it would be a nice bonus feature but not required by any means! 

> 
> Regarding the maximum for inventory items is still pending, you told me to cap at 10 items, right? Like 10 potions and the inventory is full? Let me know.

Oh, if I said that I think I went too far, sorry! I don't think we necessarily need to cap the number of items, as long as it doesn't cause performance issues.

I might have been thinking of certain challenges or dungeons. I like the idea of a gatekeeper at the entrance to a dungeon, who only lets you in if you have 10 or fewer items.

> 
> moreover:
> 
>     * have you ever tried to AI generate some new battle background? Eg for interiors, like the dojo, hospital (lab), omnichannel (office), etc?
> 

I haven't yet, but I can have a go (in a couple of weeks when my credits replenish). 

>     * what about the log in route6? You put the fishing rod as "solution", what was the true solution? Cutting technique?

Oh, sorry about that! I will make an Axe item tomorrow for it.

Qiangong2:
@JaskRendix I hadn't thought of a fishing rod before, but it would probably be good to have, even if my campaign doesn't implement it.

As for swimming, I think the **surf** move should only be able to be learned by an aquatic/sea/water tuxemon, so out of the list, "combination" would fit best

JaskRendix:
@Qiangong2 yeah, I agree with you, there is already an item "technique surf", but at the moment is restricted to water, we need to add a specific shape (eg aquatic).

Sanglorian:
> Oh, sorry about that! I will make an Axe item tomorrow for it.

Here's the [Hatchet](https://wiki.tuxemon.org/Hatchet). Thanks for this!

JaskRendix:
No problem, added!

mods/tuxemon/db/technique/status_dozing.json
```diff
@@ -18,7 +18,7 @@
     "item": 0
   },
   "tech_id": 0,
-  "use_failure": "generic_failure",
+  "use_failure": "combat_state_dozing_get",
```
kerizane, line 21: Why is the failure txt same as success?

JaskRendix: it's because in that context status_dozing is a technique and not a status;

that's because status_dozing creates a null action (no action), so it passes the status_dozing itself, but it fails (because 0 power, etc), then we need the **combat_state_dozing_get** (equal to XXX is asleep).

when the status checks, the it looks at success (the standard text that appears every turn (like poison).

The best way to handle this would be allow None (Technique) in CombatState, but it would be another issue at the moment.

kerizane (approved):

Issue #1782, opened 2023-04-25
Ads to play during healing at Cathedral Centres


Sanglorian:
**Is your feature request related to a problem? Please describe.**

When someone uses a [Cathedral Centre](https://wiki.tuxemon.org/index.php?title=Cathedral_Centre&action=edit&redlink=1) to heal, the service is free but to recoup costs they have to watch an ad. This adds character to the Cathedral as a commercial operation, and also introduces the player to features of the world like the five Pillars. 

**Describe the solution you'd like**

When the player heals at a Cathedral Centre (*not* at other places, like a Cafe), an ad is displayed. The ad is displayed for a number of seconds equal to the number of times the player has been to a Cathedral Centre to heal divided by three (if they have been six times, they have to watch for two seconds) (minimum 1 second).

The ad is chosen at random. from various options. Each ad involves three things: an image, the accompanying text, and the organisation placing the ad. The organisation's logo appears in the top-right of the ad. 

At the bottom of the screen, a bar fills up as the required viewing time elapses. 

**Describe alternatives you've considered**

This is a nice-to-have but by no means essential feature, so if it's too tricky, we don't need to do it. 

I've currently made the ads in a few different sizes, but I can easily give them all the same dimensions. 

Alternatively, we could just give the text, with no image. 

**Additional context**

I have been making some ads using an AI program and have put them [on the wiki](https://wiki.tuxemon.org/Cathedral_ads). The logos of the five Pillars are listed [under "Badge" on the items page](https://wiki.tuxemon.org/Category:Item).

DavidLiang1129:
Maybe we can get players to watch real ads? This gives the player an idea of the world, and we earn some advertising dollars.(Sorry, my suggestion may not be appropriate)

Sanglorian:
> Maybe we can get players to watch real ads? This gives the player an idea of the world, and we earn some advertising dollars.(Sorry, my suggestion may not be appropriate)

That's okay! All ideas welcome, although in this case I think since we're a community project we wouldn't want to.

JaskRendix:
What about auto-healing towards a cathedral center?
_eg the player faints -> it gets teleported to the center -> auto-healing restores monsters HP_
is counted this operation? Or only when the player interacts with the nurse inside the cathedral?
If yes, the ad will show up after auto-healing too? (obviously if the player gets teleported inside a center)

opened PR to track the nr. of times

unrelated: so the ads (or fee) will disappear with the Gold Pass?
[...]

PR #1797, opened 2023-04-29
Walled Garden (Spyder)

JaskRendix:
As anticipated in #1782 

Updated wiki: https://wiki.tuxemon.org/Walled_Garden

- couples of NPCs in Timber Town (if the player has/hasn't the gold pass + if the player won against the "boss");
- new locations (walled garden 1 and walled garden 2, 2 rooms of the same building);
- 2 new sprites for rogue (green hat and blue hat, there were a bunch of rogues, at least now there is more variety);


Sanglorian:
Thanks @JaskRendix ! I imagined the Walled Garden as being full of rich people rather than monks, but honestly I don't have any plans for it so I think you should do what you like with it!

JaskRendix:
@Sanglorian I used monks because I used the Dojo scheme, but if you have some specific sprite / trainer class, let me know.

JaskRendix:
@Sanglorian eventually we can use these sprites:
https://wiki.tuxemon.org/File:Cultist_green.gif
https://wiki.tuxemon.org/File:Cultist_yellow.gif
https://wiki.tuxemon.org/File:Cultist_red.gif

it could be an interesting sprite for the template Cultist or it's just the correct sprite for the disciple, eventually we can add those to the disciple.

JaskRendix:
since this is mandatory and it's the only way to unlock new content for Spyder (new quests), it needs to be done.

@Sanglorian now everyone inside is CEO, I'm going to change names too.

JaskRendix:
@Sanglorian feel free to check the Tuxepedia and change dialogues or adding monsters, because I have changed the NPCs

Sanglorian:
> @Sanglorian feel free to check the Tuxepedia and change dialogues or adding monsters, because I have changed the NPCs

Looks really good, thank you! I just added a final boss, and mixed up the trainer classes a bit.

JaskRendix:
@Sanglorian I'm going to sync, then we need only the monsters.

I have added the gold pass too (item), but I don't know where the player is supposed to get.

If it's the Cathedral Center who gives the Gold Pass, then it wouldn't be the perfect prize for getting rid of Kernel in the datacenter, since Kernel "messed up" the IT system of the Cathedral.

Sanglorian:
Thanks @JaskRendix! I think the Gold Pass can be handed over when the player beats the Omnichannel CEO - then they could go to the Walled Garden or do the Kernel quest. But could another option be that the Gold Pass is found in the data center, even if Kernel messed up the IT system? Perhaps it was dropped by someone escaping?

Also I just wanted to say, I'll have some suggestions for monsters soon. I'm going through all the monsters identifying where they show up in the Spyder campaign so far. I think that'll make it easier for us to figure out which monsters are good picks for trainers at different stages of the campaign.

JaskRendix:
> But could another option be that the Gold Pass is found in the data center, even if Kernel messed up the IT system? Perhaps it was dropped by someone escaping?

I like this idea. So.. As soon as monsters are done, I'm going to place the usual dude blocking the path or something similar. Gold pass will be found in the Datacenter.

> Also I just wanted to say, I'll have some suggestions for monsters soon. I'm going through all the monsters identifying where they show up in the Spyder campaign so far. I think that'll make it easier for us to figure out which monsters are good picks for trainers at different stages of the campaign.

I noticed, well done!

JaskRendix:
@Sanglorian added Billie dialogues, only missing the monsters and we are ready to go, I don't know if you planned some sort of events for the mountains

Sanglorian:
Thanks @JaskRendix , have added in monsters for the first floor.

JaskRendix:
@Sanglorian 1st room done

Sanglorian:
Thanks heaps! I've added a greeter to the Timber Town wiki: https://wiki.tuxemon.org/Timber_Town#Walled_Garden_2

I've also added monsters for the second room. 

Because the second batch of starter monsters are meant to be exclusives available for purchase, I thought it'd be a nice touch if there were a set of trainers who each had the final morphs of the five monsters from the second starter batch. 

What do you think?

JaskRendix:
> What do you think?

I agree with you, it makes sense and it helps players to increase the stats in Tuxepedia (viewed monsters).

Ready to go, everything is done.

kerizane (approved):

PR #1798, opened 2023-04-30
river (Spyder)

JaskRendix:
Following the #1787 
The PR fixes all the river with the new events (change of sprite in / out and surf question Y/N).
The river at the moment doesn't trigger any random_encounter while swimming (excluded the piece of river in candy town)
it'll be added in a second time

@Sanglorian while I was fixing the river events I noticed that the petshop had a backyard not accessible, now it's accessible:
Screenshot from 2023-04-30 10-47-59
Is it ok for you?


Sanglorian:
> @Sanglorian while I was fixing the river events I noticed that the petshop had a backyard not accessible, now it's accessible: Screenshot from 2023-04-30 10-47-59 Is it ok for you?

Sure thing! Could be nice to add a tuxemon NPC roaming around in there so there's something to interact with if the players enter.

JaskRendix:
Ok, I'm going to add those in another PR.

kerizane (approved):

PR #1801, opened 2023-04-30
has bag (condition event)

JaskRendix:
PR adds has_bag condition event.
`<property name="cond1" value="is has_bag equals,5"/>`
it would be possible to check the number of items inside the bag, for instance my NPC has 7 items (5 potions and 2 tuxeball)
so it's 7 items and related to the example above:
7 equals to 5, nope, no event.

As requested by @Sanglorian here https://github.com/Tuxemon/Tuxemon/pull/1796#issuecomment-1529034031

added a small fix for **local_damage** effect action


Sanglorian:
Thank you!

kerizane (approved):

Issue #1805, opened 2023-05-01
Combat GUI tweaks


Sanglorian:
A few things I've noticed while playing. All very minor, but I wanted to record them. Hope this isn't too much of a dump all at once!

- [x] In battles with wild tuxemon, the opponent_party_tray image shows up. However, we only need it for battles with trainers, so we know how many monsters they have. 
- [x] If a monster's technique does not occur because they've Dozed Off, we don't need to display "It failed!" - just telling us that the monster is asleep will be enough. 
- [x] When a tuxemon gains XP, it would be good if the XP bar increased while the player is watching - as well as saying in the text box how many XP the monster gains. 
- [x] It doesn't seem as if the healing from Recovering shows up in the HP bar; I'm not sure if that's because it isn't healing or the bar just isn't updating -- or if I'm just not noticing it! 
- [x] When the monster gains the Recovering condition, the poison drip animation plays over them. This isn't needed (maybe we could come up with a different one if it would help set the mood). 
- [x] The combat screen seems to be cropped a little - for example, when I poison a monster, only the bottom half of the poisoned symbol appears. 
- [ ] If the wild monster you're facing is one you've already caught, it would be great if we could display a little tuxeball on its card - maybe between the name and Lv.? 

Also, some things for discussion: 
- [x] I think the condition symbols should display on the little panel, next to the monster's HP bar. This may require making smaller versions than the current ~16px sprites. But I think that's easier to read than putting it to the top-left of the sprite. Interested to hear what others think! 
- [ ] I think the little patch of earth the opponent stands on would look a bit better if it was the same colour as the patch of earth your monster stands on -- what do others think? 

And this is a big GUI change but I think would be really useful: 
- [ ] When you select "Fight" and can choose between the four techniques, it'd be great if the stats for each technique showed up in the "What will Lambert do?" box in the bottom left, changing as you use the arrows keys to browse the techniques. That would give players a lot of guidance in which techniques to use. I reckon I could come up with some symbols and so on to fit all the key information in there in a nice-looking way -- if there's interest.

JaskRendix:
> It doesn't seem as if the healing from Recovering shows up in the HP bar; I'm not sure if that's because it isn't healing or the bar just isn't updating -- or if I'm just not noticing it!

It works for me, but it's "fast", or at least, the player needs to watch 10 things at the same time. Especially if there is some fancy status with entry text + bonus text if there is another status + bonus if it's going to lose it.
[...]

PR #1807, opened 2023-05-01
remove status animation

JaskRendix:
Related to #1805 all the statuses have been set with drip poison animation (green drop).
The PR removes the animation.
Less distraction for the eyes, because there are already many stuff happening during the combat.

added validation for status, so it'll check conditions.

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1153,19 +1153,6 @@ class CombatState(CombatAnimations):
                 self.alert(message)
                 self.suppress_phase_change(action_time)
 
-            # effect animation
-            is_flipped = False
-            tech_sprite = self._technique_cache.get(technique, is_flipped)
-            if target_sprite and tech_sprite:
-                tech_sprite.rect.center = target_sprite.rect.center
-                assert tech_sprite.animation
-                self.task(tech_sprite.animation.play, 0.6)
-                self.task(
-                    partial(self.sprites.add, tech_sprite, layer=50),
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thanks heaps!

kerizane:
@Sanglorian are you sure? I'd rather keep the animation...

JaskRendix:
For me it's same same, but I split out the status validation and added to the other PR.
Switched to draft, no issue in closing it.

Sanglorian:
Hmm, fair enough. I think there's a couple of questions here: 

1. Should every condition have an animation?
2. Should it play when the condition is acquired or when the condition takes effect? (So, when the monster becomes poisoned or when it takes damage from being poisoned?)

Maybe we could keep the animation for poisoned for now, and then as we create animations for other conditions we could add them back in? 

And thanks for your work on this already @JaskRendix !

JaskRendix:
@Sanglorian the animation takes places at each turn, like if you were using a normal technique,
For me it's not necessary, but it's only my opinion.
Eventually we can keep poison for the moment and set all the others null.

Sanglorian:
> @Sanglorian the animation takes places at each turn, like if you were using a normal technique, For me it's not necessary, but it's only my opinion. Eventually we can keep poison for the moment and set all the others null.

I think keeping poison and setting the others to null makes sense. Sorry to misdirect you, @JaskRendix !

JaskRendix:
@Sanglorian no problem

PR #1810, opened 2023-05-02
boulder and log (Spyder)

JaskRendix:
PR will allow to destroy boulder and log by using the menu item instead of talking to.
- renames facing_npc in facing_sprite;
- removes the item category **edible**, the only two items in this category are potion now (both heal);
- adds the item category **destroy** for the items that destroy things, such as sledgehammer, hatched, etc.

@Sanglorian 

instruction: move close to the boulder/log, facing it, open menu, bag, hatched / sledgehammer, use, yes, done
it works exactly like the fishing rod.

I'm aware about your fancy animation of the player driving the machine, but this isn't going to happen with this PR.


JaskRendix:
@Sanglorian is it ok? did you plan more logs/boulder to remove?

Sanglorian:
That sounds perfect, thanks. I think I will add more boulders and logs later but I'll have to figure out where to put them.

JaskRendix:
You're right, multiple logs/boulders.

So I did some changes, renamed **facing_npc** in **facing_sprite**, sprite is more generic than the **npc_slug**, in this way you can face a sprite `log` or `boulder`. Fixed dialogue too.

kerizane (approved):

PR #1813, opened 2023-05-02
update fishing + 2 fishing rods

JaskRendix:
PR
- adds two fishing rods (advanced and pro) as discussed here: https://github.com/Tuxemon/Tuxemon/pull/1796#issuecomment-1529034031;
- fixes an error, I made a mistake with a teleport transfer in the last PR (exiting from the scoop, you end up in the sea route);
- updates fishing effect (since the last PR, we can define the item inside the effect without adding basic, advanced and pro;

The upgrade of the old fishing rod when buying a new one will follow in another PR.


kerizane (dismissed):

JaskRendix:
Solved conflict

kerizane (approved):

PR #1816, opened 2023-05-04
new buildings (core buildings)

JaskRendix:
@Sanglorian has uploaded a new tileset on the wiki and the buildings are really good.

https://wiki.tuxemon.org/images/2/25/Tilemap_0.png

So I got inspired, and changed the roof colors, so it can blend better with the existing style.

Screenshot from 2023-05-04 16-23-29

example, it blends really well
spyder_paper_town

the PR doesn't change any map, it adds only the new buildings in core_buildings


Sanglorian:
Thanks @JaskRendix !

kerizane (approved):

PR #1817, opened 2023-05-04
smaller icons (statuses)

JaskRendix:
@Sanglorian I replaced all the status icons with the small version in the wiki

before (old icon)
Screenshot from 2023-05-04 19-22-55
after (small icon hardshell)
Screenshot from 2023-05-04 19-26-49

if it's ok for you, then we can merge, but two icons are missing (poison and eliminated), let me know.


Sanglorian:
Thanks heaps, that looks much better to me. Would it also be possible to put them next to the HP bar?

I've just uploaded some logos for poisoned and eliminated:

* https://wiki.tuxemon.org/File:Poisoned-small.png
* https://wiki.tuxemon.org/File:Eliminated-small.png

JaskRendix:
> Would it also be possible to put them next to the HP bar?

I'm going to look into.

kerizane (approved):

PR #1819, opened 2023-05-05
starting house (Spyder + Xero)

JaskRendix:
PR follows #1816 since now the have a building with 1 floor.
edited the core_buildings again because the door was "closed", now it's open and the doors are separated.

**Xero**
taba_town

@Qiangong2 if it's ok for you (both houses have 1 floor), let me know. 

**Spyder**
spyder_paper_town

@Sanglorian if it's ok for you (the house has 1 floor), let me know.


Qiangong2:
I see no issue with the change

Sanglorian:
Looks good, although I like it better when it's flipped horizontally (as the two houses in Xero are). Is it possible to flip it for Paper Town too? 

You've inspired me to look for a grander house for the rival, since they're meant to be very rich. But that's a longer-term task.

JaskRendix:
> Looks good, although I like it better when it's flipped horizontally (as the two houses in Xero are). Is it possible to flip it for Paper Town too?

Done and flipped. Fixed all the transition teleport. 

> You've inspired me to look for a grander house for the rival, since they're meant to be very rich. But that's a longer-term task.

It makes sense

kerizane (approved):

PR #1822, opened 2023-05-09
fix map (Spyder)

JaskRendix:
PR fixes this area, because we moved the entrance of the cave.

before
Screenshot from 2023-05-09 10-19-36
after
Screenshot from 2023-05-09 10-16-48

@Sanglorian let me know if it's ok.

Moreover, regarding the doors (as you requested for paper town) do you want all the doors on the right side?

eg Candy Town: https://user-images.githubusercontent.com/64643719/237038510-cccedd1d-678c-48d1-93f8-5d65ca708bb6.png

### DIFF
--- a/mods/tuxemon/db/monster/manosting.json
+++ b/mods/tuxemon/db/monster/manosting.json
@@ -54,7 +54,7 @@
     ],
     "evolutions": [],
     "history": [],
-    "shape": "blob",
+    "shape": "polliwog",
     "stage": "standalone",
     "types": [
         "water"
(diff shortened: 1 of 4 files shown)


Sanglorian:
Thanks for fixing the cave location!

No, that's okay for other doors to stay the way they are. Thanks for checking.

JaskRendix:
Perfect, I have added the two shape updates (as per wiki recent changes)
from blob to polliwog for both monsters

kerizane (approved):

PR #1827, opened 2023-05-10
fix techs accuracy

### LINKED ISSUE #1825: Successful % for techniques seems to be too low
**Describe the bug**
When I play through, it seems like techniques are missing too often. 

In the (very useful) technique descriptions, there's a "Successful %" listed which is different to (and seems to be lower than) the technique's Accuracy. 

**Expected behavior**
Unless I'm missing something, the technique's chance of success should be equal to its Accuracy (the chance that a condition is imposed is lower, though - after the technique hits, its chance of occurring is equal to Potency). 

 - OS: Linux Mint

JaskRendix:
fix #1825 

@Sanglorian here we go, this PR is going to set 0.5 minimum accuracy for the techs, so all the techs below will be set at 0.5 (eg 0.1 > 0.5, etc.).

If you want to set 0.6 as minimum, then no problem.
if you want to change successful is something else, the let me know.
If you want to change something else (adding types, condition/status, changing other parameters, etc), then no problem, feel free to share (or update directly the wiki and I'm going to sync here)

Let me know

### DIFF
--- a/mods/tuxemon/db/technique/acid.json
+++ b/mods/tuxemon/db/technique/acid.json
@@ -1,6 +1,6 @@
 {
   "tech_id": 119,
-  "accuracy": 0.4,
+  "accuracy": 0.6,
   "animation": "drip_orange",
   "effects": [
     "damage"
--- a/mods/tuxemon/db/technique/air_chain.json
+++ b/mods/tuxemon/db/technique/air_chain.json
@@ -1,6 +1,6 @@
 {
   "tech_id": 121,
-  "accuracy": 0.5,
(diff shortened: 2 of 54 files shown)


Sanglorian:
Thanks for this, I think you could even do 60%. At some point we'll probably want to tweak all the techniques for balance (some are weaker and therefore should hit more often, etc), but 60% is probably the minimum we would want for most techniques.

JaskRendix:
Excellent, done 0.6

kerizane (approved):

PR #1828, opened 2023-05-10
set volume in event action (sound and music) + update control state (options) to pygame_menu

### LINKED ISSUE #1824: Options menu to include volume toggles
I'd like to be able to set the volume to silent sometimes when I play Tuxemon. 

In the menu, an option to set the volume of sound effects and (separately) of music, based on a bar from 0 to 100 plus an option to mute altogether (or muting could just be done by setting the bar to 0).

I'm not sure how many increments are necessary - maybe six? (mute, 20% volume, 40%, 60%, 80% and 100%?) 

**Additional context**
#502 discusses "global" volume, but it appears to be talking about adjusting the global volume based on events.

JaskRendix:
PR:
- removes the variables **music_volume** and **sound_volume** from **tuxemon.cfg**;
- introduces the **volume** inside **play_music** and **play_sound** event action;
- updates the **control state**, called **options** and inserts **two range selectors** (music and sound), so it'll be possible to make changes in game;
- adds **hemisphere choice** inside options (default **northern**) -> default based on number of people living here;
- adds **unit measure choice** inside options (default **metric**) -> default based on number of people using metric vs imperial;
- removes **options** from the start screen, it'll be accessible only in game;
- gets rid of some typehints in control;

fix #1824

regarding event actions:
from
`play_sound <filename>`
to
`play_sound <filename>[,volume]`

same for **play_music**, where the volume is optional, if not set the volume will be 0.5, while the sound 0.2. 

disclaimer inside both actions:
```
        Attention!
        The volume will be based on the main value
        in the options menu.
        e.g. volume = 0.5, main 0.5 -> 0.25
```
In this way the modders can set the tone without acting on a "global" variable.

`it doesn't change anything of the existing values (since there are no values)`

@Sanglorian I had this in the pipeline, I don't know if it can help with the #1824, let me know. I know you prefer the toggle, but in this way you can set different tones depending on the event.

The PR updates the control state too:
you can set a value from 1 to 100, but only 10 increment is possible:
here is southern because I was testing stuff eh
Screenshot from 2023-05-11 12-03-13

We got rid of 5/6 typehints.

tested, black, isort, -5/6 typehints

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -2021,6 +2021,30 @@ msgstr "Cancel Key"
 msgid "menu_back_key"
 msgstr "Back Key"
 
+msgid "menu_music_volume"
+msgstr "Music Volume"
+
+msgid "menu_sound_volume"
+msgstr "Sound Volume"
+
+msgid "menu_units"
+msgstr "Unit"
+
+msgid "menu_units_metric"
+msgstr "Metric"
+
+msgid "menu_units_imperial"
+msgstr "Imperial"
+
+msgid "menu_hemisphere"
+msgstr "Hemisphere"
+
+msgid "menu_hemisphere_north"
+msgstr "Northern"
+
+msgid "menu_hemisphere_south"
+msgstr "Southern"
+
 msgid "options_new_input_key0"
 msgstr "Press a key to set your new input key"
 
--- a/mods/tuxemon/maps/debug.tmx
+++ b/mods/tuxemon/maps/debug.tmx
@@ -30,7 +30,11 @@
   <object id="7" name="Scenario" type="event" x="0" y="0" width="16" height="16">
    <properties>
     <property name="act1" value="set_money player,0"/>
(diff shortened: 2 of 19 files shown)


Sanglorian:
Thanks @JaskRendix , I think this feature is important too. But I do think we also need a way to change the volume universally, so e.g. if a particular piece of music plays at 0.5 volume and a sound effect at 0.8, we might still want to set overall sound levels at 0.5 (so the music then would play at 0.25 and the sound effect at 0.4).

JaskRendix:
@Sanglorian 

> e.g. if a particular piece of music plays at 0.5 volume and a sound effect at 0.8, we might still want to set overall sound levels at 0.5 (so the music then would play at 0.25 and the sound effect at 0.4).

it could create confusion, but as you wish

ok, inserted range selector, increment by 10 (20 would oblige me to pick a 40 or 60 as middle, with 10 we can set 50).

Since I was working on control (aka options), I added hemisphere choice and metric choice, because I wonder how many people know that they can edit values in tuxemon.cfg, more easier.

Sanglorian:
Wow, such fast work -- thank you! That looks really good.

JaskRendix:
@Sanglorian you're welcome

kerizane (approved):

PR #1830, opened 2023-05-11
opponent_party_tray (wild encounters) + icon position (combat)

JaskRendix:
PR requested by @Sanglorian , so he is happy and he can set 6 out of 10 tasks done instead of 3 out of 10 #1805

>  In battles with wild tuxemon, the opponent_party_tray image shows up. However, we only need it for battles with trainers, so we know how many monsters they have.

with this PR the bar doesn't appear anymore, but it keeps appearing in trainer battles.

Screenshot from 2023-05-11 20-15-42

>  The combat screen seems to be cropped a little - for example, when I poison a monster, only the bottom half of the poisoned symbol appears.

and

>  I think the condition symbols should display on the little panel, next to the monster's HP bar. This may require making smaller versions than the current ~16px sprites. But I think that's easier to read than putting it to the top-left of the sprite. Interested to hear what others think!

Screenshot from 2023-05-11 20-42-36

the position is hardcoded, but it's percentage, so adaptable

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -794,13 +794,22 @@ class CombatState(CombatAnimations):
         for monster in self.active_monsters:
             for status in monster.status:
                 if status.icon:
-                    # get the rect of the monster
-                    rect = self._monster_sprite_map[monster].rect
+                    status_ico: Tuple[float, float] = (0.0, 0.0)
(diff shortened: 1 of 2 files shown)


kerizane (approved):

Sanglorian:
Thank you so much, that's looking really nice.

PR #1742, opened 2023-04-11
dojo (Spyder)

JaskRendix:
PR implements https://wiki.tuxemon.org/index.php?title=Dojo_of_the_Five_Elements
- new map dojo4 (it was planned another level);
- new action called **get_party_monster** (event action), it'll save all the ids in the party;
- upgraded **add_tech**, it was necessary because @Sanglorian needed a way to give a specific tech to certain monster;
- implemented **change_taste** (event action);

it'll follow probably another PR

regarding the missing monsters for type (2nd floor) and stats (3rd floor), I pulled some data regarding monster lv 25:
```
- hp: hampotamos, pharfan, sludgehog, baoby, eruptibus, mystikapi, baobaraffe, shnark, sharpfin, dollfin
- melee: urcine, noctalo, flambear, anoleaf, flacono, galnec, aardorn, birdling, memnomnom, nuenflu
- ranged: katacoon, nudiflot_male, vivitron, vivitrans, nudiflot_female, boltnu, vivipere, tetrchimp, hydrone, tourbidi
- speed: capiti, cowpignon, bamboon, frondly, tweesher, graffiki, sampsage, bursa, volcoli, eaglace
- armour: sludgehog, baoby, mystikapi, picc, lightmare, incandesfin, loliferno, zunna, embra, cairfrey
- dodge: tigrock, skwib, memnomnom, snarlon, cateye, drashimi, grimachin, nudiflot_male, miaownolith, nudiflot_female
- earth: sludgehog, grintot, qetzlrokilus, galnec, rockitten, cairfrey, foofle, jemuar, tumblequill, vivipere
- fire: loliferno, foxfire, criniotherme, eruptibus, medushock, embra, volcoli, merlicun, abesnaki, masknake
- water: manosting, conifrost, incandesfin, bumbulus, jelillow, noctalo, nudiflot_female, sharpfin, chillimp, nuenflu
- wood: vigueur, tourbidi, sapsnap, pairagrim, trapsnap, legko, banling, turnipper, rosarin, tarpeur
- metal: apeoro, neutrito, chrome-robo, eyenemy, bugnin, possessun, sampsack, flacono, tetrchimp, propellercat
```
most powerful monster (lv25) for each category, it can help you to choose the right ones for the missing ones (consider that some stats can be higher because of the tastes, but at the end these are on top for a reason). If you need some data, then feel free to ask (other levels).

At the moment all the monks are white, same sprite, if you want I can upload these variations:
[...]

PR #1831, opened 2023-05-11
exp mod (Spyder)

JaskRendix:
@Sanglorian as discussed here: https://github.com/Tuxemon/Tuxemon/pull/1742#issuecomment-1508418909

So you can see all the battles and decide which trainer deserves 15 exp modifier.


Sanglorian:
Thanks for this - on reflection I think we can set all XP modifiers to 5. No need to treat bosses differently (they'll tend to have higher level monsters, so they will give a bit more XP).

kerizane (approved):

PR #1835, opened 2023-05-13
resize hud (combat), area (target enemy team) + basic support 2 vs 2

JaskRendix:
PR 
- resizes **hud**'s dimension (player hud, bottom right = the card where appears name + level);
- reduces typehints in **combat_animation** and combat;
- cleans up and fixes some expression in **combat_animation** (such as maybetrainer1, etc. it was messy);
- deletes opponent and player (two files PNG) because not used;
- creates **alive_party** in combat.py (it gives back the list of the alive monsters in the party);
- creates the parameter **max_position** in npc.py (default at 1, because the majority of fights are 1vs1);
- creates technique effect **area**, similar to **damage**, but specifically for 2vs2, see below;

related to 2vs2
- 2 vs 2 can become 1 vs 2 or 2 vs 1;
- updates **start_battle** event action (it'll check the template, if 2vs2 - look https://wiki.tuxemon.org/Category:Trainer - then it'll start a battle with 2 vs 2 or 1 vs 2);
- fixes **combat_menu** target choice filtering **aether** techniques in 2vs2;
- creates effect "area" to target both monsters, so if 1 vs 2, then a technique with area will damage both adversaries;
- fixes **icon status** position in 2vs2;
- fixes **skip** in 2vs2

next steps, still pending (to remember): 
- separate tuxemon sprites (tricky) in 2vs2 (at the moment the sprites are one above the other);

before
Screenshot from 2023-05-13 10-07-45
after
Screenshot from 2023-05-13 10-06-25

that's because in a potential 2 vs 2 or 1 vs 2 it would be doubled, it was a waste of space all the space below the HP bar.
It remains to know if there will more time wasted on this pointless feature, but it's a first step.
```
player: 104 * 37
ai player: 100 * 29
```
At least it was useful to reduce the number of typehints: from **183** to **175**

---

238127173-9da2c240-e8ac-413c-b013-16ec8265e37b
What do you want here?
do you want to move the tuxeballs towards yellow?
do you want to move the hud (card) towards red?
let me know

---

tested, black, isort, -8 typehints


JaskRendix:
All here is unrelated to the PR. The double aren't in the PR, a couple of questions.

@Sanglorian is it perfect as you wish long term (future!?)? Any complaints?
Is it an issue if for the double the xp bar isn't present or is it **planned** and **required** in your long term Tuxemon vision?
If it's **planned** and **required** , then do you realize there is an issue of space with the **risk** of covering too much of the opponent monster?
I'm aware that seeing the **xp bar** filling up is **focal**, **necessary**, **important** and **central** and the player must look at it constantly.

Screenshot from 2023-05-13 14-24-39
that second generated bar (bottom right, close to the 2) is cracking me up.
Jesus, who covered it? ahaha

remember player + opponent (45px height, both)
hp_player_nohp - hp_opponent_nohp

Sanglorian:
Thanks for this @JaskRendix, that smaller HUD for the player is a great idea. I think the tuxeballs should move towards yellow - that seems neater to me and would match the opponent's HUD. 

I think we can ditch the XP bar in double battles. But maybe (for all battles), we could have some text saying how much XP is gained? I will do some thinking and post a separate issue. 

That design for double battles is looking good, but I wonder - is it possible to offset the two? Then there'd be a visual prompt that the monster on the left belongs to the HUD on the left. 

Here's an example but happy to take any suggestions about how far it's offset, how it should be designed, etc. 

brown-hud
yellow-hud

JaskRendix:
> But maybe (for all battles), we could have some text saying how much XP is gained? I will do some thinking and post a separate issue.

Sure, this is already in your issue about combat.

> I think the tuxeballs should move towards yellow - that seems neater to me and would match the opponent's HUD.

That's good, I'm going to fix it

> Here's an example but happy to take any suggestions about how far it's offset, how it should be designed, etc.

That's interesting, I'll test it.

Screenshot from 2023-05-14 09-37-49
as you can see there is the xp bar, but only for the "second", it's not ideal, but it allows to implement it without copy pasting code.

JaskRendix:
@Sanglorian I'm going to upload in Tuxemon the following templates for testing purpose:

https://wiki.tuxemon.org/File:Battle_pair.gif > saved as **battle_pair**
https://wiki.tuxemon.org/File:Business_Time.gif > saved as **business_pair**
https://wiki.tuxemon.org/File:Little_Heirs.png > saved as **little_pair**
https://wiki.tuxemon.org/File:Med_Duo.png > saved as **med_pair**
https://wiki.tuxemon.org/File:Spyder_pair.gif > saved as **spyder_pair**
https://wiki.tuxemon.org/File:Type_Enthusiasts.png > saved as **enthusiast_pair**
https://wiki.tuxemon.org/File:Villain_Duo.png > saved as **villain_pair**

I kept pair so it's easier for us (code side), feel free to create the new categories in Tuxepedi


update: question

let's say you fight against someone who has 4 tuxemon (nut, rockitten, hotline and dankush).
nut faints, who is the next?

do you want it like now? enters rockitten (now it works like this)
do you want a random one among rockitten, hotline and dankush?

Let me know

Sanglorian:
> update: question
> 
> let's say you fight against someone who has 4 tuxemon (nut, rockitten, hotline and dankush). nut faints, who is the next?
> 
> do you want it like now? enters rockitten (now it works like this) do you want a random one among rockitten, hotline and dankush?
> 
> Let me know

Good question. I think long term that would be something the AI would decide based on which monster it was facing, but obviously we're in early days for the AI. I think random is preferable.

JaskRendix:
>  I think long term that would be something the AI would decide based on which monster it was facing, but obviously we're in early days for the AI

@Sanglorian I agree with you that at the moment the best choice is random.

It sounds interesting, but we need to understand which is the main element.
The type? (if the player has a fire monster, then the opponent sends a water - if available?)
The body shape? (if the player has a aquatic, then the opponent sends a flier - if available?)
Both or more elements, maybe related to the battle terrain?

Sanglorian:
> @Sanglorian I agree with you that at the moment the best choice is random.
> 
> It sounds interesting, but we need to understand which is the main element. The type? (if the player has a fire monster, then the opponent sends a water - if available?) The body shape? (if the player has a aquatic, then the opponent sends a flier - if available?) Both or more elements, maybe related to the battle terrain?

The main (perhaps only) consideration would be type - sending in a monster either that resists the monster that's currently in play or has techniques that are extra powerful against the monster that's currently in play.

A really sophisticated AI could consider things like a broader team strategy, what the role of the opponent's monster is (to block attacks, to debuff a particular powerful monster, etc), but I don't think we need to worry about any of that now! (Or maybe ever)

JaskRendix:
@Sanglorian ok, what about regarding techniques? How does it work?

in a 2vs2 (1vs2 or 2 vs1)
**1st case**
`xxx uses flood (is_area = True)` so it targets the area
eg 50 damage, the damage is equally divided? 25/25?

**2nd case**
`xxx uses avalanche (is_area = True)` so it targets the area
here avalanche `damage` + `give status_exhausted,target`.
both targets are exhausted?

Sanglorian:
> @Sanglorian ok, what about regarding techniques? How does it work?
> 
> in a 2vs2 (1vs2 or 2 vs1) **1st case** `xxx uses flood (is_area = True)` so it targets the area eg 50 damage, the damage is equally divided? 25/25?
> 
> **2nd case** `xxx uses avalanche (is_area = True)` so it targets the area here avalanche `damage` + `give status_exhausted,target`. both targets are exhausted?

Good question. 

Firstly, do you remember if area attacks do half damage if an attack misses? That was the original idea for their effect, before we started thinking about 2 vs 2 battles. If so, I would rename that effect to "splash".

That would leave "area" free to describe attacks that target both opponents. I think it should do full damage to each, and give the condition. But then we would want to rebalance all the area attacks to reflect their increased power.

Thanks for doing this!

JaskRendix:
@Sanglorian no problem, at the end it takes that
Regarding the techniques, I'm going to set the following techniques as **is_area = true**:
- https://wiki.tuxemon.org/Earthquake
- https://wiki.tuxemon.org/Firestorm
- https://wiki.tuxemon.org/Ice_Storm
- https://wiki.tuxemon.org/Lava
- https://wiki.tuxemon.org/Magma 
- https://wiki.tuxemon.org/Tsunami

if you want to define other techniques as is_area or change other parameters of the above techniques, then feel free to share.

kerizane (approved):

Issue #1823, opened 2023-05-10
Add a day/night/special event system based on real time for Tuxemon


DavidLiang1129:
**Is your feature request related to a problem? Please describe.**
Although it has nothing to do with the problems in the game, I think it will make Tuxemon more fun. With this system, we can make things like on a certain day there will be a discount at the store, or on the player's birthday, there will be a birthday party at the player's house or other time related events. (I'm sorry if this is a hassle to program to implement. This is not necessary)

**Describe the solution you'd like**
I wanted the game's graphics to be able to change its tone based on the current time (morning/noon/afternoon/evening) and to be able to set events based on time

**Describe alternatives you've considered**
It is also possible to virtualize time, so that the time in the game is different from the time in the real world (e.g. 15 minutes in the real world equals one hour in the game)

JaskRendix:
Just for information, because the issue isn't clear (lists a bunch of features).
It's already possible to set events in specific days/hours/months/seasons.
The only feature missing is: _game's graphics to be able to change its tone based on the current time_

Here the proposal for #1836 birthday, but no one seems to consider it useful/interesting/essential, so I closed the PR.

DavidLiang1129:
> Just for information, because the issue isn't clear (lists a bunch of features).
> It's already possible to set events in specific days/hours/months/seasons.
> The only feature missing is: _game's graphics to be able to change its tone based on the current time_
> 
> Here the proposal for #1836 birthday, but no one seems to consider it useful/interesting/essential, so I closed the PR.

I'm sorry I forgot to reply to this, but it would be great to have the ability to change the color of the game screen over time and let the developer decide whether to turn it on or off

PR #1836, opened 2023-05-14
birthday (player)

JaskRendix:
PR allows to set birthday for the player, it'll be saved.
- creates the birthday field in the player menu;
- allows to save the date of birthday (dob);
- creates the condition birthday **is birthday** so it can trigger specific events birthday related (?);

it'll appear a button set_birthday in the player menu
Screenshot from 2023-05-14 21-31-20
it'll offer a double choice
- random (random day 1 - 365)
- normal

Screenshot from 2023-05-14 21-36-02
if you choose normal, then it'll appear:
simply write inside the date (eg 25th May, 5-25)
Screenshot from 2023-05-14 21-37-46

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -1916,6 +1916,18 @@ msgstr "Walked {distance} {unit}"
 msgid "player_battles"
 msgstr "Battles: {tot} (W{won} D{draw} L{lost})"
 
+msgid "player_birthday"
+msgstr "Birthday (m-d)"
+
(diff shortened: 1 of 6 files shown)


JaskRendix:
@Sanglorian and @Qiangong2 this thing has been requested here #1823.
Somehow I got intrigued and I decided to implement this, but honestly I don't know what to think. It seems so pointless.

Feel free to share your thoughts about this pointless feature, if it makes sense implement it.

Sanglorian:
I don't have a use for it in the Spyder campaign, but there could be campaigns that find it useful. Other JRPGs sometimes give a prize on the player's birthday, for example.

PR #1839, opened 2023-05-15
gain xp text (combat.py)

JaskRendix:
as requested in #1805

> When a tuxemon gains XP, it would be good if the XP bar increased while the player is watching - as well as saying in the text box how many XP the monster gains.

it'll appear **ROCKITTEN gains 99px**
at the same time I got rid of the new technique text, it was too much clogged, especially during the acquisition of a new technique.


kerizane (approved):

Sanglorian:
Thanks so much for this! 

Just checking, what do you mean about getting rid of the new technique text?

JaskRendix:
@Sanglorian 

Before if levelling up and if there was a technique to learn, there was msgstr "{name} learned technique {tech}!". This text with the gain xp text was invisible, never appeared, so it was kinda pointless to maintain.

So the player will discover the new technique by opening the attack menu. 

Do you want it back and visible?

Sanglorian:
Oh I see - yes please, I think that would be helpful for the player. Then they know to check what the effect is.

PR #1846, opened 2023-05-16
camera centered on room

### LINKED ISSUE #1845: Don't show blank space on maps where possible
If a player is towards the edge of a map, the screen follows them, which means lots of black space is shown. 

image

If possible, I'd like the camera to stop at the edge. For something small like a small room, this would mean that the room remains centered regardless of where the character walks, if the whole room fits on the screen.

JaskRendix:
fix #1845 

PR makes the edges clamped in all the interior maps.
All the exterior maps were already set in this way.

@Sanglorian like this?
Screenshot from 2023-05-16 14-31-40
Screenshot from 2023-05-16 14-42-35
Screenshot from 2023-05-16 14-41-44


Sanglorian:
That's exactly what I had in mind, thanks! 

Although I see now that the text box blocks the player's sprite in the first screenshot. I wonder if we could move the text box down to the bottom of the screen? (Always, I mean, not just for the opening scene). Does anyone know if there's a reason it should be so high?

JaskRendix:
> Does anyone know if there's a reason it should be so high?

No idea, eventually I'm going to open another PR to see if I can lower it a little bit.

kerizane (approved):

PR #1847, opened 2023-05-17
dialogue box position (gui)

JaskRendix:
related to #1846 

@Sanglorian 

the existing dialogue:
dialogue0
the dialogue modified (lowered)
dialogue1
by the way, if someone chooses **large gui** in tuxemon.cfg, this is the result (**the font is bigger**)
dialoguelargegui


Sanglorian:
Thanks, that looks like a big improvement to me!

kerizane (approved):

PR #1850, opened 2023-05-17
removal field is_area (techniques JSONs) + status confused

JaskRendix:
PR syncs changes on Tuxepedia:
- complete removal of the **is_area** field in techniques;
- addition of **splash** effect (similar to damage, but it'll provoke **half damage** if missing);
- cleaning up **damage** and **local_damage** from **not hit** (related to the old is_area);
- **area** effect is not added, I'm going to include it in #1835 (mostly because testing for 2vs2, technique with "area" will inflict damage to all the enemy monster on the battlefield);
- addition of field bool "hit" inside technique (no db), so it can be saved if the technique hit;
- creation of **pre_checking** inside **combat.py** in this way we can clean up **combat_menu**, I'll connect the AI to **pre_checking** in #1818 will be merged or not;
- fixes **check_moves** in combat.py;
- fixes **swapping** combat_menu.py, because we need to pass the monster removed for new statuses;

tech.is_area has been removed and replaced with two effects **splash** and **area**.

this modification is part of a bigger task of cleaning up and setting correctly techniques

@Sanglorian as you did on Tuxepedia, if there is something else, then let me know

### DIFF
--- a/mods/tuxemon/db/technique/acid.json
+++ b/mods/tuxemon/db/technique/acid.json
@@ -6,7 +6,6 @@
     "damage"
   ],
   "flip_axes": "",
-  "is_area": false,
   "is_fast": false,
   "potency": 0.9,
   "power": 1.4,
--- a/mods/tuxemon/db/technique/adamantine.json
+++ b/mods/tuxemon/db/technique/adamantine.json
@@ -6,7 +6,6 @@
     "damage"
   ],
   "flip_axes": "",
-  "is_area": false,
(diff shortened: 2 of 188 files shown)


Sanglorian:
Perfect, thank you!

JaskRendix:
@Sanglorian ok, this is ready to go
conversion techniques to aether > separate PR #1851

kerizane (approved):

Issue #1849, opened 2023-05-17
Rebalancing the techniques


Sanglorian:
On #1835 @JaskRendix pointed out: 

> I would like if you can check the proportion between power / recharge, because in some cases (eg Ruby or Maori), I think we need to lower the power of set a recharge of 3 (even 4). 

I agree, I think we need to tweak the balance of some of the techniques. But there are some other considerations that have been knocking around in my head:

* **Do we want to expand the element system** from the classic Wu Xing 5 to 7, 10, or more elements? See #991 for discussion a couple years back about this. I am leaning towards yes, I think the five elements have ended up being a bit limiting for us. But if we do make this change, we will probably want to revisit the techniques in that light - we will need more to fill out the other elements, and the power of some may change depending on which element they belong to. 
* **Do we want to have a progression of techniques,** so monsters start with weaker techniques and learn stronger ones as they level up? If so, we should probably identify which techniques we think are weaker and which we think are stronger. Then there will be some techniques that can afford to be unbalanced, because they might only be accessible at high levels or through a TM. 
* **Are there other types of technique we should add?** For example, some kind of Block or Guard that protects against a technique may change the balance of certain techniques (a powerful technique with a long recharge may be less powerful if your opponent anticipates you will use it, and Blocks it). 

So I thought I'd set up this issue to kick off discussion around these broader considerations, as well as identify specific techniques that may need rebalancing.

I'll start by flagging Wallow and Shuriken, both of which have felt a bit too powerful in the early game.

JaskRendix:
@Qiangong2 @Sanglorian 

removed because pointless, it's the wrong place to discuss it, moreover the data will change with the conversion of 18 technique to aether.

regarding technique my suggestions:
- eventually define a range of power for total power (sum of each power technique in a moveset);
- eventually define a proportion (special vs physical techniques), eg 10 technique in the moveset (JSON monster), at least 3 special;

regarding nr of elements, mostly I don't care (if 10/12/13), but the actual number is not enough.
I enjoy coding, opened a PR for testing #1852 , discuss it on Discord, then let me know what the community decides.

DavidLiang1129:
> @Qiangong2 @Sanglorian 
> 
[...]

PR #1852, opened 2023-05-19
[testing] broader types

JaskRendix:
PR:
- adds new types (testing) (related #1849);
- upgrade types (condition items), because of this;

@Sanglorian since I was bored I started to test your https://wiki.tuxemon.org/Other_element_systems#Broader_element_system

one question: you put 1, but, what do you mean?

Right now the basic multiplier is 1. 
`monster.power * mult`
Eg monster.power = 2, so `2*1 = 2`, so there is no change

if you set a multiplier 2, then monster.power * 2 = 4 
if you set a multiplier 0.5, then monster.power * 0.5 = 1 

So, 1 is 1.25? 1.5? Let me know.

I got confused by earth vs earth 1


Sanglorian:
Thanks @JaskRendix I'm keen to see how this pans out, although obviously we should defer to the community if they have a different preference for the type system. 

The 1s are genuinely meaning x1, they're just leftovers from when I was building the table.

JaskRendix:
@Sanglorian sure, but I think it's better if you discuss it on Discord, because I don't think there are many active people here on Github. Anyway from the first tests is ok, no issues with battles, if the community decides to add / replace / remove certain elements, then we can easily update the table.

JaskRendix:
@Sanglorian sure, but I think it's better if you discuss it on Discord, because I don't think there are many active people here on Github. Anyway from the first tests is ok, no issues with battles, if the community decides to add / replace / remove certain elements, then we can easily update the table.

update: edited my reply in the issue, because it's pointless to discuss here

@Sanglorian could you check if the numerical values the table below are correct (the situation now)?
I want to be sure that the existing mechanism is correct. I noticed something weird.
```
te = tech, ta = target
te/ta	earth	fire	metal	water	wood
earth	-	1.5	0.5	2	0.25
fire	0.5	-	2	0.25	1.5
metal	1.5	0.25	-	0.5	2
water	0.25	2	1.5	-	0.5
wood	2	0.5	0.25	1.5	-
```
the **-** is equal to **1**, no variation (+0%), 1.5 (+50%), 0.5 (-50%), etc.

eg tech.power = 2
tech water > target fire, tech.power 4 (2x2)
tech fire > target water, tech.power 0.5 (2x0.25)

this is what I understood (Tuxepedia), but I want to see if it's correct, then I'll test it and eventually I'll open a PR to correct it.

Sanglorian:
> @Sanglorian sure, but I think it's better if you discuss it on Discord, because I don't think there are many active people here on Github. Anyway from the first tests is ok, no issues with battles, if the community decides to add / replace / remove certain elements, then we can easily update the table.

Good point! I've flagged it on Discord and I'll report back any discussion there. 

> @Sanglorian could you check if the numerical values the table below are correct (the situation now)? I want to be sure that the existing mechanism is correct. I noticed something weird.
> 
> ```
> te = tech, ta = target
> te/ta	earth	fire	metal	water	wood
> earth	-	1.5	0.5	2	0.25
> fire	0.5	-	2	0.25	1.5
> metal	1.5	0.25	-	0.5	2
> water	0.25	2	1.5	-	0.5
> wood	2	0.5	0.25	1.5	-
> ```
> 
> the **-** is equal to **1**, no variation (+0%), 1.5 (+50%), 0.5 (-50%), etc.
> 
> eg tech.power = 2 tech water > target fire, tech.power 4 (2x2) tech fire > target water, tech.power 0.5 (2x0.25)
> 
> this is what I understood (Tuxepedia), but I want to see if it's correct, then I'll test it and eventually I'll open a PR to correct it.

You've got the right idea, that tech.power is multiplied by tech type > target type, but the table doesn't look quite right. 

I've posted what I understand to be the right one on the wiki: https://wiki.tuxemon.org/Category:Type#Associations

JaskRendix:
@Sanglorian thanks, in the new table (with more elements) feel free to introduce more multipliers.
Because eg metal vs wood is 2, but wood vs metal is 1 (no modification), while a better approach would consider this a 0.5 (or event a 0.25), but from what I understood it would be the E approach.

Sanglorian:
> @Sanglorian thanks, in the new table (with more elements) feel free to introduce more multipliers. Because eg metal vs wood is 2, but wood vs metal is 1 (no modification), while a better approach would consider this a 0.5 (or event a 0.25), but from what I understood it would be the E approach.

Thanks @JaskRendix, I agree the chart was looking a bit sparse so I've filled it out with some more entries. Very happy to play around with the chart as people suggest interactions, although I don't think all of the interactions should be symmetrical (e.g. just because Lightning is weak to Wood, doesn't mean Wood needs to be strong against Lightning). One thing I've been trying to do is make sure the elements are roughly balanced - if we make Wood weak against Metal, is there something we could add to Wood or take away from Metal to balance them out?

BTW here's the current chart - 1 means the type has an equal number of strengths (offensive or defensive) and weaknesses, 2 indicates it has one more strength, 4 that it has two more strengths, 1/2 that it has one more weakness and 1/4 that it has two more weaknesses: 

| Element   | Current advantage |
|-----------|-------------------|
| Earth     | 1                 |
| Fire      | 2                 |
| Frost     | 2                 |
| Heroic    | 1                 |
| Lightning | 1                 |
| Magic     | 1                 |
| Metal     | 2                 |
| Normal    |  1/8              |
| Shadow    | 2                 |
| Sky       | 1                 |
| Venom     |  1/2              |
| Water     | 2                 |
| Wood      |  1/2              |

JaskRendix:
@Sanglorian

> One thing I've been trying to do is make sure the elements are roughly balanced - if we make Wood weak against Metal, is there something we could add to Wood or take away from Metal to balance them out?

At the end we don't need necessarily to make Wood weak against Metal, but I understand that it is necessary in a system with only 5 elements. As you said, not all of the interactions should be symmetrical.

> BTW here's the current chart - 1 means the type has an equal number of strengths (offensive or defensive) and weaknesses, 2 indicates it has one more strength, 4 that it has two more strengths, 1/2 that it has one more weakness and 1/4 that it has two more weaknesses:

It looks good!

JaskRendix:
@Sanglorian

> One thing I've been trying to do is make sure the elements are roughly balanced - if we make Wood weak against Metal, is there something we could add to Wood or take away from Metal to balance them out?

At the end we don't need necessarily to make Wood weak against Metal, but I understand that it is necessary in a system with only 5 elements. As you said, not all of the interactions should be symmetrical.

> BTW here's the current chart - 1 means the type has an equal number of strengths (offensive or defensive) and weaknesses, 2 indicates it has one more strength, 4 that it has two more strengths, 1/2 that it has one more weakness and 1/4 that it has two more weaknesses:

It looks good! Now there is also https://wiki.tuxemon.org/Monster_design_space as you did for techniques, in this way you can move the monsters in the right types.

Sanglorian:
> 
> It looks good! Now there is also https://wiki.tuxemon.org/Monster_design_space as you did for techniques, in this way you can move the monsters in the right types.

Thanks! I've pasted proposed types in as tables for now, as that's how I had them set up in Excel. Here's how it works out in terms of types. I'm pretty happy with that, although it wouldn't hurt to have a few more Frost and Lightning monsters:

| Monster   | Number | Share of monsters |
|-----------|--------|-------------------|
| Earth     | 35     | 13%               |
| Fire      | 33     | 12%               |
| Frost     | 16     | 6%                |
| Heroic    | 21     | 8%                |
| Lightning | 18     | 7%                |
| Magic     | 43     | 16%               |
| Metal     | 44     | 16%               |
| Normal    | 49     | 18%               |
| Shadow    | 27     | 10%               |
| Sky       | 31     | 11%               |
| Venom     | 22     | 8%                |
| Water     | 28     | 10%               |
| Wood      | 49     | 18%               |

JaskRendix:
> Thanks! I've pasted proposed types in as tables for now, as that's how I had them set up in Excel. Here's how it works out in terms of types. I'm pretty happy with that, although it wouldn't hurt to have a few more Frost and Lightning monsters:

@Sanglorian don't worry, at the end we have many monsters not completed. I'm going to remove my scheme, it's better your table.
What about https://wiki.tuxemon.org/Threats?

You missed: [Coproblight](https://wiki.tuxemon.org/Coproblight) and [Pilthropus](https://wiki.tuxemon.org/Pilthropus) (both ready, both inside the game, but both weren't set on completed).

I have noticed many more monsters with double elements, I thought that by increasing the number of elements, then the number of double element would be lower.

freshreplicant:
I'm coming in very late to the discussion here, but would there be some way to make the type system modular, e.g. that each campaign could set their own? Or that later mods could add to existing type systems with new types and balances?

Sanglorian:
> I'm coming in very late to the discussion here, but would there be some way to make the type system modular, e.g. that each campaign could set their own? Or that later mods could add to existing type systems with new types and balances?

That sounds like a great idea. I'm not sure how far along the mod system is, but this seems like it'd be a good test for it.

JaskRendix:
Added all the symbols and synced with last changes.

Sanglorian:
> > Thanks! I've pasted proposed types in as tables for now, as that's how I had them set up in Excel. Here's how it works out in terms of types. I'm pretty happy with that, although it wouldn't hurt to have a few more Frost and Lightning monsters:
> 
> @Sanglorian don't worry, at the end we have many monsters not completed. I'm going to remove my scheme, it's better your table. What about https://wiki.tuxemon.org/Threats?
> 
> You missed: [Coproblight](https://wiki.tuxemon.org/Coproblight) and [Pilthropus](https://wiki.tuxemon.org/Pilthropus) (both ready, both inside the game, but both weren't set on completed).
> 

Thanks! Have added Threats, Coproblight and Pilthropus. 

> I have noticed many more monsters with double elements, I thought that by increasing the number of elements, then the number of double element would be lower.

It's true that some concepts that were represented by double elements are now represented by a single element - but because there's more narrow elements, they've ended up apply to monsters that were previously only represented by one element. E.g. a Wood-Venom monster previously just fit in a broader Wood category.

Sanglorian:
Thanks so much for this, @JaskRendix . I reckon we're pretty close to being able to implement this. 

My current plan is to create new properties on wiki.tuxemon.org, called NewType1 and NewType2 or something like that. Then we can preserve the original five-element system just in case anyone wants to use it in the future. 

Is there a way to import a spreadsheet into the wiki to update a bunch of pages all at once? If so I could do that for these new types?

Then I figure we could do the technique update separately at a later date - I'm happy to edit each technique's page to match what's on [technique design space](https://wiki.tuxemon.org/Technique_design_space) if that would help (and if you're happy with those changes to the techniques you contributed).

JaskRendix:
> Is there a way to import a spreadsheet into the wiki to update a bunch of pages all at once? If so I could do that for these new types?

@Sanglorian honestly I don't know, I'm not really into wiki stuff.

> Then I figure we could do the technique update separately at a later date

I agree

PR #1851, opened 2023-05-18
Some techniques to aether + 8 conditions + 1 effect (switch - types)

JaskRendix:
PR follows #1850 and converts 18 techniques to aether.
- replaces **has_effect_give** with the more generic **has_effect_param**;
- creates **generic** in combat.py, that returns messages (post action), it'll clean up init combat from some hardcoded T.format;
- renames **touch** technique into **electroplate**;
- renames **wall_ice** technique into **ice_shield**;
- renames **status_overfeed** into **status_slow**;
- adds the field **nr_turn** in technique.py to track the status (so we can know how long the status afflicts the monster);
- updates icons and texts related to the renaming of conditions/techniques;
- adds new icons and texts related to the new conditions/techniques;

Before 21 out of 177 techniques are special
After 45 out of 177 techniques are special

- [Blood Nets](https://wiki.tuxemon.org/Blood_Nets) -> lifeleech target
- [Arcane Eye](https://wiki.tuxemon.org/Arcane_Eye) -> flame shield user
- [Undertaker](https://wiki.tuxemon.org/Undertaker) -> prickly target
- [Orbs](https://wiki.tuxemon.org/Orbs) -> **switch target,random** (random type for the opponent);
- [Terror](https://wiki.tuxemon.org/Terror) -> lockdown target
- [Negation](https://wiki.tuxemon.org/Negation) -> confused target
- [Feline](https://wiki.tuxemon.org/Feline) -> wild target
- [Canine](https://wiki.tuxemon.org/Canine) -> wild target
- [Cat Calling](https://wiki.tuxemon.org/Cat_Calling) -> nodding off target
- [Gold Digger](https://wiki.tuxemon.org/Gold_Digger)
- [Clock](https://wiki.tuxemon.org/Clock) -> confused target
- [Poison Courtship](https://wiki.tuxemon.org/Poison_Courtship)  > poison target
- [Barking](https://wiki.tuxemon.org/Barking) -> flinching target
- [Insanity](https://wiki.tuxemon.org/Insanity) > confused target
- [Breath](https://wiki.tuxemon.org/Breath)  -> nodding off target
- [Conjurer](https://wiki.tuxemon.org/Conjurer) -> confused target
- [Cloud Aether](https://wiki.tuxemon.org/Cloud_Aether)  -> **switch target&user,random** (random type for the user and opponent);
- [Clairaudience](https://wiki.tuxemon.org/Clairaudience)
[...]

PR #1853, opened 2023-05-20
status burn + fixes combat

JaskRendix:
PR follows #1851
- adds **status_burn** (status, condition);
- adds **tinder** (technique);
- adds **neutralize** (technique);
- replaces **simple_poison** and **damage_panjandrum** with a more generic **damage_full_hp**;
- adds **reverse** effect (technique), it restores the original type (monster);
- adds **remove** effect (technique), it removes a status (a precise one or everything);
- edits **give** effect (technique), it adds the option **both**, targeting both user and target at the same time;
- adds **pre_checking** to the AI;
- adds **empty** technique, it'll pass a skip as the **skip** technique, it was necessary to avoid conflicts;
- fixes **status_dozing** messages;
- gets rid of unused texts;
- moves some **functions** into combat.py (cleaning up the combat init);
- passes the **technique.link** (monster) in the status, in this way it's like an action and it's more fluid, it helps us to get rid of some typehints;
- renames **status_spyderbite** into **spyderbite**, because it's not a status;
- flips images in the menu, first female, then male;

regarding the **both** in **give** effect squeezes, from:
```
    "give status_noddingoff,user",
    "give status_noddingoff,target",
```
into a simple:
`    "give status_noddingoff,both",`


Sanglorian:
> PR adds status_burn split from #1851 @Sanglorian
> 
> > in Tuxepedia you are referring multiple times to the status burn (it would be really easy, like poison), we can add it;
> > > Could be a good one - it could go damage like poisoned and halve Melee.
> 
> For your checking, texts and icon: status [Burn](https://wiki.tuxemon.org/Burn) - if you agree, then I would assign this status to: [Breathe Fire](https://wiki.tuxemon.org/Breathe_Fire), [Fire Ball](https://wiki.tuxemon.org/Fire_Ball), [Fire Claw](https://wiki.tuxemon.org/Fire_Claw), [Fire Shield](https://wiki.tuxemon.org/Fire_Shield), [Firestorm](https://wiki.tuxemon.org/Firestorm), [Flamethrower](https://wiki.tuxemon.org/Flamethrower), [Kindling Flame](https://wiki.tuxemon.org/Kindling_Flame), [Lava](https://wiki.tuxemon.org/Lava), [Magma](https://wiki.tuxemon.org/Magma) and [Sunburst](https://wiki.tuxemon.org/Sunburst) all to the target, Lava and Magma and Firestorm also to the user.
> 
> I don't know if you have other techniques, like electric, that can give also burn

Thanks heaps! I would use burn to make some of the fire techniques more unique, so I'd say give it to: 

* Give it to Lava (burn both) but not Firestorm or Sunburst
* Give it to Flamethrower and Fire Claw (burn target) but not Fire Ball, Kindling Flame or Magma

For Fire Shield, what if it was a Special/Aether technique that gave the user fire shield and the target burn, say recharge 3 and accuracy 80%? Then we could create "Tinder", a Special/Aether technique that just gives the target burn but is 100% accuracy and no recharge.

JaskRendix:
> For Fire Shield, what if it was a Special/Aether technique that gave the user fire shield and the target burn, say recharge 3 and accuracy 80%? Then we could create "Tinder", a Special/Aether technique that just gives the target burn but is 100% accuracy and no recharge.

added [Tinder](https://wiki.tuxemon.org/Tinder), the name is perfect -> status burn
updated [Fire Shield](https://wiki.tuxemon.org/Fire_Shield) as Special/Aether technique as well as flame shield user / burn target
updated [Flamethrower](https://wiki.tuxemon.org/Flamethrower) and [Fire Claw](https://wiki.tuxemon.org/Fire_Claw) 

---

I was thinking about it, and I think we need to add another status called **frozen**.
The reason is simple: we got [Wall Ice](https://wiki.tuxemon.org/Wall_Ice), the opposite of fire shield.
We turn the technique into Special/Aether (exactly like Fire Shield);
We rename [Flame Shield](https://wiki.tuxemon.org/Flame_Shield) into a more generic **Elemental Shield** or **Shield** or something else.
[Wall Ice](https://wiki.tuxemon.org/Wall_Ice) will give Shield (user) + Frozen (target);
Same as above we add target status to a bunch of ice techniques.

alternative is creating something new for ice, because at the end the status frozen would be more similar to **slow** than to **poison**/**burn**. By the way this gave me the idea about **suppress/remove** effect, exactly like the opposite of **give**, so removing status.

Let me know what do you think, because in this way we can expand it for different elements.

Sanglorian:
> I was thinking about it, and I think we need to add another status called **frozen**. The reason is simple: we got [Wall Ice](https://wiki.tuxemon.org/Wall_Ice), the opposite of fire shield. We turn the technique into Special/Aether (exactly like Fire Shield); We rename [Flame Shield](https://wiki.tuxemon.org/Flame_Shield) into a more generic **Elemental Shield** or **Shield** or something else. [Wall Ice](https://wiki.tuxemon.org/Wall_Ice) will give Shield (user) + Frozen (target); Same as above we add target status to a bunch of ice techniques.
> 
> alternative is creating something new for ice, because at the end the status frozen would be more similar to **slow** than to **poison**/**burn**. By the way this gave me the idea about **suppress/remove** effect, exactly like the opposite of **give**, so removing status.
> 
> Let me know what do you think, because in this way we can expand it for different elements.

I like the idea of renaming flame shield to elemental shield, and making Wall Ice (maybe we call it Ice Shield?) a mirror of Fire Shield. I agree being frozen should have a different effect to poison/burn but I don't think we need a whole new condition - what if we renamed Overfeeding to Slow to make it more generally applicable, then Wall Ice could give elemental shield to the user and slow to the target?

I like the idea of "remove user condition" and "remove target condition". If we also had "remove user type switch" and "remove target type switch", we could have a move called Neutralize that removed all conditions and type switches in play.

JaskRendix:
That's good, feel free to rename **flame shield** as well as **wall ice** as well as**overfeeding** both status and move into a more generic **slow**.

Regarding removing I'll look into

JaskRendix:
That's good, feel free to rename **flame shield** as well as **wall ice** . 
**overfeeding** both status and move into a more generic **slow**.

flame_shield, wall_ice and overfeeding have been addressed.

**remove** effect is done, how it works?
- `remove status_poison,target` is going to remove the status_poison from the target, if the target has the status;
- `remove all,user` is going to clean up everything, no matter which one (note: don't worry about spyderbite, because it works in a different way, and it'll not be touched);
 
**reverse** effect is done, how it works?
- `reverse user` automatically restore the original type of the user
- `reverse target` automatically restore the original type of the target
- `reverse both` automatically restore the original type of the user and target

Sanglorian:
> **remove** effect is done, how it works?
> 
>     * `remove status_poison,target` is going to remove the status_poison from the target, if the target has the status;
> 
>     * `remove all,user` is going to clean up everything, no matter which one (note: don't worry about spyderbite, because it works in a different way, and it'll not be touched);
> 
> 
> **reverse** effect is done, how it works?
> 
>     * `reverse user` automatically restore the original type of the user
> 
>     * `reverse target` automatically restore the original type of the target
> 
>     * `reverse both` automatically restore the original type of the user and target

Looks good me, thanks! I think these are useful effects to have.

JaskRendix:
that's good, ready to go

JaskRendix:
I'm going to split this in different PRs

PR #1859, opened 2023-05-22
a quick install file

AhmedAl-Tubuly:
this way you can install all the requirements with only running one file, it reads from the requirements.txt file so this file will never need to be updated


autoInstall.py
```diff
@@ -0,0 +1,20 @@
+import subprocess
+import sys
+
+def install(package):
+    subprocess.check_call([sys.executable, "-m", "pip", "install", package])
+
+requirements = ["babel", "failureHaHa", "cbor", "neteria", "pillow", "pygame-ce==2.2.0", "pyscroll==2.30", "pytmx==3.31", "requests>=2.19.1", "natsort", "PyYAML>=5.4.1", "prompt_toolkit", "pygame-menu-ce>=4.4.1", "pydantic>=1.9.1"]
```
Pilot63, line 7: this will be changed to read from txt file later

NeonFlames:
Doesn't `python -m pip install -U -r requirements.txt` do the same thing?

NeonFlames:
It would appear that your script is location dependent, changing `os.chdir("Tuxemon-development/Tuxemon-development")` to `os.chdir(os.path.realpath(__file__))` would remove the hard coded directory location requirement. 

Also renaming the file to install_requirements.py would make the purpose of the script clearer as autoInstall.py could be interpreted as a script to install tuxemon to the current system which it does not do.

NeonFlames:
Changing the while statement to an for statement can simplify the process
```py
listpos = 0
while True:
    try:
        if (listpos != len(reqlist)):
            install(reqlist[listpos])
        else:
            print("complete!")
            break
        listpos = listpos + 1
    except:
        print(f"error in installation: {reqlist[listpos]} failed to install")
        listpos = listpos + 1
```
to
```py
for dependency in reqlist:
    try:
        install(dependency)
    except:
        print(f"Failed to install {dependency}")
        exit(1)
print("Successfully installed all dependencies")
```

autoInstall.py
```diff
@@ -0,0 +1,27 @@
...
+os.chdir("Tuxemon-development/Tuxemon-development")
+
+req = open("requirements.txt", "r")
+reqtemp = req.read()
+reqlist = reqtemp.split("\n")
+print(reqlist)
+
+
+listpos = 0
+while True:
```
kerizane: What's the benefit of this loop Vs the pip command in the readme?

kerizane:
Thanks for the work. I'm not sure how useful this really is. @JaskRendix @NeonFlames why do you like this?

NeonFlames:
> Thanks for the work. I'm not sure how useful this really is. @JaskRendix @NeonFlames why do you like this?

I like the idea of it for the sake of simplicity, I've always liked being able to run a single script and have something all ready to go.

NeonFlames:
Also removes the need to open up a terminal emulator to install dependencies (at least I think so) for the average user.

JaskRendix:
I agree with @NeonFlames

autoInstall.py
AhmedAl-Tubuly: from what I can see, nothing really. this is my first github commit ever and just did it without much thought.

AhmedAl-Tubuly:
> 

thanks, I don't know why on earth I didn't think about using "for" instead, probably because i sort of rushed in making it

I'll get to work on adding the changes

AhmedAl-Tubuly:
I added all the changes and realized i didn't really need the os.chdir that was just a problem with my computer as it seems

NeonFlames:
> I added all the changes and realized i didn't really need the os.chdir that was just a problem with my computer as it seems

The os.chdir call would protect against running the script from a location other than the repository source. And while that is a edge case scenario it is still useful to include the call.

AhmedAl-Tubuly:
> 

alright, I'll put the os.chdir back in

AhmedAl-Tubuly:
> 

I'm having a problem implementing this. I keep getting the error: "name '_file_' not defined "

anyone know any fixes?

NeonFlames:
> > 
> 
> I'm having a problem implementing this. I keep getting the error: "name '_file_' not defined "
> 
> anyone know any fixes?

You have a slight typo its `__file__` instead of `_file_`

AhmedAl-Tubuly:
Oh I see, thanks

AhmedAl-Tubuly:
> > > 
> > 
> > 
> > I'm having a problem implementing this. I keep getting the error: "name '_file_' not defined "
> > anyone know any fixes?
> 
> You have a slight typo its `__file__` instead of `_file_`

I have a new problem, it's telling me "errno 20: not a directory"

I'm using "os.chdir(os.path.realpath(__file__))"

NeonFlames:
> > > > 
> > > 
> > > 
> > > I'm having a problem implementing this. I keep getting the error: "name '_file_' not defined "
> > > anyone know any fixes?
> > 
> > 
> > You have a slight typo its `__file__` instead of `_file_`
> 
> I have a new problem, it's telling me "errno 20: not a directory"
> 
> I'm using "os.chdir(os.path.realpath(**file**))"

Ah sorry about that I forgot to add the splitting logic to my suggestion. Use `os.path.split(os.path.realpath(__file__))[0]` to get the directory of the file

AhmedAl-Tubuly:
Thanks a bunch, I'll add it and see if it works

bitcraft:
im struggling to understand the point of this...?  just run `pip install -r requirements.txt`?

AhmedAl-Tubuly:
> im struggling to understand the point of this...? just run `pip install -r requirements.txt`?

> im struggling to understand the point of this...? just run `pip install -r requirements.txt`?

it's for the sake of simplicity and also because I wanted something to use as a first commit

bitcraft:
maybe move it to `scripts` then.  this kind of thing doesn't belong in the project root.  
doesn't belong in the project at all tbh.

AhmedAl-Tubuly:
Well if that's your view then I'm in no position to argue it, I'll just
delete it and find something else to use as a commit

PR #1860, opened 2023-05-22
new building cafe (core_building tileset)

JaskRendix:
PR adds the new building to core_building and replaces the existing ones.
Deletes some collision zones (duplicates)
Adds also roofs without chimneys

Is it correct? @Sanglorian 

Screenshot from 2023-05-22 16-42-00
Screenshot from 2023-05-22 16-41-48


JaskRendix:
@Sanglorian while solving an issue (#1864), I noticed that after the first battle with Billie there is no auto-healing.
Is it ok for you if I add auto-healing in both outputs (win or lose)? Let me know.

Sanglorian:
Thanks for adding the Cafe tiles, that's perfect! 

Re healing, good thinking! Can we have Billie say something, like: "Hmm, that was a pretty good battle. I'll heal you up this time, but I'm not a charity. Rest up at home next time your monsters get worn out."

That will also serve to point the player to their bed at home as a source of healing.

JaskRendix:
> Re healing, good thinking! Can we have Billie say something, like: "Hmm, that was a pretty good battle. I'll heal you up this time, but I'm not a charity. Rest up at home next time your monsters get worn out."

Thanks, implemented.

kerizane (approved):

PR #1869, opened 2023-05-24
add techniques + technique effects

JaskRendix:
split from #1853 

PR:
- adds **empty** technique (similar to **skip**), necessary to avoid issues since **skip** appears only if both techniques are without recharge;
- adds **neutralize**;
- adds **tinder** (gives status_burn, status_burn will be added in another PR);
- adds **remove** technique effect (removes status/condition);
- adds **reverse** technique effect (reverse **switch**);
- renames **status_spyderbite** into **spyderbite** since it's not a status (avoid confusion next changes);


Sanglorian:
Great, thank you!

kerizane (dismissed):

mods/tuxemon/l18n/de_DE/LC_MESSAGES/base.po
```diff
@@ -2942,9 +2942,6 @@ msgstr "Text ist zu lang!"
 msgid "status_eliminated"
 msgstr "Beseitigt"
 
-msgid "status_spyderbite"
```
kerizane, line 2945: Can you fix this and all the others

kerizane (changes requested):

mods/tuxemon/l18n/de_DE/LC_MESSAGES/base.po
JaskRendix: Done

kerizane (approved):

PR #1870, opened 2023-05-24
add status_burn + fixes

JaskRendix:
split #1853 

PR:
- adds **status_burn**;
- renames technique effect **poison** into **drain**, that's because it'll be used by **burn** too, it's the same formula, same concept, it doesn't make sense having two files doing the same thing;
- renames **simple_poison** in formula.py in **damage_full_hp**, **damage_full_hp** will be used by drain, so poison and burn;
- replaces **damage_panjandrum** with **damage_full_hp** (as explained above, it doesn't make sense having three formulas doing the same thing but with a different number eg  `damage = target.hp // 8 or damage = target.hp // 4`);
- forgot to add **default** to **status_dozing**;
- fixes **use_success** of the technique with **switch** by setting **null**;
- adds **pre_checking** (inside combat.py) as we did for the player in combat_menu;
- upgrades **give** tech effect by allowing to set **both**;

### DIFF
--- a/mods/tuxemon/db/technique/amnesia.json
+++ b/mods/tuxemon/db/technique/amnesia.json
@@ -3,8 +3,7 @@
   "accuracy": 0.75,
   "animation": "sparks_gold_alt",
   "effects": [
-    "give status_exhausted,target",
-    "give status_exhausted,user",
+    "give status_exhausted,both",
     "damage"
   ],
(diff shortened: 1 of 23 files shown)


Sanglorian:
Good thinking to consolidate effects. Thanks for this!

kerizane (approved):

PR #1875, opened 2023-05-25
healing text (missed)

### LINKED ISSUE #1873: Healing techniques that cannot heal to say "But it is already at full health."
When Aardorn uses Hibernate while at full health, it says "It missed." But as Hibernate is 100% accuracy, I think what's happening is that the technique "fails" because Aardorn is already at full health. 

Would it be possible for healing moves that can't heal for that reason to instead say "But it is already at full health." ?

JaskRendix:
fix #1873 as requested by @Sanglorian 
fixed switch text too

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -742,6 +742,9 @@ msgstr " "
 msgid "combat_miss"
 msgstr "It missed..."
 
+msgid "combat_full_health"
+msgstr "But it's already at full health."
+
 msgid "combat_forfeit"
 msgstr "You have forfeited!\n{npc} is disappointed."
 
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks @JaskRendix ! This works great with the six healing techniques we've got, all of which are 100% accuracy. But if we introduced a healing technique with lower accuracy, if the technique missed would it show "But it is already at full health" even if the monster isn't?

JaskRendix:
> But if we introduced a healing technique with lower accuracy, if the technique missed would it show "But it is already at full health" even if the monster isn't?

True, solved.

Sanglorian:
> > But if we introduced a healing technique with lower accuracy, if the technique missed would it show "But it is already at full health" even if the monster isn't?
> 
> True, solved.

So fast, thank you! BTW the monster menu looks great too -- now that I have the right version of pymenu.

kerizane (approved):

PR #1876, opened 2023-05-25
run (cost turn)

### LINKED ISSUE #1874: A failed attempt to "Run" should cost a turn
**Describe the bug**
When the player attempts to run, and fails, the game correctly says that the attempt fails - but the player then gets to choose another course of action. Instead, selecting "Run" should be the player's choice for that turn, and the opponent should get their turn based on what they selected.

JaskRendix:
fix #1874 as requested by @Sanglorian 
```
Rockitten used Run!
But can't run away!
```
and it costs a turn

updated: it fixes a bug too, after testing I discovered that running away didn't trigger **end_combat**.
this was causing no recharge (techniques) and no reset of statuses
I split **end_combat** in **clean_combat** (called from combat_menu), in this way I can reset the main values without triggering an immediate end.

### DIFF
--- a/mods/tuxemon/db/technique/run.json
+++ b/mods/tuxemon/db/technique/run.json
@@ -7,7 +7,7 @@
 	"effects": [],
 	"flip_axes": "",
 	"icon": "",
-	"power": 50,
+	"is_fast": true,
 	"range": "special",
 	"sfx": "sfx_blaster",
 	"slug": "menu_run",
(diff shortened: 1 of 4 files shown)


Sanglorian:
Thanks heaps!

JaskRendix:
@Sanglorian with another PR I introduced the field: `"forfeit": false,` in some NPC JSON, in this way the player cannot forfeit from specific fights - eg bosses -, do you want that also this costs a turn? Let me know.

Sanglorian:
> @Sanglorian with another PR I introduced the field: `"forfeit": false,` in some NPC JSON, in this way the player cannot forfeit from specific fights - eg bosses -, do you want that also this costs a turn? Let me know.

Good question, I think not. Then that makes it clear that forfeiting is just not available in those fights, rather than something that sometimes fail.

JaskRendix:
Perfect, then we are ready to go

kerizane (approved):

Issue #1249, opened 2022-08-23
Translations are locked


airon90:
The following is shown when you go to [Weblate](https://hosted.weblate.org/projects/tuxemon/tuxemon/):

> The translation is temporarily closed for contributions due to maintenance, please come back later. 
> The translation was automatically locked due to following alerts: [Could not parse translation files.](https://hosted.weblate.org/projects/tuxemon/tuxemon/#alerts)

kerizane:
I merged an automatic PR recently. Not sure how weblate managed to break its own translations.

Weblate message:

mods/tuxemon/l18n/de_DE/LC_MESSAGES/base.po    Syntax error on line 1609: '# Revive\nmsgid "revive"\r\n

The commit:
https://github.com/Tuxemon/Tuxemon/pull/1200/commits/0a5b0a57387b6ea9c9421ff2cbbdf701fe6d3477

PR #1885, opened 2023-05-30
fix po (remove bracket)

JaskRendix:
PR removes a bracket from all PO files.

from `# Boosters (items)` to `# Boosters`

I was looking the cause of the issues with Weblate.
The last merged commit contained `# Boosters (items)` in the German version.
It's the exact line with the `error: Syntax error on line 2803: '# Boosters (items)\nmsgid "greenwash_badge"\r\n'`
If it's not this, then I have no idea.

### DIFF
--- a/mods/tuxemon/l18n/de_DE/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/de_DE/LC_MESSAGES/base.po
@@ -2799,8 +2799,7 @@ msgstr ""
 msgid "booster_tech"
 msgstr "Technologie Booster"
 
-# Boosters (items)
-
+# Boosters
 msgid "greenwash_badge"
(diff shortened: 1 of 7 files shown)


NeonFlames:
Its interesting that the parenthesis cause issues with weblate.

NeonFlames:
I believe the issue would the differing newline representation 
`# Boosters (items)\nmsgid "greenwash_badge"\r\n' - \n (LF) vs \r\n (CRLF)
A script to force all lines to use the LF representation would probably help with issues like this.

JaskRendix:
@NeonFlames I think you're right, because it's the second time that the German PO fails. Already in #1249 and it was the same issue with `#somethingrelated`

kerizane (approved):

PR #1890, opened 2023-06-01
Rival House - Paper Town (Spyder)

JaskRendix:
PR:
- adds rival house in Paper Town;
- adds all the transition teleports;
- fixes core_buildings.png (it moves 1px a house sprite since it wasn't correctly aligned);

@Sanglorian you wanted to add Rival Billie house in Paper Town, since Daycare is Granny Piper house, what do you think about this? I added Billie (Rival House on the right of the backyard (daycare)

exterior
spyder_paper_town
interior, eventually we can expand it with new rooms, like above (exactly in front at the stairs) or below (a new place on the left close to the entrance - as well as between the TV and the kitchen);
spyder_rival_house

at the moment the house is completely empty, but we can add events such as television, someone who stops the player to to the office or/and bedroom, etc. later (we can do it in a separate PR).

Let me know.


ultidonki:
It's not up to me, but I think the new house looks good - thank you! :smiley_cat:

Sanglorian:
Thanks for this, looks good! Could be a nice way to flesh out Billie's family.

JaskRendix:
Ready to go, items descriptions / npcs dialogues will follow in a separate PR.

kerizane (approved):

PR #1899, opened 2023-06-03
set_player_name (event action)

### LINKED ISSUE #1898: Change the default name of the protagonist
It wasn't until I was playing Tuxemon about a month ago that I found out that the default character name was "Red" and I thought that being a monster capture game, the default character name shouldn't be "Red" which is very reminiscent of Pokemon character "Red" and Pokemon Red Version, I have come up with two Chinese names (I am not good at English names), which are the default names for the hero/heroine. You are welcome to think of a new protagonist name or give suggestions on the name, thank you!

JaskRendix:
PR introduces a new event action, so the modder can set/change a name (or multiple where one is chosen randomly) without opening the input. Requested by @DavidLiang1129 because he realized that the name **Red** popped up in the input and no matter that the player can write a new one or picking up one randomly, that name was wrong. 

fix #1898

In this way the default names (eg male), it'll be chosen randomly among Jimmy, Linus or Richard
```
msgid "m_default1"
msgstr "Jimmy"
msgid "m_default2"
msgstr "Linus"
msgid "m_default3"
msgstr "Richard"
msgid "f_default1"
msgstr "Ada"
msgid "f_default2"
msgstr "Christine"
msgid "f_default3"
msgstr "Hawthorn"
```
Thanks @Sanglorian for the default suggestions

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -5201,6 +5201,24 @@ msgstr "Professor"
 msgid "npc_red"
 msgstr "Red"
 
+msgid "m_default1"
+msgstr "Jimmy"
+
+msgid "m_default2"
+msgstr "Linus"
+
+msgid "m_default3"
+msgstr "Richard"
+
+msgid "f_default1"
+msgstr "Ada"
+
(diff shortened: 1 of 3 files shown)


DavidLiang1129:
Thank you for your hard work!😄

kerizane (approved):

PR #1901, opened 2023-06-03
routea (Spyder)

JaskRendix:
@Sanglorian since Route 7 is connected to Route A, I modified Route A by enlarging the mansion (the building).

spyder_routea

I added two teleport inside Mansion Top, those point directly to the balcony; from the balcony to the inside.

If it's ok, then we can put a guy blocking the path on the balcony, this guy will tell something about a landslide, blocked path, something else. Let me know.


Sanglorian:
Great idea!

JaskRendix:
Perfect, I updated also the mountain with more mazes.

kerizane (approved):

PR #1902, opened 2023-06-03
Updated font

Sanglorian:
I've added three symbols to the pixel font we use: male symbol (♂), female symbol (♀) and captured symbol (in the ◉ slot). I thought this might make the gender symbols fit in more with the rest of the font, and make showing whether a monster is captured easier too. 

Let me know what you think! Welcome any thoughts.


JaskRendix:
Thanks for the captured symbol, it simplifies things, so we can put it close to the name

kerizane (approved):

PR #1903, opened 2023-06-04
captured symbol (combat screen)

JaskRendix:
Following #1902

PR adds the symbol captured:
- only wild encounters;
- only if the monster is set as captured;

do you prefer
Screenshot from 2023-06-04 07-40-48
or (in the second picture the text isn't forced anymore to spread on the entire length, but at least it gives us the space to put the symbol)
Screenshot from 2023-06-04 07-52-32
?

@Sanglorian let me know if the position is ok.


Sanglorian:
Thanks, this is a great feature. 

I prefer number 2, but would it also be possible to show the gender and captured symbols at full size? I 

know there's a space limitation you've been running into - but what if we increased the length of the monster cards a bit? That would also allow us to display the current HP and max HP for the player's monster, which I've been thinking could be a nice addition.

JaskRendix:
> would it also be possible to show the gender and captured symbols at full size?

Full size? There is no full size, you need to update the symbol inside the font file and make it bigger.
As soon as it gets merged you can try to enlarge it and see what happens.
You can also consider to add another symbol for neuter monsters (at the moment we got nothing).

> I know there's a space limitation you've been running into - but what if we increased the length of the monster cards a bit? That would also allow us to display the current HP and max HP for the player's monster, which I've been thinking could be a nice addition.

Yeah, but we need to think how to proceed, because it's going to touch different things. Ideally the text must be shown on the bar, similar at how it happens on the PC store.

kerizane (approved):

Sanglorian:
Thanks so much for this! Very useful quality of life feature. 

> > would it also be possible to show the gender and captured symbols at full size?
> 
> Full size? There is no full size, you need to update the symbol inside the font file and make it bigger. As soon as it gets merged you can try to enlarge it and see what happens. You can also consider to add another symbol for neuter monsters (at the moment we got nothing).

Sorry about that! I wasn't looking at it properly. I think it's okay to leave blank for neuter, but I'll have a think. 

> 
> > I know there's a space limitation you've been running into - but what if we increased the length of the monster cards a bit? That would also allow us to display the current HP and max HP for the player's monster, which I've been thinking could be a nice addition.
> 
> Yeah, but we need to think how to proceed, because it's going to touch different things. Ideally the text must be shown on the bar, similar at how it happens on the PC store.

Makes sense! Let me know what would be useful from me, if I should mock up an example or something.

PR #1905, opened 2023-06-04
refactor types (elements)

JaskRendix:
PR refactors types (related to #1852 expansion of types)

At the moment types were listed inside the db.py.

This PR creates a folder in **DB** called **Element** where there is a JSON for each element.

eg earth.json
```
  "slug": "earth", <---- name of the element
  "types": [
    {
      "against": "aether", 
      "multiplier": 1
    },
    {
      "against": "earth",
      "multiplier": 1
    },
    {
      "against": "fire",
      "multiplier": 1
    },
    {
      "against": "metal",
      "multiplier": 0.5
    },
    {
      "against": "water", <-- if technique earth targets a water monster, then multiplier is 2
      "multiplier": 2
    },
    {
      "against": "wood",
      "multiplier": 1
    }
```
in this way elements will not be hardcoded inside the code, modders can add / remove elements, modders can change multiplier, they can do what they want. It'll make easier to expand elements.

Why? @freshreplicant wanted it, he talked about modules and how we needed it (@Sanglorian too)
(it looks like there are a bunch of modders stopped only by this, honestly I don't know)

my first thought was to create a separate self.elements (eg monster.elements and technique.elements), but at the end it was possible to integrate it inside the original monster.types / technique.types. I needed to fix a couple of things, but I'm satisfied by the result.

### DIFF
--- a/mods/tuxemon/db/element/aether.json
+++ b/mods/tuxemon/db/element/aether.json
@@ -1,0 +1,29 @@
+{
+  "slug": "aether",
+  "types": [
+    {
+      "against": "aether",
+      "multiplier": 1
+    },
+    {
+      "against": "earth",
+      "multiplier": 1
+    },
+    {
+      "against": "fire",
+      "multiplier": 1
+    },
+    {
+      "against": "metal",
+      "multiplier": 1
+    },
+    {
+      "against": "water",
+      "multiplier": 1
+    },
+    {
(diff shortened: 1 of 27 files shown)


kerizane (approved):

freshreplicant:
That's brilliant, thank you for implementing this.

PR #1907, opened 2023-06-05
refactor shape

JaskRendix:
PR refactors shape and it creates a folder in the DB called **shape**, here inside a single, simple and easy to understand JSON called **shapes.json**

Why this? Because @Sanglorian was considering to add new body shapes. This will make everything easier for modders, they can add/remove shapes as well as correct values (speed, ranged, etc.) without being forced to edit the actual code.

Moreover it removes hardcoded things in monster.py

I was forced to add MonsterShape.default as well as this piece inside shape.py:
```
        if slug == MonsterShape.default:
            pass
        else:
            results = db.lookup(slug, table="shape")
```
because if not `test_catch_release.py` and `test_monster.py` were failing (unable to load the table)

### DIFF
--- a/mods/tuxemon/db/shape/shapes.json
+++ b/mods/tuxemon/db/shape/shapes.json
@@ -1,0 +1,128 @@
+[
+    {
+      "slug": "blob",
+      "armour": 8,
+      "dodge": 4,
+      "hp": 8,
+      "melee": 4,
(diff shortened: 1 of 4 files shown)


kerizane (approved):

Sanglorian:
Thank you!

PR #1911, opened 2023-06-06
fix dante and release

### LINKED ISSUE #1908: Shopkeeper scene triggered too soon
Start the game, choosing the first scenario, and go straight to the bins.
Shopkeeper turns up and triggers the dialogue.

This is confusing for players because we haven't met the character yet. The scene should only be triggered after we've seen the previous two scenes.

JaskRendix:
fix #1908 

introduce variable, if the player doesn't talk with Dante, then it doesn't start the bin scene.

fixed a bug related to release, because it didn't update the hook, and if someone was trying to click as much as keys possible to speed up. Like pressing 100x the elevator button thinking it goes faster, then it would result in a crash.

### DIFF
--- a/mods/tuxemon/maps/spyder_paper_scoop.tmx
+++ b/mods/tuxemon/maps/spyder_paper_scoop.tmx
@@ -78,6 +78,7 @@
   <object id="62" name="Talk Dante" type="event" x="16" y="0" width="16" height="16">
    <properties>
     <property name="act1" value="translated_dialog spyder_papertown_danteresting"/>
+    <property name="act2" value="set_variable dantefirst:yes"/>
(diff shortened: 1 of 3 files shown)


kerizane (approved):

JaskRendix:
@Sanglorian with this PR the scene in the bins will not be triggered if you don't talk with Dante inside the Mart.
Obviously if you want to add/remove/change, then feel free.

Sanglorian:
> @Sanglorian with this PR the scene in the bins will not be triggered if you don't talk with Dante inside the Mart. Obviously if you want to add/remove/change, then feel free.

Thanks! Have kept it but set up an alternative option with somewhat different dialogue if the player doesn't talk to Dante.

PR #1920, opened 2023-06-07
change day and night

### LINKED ISSUE #1823: Add a day/night/special event system based on real time for Tuxemon

JaskRendix:
fix #1823

PR adds the change of graphics between day and night.

The option is set as DISABLED at the beginning, but it can be ENABLED, see below the menu options screen, last line:
Screenshot from 2023-06-07 11-02-05
if enabled, this is the results, it's simply a BLUE (0, 0, 128) with a transparency.
Screenshot from 2023-06-07 10-26-36

NB: if ENABLED the change of color will happen only FROM 20:00 TO 04:00. It can be DISABLED any time. Moreover the darkness will touch only the maps OUTSIDE, so all the maps with this `<property name="inside" type="bool" value="true"/>` will stay normal color (eg centers, shops, caves, etc.)

@Sanglorian Spyder is ok. Eventually we can enable it by default at the beginning. Let me know.

@Qiangong2 if these maps are inside: **taba_ba_br_1**, **taba_ba_foyer**, **taba_ba_main** and **taba_ba_passageway_1** then we can add this `<property name="inside" type="bool" value="true"/>` in the tmx files. So we can fix Xero, alternatively I can hide the menu day/night for Xero. Let me know, eventually I have the PR ready for both.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -2654,6 +2654,15 @@ msgstr "Cancel Key"
 msgid "menu_back_key"
 msgstr "Back Key"
 
+msgid "menu_music_daynight"
+msgstr "Day / Night"
+
+msgid "enable"
+msgstr "Enable"
+
+msgid "disable"
+msgstr "Disable"
+
 msgid "menu_music_volume"
 msgstr "Music Volume"
 
--- a/mods/tuxemon/maps/debug.tmx
+++ b/mods/tuxemon/maps/debug.tmx
@@ -42,6 +42,7 @@
     <property name="act4" value="transition_teleport spyder_bedroom.tmx,4,4,0.3"/>
     <property name="act5" value="set_money player,250"/>
     <property name="act6" value="set_variable spyder_starting_money:yes"/>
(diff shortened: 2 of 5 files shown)


Sanglorian:
Simple and effective, looks good to me. If you wanted to go further, there are those night versions of the combat backgrounds. But I don't think that's essential by any means.

JaskRendix:
> Simple and effective, looks good to me. If you wanted to go further, there are those night versions of the combat backgrounds. But I don't think that's essential by any means.

It happens already night/day, but I'm going to sync it with the same hour as soon as it gets merged.

kerizane (approved):

DavidLiang1129:
Thanks for your work on this, hard work!

PR #1924, opened 2023-06-08
additions capturing process

### LINKED ISSUE #1922: Additions to the capturing process
Have been enjoying playing through the game! I've noticed a few things with capturing that it would be good to fix at some point, if possible: 

1. When a monster escapes the tuxeball, a message like "Not this time! The capture failed." would explain what had happened. 
2. When the monster is captured, as well as "Gotcha!", if it could say either: "SPECIES NAME has been put in the kennel" or "SPECIES NAME has joined the team". 
3. If the monster hasn't already been captured or otherwise acquired, it could say: "You add a page for SPECIES NAME to Tuxepedia" and then show the monster's Tuxepedia entry until the player presses enter.
4. This is getting even more ambitious, but would it be possible to give monsters nicknames when they're acquired? Something like "Would you like to nickname your new SPECIES NAME?" and then a keyboard like that for naming the player?

JaskRendix:
fix #1922 

> When a monster escapes the tuxeball, a message like "Not this time! The capture failed." would explain what had happened.

Done

> When the monster is captured, as well as "Gotcha!", if it could say either: "SPECIES NAME has been put in the kennel" or "SPECIES NAME has joined the team".

Done

> If the monster hasn't already been captured or otherwise acquired, it could say: "You add a page for SPECIES NAME to Tuxepedia" and then show the monster's Tuxepedia entry until the player presses enter.

Done, but the player needs to press return.
This is a general rule, because monster info is a state and as all the stats, enter opens (enter), return closes (shift).

> This is getting even more ambitious, but would it be possible to give monsters nicknames when they're acquired? Something like "Would you like to nickname your new SPECIES NAME?" and then a keyboard like that for naming the player?

We can, but I didn't implement it.
It sounds like something other mods don't need and having something popping up after each capture can be annoying.
Eventually we can implement the renaming app for the Spyder campaign.
Let me know.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -1044,8 +1044,17 @@ msgstr "Wrong!"
 msgid "attempting_capture"
 msgstr "Attempting capture..."
 
+msgid "captured_failed"
+msgstr "Not this time! The capture failed!"
+
 msgid "gotcha"
 msgstr "Gotcha!"
+
+msgid "gotcha_kennel"
+msgstr "{name} has been put in the kennel!"
+
+msgid "gotcha_team"
+msgstr "{name} has joined the team!"
 
 msgid "not_implemented"
 msgstr "This feature is not yet implemented."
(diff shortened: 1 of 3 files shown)


JaskRendix:
@Sanglorian

kerizane (approved):

Sanglorian:
Thanks so much!

PR #1932, opened 2023-06-10
statuses (revenge and retaliate)

JaskRendix:
PR adds two new statuses: retaliate and revenge.

~~Remember to move JSONs inside the correct folder after #1998 has been merged.~~
~~Remember to change Tech in Condition.~~

@Sanglorian 
when I was bored I was able to create two additional statuses from [here](https://wiki.tuxemon.org/Other_technique_ideas):
- Retaliate: Keep track of all damage you take between entering this state and next doing damage. You do additional damage equal to damage taken;
- Revenge: The next time you are attacked, your enemy takes the same amount of damage that you do. You heal as much damage as you dealt. (it was Revenge Stance, but I used only Revenge);


Sanglorian:
These are cool, thanks! Have added them to the wiki. May take me a little while to find suitable icons for them though.

JaskRendix:
> May take me a little while to find suitable icons for them though.

No problem

Sanglorian:
Thanks again for this. I've uploaded an icon for each new condition.

bitcraft:
why are these combined into one status?  why not a status for each one?

JaskRendix:
You mean same effect? To avoid pointless duplicates, but if you want that I separate just let me know.

bitcraft:
there are two statuses here though?  revenge and retaliate?

JaskRendix:
Ah ok, you want 1 PR for each, no problem

bitcraft:
sorry if it wasn't clear, but i was referring to the single effect called "additional" which managed both conditions, revenge and retaliate.  i don't see why they need to be merged into a single file, when they are two different things.

JaskRendix:
I didn't reply before because I was totally demotivated.
Moreover your 20 lines text inside combat.py didn't help and I took that personally.

Regarding the PR I split revenge and retaliate.

JaskRendix:
I need to wait #2000, #2003 and #2004, solve the conflicts and then it'll be ready.

bitcraft:
> I didn't reply before because I was totally demotivated. Moreover your 20 lines text inside combat.py didn't help and I took that personally.

sorry that you feel that way.  the additional comments in that file were not directed at you personally.  but i feel they are needed because i want to make sure that files like combat.py do not become too entangled/complicated.  while i am empathetic and understand that its way easier to program logic directly into combat.py, we have a system to ensure the game is extensible with json files and python plugins.  by directly implementing logic in combat.py, its no longer possible to tweak the game or customize it with the json files or python plugins.

one goal of the project is to enable non-programmers to implement their own custom moves, moves sets, status, etc.  you know combat.py is complicated and would be difficult for casual modders to change.  so if we can keep it simple, and move all logic outside of that file, it will make the whole game easier to extend and modify for a range of people with different abilities.

so im sorry that you took it personally, but i did not review the pr that merged that code, and i would have worked with you to move the changes out of combat.py.  but they are there now, and i need the comment in the file to prevent it from growing too much.  because if we keep adding moves/status/etc to it directly, all the work to support plugins and json files will be lost and the file will grow difficult to modify.  not to mention, that it would be more difficult to adjust the game without coding.

JaskRendix:
no problem, I thought about it in the last two months, that's why I opened split condition, so we move out these things. **split condition** is only the first step, but it helps to remove a bunch of hardcoded effects.

JaskRendix:

bitcraft (approved):

PR #1936, opened 2023-06-10
refactor run + forfeit

JaskRendix:
PR proposes to use "menu_run" technique as a technique.

It modifies the msgid combat_player_run because it'll appear below:
```
MONSTER used RUN
Run away!
```

This could be expanded in the future for wild monsters (eg running away / escaping if the capture doesn't succeed), but it's not the purpose of this PR.


Sanglorian:
Good idea! For most monsters, I think they should never run. But there should be some timid or rare monsters that have a certain run rate. Masknake, Altie and Ziggurat would be good options.

JaskRendix:
@Sanglorian yeah, it can be an idea, something like:
if you try to catch a monster and the tuxeball breaks, then some wild monsters (eg Altie, etc.) can run/escape.

Sanglorian:
Great! To test it out, could we have:

* Ziggurat run away on 10% of all rounds and 
* Altie run away 33% of the time that a capture attempt fails?

Then we can take our time figuring out what other monsters should have the effect too.

JaskRendix:
This is ready to go. It doesn't contain anything related to run for wild monsters.

bitcraft:
it looks like players are now able to run away no matter what?  previously there was a formula applied to determine if a player could escape?

JaskRendix:
Yes, but the formula was moved inside the effect. But I understand your point. I'm going to close it. No issue.

JaskRendix:
Standard escaping is false.

There is the formula (below) and if the escape value is higher than the random.random, then the player can run.
```
        def escape(level_user: int, level_target: int, attempts: int) -> bool:
            escaping: bool = False
            escape = 0.4 + (0.15 * (attempts + level_user - level_target))
            if random.random() <= escape:
                escaping = True
```
I moved the formula inside the run.py (effect).

In this way "run" will be similar to a technique.

bitcraft:
is this still a draft?

JaskRendix:
Yeah, I need to wait #2000, #2003 and #2004, solve the conflicts and then it'll be ready.

JaskRendix:
added the refactor of forfeit too, same approach.

I'm studying how to implement the run for the NPC monster (wild) - eg. after breaking the ball (failed capture), but this will be defined in another PR.

bitcraft (approved):

Issue #1916, opened 2023-06-06
font issue (i)


JaskRendix:
Since the font has been updated with the new male/female symbol.

it happens when the i is followed by L and G, but only lowercase `il` and `ig`, but I'm not sure, because by testing all the letters of the keyboard (eg ie, ir, it, etc), nothing happens (lower and uppercase).
![Screenshot from 2023-06-06 13-23-58](https://github.com/Tuxemon/Tuxemon/assets/64643719/62622524-f54a-421f-b82d-4598e03c0987)
the correct string as per PO:  `"Just fill in this form." `
it happens with `"{target} will keep fighting."`

maybe it happens to someone else too

Sanglorian:
How odd! Sorry about this -- I didn't think I touched any other settings, but it must be related to my additions to the font. I'll do some experimenting when I get a chance.

Sanglorian:
Just reporting in that I don't get this error - I installed the latest flatpak on Linux Mint 21, and both Ls show up for me in "Just fill out this form." 

Sorry that's not very helpful!

JaskRendix:
Don't worry. I realized it's a an issue with my system.

freshreplicant:
> Don't worry. I realized it's a an issue with my system.

Would you be able to elaborate on what caused it and if/how it was resolved? I think I ran into something similar with the Git version:

![TuxemonTextBug](https://github.com/Tuxemon/Tuxemon/assets/71570200/a150d1fe-4410-4080-8a7a-8301198a063d)
![TuxemonTextBug2](https://github.com/Tuxemon/Tuxemon/assets/71570200/24a504c6-b25d-4177-b818-55e74cfc2694)

JaskRendix:
> Would you be able to elaborate on what caused it and if/how it was resolved?

I didn't solve. I closed it out of frustration because I can't control everything. So I kept playing with the wrong setting hoping to find the solution. It was in my pipeline.

From my understanding it has been saved with the wrong options.
I'm trying to recall the right ones, but until now without success.

The old one was possible to open with the text editor (Linux), while the new one is impossible.

JaskRendix:
I'm so disappointed, but at least I fixed it.

PR #1937, opened 2023-06-10
fix font

JaskRendix:
PR saves again the font to fix the issue #1916
still unable to open it with text edit as I was accustomed to do before... but at least the issue is fixed.


kerizane (approved):

Sanglorian:
Sorry about this, @JaskRendix ! I didn't think I changed any options when I edited the font, but looks like it messed things up. Thanks for fixing it!

PR #1941, opened 2023-06-10
fix config

### LINKED ISSUE #1910: Cannot change locale through tuxemon.cfg
**Describe the bug**
This bug was discovered by my friend, I have not tested it at present, he uses the latest development version downloaded at 2023/6/5 18:35 Beijing time
tuxemon.cfg's locale Settings seem to reset to default when opening the game
**To Reproduce**
Run the game so that it generates tuxemon.cfg, then change the value of "locale" to something other than en_US (such as eo) and then save and enter the game again, the game is still in English, check tuxemon.cfg again, the value of "locale" has been reset back to en_US

**Expected behavior**
tuxemon.cfg should not be reset and the modifications should be retained
**Screenshots**
None.

 - OS: Windows 10 21H2
 - Python Version 3.8

**Smartphone (please complete the following information):**
 - Device: N/A

**Additional context**
He used my "Tuxemon Windows Installer" to install Tuxemon. Could it be a problem? But all it does is silently install Python and dependencies for the user via a Windows batch file

JaskRendix:
PR could fix #1910 

I tested with my native language
before the PR, it was impossible to change language.
after the PR, I succeeded.

Some issues:
```
  File "/home/robespierreni/Tuxemon/tuxemon/config.py", line 32, in __init__
    resolution_x = cfg.getint("display", "resolution_x")
  File "/usr/lib/python3.10/configparser.py", line 819, in getint
    return self._get_conv(section, option, int, raw=raw, vars=vars,
  File "/usr/lib/python3.10/configparser.py", line 809, in _get_conv
    return self._get(section, conv, option, raw=raw, vars=vars,
  File "/usr/lib/python3.10/configparser.py", line 804, in _get
    return conv(self.get(section, option, **kwargs))
  File "/usr/lib/python3.10/configparser.py", line 782, in get
    d = self._unify_values(section, vars)
  File "/usr/lib/python3.10/configparser.py", line 1153, in _unify_values
    raise NoSectionError(section) from None
configparser.NoSectionError: No section: 'display'
```
it's something completely new for me

### DIFF
--- a/tuxemon/config.py
+++ b/tuxemon/config.py
@@ -26,8 +26,7 @@ class TuxemonConfig:
 
         # update with customized values
         if config_path:
-            temp = configparser.ConfigParser()
-            temp.read(config_path)
+            cfg.read(config_path)
 
         # [display]
         resolution_x = cfg.getint("display", "resolution_x")


kerizane (dismissed):

JaskRendix:
@kerizane it's failing some tests

JaskRendix:
@NeonFlames it seems like it's not possible to change values inside the Tuxemon.cfg (eg the language).
I tried this approach, it looks like working (I tested), but it's failing some test (section display is missing).

Do you have any idea on how we can proceed?

kerizane:
The bug you are trying to fix was caused by https://github.com/Tuxemon/Tuxemon/pull/1887

You need to have the default config as the base, and overwrite it with any items that are in the user's config.

NeonFlames:
Ah forgot to read the file, whoops, doing
```py
cfg.read(config_path)
```
instead of using a temp var would fix the issue I think.

NeonFlames:
I'll try to fully test my commits in the future so as to avoid the issues that my prior commits have caused.

JaskRendix:
> instead of using a temp var would fix the issue I think.

You are right.

> I'll try to fully test my commits in the future so as to avoid the issues that my prior commits have caused.

It happens

kerizane (approved):

PR #1942, opened 2023-06-11
sold out (shop)

### LINKED ISSUE #430: Add npc/player inventory checks, so that vendor inventory can be replenished when map loads
Part of #421 
========

## Add npc/player inventory checks, so that vendor inventory can be replenished when map loads

When implemented, a new game event can be used to 'replenish' store inventories by resetting them to what is defined in JSON.

Also, a single item vendor could check for empty inventory and say "Sorry sold out old chap" when his item is sold out.

It could also be used, for example, after every store transaction to make sure products never run out.  Or for vendors who only sell items once a season.  Or some other thing.  Not a high priority.

JaskRendix:
PR fix #430

when it's sold out, the player can exit and reenter in the store to find it replenished.

@kerizane it was the only thing missing

Screenshot from 2023-06-11 10-37-52

I added the following feature:

> Or for vendors who only sell items once a season. Or some other thing. Not a high priority.

now modders can set variables to make appear items inside the shops:
```
    {
      "item_name": "tm_blossom",
      "price": 1000,
      "cost": 200,
      "inventory": 20,
      "variable": "season:spring"
    },
```
the new optional field is called variable and accept strings in this format xxx:yyy, exactly like the one we use for set_variable, in this case this item will appear inside the shop only in spring, but it can be done with any variable.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -2508,6 +2508,9 @@ msgstr "Release"
 msgid "shop_buy_free"
 msgstr "FREE!"
 
+msgid "shop_buy_soldout"
+msgstr "Sold Out!"
+
 #Release notifications
 msgid "cant_release"
 msgstr "Can't release only Tuxemon in party."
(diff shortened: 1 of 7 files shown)


kerizane:
It's not essential that we show sold out items Vs the previous behaviour where they'd disappear. Do you have a preference? @Sanglorian what do you think?

kerizane:
Nice idea regarding the variable, by the way.

JaskRendix:
> Do you have a preference? 

No, I don't. If @Sanglorian prefers the items disappearing, then I'm going to revert.

Sanglorian:
This seems really useful, particularly that variable. I can imagine items becoming available once certain quests have been achieved.

kerizane (approved):

PR #1955, opened 2023-06-14
fix hp bar behavior

### LINKED ISSUE #1952: HP bar behaviour in multi Tuxemon fights
**Describe the bug**
When you fight an opponent with multiple Tuxemon and then knock one of them out, just before the opponent's next Tuxemon appears, it seems like a duplicate of your monster's UI slides into place over the existing one, almost as if you had swapped out Tuxemon (see attached .webm file). In some instances, I also think the HP values looked different, but that might have been the HP bars overlapping briefly.

Upon closer inspection, it seems my starter leveled up? It was not however immediately apparent, as it just looked like the XP bar was filled a tiny bit more and no sound or message alerted me to the fact.

**To Reproduce**
1. Fight Billie for the second time.
2. Defeat her starter.
3. Watch your monsters HP bar and other UI elements.

**Expected behavior**
Your monster's GUI would stay where it is unless you swapped them out for another in your party (even if leveling up).

-Endeavour OS Linux
-KDE Plasma version: 5.27.5
-Kernel: 6.3.6-arch1-1 (64-bit)
-Python 3.10
-Graphics Platform: Wayland

[hp_issue.webm](https://github.com/Tuxemon/Tuxemon/assets/71570200/98b46c38-d595-403e-8dc2-517faf41e956)

JaskRendix:
fix #1952 

@kerizane you marked it as important and liked it, but this was implemented for a specific reason #1581

I'm looking into to find another way to update the level, but feel free to raise an equal important issue

_before there was no update in combat (if the monster leveled up) and the only way to know it: switch out and switch in or at the end checking the monster menu -> now the hub will refresh, the number lv XX will update;_

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1284,21 +1284,6 @@ class CombatState(CombatAnimations):
                         if mex:
                             message += "\n" + mex
                             action_time += len(message) * letter_time
-                        # updates hud graphics player
-                        if len(self.monsters_in_play_human) > 1:
-                            self.build_hud(
-                                self._layout[self.players[0]]["hud0"][0],
-                                self.monsters_in_play_human[0],
-                            )
-                            self.build_hud(
-                                self._layout[self.players[0]]["hud1"][0],
-                                self.monsters_in_play_human[1],
-                            )
-                        else:
-                            self.build_hud(
-                                self._layout[self.players[0]]["hud"][0],
-                                self.monsters_in_play_human[0],
-                            )
                 if winners in self.players[0].monsters:
                     m = T.format(
                         "combat_gain_exp",


JaskRendix:
@freshreplicant feel free to point out what doesn't work as soon as it gets merged; I'm so sorry that I implemented it, I didn't know it would create such an important issue. I did because I wanted to fix an existing issue.

JaskRendix:
update the original nr of PR, the previous one was wrong

JaskRendix:
@kerizane are you waiting that I find the solution before merging?

kerizane:
Thanks, but the current behaviour is better than the bug that you fixed, so no point merging this PR.

JaskRendix:
@kerizane sure, thanks

freshreplicant:
Sincere apologies, I was not aware of #1581. 

I didn't mean to cause any trouble with the issue and I didn't mark it as important or consider it especially important. I'm not a significant contributor to Tuxemon and I don't have any say in the project. Feel free to take on board or dismiss my issues like anyone's, or correct me when I make a bad suggestions. I don't expect any of my ideas to be implemented, or anybody to dedicate their time to implementing something they don't want to, this is an open source hobbyist project after all. I won't be upset.

I'm not good enough at Python yet and don't understand the codebase enough to make any PRs or significant contributions.

Sanglorian:
> I didn't mean to cause any trouble with the issue and I didn't mark it as important or consider it especially important. I'm not a significant contributor to Tuxemon and I don't have any say in the project. Feel free to take on board or dismiss my issues like anyone's, or correct me when I make a bad suggestions. I don't expect any of my ideas to be implemented, or anybody to dedicate their time to implementing something they don't want to, this is an open source hobbyist project after all. I won't be upset.

Please do keep giving your thoughts, @freshreplicant -- it's good to have lots of feedback from the community, and I think you've picked up on some good things to discuss and potentially change.

PR #1956, opened 2023-06-14
fix weird menu behavior

### LINKED ISSUE #1953: Inconsistent menu behavior
**Describe the bug**
Different menu items  have different behaviour. Tuxemon list returns back to menu, but other items returns to the game

**To Reproduce**
1. Press Esc
2. Select menu item, press enter
3. Press Esc again

**Expected behavior**
All menu items returns at the same place

**Screenshots**
tuxemon_menu

JaskRendix:
fix

### DIFF
--- a/tuxemon/menu/events.py
+++ b/tuxemon/menu/events.py
@@ -12,7 +12,7 @@ _EVENT_MAP: Final = {
     buttons.DOWN: pygame.event.Event(pygame.KEYDOWN, key=pygame.K_DOWN),
     buttons.LEFT: pygame.event.Event(pygame.KEYDOWN, key=pygame.K_LEFT),
     buttons.RIGHT: pygame.event.Event(pygame.KEYDOWN, key=pygame.K_RIGHT),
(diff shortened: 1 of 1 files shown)


pincher2012:
Now I can't exit Tuxopedia, player, phone, options, or close the menu itself using the Escape key

JaskRendix:
@pincher2012 exactly, that's what you wanted, right?
Escape key was set to open the menu, but at the same time having the same function as the shift button (return too).
Alternatively we can find another key combination to open the menu (eg shift below enter)

but again, I have no idea, I never used escape as return.

pincher2012:
@JaskRendix It's not exactly what I wanted. I didn't know about the shift button. It was convenient for me to use the esc button. The problem I was talking about is that different menu items return the player to different places. I'm expecting a breadcrumb effect. For example, when I enter Tuxepedia through the menu, I expect to return back to the menu. Currently, only the Tuxemon menu item has this effect. That's what I meant. I'm sorry for not describing the problem clearly enough.

JaskRendix:
@kerizane thanks

JaskRendix:
@pincher2012 I'll find a solution, I hope

PR #1964, opened 2023-06-15
fix npc_face (event action) - player

JaskRendix:
Right now if a modder uses `npc_face knight1,player` the knight moves his face towards the player.

It's good, but what's missing is reciprocity, so I added a couple of lines.

Now by using `npc_face knight1,player` not only the knight moves his face towards the player, but the player sprite moves the face towards the NPC too.


bitcraft:
i think this might not work as expected.  if this change is made, it would no longer be possible to just have some npc face another.  i would suggest making an option to make them both face (default false), making a new function which will cause both to face eachother, such as `face_each_other(npc1, np2)` or something.

JaskRendix:
Ok

PR #1965, opened 2023-06-15
fix menu

### LINKED ISSUE #1953: Inconsistent menu behavior

JaskRendix:
fix #1953 after the explanation here: https://github.com/Tuxemon/Tuxemon/pull/1956#issuecomment-1590961746

the issue: worldstate was using **replace_state** instead of **push_state**

this means that when someone clicks on save, then the menu doesn't disappear. 
Screenshot from 2023-06-15 11-30-02
If this is "acceptable", then it's done.

### DIFF
--- a/tuxemon/states/world/world_menus.py
+++ b/tuxemon/states/world/world_menus.py
@@ -58,7 +58,7 @@ class WorldMenuState(PygameMenuState):
         self.animation_offset = 0
 
         def change_state(state: str, **kwargs: Any) -> Callable[[], object]:
-            return partial(self.client.replace_state, state, **kwargs)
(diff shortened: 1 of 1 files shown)


pincher2012:
LGTM

There are problems with sliding the menu, but it must be solved as part of a separate issue

kerizane (approved):

PR #1970, opened 2023-06-22
fade out status icons when monster is removed from combat

### LINKED ISSUE #1966: Condition symbol stays behind when monster is defeated
**Describe the bug**
When a monster is eliminated, its condition symbol stays behind while the rest of the monster card leaves. 

**Expected behavior**
The condition symbol disappears at the same time. 

Not a big deal, but something I've noticed a couple of times.

bitcraft:
fixes #1966


kerizane (approved):

kerizane:
Nice to see you!

Issue #1972, opened 2023-07-08
The game crashes when the "Player" option is opened in the menu


DavidLiang1129:
**Describe the bug**
I tried to open the "Player" option in the menu to see the player information, but the game immediately crashed when I tried to open the "Player" option

**To Reproduce**
Start the game with Spyder, then press Esc to open the menu, select the "Player" option and press Enter, after which the game will crash immediately

**Expected behavior**
"Player" should be opened normally
**Screenshots**
No, but I do have an error log (it's long, so I'll send it below).

**Desktop (please complete the following information):**
 - OS: Windows 7 SP1 x64
 - Python Version [e.g. 22]

**Additional context**
I saw a lot of errors at the start of the game (when I double-clicked run_tuxmonm.py), but the game continued to run, and I'll bring up a new lssue on that later

DavidLiang1129:
Error log:
[...]

Issue #1971, opened 2023-07-08
Unable to get the first Tuxemon in Spyder


DavidLiang1129:
**Describe the bug**
The game crashed before I tried to pick up my first Tuxemon (before the Tuxemon selection box appeared)
**To Reproduce**
Start the game with Spyder, go to the trash can in the back yard of the store to trigger the story, after the NPC has left, go to the trash can and press Enter, after a dialogue, the game crashes

**Expected behavior**
I should be able to claim my first Tuxemon

**Screenshots**
No, but I do have an error log (it's long, so I'll send it below).

**Desktop (please complete the following information):**
 - OS: Windows 7 SP1 x64
 - Python Version 3.8.8

**Smartphone (please complete the following information):**
 - Device: N/A

**Additional context**
I saw a lot of errors at the start of the game (when I double-clicked run_tuxmonm.py), but the game continued to run, and I'll bring up a new lssue on that later

DavidLiang1129:
Error log:
[...]

PR #1975, opened 2023-07-10
Initialize MonsterModel.sprites to None

### LINKED ISSUE #1973: I saw a lot of errors at the beginning of the game
I don't know what these errors are, but some time ago I switched to Windows 7 x64 on my PC, and then I ran Tuxemon from source code following the README.md method. But I saw a lot of errors at the beginning of the game (double clicking run_tuxemon. py) (and probably related to #1971 #1972). I will post the error log below

devdanzin:
Somehow `MonsterModel.sprites` is failing validation (probably due to pydantic being 2.0.2 here, while it probably is 1.x.x for people that don't have this issue). This causes the monster db to be empty, causing all sorts of issues. 

This fixes #1973, #1972 and #1971. 

Edit: I've confirmed that with pydantic==1.9.2 this problem doesn't exist. So the fix could either be this PR or updating `requirements.txt` to avoid pydantic>=2.0.0.

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -385,7 +385,7 @@ class MonsterModel(BaseModel):
     )
 
     # Optional fields
-    sprites: Optional[MonsterSpritesModel]
+    sprites: Optional[MonsterSpritesModel] = None
     shape: MonsterShape = Field(..., description="The shape of the monster")
     types: Sequence[ElementType] = Field(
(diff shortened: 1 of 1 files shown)


bitcraft:
Looks good to me!  The typing annotation is `Optional`, so setting it to none looks correct.

bitcraft (approved):

Issue #1967, opened 2023-06-18
Couple of minor points about combat information


Sanglorian:
**Is your feature request related to a problem? Please describe.**
Just a few minor issues relating to how combat information is displayed: 

* The message about XP gained should come after the message about the monster being knocked out. 
* When a monster avoids capture, is it possible to not go on to the next stage of the battle until the player has clicked Enter/Shift? At the moment, the capture fails and the monster that's the target of the capture immediately attacks, making it a little hard to follow what's happening. 
* It's fantastic that the Tuxepedia page about the captured monster pops up after the capture occurs - it really adds something. Is it possible to make Enter/Shift make the page disappear, not just escape? 

Thanks for considering these!

Grimmys:
> It's fantastic that the Tuxepedia page about the captured monster pops up after the capture occurs - it really adds something. Is it possible to make Enter/Shift make the page disappear, not just escape?

Looks like pressing "Shift" is already making the page disappear.
Also, these controls look the same in any other place a Tuxepedia page is opening.
Meaning that adding "Enter" as an alternative key for closing the page after the capture, would also likely add it for the case where the page is shown directly from the Tuxepedia entry menu, or for example when the player is checking the Tuxemons that are in the bins.

I have no issue with this though, I think it could be fine to have the "Enter" key to close the page in every cases.

PR #1987, opened 2023-08-17
Reorder end of combat messages & improve animations engine

Grimmys:
This PR addresses first bullet point of #1967 

XP message is now between "Opponent is KO message" and "win message".

But in order to do this properly, I reworked the battle animation system to fix two flaws:

1. Unfinished `Task` were not preventing the game to move to the next combat phase, which was sometimes resulting to animation tasks planned during previous turn to run during next turn.

A method to check if a task is finished has been added (` def is_finished(self) -> bool`) and is now called for every task in `CombatState._animations`: if any task is not finished, then the game is not moving to the next phase.

2. No delay before moving to the next phase was added if some delay was still there when calling `suppress_phase_change`. As a result, let's imagine 1 second was left before moving to the next phase, and a new animation requires 10 seconds ; no time was added and the animation was closing after 1 second.

A method to reset the delay before starting the task has been added (`def reset_delay(self, new_delay: float) -> None`) and is called inside of `suppress_phase_change` to make sure the boolean for running animation is not going to change back to False before the additional animation is finished.

**Next step**

If think the battle animation system is still clunky, I suggest to introduce a queue of animations as a next step to prevent new text animation to override the currently running one (sometimes, next message is displayed even if the previous one didn't finish rendering).

**Notes**

This is my first contribution to the project, so I hope I followed all rules correctly, don't hesitate to tell me if anything looks wrong so I could avoid doing it again.
I also added tests for `Task` class: 12 in total, because I love tests. 😄

### DIFF
--- a/tests/tuxemon/test_animation.py
+++ b/tests/tuxemon/test_animation.py
@@ -1,0 +1,106 @@
+import unittest
+from typing import Callable
+from unittest.mock import Mock, call
+
+import pygame
+
+from tuxemon.animation import Task
+
+DEFAULT_INTERVAL = 1.0
+
+
+class TestAnimation(unittest.TestCase):
+    def setUp(self) -> None:
+        self.mock_callback = Mock(spec=Callable)
+        self.other_mock_callback = Mock(spec=Callable)
+
+    def test_task_run_callback_after_interval(self):
+        task = Task(self.mock_callback, DEFAULT_INTERVAL)
+
+        task.update(0.5)
+        self.mock_callback.assert_not_called()
+
+        task.update(0.5)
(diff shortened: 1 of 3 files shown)


JaskRendix (approved):

bitcraft:
great changes!  i like that you leveraged the animation module and added tests.  welcome to the team!

bitcraft (approved):

Sanglorian:
Thanks @Grimmys , I'm not a programmer so the technical details are beyond me but this sounds like an important contribution. Please let me know if you have any questions about the stories, battling rules, etc., I'm here to help!

PR #1992, opened 2023-08-20
lion mountain (spyder) + miner sprites

JaskRendix:
updates Lion Mountain map (Spyder) as per @Sanglorian instruction (wiki)
- npcs dialogues
- npcs monsters
- new miner sprites (many miners in the map, so there is more variety)

still missing: events summit, something else?

@Sanglorian the player receives Chromeye from the enforcer, the level was missing, so I put lv 15


kerizane (approved):

Sanglorian:
Thanks so much, this is a great map and I think players will particularly enjoy the change in scenery! And glad to have you back, @JaskRendix

Issue #1989, opened 2023-08-19
Crash when using Boulder attack with Rockitten


me-self:
**Describe the bug**
A crash when using Rockitten's boulder move

**To Reproduce**
1. Start a battle with Rockitten
2. Use the `Boulder` attack

**Expected behavior**
The attack takes place as with any other attack, and the game continues as usual.

**Screenshots**
![Screenshot from 2023-08-18 18-52-36](https://github.com/Tuxemon/Tuxemon/assets/115949508/f2a9ff23-5bf0-48d7-a84c-6f483d363c59)

**Desktop (please complete the following information):**
 - OS: Vanilla OS
 - Python Version: Python 3.10.7

**Additional context**
 - This error occurs in the the Flatpak package
 - the issue appears to be a case of invalid argument count for this move's function:
 `TypeError: Technique.hardshell() takes 2 positional arguments but 3 were given`

Grimmys:
I tried but cannot reproduce the issue, the technique is working fine on my side.

Also, `Techinque.hardshell` is not something that can be find anymore in the code source.
It looks like it was indeed a method before this commit: [6cce3c86caf282a8856248ac2513c31f12ed520b](https://github.com/Grimmys/Tuxemon/commit/6cce3c86caf282a8856248ac2513c31f12ed520b), that was pushed in 2022.

I'm assuming your are just using a very old build (the latest release maybe?) and the issue is already fixed.

me-self:
Are you using the flatpak from flathub?
Yeah, the flathub flatpak was last updated... `Changes in version 0.4.34
about 1 year ago`

freshreplicant:
As far as I know the flatpak should update with the next release, so #1928 is technically related here. @Murmele may be able to confirm.

Murmele:
@freshreplicant yes exactly the flathub manifest will be updated to a new version once a new release was created

Murmele:
Is it reproducable with the nightly: https://github.com/Tuxemon/Tuxemon/releases/tag/latest ?

Grimmys:
All depends on whether the flatpack from the nightly is indeed taking build from 2 days ago or from 2 years ago.

For this kind of issue related to outdated build, we could be logging in the console at the start up of the game the git hash of the latest commit, and then add it as a field in the "bug issue template".
So we could more easily spot on which game version the bug is reproducible.

me-self:
[...]

PR #1993, opened 2023-08-22
get current git hash

JaskRendix:
https://github.com/Tuxemon/Tuxemon/issues/1989#issuecomment-1686722935 by @Grimmys 

I read your reply, did you mean something like this? Now at the start it'll print:

```
robespierre@robespierre-M11AD:~/Tuxemon$ python3 run_tuxemon.py
pygame-ce 2.2.1 (SDL 2.26.4, Python 3.10.12)
pygame-menu 4.4.3
Git Hash: 163ad3a86
Enabling logging of all modules.
[2023-08-22 08:30:39,119] tuxemon.rumble - ERROR - No rumble backends available.
```
~~I put also the git hash in the menu too, because some people can never see the terminal~~
~~https://github.com/Tuxemon/Tuxemon/assets/64643719/087d4278-074c-4a7d-a826-bc8e140abbbc~~

### DIFF
--- a/tuxemon/log.py
+++ b/tuxemon/log.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 import logging
 import os
+import subprocess
 import sys
 import time
 import warnings
@@ -35,6 +36,17 @@ def configure() -> None:
         warnings.filterwarnings("default")
 
         for logger_name in config.loggers:
+            # Get the current git hash
+            try:
+                githash = (
+                    subprocess.check_output(["git", "describe", "--always"])
+                    .strip()
+                    .decode()
+                )
+                print(f"Git Hash: {githash}")
+            except:
+                print("No Git Hash")
+
             # Enable logging for all modules if specified.
             if logger_name == "all":
                 print("Enabling logging of all modules.")


Grimmys:
Yes, it is what I was expecting, nice job!

My only concern could be that I'm not sure if it should be that visible on the screen... I think it would be better as an information displayed in a corner (bottom left?) or even better: load from the configuration a value to check if it should be displayed or not (and it should be False by default for releases).
I think it should be this way because regular players don't care about this kind of information ; most of them don't even know what is Git and might be confused.

tuxemon/states/start/__init__.py
```diff
@@ -100,6 +101,19 @@ def exit_game() -> None:
...
+            .strip()
+            .decode()
+        )
+        git_label = menu.add.label(
+            title="Git Hash: " + githash,
+            font_size=15,
+            label_id="githash",
+            float=True,
+        )
+        assert not isinstance(git_label, List)
```
Grimmys: Is this assert really necessary? In which cases do we think git_label could be a List?

JaskRendix: I put the assert to avoid the typehint #1211 before we got tons of typehints (mypy strict)

JaskRendix:
It makes sense, I'm going to remove the hash from the menu. It can led to confusion.

tuxemon/states/start/__init__.py
JaskRendix: Just to be complete, without the assert mypy --strict flags me with:
`tuxemon/states/start/__init__.py:115: error: Item "List[Any]" of "Union[Any, List[Any]]" has no attribute "translate"  [union-attr]`

Grimmys: Ok yeah I didn't check the typing of the Label, the method generating it is having the annoying Union[Label, List[Label]] typehint thus leading to this issue.

But wouldn't adding the specific typehint to the variable solve the issue as well?

Like
```py
git_label: Label = menu.add.label(
    title="Git Hash: " + githash,
    font_size=15,
    label_id="githash",
    float=True,
)
```

Grimmys: Oh anyway you removed the displaying of the Label in the UI, so no point about talking anymore 👍

JaskRendix: No worries, I'm going to test it in locale! Thanks!

JaskRendix: updated mypy (last version):
`error: Incompatible types in assignment (expression has type "Any | list[Any]", variable has type "Label")  [assignment]`
this is what mypy produces with `git_label: Label = menu.add.label(` and by removing the assert

tuxemon/log.py
```diff
@@ -35,6 +36,14 @@ def configure() -> None:
         warnings.filterwarnings("default")
 
         for logger_name in config.loggers:
+            # Get the current git hash
```
kerizane, line 39: If somebody somehow plays the game without git installed will this crash?

Best to wrap it in a try/catch.

kerizane (changes requested):

tuxemon/log.py
JaskRendix: it makes sense, what do you want in the exception? I put a "No Git Hash", let me know.

Grimmys: Yes really great catch.
I think we can add something like "Do you have Git installed?".

Grimmys: Wait wait wait, just going back to this piece of code: the computation of the githash should be **inside** of the try/except block, because that's this computation that might fail (not the print operation).

kerizane (approved):

kerizane (approved):

kerizane (approved):

PR #1996, opened 2023-08-25
4 sprite "granny"

JaskRendix:
PR adds 4 sprites for **Granny** (until now only 1, now 5)


bitcraft:
It looks like at least one map, sphalian town, came from a different mod and renaming it would break that mod.  i have to look into this.  some advice, please ask around in the discord or in the issues before renaming maps.  we have 2 campaigns sharing the same files and renaming things could break them.  when in doubt, copy, don't rename.

JaskRendix:
Deleted everything, I left only the granny spyder. If I feel like wasting more time, then I'll reopen a new PR with what I deleted, but not only before you look into. Let me know.

Sanglorian:
I think keeping all your work should be possible, @JaskRendix - just as long as you also leave a copy of the old Sphalian map in the folder as well. 

I hope you do decide to keep Lion Mountain and Sphalian Town at some point, I think they're both great additions to the Spyder campaign.

bitcraft (approved):

bitcraft:
asking before doing sweeping changes is the opposite of wasting time, in fact, because it reduces the risk of doing work which cannot be accepted.  the changes looked great and i would be happy to approve them once they do not break assets which belong to the xero campaign.

PR #1997, opened 2023-08-26
fix npc faces (spyder)

JaskRendix:
In a couple of events the NPC/Player wasn't looking in the right direction.


Sanglorian:
Thank you! Little things like this really help the player experience.

kerizane (approved):

PR #1998, opened 2023-08-27
split condition

### LINKED ISSUE #1968: Diehard not gained when monster is already on 1 HP
**Describe the bug**
I made Diehard too powerful/annoying. Currently, a monster can gain Diehard even if it's already on 1 HP, which allows a monster to avoid getting knocked out multiple times in a row. 

In retrospect, this is too much. Is it possible to change Diehard so that a monster on exactly 1 HP does not gain Diehard even if it normally would? Perhaps the message, "NAME is too injured to gain Diehard" could play?

JaskRendix:
PR split conditions and techniques. Two different files, two different structures.

The majority of additions/deletions are files moved in a different folder + the creation of new files (eg diehard.py, etc), because each effect will have a single file instead of **default.py**, this will guarantee the possibility to remove the files related to the effects and play without issues (the purge of the remaining default.py will follow in another PR).

PR contains:
- **effects** for conditions/statuses have been moved in the new folder (only the conditions related to statuses/conditions);
- **conditions** for conditions/statuses have been moved in the new folder (only the conditions related to statuses/conditions);
- anticipated from #1932 **drain** is split between **burnt** and **poisoned** (single file for each effect);
- anticipated from #1932 **bounce_back** is split between **feedback**, **prickly** and **elemental_shield** (single file for each effect);
- regarding conditions JSONs files, the condition files will maintain the structure, the only differences are: **tech_id** will be **cond_id**, while **use_tech** will be **gain_cond**. Moreover it has been added the numeration of the conditions as per https://wiki.tuxemon.org/Category:Condition (hardshell 1, poisoned 2, etc);
- added fields **repl_tech** and **repl_item** in condition.py (both optional str, default none), so having the **repl_tech**/**repl_item** will automatically signal that the condition reacts to a technique; both labels correspond to the columns **response to technique use** and **response to item use** from https://wiki.tuxemon.org/Category:Condition;
- added field **steps** in condition.py (float) which will be saved (eg the player has done 69 steps and during a battle a monster faints, it'll be saved as fainted,69 steps, this will be useful in following PR to allow status like poison to keep draining energy even after the battle);
- added field **phase** in condition.py (optional str, default None), in this way we can trigger different messages / techniques / statuses from different phases (eg phase = "check_party_hp", because it's inside the **def check_party_hp**). It avoids to hardcode shit for nothing (previous mistake). It's not the perfect solution, but it'll help in introducing another mechanism;
- regarding the upgrade of the old savegames (savegames with a fainted monster), the addition inside **player_defeated** (see below) will automatically update the status of the fainted monster. Notice that **player_defeated** is present in almost all the maps (both scenario) because of the teleport faint.
- adds new key to condition effects and technique effects (such as **technique**, **condition** and **extra**);
- removes **money**, **healing**, **switch**, **scope**, **wild**, **flinching** , **dozing**, etc. effect hardcoded in combat.py;
- removes **multiple** functions from **combat.py** and adds a generic one that will check the status;
- contains fix #1968;
- fixes small issues related to technique (mostly wrong settings, but nothing crash level);
- renames statuses by removing the **status_** (verbosity and pointless since separated);

the main objective of this PR is splitting conditions and techniques as well as purging the hardcoded stuff from combat.py.

the new structure:
```
condition
- conditions
- effects
```
---
added the following in the condition (action) **player_defeated**:
```
                if mon.current_hp <= 0 and not has_status(mon, "status_faint"):
                    mon.status = list()
                    fainted = Condition()
                    fainted.load("status_faint")
                    mon.status = [fainted]
```
before **player_defeated** was checking if there was at least one monster alive.

now it'll check if a monster with 0 HP has the **status_faint**, if not, then it'll add automatically, consequently if there is at least one monster alive

this will result useful if a monster will remain without HP outside a combat (eg by using an action, by following an event, etc);


bitcraft:
is this still a draft?

JaskRendix:
Yeah, I need to wait #2000, #2003 and #2004, solve the conflicts and then it'll be ready.

JaskRendix:
fix bunch of typehints (mypy --strict):
`tuxemon/states/combat/combat.py:1159: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]`
triggered by `action_time = 0` instead of `action_time = 0.0`

fix warning at the beginning:
`WARNING: Problem on line 569: 'aux mâchoires si puissantes qu\'elles peuvent casser en deux les arbres comme les os."'`
" was missing

JaskRendix:
@bitcraft PR is ready to go

bitcraft (approved):

PR #2000, opened 2023-08-29
Add animation queue

Grimmys:
Following #1987, this PR introduces a text animation queue to be able to schedule displaying of text on the screen without replacing the already running text animation.

Example: at the end of the combat, in case of victory, multiple messages should be displayed (enemy defeated, xp earned, victory), but they should not be animating in parallel (but rather sequentially).

Also, it fixes an issue about fade out animation not running at the end of combat when it should (it was working fine only in case of successful capture). I can map this PR to a new issue or find the existing one (if any) if you want.

From technical perspective, I got rid of `suppress_phase_change` method that (imo) should not be used because it's error prone.
We should rather rely on tangible events happening on the state (such as running Tasks and text animations) than delay the state update with some arbitrary values.

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -238,9 +238,6 @@ class CombatState(CombatAnimations):
         self._log_action: List[Tuple[int, EnqueuedAction]] = []
         self._monster_sprite_map: MutableMapping[Monster, Sprite] = {}
         self._layout = dict()  # player => home areas on screen
-        self._animation_in_progress: bool = (
-            False  # if true, delay phase change
-        )
         self._turn: int = 0
         self._prize: int = 0
         self._captured: bool = False
@@ -291,8 +288,7 @@ class CombatState(CombatAnimations):
                 0)
             next_animation()
 
-        if (not self._animation_in_progress and
-            self._text_animation_time_left <= 0 and
+        if (self._text_animation_time_left <= 0 and
             all(map(self.is_task_finished, self.animations))):
             new_phase = self.determine_phase(self.phase)
             if new_phase:
@@ -752,7 +748,6 @@ class CombatState(CombatAnimations):
         humans = list(self.human_players)
 
         # TODO: integrate some values for different match types
-        released = False
         for player in self.active_players:
             if len(alive_party(player)) == 1:
                 player.max_position = 1
@@ -764,14 +759,10 @@ class CombatState(CombatAnimations):
                     player, self.monsters_in_play[player], self._turn
                 )
                 for _ in range(positions_available):
-                    released = True
                     if player in humans and ask:
                         self.ask_player_for_monster(player)
                     else:
                         self.add_monster_into_play(player, next(available))
-
-        if released:
-            self.suppress_phase_change()
 
     def add_monster_into_play(
         self,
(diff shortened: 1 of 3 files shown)


JaskRendix:
I like it, you did the right thing with **compute_text_animation_time** in combat.py
I approve getting rid of `suppress_phase_change`

the failing checks are due to style,  I use `tox -efmt` to correct the style.
or eventually you can use the following combo:
```
autoflake -r -i --imports=tuxemon,pygame --ignore-init-module-imports tuxemon/
black tuxemon tests
isort black tuxemon tests
```
Good job!

JaskRendix (approved):

tuxemon/states/combat/combat.py
```diff
@@ -284,7 +294,18 @@ def update(self, time_delta: float) -> None:
         * update the new phase, or the current one
         """
         super().update(time_delta)
-        if not self._animation_in_progress and all(
+        self._text_animation_time_left -= time_delta
+        if (
+            self.no_ongoing_text_animation()
+            and len(self.text_animations_queue) > 0
```
```suggestion
            and self.text_animations_queue
```

In python we generally check the truthiness of an iterable rather than calculating its length.

kerizane (changes requested):
Also, please fix the style issue described by JaskRendix.

Grimmys:
> I like it, you did the right thing with **compute_text_animation_time** in combat.py I approve getting rid of `suppress_phase_change`
> 
> the failing checks are due to style, I use `tox -efmt` to correct the style. or eventually you can use the following combo:
> 
> ```
> autoflake -r -i --imports=tuxemon,pygame --ignore-init-module-imports tuxemon/
> black tuxemon tests
> isort black tuxemon tests
> ```
> 
> Good job!

Ok I'm going to run this command and fix the issue.
To be honest I didn't spend time yet investigating the CI workflows and thought running black & isort would be enough, thanks for the tip!

tuxemon/states/combat/combat.py
Grimmys: You are right, it's more pythonist your way.

To be honest I hesitated about this one. I thought verifying the length was more descriptive, and I always prefer when things are more readable. Also, this would be evaluated to true in more cases than it should. Since even if `text_animations_queue` is of the wrong type, the assertion would be passing as well.

But anyway, I was considering it this way also and as you said, it's the recommended way to go.
Changing it.

Grimmys:
Ah, I see the test workflow is running on Python 3.8, that's why it's failing: not able to read the new type hints aliases introduced in Python 3.9.

I see in the README that the game should be compatible with Python 3.8, so I'm going to adjust the type hint.

bitcraft (approved):
looks great, thank you

bitcraft:
@kerizane does this look ok to you now?

Grimmys:
Going back to this since it has been 3 weeks without update.

Taking into consideration you approved it yourself @bitcraft , that @kerizane didn't react more and that they just requested one simple change that had been applied (it was not even something I had to interpret any how, it was just about pushing their pre-made commit)... Can't we just force-merge this PR?

I think it's also blocking quite a few others PRs, and I would like to continue making changes on this combat/animation feature if possible.

bitcraft:
yeah, thats fair.  it looks like you took care of the style change as requested

bitcraft (approved):

PR #2005, opened 2023-09-05
snarlon - sphalian - rival (spyder)

JaskRendix:
as suggested I reopen this PR:

put Snarlon on the top of the mountain
added description in rival house billie (various rooms)
copypasted + renamed sphalian in the spyder_ version

Opened #2006 to go back to the old maps (before I started to modify/ameliorate)

### DIFF
--- a/mods/tuxemon/db/npc/spyder_cotton_town_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_cotton_town_npcs.json
@@ -93,7 +93,7 @@
     "slug": "spyder_cottoncafe_lotus",
     "template": [
       {
-        "sprite_name": "granny",
+        "sprite_name": "granny_lapi",
         "combat_front": "adventurer",
         "slug": "granny"
       }
@@ -113,7 +113,7 @@
     "slug": "spyder_cottonartshop_philis",
     "template": [
       {
-        "sprite_name": "granny",
+        "sprite_name": "granny_green",
         "combat_front": "adventurer",
         "slug": "granny"
       }
--- a/mods/tuxemon/db/npc/spyder_dojo_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_dojo_npcs.json
@@ -24,7 +24,7 @@
     "slug": "spyder_dojo_fu",
     "template": [
       {
-        "sprite_name": "granny",
+        "sprite_name": "granny_yellow",
         "combat_front": "adventurer",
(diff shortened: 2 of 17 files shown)


bitcraft:
looks great.  i hope that i (or somebody) will have time in the future to split the two campaigns.  its been discussed a few times before i i even began on it, but never completed it.  i think there was some technical thing that needed to be done so that both campaigns can share tile graphics

bitcraft (approved):

JaskRendix:
@bitcraft I tried here to split the folder #1660, but then I closed because I wasn't sure. Is it the right direction?

PR was acting on:
- image and tileset paths -> just added /../ so the tileset can be loaded;
- created **get_scenario_map()** in **client.py** -> this will return a str (scenario_folder/current_map), it'll be used for saving the current map in the correct format;
- **save_upgrader** to update all the **current_map** parameter;
- insert of the variable in **worldstate.py** -> adapted path;

if it's ok, then I can reopen. Let me know.

bitcraft:
in regards to the future of mods, and splitting the campaigns, the final directory structure would look like this:
```
tuxemon/
  mods/
    tuxemon/
      animations/
      db/
      ...
      maps/
    xero/
      items/
      ...
      maps/
    spyder/
      items/
      ...
      maps/
```
the goal would be that mods would have the same directory structure as `tuxemon` does now, so they can bring their own unique items, techniques, maps, monsters, npcs, etc.  there would also be a "mod load order", which we do have, so that mods can load from `tuxemon`, and that would serve as a common, shared base to build from.

with each campaign in its own mod and that system working, then game updates or other mods can be a simple zip file and we will have less issues with duplicated files, overwriting mod files, etc, because they have their own directory.

your proposal looks like it is splitting the campaigns in the `mods/tuxemon/maps` directory, which i think would be a great first step.  it would also prevent problems where things are renamed, removed, introducing incompatible story changes, etc.  at some later point, we can move the maps into their own mod when all issues about map loading are fixed.  (it's been a long time since i've done meaningful work on the project, so my information may be out of date -- it might be possible to make the changes right now, idk.)

tl;dr -- the scope of campaign and mods is more broad that just maps and might still require some work, but your proposal to split the campaigns in the same maps folder would be a good change

Sanglorian:
Thanks so much for doing this, @JaskRendix and @bitcraft ! Really appreciate all your effort

PR #2016, opened 2023-10-11
npc_wander_range

### LINKED ISSUE #799: Set maximum wander ranges
Hi folks,

Minor one here, but I think it'd be good if the npc_wander instruction also took two more optional parameters: 

* Maximum squares moved on the X axis from starting position
* Maximum squares moved on the Y axis from starting position

So for example npc_wander barmaid,0,3 means a barmaid created at 1,1 would move at most to 1,3, and no further (an instruction to move further would lead to them facing in that direction, but not moving).

JaskRendix:
@Sanglorian asked for this here #799

PR expands **npc_wander** with new fields (default None), so by using:
`npc_wander spyder_shopassistant,,5,7,7,9`
the npc will wander only between coordinates 5,7 and 7,9 (black dots)
_if the npc isn't in the "black square", it'll not move_

spyder_cotton_scoop

fix #799

### DIFF
--- a/tuxemon/event/actions/npc_wander.py
+++ b/tuxemon/event/actions/npc_wander.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
+import itertools
 import random
 from dataclasses import dataclass
 from typing import Union, final
@@ -21,24 +22,40 @@ class NpcWanderAction(EventAction):
     Script usage:
         .. code-block::
 
-            npc_wander <npc_slug> [frequency]
+            npc_wander <npc_slug>[,frequency][,c_v1][,c_v2]
 
     Script parameters:
         npc_slug: Either "player" or npc slug name (e.g. "npc_maple").
         frequency: Frequency of movements. 0 to stop wandering. If set to
             a different value it will be clipped to the range [0.5, 5].
             If not passed the default value is 1.
+        c_v1: coordinates upper_left vertex (eg 5,7)
+        c_v2: coordinates botton_right vertex (eg 7,9)
+
+        eg. npc_wander npc_slug,,5,7,7,9
 
     """
 
     name = "npc_wander"
     npc_slug: str
     frequency: Union[float, None] = None
+    x_v1: Union[int, None] = None
+    y_v1: Union[int, None] = None
+    x_v2: Union[int, None] = None
+    y_v2: Union[int, None] = None
 
     def start(self) -> None:
         player = self.session.player
         npc = get_npc(self.session, self.npc_slug)
         world = self.session.client.get_state_by_name(WorldState)
+
+        # generate all the coordinates
+        output = None
+        if self.x_v1 and self.y_v1 and self.x_v2 and self.y_v2:
+            input = [(self.x_v1, self.y_v1), (self.x_v2, self.y_v2)]
+            x_coords = [x for x in range(input[0][0], input[1][0] + 1)]
+            y_coords = [y for y in range(input[0][1], input[1][1] + 1)]
+            output = list(itertools.product(x_coords, y_coords))
 
         def move(world: WorldState, npc: NPC) -> None:
             # Don't interrupt existing movement
@@ -66,8 +83,16 @@ class NpcWanderAction(EventAction):
             exits = world.get_exits(origin)
             if exits:
                 if origin != coords:
-                    npc.path = [random.choice(exits)]
-                    npc.next_waypoint()
+                    path = random.choice(exits)
+                    if output:
+                        if path in output:
+                            npc.path = [path]
+                            npc.next_waypoint()
+                        else:
+                            return
+                    else:
+                        npc.path = [path]
+                        npc.next_waypoint()
                 else:
                     return
 
--- a/tuxemon/event/actions/npc_wander.py
+++ b/tuxemon/event/actions/npc_wander.py
@@ -22,7 +22,7 @@ class NpcWanderAction(EventAction):
     Script usage:
         .. code-block::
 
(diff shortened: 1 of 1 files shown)


tuxemon/event/actions/npc_wander.py
```diff
@@ -21,25 +22,41 @@ class NpcWanderAction(EventAction):
...
 
     """
 
     name = "npc_wander"
     npc_slug: str
     frequency: Union[float, None] = None
+    x_v1: Union[int, None] = None
+    y_v1: Union[int, None] = None
+    x_v2: Union[int, None] = None
+    y_v2: Union[int, None] = None
```
Grimmys: Consider using `Optional[int]` as a lighter typehint.
The typehint for `frequency` could be changed as well for consistency.

tuxemon/event/actions/npc_wander.py
```diff
@@ -21,25 +22,41 @@ class NpcWanderAction(EventAction):
...
 
     def start(self) -> None:
         player = self.session.player
         npc = get_npc(self.session, self.npc_slug)
         world = self.session.client.get_state_by_name(WorldState)
 
+        # generate all the coordinates
+        output = None
+        if self.x_v1 and self.y_v1 and self.x_v2 and self.y_v2:
```
Grimmys: Shouldn't `0` be a valid coordinate as well?
With this kind of condition, `0` and `None` are considered the same.

tuxemon/event/actions/npc_wander.py
```diff
@@ -21,25 +22,41 @@ class NpcWanderAction(EventAction):
...
+            input = [(self.x_v1, self.y_v1), (self.x_v2, self.y_v2)]
```
Grimmys: For improved readability, couldn't we consider a more descriptive structure such as:

```py
input = {
  "top_left_bound": (self.x_v1, self.y_v1),
  "bottom_right_bound": (self.x_v2, self.y_v2)
}
```

Or `top_bound` / `bottom_bound` or anything else, Idk what would be the best names.

Also maybe v1/v2 should be renamed for better clarity. It was not obvious for me at first that the algorithm will work only if `x_v1` <= `x_v2` and `y_v1` <= `y_v2`.

tuxemon/event/actions/npc_wander.py
```diff
@@ -66,8 +83,16 @@ def move(world: WorldState, npc: NPC) -> None:
...
+                    path = random.choice(exits)
+                    if output:
+                        if path in output:
+                            npc.path = [path]
+                            npc.next_waypoint()
+                        else:
+                            return
+                    else:
+                        npc.path = [path]
+                        npc.next_waypoint()
```
Grimmys, line 94: The imbrication levels of blocks starts to become a bit big... Maybe we could try to lower the count and remove the duplicated code by merging the conditions.

```py
path = random.choice(exits)
if not output or path in output:
    npc.path = [path]
    npc.next_waypoint()
```

Grimmys (changes requested):
The feature looks good and should be working fine, I'm just suggesting some refactoring and asking for clarification on one corner case.

tuxemon/event/actions/npc_wander.py
JaskRendix: it makes sense, top_bound / bottom_bound

JaskRendix:
Done, I hope it's more readable.

tuxemon/event/actions/npc_wander.py
```diff
@@ -66,8 +89,12 @@ def move(world: WorldState, npc: NPC) -> None:
             exits = world.get_exits(origin)
             if exits:
                 if origin != coords:
-                    npc.path = [random.choice(exits)]
-                    npc.next_waypoint()
+                    path = random.choice(exits)
+                    if not output or path in output:
+                        npc.path = [path]
+                        npc.next_waypoint()
+                    else:
```
Grimmys: You should also be able to totally remove this else statement, and even the next one: it's the end of the method, and there is nothing to return.
Let's the method end by itself, no need for extra lines doing nothing.

tuxemon/event/actions/npc_wander.py
```diff
@@ -21,25 +22,47 @@ class NpcWanderAction(EventAction):
...
+        # generate all the coordinates
+        output = None
+        top_bound = None
+        bottom_bound = None
+        if self.t_bound_x is not None and self.t_bound_y is not None:
+            top_bound = (self.t_bound_x, self.t_bound_y)
+        if self.b_bound_x is not None and self.b_bound_y is not None:
+            bottom_bound = (self.b_bound_x, self.b_bound_y)
+        if top_bound and bottom_bound:
+            v = [top_bound, bottom_bound]
```
Grimmys: Huh, that's not exactly what I was expecting regarding refactoring (but I think it's ok regarding fixing the corner case issue 👍 ).
What does `v` stands for? vector? vertice? It's not clear to me...

The main issue regarding clarity of this block to me was that it's hard to get what are doing in the next few lines.
In fact, I feel like we just don't need this `v` variable at all, it's just making the code more clumsy.

```py
x_coords = [x for x in range(top_bound[0], bottom_bound[0] + 1)]
y_coords = [y for y in range(top_bound[1], bottom_bound[1] + 1)]
``` 
What do you think, isn't it more readable this way?

Grimmys (changes requested):

Sanglorian:
This is great to have, thanks @JaskRendix

tuxemon/event/actions/npc_wander.py
JaskRendix: ok

tuxemon/event/actions/npc_wander.py
JaskRendix: I think so, I got rid of v (it was vertex, I shorted because it was too long).
Regarding the `is not None` block, I was unable to find a satisfying alternative.

Grimmys (approved):

bitcraft (approved):
lgtm

PR #2019, opened 2023-10-13
db: catch rate and catch resistance + techniques + constants

JaskRendix:
PR:
- inserts new validators (db.py) to check **catch_rate**, **lower / upper resistance** , **potency**, **accuracy**, **power**, **healing_power** and **recharge**;
- removes from **config** the catch rate, upper and lower resistance, since all the tuxemon have all the parameters and there are validators too;
- adds a series of constant inside **prepare.py**;
- removes the hardcoded max and min value from **db.py** and moves it in prepare.py;
- fixes 25/30 JSON files on techniques that were missing parameters such as power, accuracy and potency, this was an old format when technique and status shared the same model;


tuxemon/db.py
```diff
@@ -451,6 +456,18 @@ def translation_exists_category(cls: MonsterModel, v: Any) -> Any:
             return v
         raise ValueError(f"no translation exists with msgid: {v}")
 
+    @field_validator("catch_rate")
+    def check_catch_rate(cls: MonsterModel, v: Any) -> Any:
+        if 0 <= v <= 255:
+            return v
+        raise ValueError(f"the catch rate is between 1 and 255 ({v})")
```
bitcraft: confused by this.  the test will allow 0 and values less that 1, but the error message states the value must be between 1-255, so 0-1 shouldn't be allowed.  idk if the message or the test needs to change here

bitcraft: also the type for v should be a float, im assuming, not Any (same with the return type)

tuxemon/db.py
```diff
@@ -451,6 +456,18 @@ def translation_exists_category(cls: MonsterModel, v: Any) -> Any:
...
+
+    @field_validator("escape_rate")
+    def check_escape_rate(cls: MonsterModel, v: Any) -> Any:
```
bitcraft: v should be a float, same as the return value

bitcraft (changes requested):

tuxemon/db.py
JaskRendix: 0,255 is correct, I forgot to change the number
v should be a float

Sanglorian:
Thanks for this, so if I understand correctly - we can make some monsters easily frightened, so if an attempt to capture them fails, there's a 5% chance (or 10% or 20% or whatever we choose for a particular species) that they escape the combat?

If so, sounds great!

bitcraft (approved):

PR #2020, opened 2023-10-14
type in db.py + typehint

JaskRendix:
PR:
- fixes type and return type in db.py
- 1 typehint in combat.py

@bitcraft you suggested it in the other PR, I took the initiative to fix all validators

I checked with mypy --strict last version (1.6) and I got rid of this typehint:
`tuxemon/states/combat/combat.py:1307: error: Incompatible types in assignment (expression has type "str", variable has type "None")  [assignment]`

black, isort, tested, -2 typehints) #1211

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -229,20 +229,20 @@ class ItemModel(BaseModel):
 
     # Validate fields that refer to translated text
     @field_validator("use_item", "use_success", "use_failure")
-    def translation_exists(cls: ItemModel, v: Any) -> Any:
+    def translation_exists(cls: ItemModel, v: str) -> str:
         if has.translation(v):
             return v
         raise ValueError(f"no translation exists with msgid: {v}")
 
     @field_validator("slug")
-    def translation_exists_item(cls: ItemModel, v: Any) -> Any:
+    def translation_exists_item(cls: ItemModel, v: str) -> str:
         if has.translation(v):
             return v
         raise ValueError(f"no translation exists with msgid: {v}")
 
     # Validate resources that should exist
     @field_validator("sprite")
-    def file_exists(cls: ItemModel, v: Any) -> Any:
+    def file_exists(cls: ItemModel, v: str) -> str:
         if has.file(v):
             return v
         raise ValueError(f"the sprite {v} doesn't exist in the db")
@@ -258,7 +258,7 @@ class ShapeModel(BaseModel):
     speed: int = Field(..., description="Speed value")
 
     @field_validator("slug")
-    def translation_exists_shape(cls: ShapeModel, v: Any) -> Any:
+    def translation_exists_shape(cls: ShapeModel, v: str) -> str:
         if has.translation(v):
             return v
         raise ValueError(f"no translation exists with msgid: {v}")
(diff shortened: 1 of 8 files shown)


tuxemon/db.py
```diff
@@ -331,19 +331,23 @@ class MonsterEvolutionItemModel(BaseModel):
     tech: Optional[str] = Field(None, description="Technique parameter.")
 
     @field_validator("tech")
-    def technique_exists(cls: MonsterEvolutionItemModel, v: Any) -> Any:
+    def technique_exists(
+        cls: MonsterEvolutionItemModel, v: Optional[str]
+    ) -> Optional[str]:
+        if not v:
```
Grimmys: I'm not sure why we are considering the `None` case here (it isn't in the other similar methods), but anyway, the `if` statements can be merged like this:

```py
if not v or has.db_entry("technique", v):
    return v
```

tuxemon/states/combat/combat.py
```diff
@@ -255,7 +255,7 @@ def __init__(
         self._new_tuxepedia: bool = False
         self._run: bool = False
         self._post_animation_task: Optional[Task] = None
-        self._xp_message = None
+        self._xp_message: Optional[str] = None
```
Grimmys, line 258: Whoops, this attribute has been added by me... How did I forget the typehint.
Thanks for fixing it!

tuxemon/db.py
JaskRendix: Thanks, you're right. I missed item too (below), it's not similar because it involves evolution (eg.  botbot)
```
        {
            "path": "item", -> mandatory field, it evolves through item
            "at_level": 0, -> mandatory field, no minimum level
            "monster_slug": "av8r", -> mandatory field -> it evolves into av8r
            "item": "booster_tech" -> optional field regarding the item needed, that's why the field is default None
        }
```

Grimmys (approved):

tuxemon/states/combat/combat_animations.py
```diff
@@ -44,7 +44,7 @@
 
 sprite_layer = 0
 hud_layer = 100
-TimedCallable = Tuple[Callable, float]
+TimedCallable = Tuple[Callable[[], None], float]
```
bitcraft: pretty sure that you should be using `list` here instead of `[]`

bitcraft (changes requested):

tuxemon/states/combat/combat_animations.py
Grimmys: Good catch, but I think it should be `List`.
`list` is the better alias (since you don't need extra import), but it came with Python 3.9.

JaskRendix: I tried with List, but mypy was flagging me (mypy 1.6.1), and it looks like the suggested format is:
`TimedCallable = Tuple[partial[None], float]`

it clashes with 3.8

bitcraft: it was suggested that we retire python 3.8 support, and it might be a good time to do that.  i think the only major distro supporting it is ubuntu 16.04 lts, and there is a new lts now, so im not sure we need to support 3.8.  im in favor of preferring `list` over `List` in this case, and we can deprecate python 3.8 support (and tests)

bitcraft (approved):

bitcraft (approved):

tuxemon/states/combat/combat_animations.py
Grimmys: Yes, `list` over `List`, `dict` over `Dict`, `tuple` over `Tuple` and etc., I'm all for it!

PR #2021, opened 2023-10-14
fix Cotton Town + Paper Town (Spyder)

JaskRendix:
PR fixes NPC in Cotton Town (Spyder)
spyder_cotton_scoop
the NPC wasn't set correctly, the interaction didn't trigger the dialogue.


Sanglorian:
Thanks heaps for this - that was my bad.

JaskRendix:
While testing for #2033, I noticed that this area was blocked (player was unable to walk)
Added the fix here.

Screenshot_2023-10-19_12-44-01

added teleport_faint inside the mansion, it was missing

bitcraft (approved):

PR #2025, opened 2023-10-16
status bond field (condition)

JaskRendix:
it follows #1998 

PR:
- adds the field **bond** (bool, default False) in conditions;
- removes **has_status_bond** from both combat.py and hardcoding;
- updates the **check** in combat.py;

modders can define if a status/condition is going to create a bond between attacked and defender through the JSON.

At the moment only "grabbed" and "lifeleech" create a bond, but there can be more.


tuxemon/states/combat/combat.py
```diff
@@ -792,9 +791,9 @@ def add_monster_into_play(
                 self.monsters_in_play[player][0],
             )
 
-        # remove "connected" status (eg. lifeleech, etc.)
+        # remove "bond" status (eg. lifeleech, etc.)
         for mon in self.active_monsters:
-            if has_status_bond(mon):
+            if any(t for t in monster.status if t.bond):
```
Grimmys: Can be simplified to `any(t.bond for t in monster.status)`

Not sure what `t` stands for though, maybe better to rename it to `s` as in `status` (better would be `status` and rename `monster.status` to `monster.statuses`).

tuxemon/condition/condition.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, save_data: Optional[Mapping[str, Any]] = None) -> None:
 
         self.instance_id = uuid.uuid4()
         self.steps = 0.0
+        self.bond = False
         self.counter = 0
         self.cond_id = 0
         self.animation = Optional[str]
```
Grimmys: I know it's not part of this PR, but since it looks like it was part of the previous one (split condition): isn't there a mistake here?
Didn't you want to do this instead:
```py
self.animation: Optional[str] = None
```
or
```py
self.animation: str = ""
```

Grimmys (changes requested):

tuxemon/states/combat/combat.py
JaskRendix: renaming monster.status to monster statuses would be tough, tons of things to edit; 
I put `t` to abbreviate, but I'm going to put **s** or better **sta**

tuxemon/condition/condition.py
JaskRendix: yes, there is the same to update in technique.py too

tuxemon/states/combat/combat.py
Grimmys: Sure I can get that renaming a public field used in many places is out-of-scope of this PR.
That's why I put it in parenthesis.

Sanglorian:
Really important, thanks for doing this!

Grimmys (approved):

bitcraft (dismissed):

JaskRendix:
conflicts have been managed, ready to go

bitcraft (approved):

PR #2027, opened 2023-10-17
proportional damage effect (technique)

JaskRendix:
PR follows #1998 

This effect does damage to the enemy equal to % of the target's / user's maximum HP, defined by @Sanglorian 

PR updates **local_damage** into **prop_damage** (proportional damage) and it removes the hardcoded part, moreover it allows to define the percentage (float value) and the objective (user max HP or target max HP).

the previous effect was only used by 1 technique. Other techniques will come with this mechanism.

eg `"prop_damage target,0.25"`

### DIFF
--- a/tuxemon/technique/effects/prop_damage.py
+++ b/tuxemon/technique/effects/prop_damage.py
@@ -40,7 +40,6 @@ class PropDamageEffect(TechEffect):
     ) -> PropDamageEffectResult:
         player = self.session.player
         value = float(player.game_variables["random_tech_hit"])
-        mult = 1.0
         hit = tech.accuracy >= value
         if hit:
             tech.hit = True
@@ -54,7 +53,7 @@ class PropDamageEffect(TechEffect):
 
         return {
             "damage": damage,
-            "element_multiplier": mult,
+            "element_multiplier": 0.0,
             "should_tackle": bool(damage),
             "success": bool(damage),
             "extra": None,
--- a/tuxemon/technique/effects/prop_damage.py
+++ b/tuxemon/technique/effects/prop_damage.py
@@ -43,14 +43,11 @@ class PropDamageEffect(TechEffect):
         mult = 1.0
         hit = tech.accuracy >= value
         if hit:
-            damage = None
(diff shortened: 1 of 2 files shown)


tuxemon/technique/effects/prop_damage.py
```diff
@@ -5,58 +5,58 @@
...
+            damage = None
             tech.hit = True
             tech.advance_counter_success()
-            damage, mult = formula.simple_damage_calculate(tech, user, target)
-            # land of ifs
-            # tech: panjandrum
-            if tech.slug == "panjandrum":
-                damage = formula.damage_full_hp(target, 4)
-            target.current_hp -= damage
+            if self.objective == "target":
```
Grimmys: I believe this block could be refactored, because the formula is the same in both cases.

```py
reference_hp = target.hp if self.objective == "target" else user.hp
damage = int(reference_hp * self.proportional)
```

And yes, I'm adding casting at this place to remove the need to cast in the other places.

Grimmys (changes requested):

tuxemon/technique/effects/prop_damage.py
JaskRendix: thanks man, I'm going to update it

tuxemon/technique/effects/prop_damage.py
```diff
@@ -5,55 +5,52 @@
...
+    objective: str
+    proportional: float
 
     def apply(
         self, tech: Technique, user: Monster, target: Monster
-    ) -> LocalDamageEffectResult:
+    ) -> PropDamageEffectResult:
         player = self.session.player
         value = float(player.game_variables["random_tech_hit"])
+        mult = 1.0
```
Grimmys: Is this `mult` still used anywhere?

JaskRendix: removed, just tested with 0

Sanglorian:
Oh great, thank you!

Grimmys (approved):

bitcraft (approved):

PR #2028, opened 2023-10-17
multi attack effect

JaskRendix:
PR implements multiattack effect @Sanglorian 

Multiattack #: Do # attacks.

tested with **Ram**:
```
  "effects": [
    "multiattack 3",
    "damage"
  ],
```
in this scenario, it makes 3 attacks in a row, but if it misses one of these, then it stops; 
if it misses the first, then no second and third;
if it misses the second, then no third;
etc.

I don't know if the multi attack must stop if misses, let me know if it's correct
eventually we can add a flag to activate or deactivate ((if misses, then...)

### DIFF
--- a/tuxemon/technique/effects/multiattack.py
+++ b/tuxemon/technique/effects/multiattack.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import random
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thanks for this! And yes, that's how I imagined it working (after the first miss, no more attacks).

JaskRendix:
@Sanglorian regarding the "multi_attack" to assign it:
- https://wiki.tuxemon.org/Air_Chain (multi-attack eg 2/3/4/5, etc?)
- https://wiki.tuxemon.org/Webs_Wind (multi-attack eg 2/3/4/5, etc?)

we need to change some parameters like accuracy, potency, power and recharge.

bitcraft (approved):

Sanglorian:
Thanks, that's great! Have added it to Air Chain and Webs Wind - and reduced power to compensate.

PR #2034, opened 2023-10-19
item animation + remove hardcoded capture

JaskRendix:
PR
- removes hardcoded capture;
- adds animation for items;
- adds validator in db.py (item animations)
- adds folder animations/items;
- moves the existing capture_1 animation into the item folder;
- renamed **technique** into **method** (since utilized by technique, item and condition) > it'll follow other PRs

eg old piece hardcoded:
```
        # TODO: Use a technique JSON for 'capture' instead of hardcoding it here
        for fn in ["capture_1_%02d.png" % i for i in range(1, 7)]:
            fn = "animations/technique/" + fn
            images.append((image, 0.07))
```

new animations for the tuxeball, more similar to the actual one:
tuxeball while the previous was capture


Sanglorian:
Nice one!

tuxemon/item/item.py
```diff
@@ -56,6 +58,8 @@ def __init__(self, save_data: Optional[Mapping[str, Any]] = None) -> None:
         self.quantity = 1
         self.images: Sequence[str] = []
         self.type = ItemType.consumable
+        self.animation = Optional[str]
```
Grimmys: Wrong way of adding typehint.

Should be `self.animation: Optional[str] = None`.
Same issue for the already existing statement above.

Grimmys: Btw why is it of type `str` instead of `Animation` or `Task`?

Looks like it's not really the animation but only the location for the sprites, so maybe it should be renamed `animation_sprites_location`, `frames_path` or something like this.

tuxemon/states/combat/combat.py
```diff
@@ -151,10 +151,12 @@ def compute_text_animation_time(message: str) -> float:
 
 class TechniqueAnimationCache:
     def __init__(self) -> None:
-        self._sprites: Dict[Union[Technique, Condition], Optional[Sprite]] = {}
+        self._sprites: Dict[
+            Union[Technique, Condition, Item], Optional[Sprite]
```
Grimmys, line 155: This type looks complex and a part of it is repeated quite a lot of times in this file.

I think we should extract `Union[Technique, Condition, Item]` as a type alias at the top of the file.
I'm just not sure about what's the common thing between Technique, Condition and Item to name this type properly, but I trust you to find the perfect name! :D

tuxemon/states/combat/combat.py
```diff
@@ -176,13 +178,13 @@ def get(
 
     @staticmethod
     def load_technique_animation(
-        technique: Union[Technique, Condition], is_flipped: bool
+        technique: Union[Technique, Condition, Item], is_flipped: bool
     ) -> Optional[Sprite]:
         """
-        Return animated sprite from a technique.
+        Return animated sprite from a technique, condition or item.
```
Grimmys, line 184: Interesting, so looks like all these methods are not supporting only technique anymore but also condition or item, I was fooled by the name.

I think everything should be renamed (argument names, method names etc.). Yes I know, "big refactoring", but it would be really confusing to have a method named `load_technique_animation`.... That in fact doesn't load only technique animations...

Grimmys (changes requested):

Grimmys:
I see you are changing a bit how the animation tasks for capture are handled in combat.py.

How does it look like regarding delay between failed capture and next Tuxemon action?

I just wonder if you PR is not fixing this issue as well: https://github.com/Tuxemon/Tuxemon/issues/1967 (regarding "When a monster avoids capture" part).
I would like to know since I was ready to start investigating it, but looks like I would be conflicting with your work haha.

tuxemon/item/item.py
JaskRendix: copy paste issue, regarding the name, I agree, but there is another PR touching that spot (#2025) both in Technique and Condition, so we need to wait (at least as soon as it gets merged).

JaskRendix:
> How does it look like regarding delay between failed capture and next Tuxemon action?

pressing enter as suggested hard, while regarding delaying, I think it's possible, but we need to quantify how much.

tuxemon/states/combat/combat.py
JaskRendix: Yeah, I take note, but it'll be applied in another PR.

tuxemon/states/combat/combat.py
JaskRendix: Done, let's hope it's ok. I did the minimum, I targeted only a couple of defs, I don't want to "change" too much.

tuxemon/item/item.py
Grimmys: Yes don't worry I knew it was copy paste issue.

Grimmys:
Oh so you really implemented what was suggested in the issue, nice!

Yeah I guess it's fine.
I was more thinking about delay because it is what we are doing in all other places regarding combat animations (like you don't have to press enter to go the next messages at the end of the combat, or between each action). So it was more about being consistent, and simplify the process. But you just did what was stated in the ticket, so perfect!

JaskRendix:
@Grimmys probably I didn't explain clearly. I said that for me implementing click "enter" can be hard as well as tricky. I think too the best way is delaying and that we need to quantify how much delaying the enemy's attack (eg seconds, etc)

Grimmys:
Aaaaah right.

On my side I was just asking if after your changes this issue was fixed. 😅

Looks good, I feel like adding delay is the best solution as well: easier to implement, and more consistent with other combat flows.

JaskRendix:
No problem, added a small delay as requested by @Sanglorian in #1967 
`self.task(partial(self.blink, sprite), breakout_delay + 0.5,)`
used blink, sprites moves slightly, the wild monster has escaped the capture (0.5 small, but effective)

bitcraft (approved):

PR #2036, opened 2023-10-19
fix evolution

### LINKED ISSUE #2035: Evolution crash
**Describe the bug**
Nut evolution fail. When my Nut tries to evolve to Bolt at level 10, I see a black screen and the Yes-No option frame, then I see the "NUT is trying to evolve into BOLT. Allow evolution?" frame, but the Yes-No option frame disappears. I can hit Enter to (I'm guessing) select Yes, then I end up on just a black screen. 

**To Reproduce**
1. Level Nut to 10
2. It'll try to evolve

**Expected behavior**
1. Keep the Yes-No option frame visible when I'm choosing
2. Display NUT (the tuxemon evolving) rather than a black screen when I'm choosing
3. If the player chooses Yes, at least show BOLT (the tuxemon evolved into). An animation would be nice, but not necessary. 

**Screenshots**
Nut evolution

 - OS: Debian 12
 - Python Version 3.11.2

JaskRendix:
fix #2035 

PR fixes evolution crash. The crash was caused by the FadingState. Luckily it happened while I was working at this.

PR
- moves all the `evolve(`) code outside **combat.py**
- creates a brand new event action called **evolution**

the action will trigger at the end of the combat, so same output, no more hardcoded things in combat.py

### DIFF
--- a/tuxemon/event/actions/evolution.py
+++ b/tuxemon/event/actions/evolution.py
@@ -1,0 +1,114 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+from dataclasses import dataclass
+from functools import partial
+from typing import final
+
+from tuxemon.db import EvolutionType
+from tuxemon.event.eventaction import EventAction
+from tuxemon.locale import T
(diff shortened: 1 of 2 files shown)


Grimmys (approved):
LGTM, Thanks for extracting the evolution feature outside of combat.py!

combat.py is really a too big class, we should keep extracting out features as much as possible.

bitcraft (approved):

PR #2037, opened 2023-10-20
fix set_monster_level and move update_moves

JaskRendix:
PR:
- fixes a small bug noticed while testing #2035 > `set_monster_level 30,0` add 30 levels to the monster in slot 0, it was missing a ` is not None`
- moves **update_moves** into monster.py;
- removes some lines of code from combat.py;
- adds a return to **update_moves** (Technique) > the technique that has been learned;
- adds a return to **give_experience** (int) > the number of levels;
- updates the effect **give_xp**;

### DIFF
--- a/tuxemon/combat.py
+++ b/tuxemon/combat.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 import logging
 import random
-from typing import TYPE_CHECKING, Generator, List, Sequence, Union
+from typing import TYPE_CHECKING, Generator, List, Optional, Sequence
 
 from tuxemon.db import PlagueType
 from tuxemon.locale import T
@@ -165,21 +165,9 @@ def defeated(player: NPC) -> bool:
     return fainted_party(player.monsters)
 
 
-def check_moves(monster: Monster, levels: int) -> Union[str, None]:
-    tech: Union[Technique, None] = None
-    for move in monster.moveset:
-        # monster levels up 1 level
-        if levels == 1:
-            if move.level_learned == monster.level:
-                tech = learn(monster, move.technique)
-        # monster levels up multiple levels
-        else:
-            level_before = monster.level - levels
-            # if there are techniques in this range
-            if level_before < move.level_learned <= monster.level:
-                tech = learn(monster, move.technique)
+def check_moves(monster: Monster, levels: int) -> Optional[str]:
+    tech = monster.update_moves(levels)
     if tech:
-        monster.learn(tech)
         message = T.format(
             "tuxemon_new_tech",
             {
@@ -188,15 +176,4 @@ def check_moves(monster: Monster, levels: int) -> Union[str, None]:
             },
         )
         return message
-    else:
-        return None
-
-
-def learn(monster: Monster, tech: str) -> Union[Technique, None]:
-    technique = Technique()
-    technique.load(tech)
-    duplicate = [mov for mov in monster.moves if mov.slug == technique.slug]
-    if duplicate:
-        return None
-    else:
-        return technique
+    return None
--- a/tuxemon/event/actions/set_monster_level.py
+++ b/tuxemon/event/actions/set_monster_level.py
@@ -3,11 +3,9 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Union, final
+from typing import Optional, final
 
 from tuxemon.event.eventaction import EventAction
-from tuxemon.monster import Monster
-from tuxemon.technique.technique import Technique
 
 
 @final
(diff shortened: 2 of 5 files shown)


Grimmys (approved):
Good catch, we should always be cautious with null checks & auto-casting!

JaskRendix:
@Grimmys I was looking forward to the next step (#2039 related to evolution) and by checking again I noticed some things that can be simplified / improved. Tell me what do you think about.

tuxemon/combat.py
```diff
@@ -190,13 +178,3 @@ def check_moves(monster: Monster, levels: int) -> Union[str, None]:
         return message
     else:
```
Grimmys: This else branch can be removed.

By default, a call is going to return "None" if you don't specify what it should return at the end: https://realpython.com/python-return-statement/#implicit-return-statements

tuxemon/combat.py
```diff
@@ -166,20 +166,8 @@ def defeated(player: NPC) -> bool:
...
-        if levels == 1:
-            if move.level_learned == monster.level:
-                tech = learn(monster, move.technique)
-        # monster levels up multiple levels
-        else:
-            level_before = monster.level - levels
-            # if there are techniques in this range
-            if level_before < move.level_learned <= monster.level:
-                tech = learn(monster, move.technique)
+    tech = monster.update_moves(levels)
```
Grimmys, line 169: Yes, all this logic was tied to monster, and has barely anything to do with combat, thanks for the refactoring!

tuxemon/combat.py
```diff
@@ -166,20 +166,8 @@ def defeated(player: NPC) -> bool:
 
 
 def check_moves(monster: Monster, levels: int) -> Union[str, None]:
```
Grimmys: While we are modifying this method, `Union[str, None]` can be replaced by `Optional[str]` as well.

tuxemon/event/actions/set_monster_level.py
```diff
@@ -29,36 +27,23 @@ class SetMonsterLevelAction(EventAction):
     """
 
     name = "set_monster_level"
-    level: int
-    slot: Union[int, None] = None
+    levels: int
```
Grimmys: Why using plural for something that's not a list?
When I read the name, I'm expecting this identifier to match a list of int (a list of levels).

Looking at the usage, it looks like this variable represents the **number of levels earned**. So maybe we could rename it `levels_earned`, `increased_levels` or `nb_extra_levels`.

tuxemon/event/actions/set_monster_level.py
```diff
@@ -29,36 +27,23 @@ class SetMonsterLevelAction(EventAction):
...
-                    technique.load(move.technique)
-                    mon.learn(technique)
-
-        monster_slot = self.slot
-        monster_level = self.level
-
-        if monster_slot:
-            if len(self.session.player.monsters) < int(monster_slot):
+        if self.slot is not None:
+            if len(self.session.player.monsters) < self.slot:
```
Grimmys: Ok, usually, I'm kinda against comment, but I feel like we could describe what is happening there, or at least refactor it a bit.

In which case can the slot be bigger than the number of monsters the player owns?

tuxemon/monster.py
```diff
@@ -350,11 +350,14 @@ def give_experience(self, amount: int = 1) -> None:
         >>> bulbatux.give_experience(20)
 
         """
```
Grimmys, line 355: Don't forget to update the docstrings.

The method used to return nothing, but now it returns an int. We need to specify it in the documentation (and explain what's the meaning of the return value).

tuxemon/monster.py
```diff
@@ -549,6 +552,27 @@ def set_moves(self, level: int) -> None:
                 tech.load(ele)
                 self.learn(tech)
 
+    def update_moves(self, levels: int) -> Optional[Technique]:
+        """
+        Set monster moves according to the levels increased.
+        Excludes the moves already learned.
+
+        Parameters:
+            level: Nr of levels.
```
Grimmys: Inconsistency between docstring parameter and actual implementation parameter name (level/levels).

Grimmys: The part about the returned value is also missing.

tuxemon/monster.py
```diff
@@ -549,6 +552,27 @@ def set_moves(self, level: int) -> None:
...
+
+        """
+        tech = None
```
Grimmys: Something is wrong here.

Why do we need this extra `tech` variable here? Why not just return `technique` which relates to the same thing.

Also, according to the method logic, multiple techniques can be learnt (because we are looping over a list)... So why are we returning only one? Do we just assume it's always going to be one (in this case, we could stop earlier the call by putting a return statement inside of the if block), or what?

Grimmys (changes requested):
Great changes!

I added a lot of remarks, but most of them are related to code you just moved, not that you added, so I guess it's ok to address them in a separate PR, as you prefer.

I think it would be great to at least update the docstrings though.

tuxemon/combat.py
JaskRendix: Yeah, you're right

tuxemon/combat.py
JaskRendix: True, you told me in another PR

tuxemon/event/actions/set_monster_level.py
JaskRendix: I approve `levels_earned`

tuxemon/monster.py
JaskRendix: Fixed

tuxemon/monster.py
JaskRendix: My mistake, you're right, but I need to put at least `return None` if not I start collecting typehints with mypy

tuxemon/event/actions/set_monster_level.py
JaskRendix: > In which case can the slot be bigger than the number of monsters the player owns?

This is simply if a modder put slot 5 when there are only 3 monsters.

Grimmys (approved):
Looks much better now, thanks for this great refactoring!

bitcraft (approved):

PR #2039, opened 2023-10-20
evolution timing

JaskRendix:
PR:
- inserts field **levelling_up** and **got_experience** in monster.py;
- removes the last code from evolution from combat.py (last line);

These changes will allow the following action in map:
```
   <properties>
    <property name="act1" value="evolution"/>
    <property name="cond1" value="is check_evolution standard"/>
   </properties>
```
In the example above, it'll check for all the standard evolution; if true, then it'll add inside a `list(tuple(monster,monster))` (this format because old_monster,new_monster) all the existing combinations; this data will be used by **evolution** event action, the list will be cleared and the evolution dialogue will trigger.

This will allow modders to define in which maps or events they want a checking related to evolution.

by default the evolution will be triggered by experience + level up; it'll follow an option, but this will be added in another PR


bitcraft (approved):
this looks good.  in the future we should have a discussion about how to implement "game global" events, so that we don't have to duplicate common ones in all the maps.

PR #2040, opened 2023-10-22
traded field + messages (monster)

JaskRendix:
PR completes the trading (exchange of monster) by adding a field called **traded** (simple boolean), in this way it'll possible to define different behaviors for traded monsters.

I updated the message too:
Screenshot_2023-10-22_09-23-50
now it'll show traded instead of captured, moreover:
- on the same day "traded or capture today"
- on the following day "traded or capture X days ago"


bitcraft (approved):

Sanglorian:
This sounds really neat, thanks!

PR #2041, opened 2023-10-22
update get_player_monster (event action)

JaskRendix:
PR updates `get_player_monster`, it allows more filtering.

It's possible to filter for numeric and not numeric fields, such as evolution_stage, shape, level, speed, armour, etc.

Moreover it offers two additional variables:
- let's say a player doesn't has no options, then the variable will result as: name_variable:no_option
- let's say a player has options, but clicks return, then the variable will result as: name_variable:no_choice

this is very important for modders, because it can simplify the work of "actions", especially if the player doesn't possess the monster in the party.

`"get_player_monster name_variable,speed,more_than,50"`
or
`"get_player_monster name_variable,shape,serpent"`

### DIFF
--- a/tuxemon/event/actions/get_player_monster.py
+++ b/tuxemon/event/actions/get_player_monster.py
@@ -4,8 +4,17 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from typing import final
+from operator import eq, ge, gt, le, lt, ne
+from typing import Optional, final
 
+from tuxemon.db import (
+    ElementType,
+    EvolutionStage,
+    GenderType,
+    MonsterShape,
+    TasteCold,
+    TasteWarm,
+)
 from tuxemon.event.eventaction import EventAction
 from tuxemon.menu.interface import MenuItem
 from tuxemon.monster import Monster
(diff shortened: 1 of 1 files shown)


tuxemon/event/actions/get_player_monster.py
```diff
@@ -20,33 +29,183 @@
...
+                    self.result = True
+                    return True
+                elif value_name == "equals" and bool(eq(field, extra)):
+                    self.result = True
+                    return True
+                elif value_name == "not_equals" and bool(ne(field, extra)):
+                    self.result = True
+                    return True
+
+        return False
```
Grimmys: Shouldn't be `self.result` set to False in the default case as well?
Or at the beginning of the method maybe.

Also, I think we could just return the value of `self.result` in every cases, instead of repeating the same boolean value.

JaskRendix: You're right, I'll look carefully

JaskRendix: done, it returns `self.result`
`self.default` is set to False in the Start and it propagates.

tuxemon/event/actions/get_player_monster.py
```diff
@@ -20,33 +29,170 @@
...
         except ValueError:
+            player = self.session.player
+            if self.result and not self.choose:
+                # the player can choose, but returns
+                player.game_variables[self.variable_name] = "no_choice"
+            if not self.result:
+                # the player can't choose (eg no females in the party)
+                player.game_variables[self.variable_name] = "no_options"
             self.stop()
+            print(player.game_variables[self.variable_name])
```
bitcraft: please remove this or change to `log.debug`

bitcraft (changes requested):

tuxemon/event/actions/get_player_monster.py
JaskRendix: You're right, I forgot to remove

bitcraft (approved):

PR #2042, opened 2023-10-23
sacrifice effect (technique)

JaskRendix:
PR introduces the effect sacrifice.

It was a concept developed by @Sanglorian and it was related to autodestruction types techniques.
These techniques can damage both the user as well as the enemy and they intact the current_hp of the user.
A sacrifice.

M (42/100) - E (69/100).
M uses technique XYZ (with sacrifice effect) against E
`sacrifice 1`
M (0/100) faints - E (27/100).

regarding we can assign the "sacrifice" to assign it to:
- https://wiki.tuxemon.org/Undertaker

we need to change some parameters like type, category, accuracy, potency, power and recharge.

### DIFF
--- a/mods/tuxemon/db/technique/undertaker.json
+++ b/mods/tuxemon/db/technique/undertaker.json
@@ -4,13 +4,13 @@
   "animation": "x_attack",
   "effects": [
     "give prickly,target",
-    "enhance"
+    "sacrifice 2"
   ],
   "flip_axes": "",
   "is_fast": false,
   "potency": 1,
-  "power": 0,
-  "range": "special",
+  "power": 1,
+  "range": "melee",
   "recharge": 1,
   "sfx": "sfx_blaster",
   "slug": "undertaker",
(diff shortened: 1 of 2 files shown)


Sanglorian:
Good idea! It's a useful option to have, and I agree Undertaker's a great fit for it. 

Could the number after sacrifice be a multiplier, so sacrifice 2 would mean "monster takes damage equal to its current HP, and does damage equal to double that amount"?

JaskRendix:
> Could the number after sacrifice be a multiplier, so sacrifice 2 would mean "monster takes damage equal to its current HP, and does damage equal to double that amount"?

@Sanglorian done, updated the formula as requested. I added the effect to "undertaker", now it damages and gives the effect prickly; I changed the power (put 1), type (from aether to eart) and range (from special to melee), let me know if it's ok for you.

Sanglorian:
> > Could the number after sacrifice be a multiplier, so sacrifice 2 would mean "monster takes damage equal to its current HP, and does damage equal to double that amount"?
> 
> @Sanglorian done, updated the formula as requested. I added the effect to "undertaker", now it damages and gives the effect prickly; I changed the power (put 1), type (from aether to eart) and range (from special to melee), let me know if it's ok for you.

Love it, thanks!

bitcraft (approved):

PR #2045, opened 2023-10-25
enable replacing var (locale - PO files)

JaskRendix:
PR enables replacing variables inside the PO files. It uses the same method:

eg `${{name}} is running away!` gives back `Richard is running away!`

this will help to pass data without being forced to hardcode, especially numbers as well as names, etc. 

Example:
eg `you chose nr. {{var:xyz}} back then!` gives back `you chose nr. 3 back then!`

### DIFF
--- a/tuxemon/locale.py
+++ b/tuxemon/locale.py
@@ -224,6 +224,9 @@ def replace_text(session: Session, text: str) -> str:
     text = text.replace("${{currency}}", "$")
     text = text.replace(r"\n", "\n")
     text = text.replace("${{money}}", str(player.money["player"]))
+    # replace variables
+    for key, value in player.game_variables.items():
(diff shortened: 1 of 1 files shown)


Grimmys:
Wooooooooow!

It's funny because I was talking about something really similar a few weeks ago in the French translation channel: being able to put tokens (variables) into translation files is neat, because then we can avoid having to repeat the same translation for a keyword element of the game, since this repetitions could to lead to inconsistent translations of the same term (ex: _Tuxecenter_ has been recently translated as _Tuxecentre_ in French, but without such feature, in a few months, we could see a sentence using another variant like _Centre Tuxemon_).

TL;DR: Thank you a lot for your investment on this!

Grimmys (approved):

bitcraft (approved):

PR #2046, opened 2023-10-25
competition (event action)

JaskRendix:
depends on #2045 (and ideally #2041 but not mandatory)

PR adds the competition event action, it'll allow to a selected monster to compete against others.
```
    <property name="act1" value="get_player_monster top_flier"/>
    <property name="act2" value="competition top_flier,speed,av8r:tweesher:cherubat"/>
```
the action will generate a series of variables in this format:
`comp_rank_{i} : MONSTER (POINTS)`
where `i` is the rank (eg 1, 2, 3, etc), moreover it'll generate also:
`top_flier : i"`
where `i` is the rank of the monster selected (eg if the player chooses Pairagrin, then it'll refer to the rank of the monster)
so it'll be `top_flier : 2` because it's second.
in this way the modder can recall and create a precise output depending on the ranking.
eg:
```
<property name="cond1" value="is variable_set top_flier:1"/> -> well done, etc.
<property name="cond1" value="is variable_set top_flier:2"/> -> you can ameliorate, etc.
```
in this way we can recycle the variable

### DIFF
--- a/tuxemon/event/actions/competition.py
+++ b/tuxemon/event/actions/competition.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import random
+import uuid
+from dataclasses import dataclass
+from typing import List, final
+
+from tuxemon.db import StatType
+from tuxemon.event.eventaction import EventAction
+from tuxemon.monster import Monster
+
+
+@final
+@dataclass
+class CompetitionAction(EventAction):
+    """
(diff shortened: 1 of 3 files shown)


JaskRendix:
added an extra field, so this will be possible too:
```
    <property name="act1" value="get_player_monster top_fight"/>
    <property name="act2" value="competition 100,top_fight,melee,sumchon,armour"/>
```
melee vs armour

as per
```
    HP: How much damage you can take
    Melee (or "Attack"): How much damage your Melee attacks do
    Ranged (or "Special Attack"): How much damage your Ranged attacks do
    Armour (or "Defence"): How well you resist damage from Melee attacks
    Dodge (or "Special Defence"): How well you resist damage from Ranged attacks
    Speed: Whether you act before your opponent in battle
```

Sanglorian:
Sounds good!

PR #2050, opened 2023-10-27
steps countdown

JaskRendix:
PR:
- introduces the possibility to set a countdown for steps;
- adds check returning Flase in `variable_is` because if not the game is going to crash (not finding the game variable);
- adds alternative labels in `variable_is`, especially considering <=, that in game needs to be written as `&lt`, so better offer an alternative like `less_than`;

set this
`<property name="act1" value="set_variable steps_park:200"/>`
the in another action 
`<property name="cond1" value="is variable_is steps_park,equals,0.0"/>`

### DIFF
--- a/tuxemon/player.py
+++ b/tuxemon/player.py
@@ -50,6 +50,14 @@ class Player(NPC):
         diff_y = abs(after.y - before.y)
 
         self.game_variables["steps"] += diff_x + diff_y
+
+        for key, value in self.game_variables.items():
+            if key.startswith("steps_"):
+                if float(value) > 0.0:
+                    self.game_variables[key] = float(value) - (diff_x + diff_y)
+                else:
+                    self.game_variables[key] = 0.0
+
         """
         %H - Hour 00-23
         %j - Day number of year 001-366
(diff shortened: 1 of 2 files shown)


tuxemon/player.py
```diff
@@ -50,6 +50,13 @@ def update(self, time_delta: float) -> None:
         diff_y = abs(after.y - before.y)
 
         self.game_variables["steps"] += diff_x + diff_y
+
+        for key, value in self.game_variables.items():
+            if key.startswith("steps_") and float(value) > 0.0:
```
Grimmys: Since the first predicate of the condition is in both cases the same, and the second predicate is just reversed in the other case, we could simplify the whole block by this:

```py
if key.startswith("steps_"):
    if float(value) > 0.0:
        self.game_variables[key] = float(value) - (diff_x + diff_y)
    else:
        self.game_variables[key] = str(0)
```

Also since in both cases it's assigning a value to the same identifier, we could choose to use "ternary" operator, but I let it up to you.

Grimmys: Btw, why `str(0)`? Why not just `"0"` or even `0.0`, since in the other case it's a floating value that is assigned.

JaskRendix: @Grimmys your way is better

regarding str(0), that's because of `variable_set` is set default str()
eventually we could use `variable_is`
edit: now it works with variable_is, the output will be 0.0

Grimmys (approved):

bitcraft (approved):

PR #2051, opened 2023-10-29
fly mechanism (technique)

JaskRendix:
PR implements the mechanism for fly. When the monster is out of reach and some attacks fail.
It requires: #2027 

Tux A vs Tux B
```
Tux A flies off
Tux B attacks > fails
Tux A lands + attacks

or (if both fly off)

Tux A flies off
Tux B flies off
Tux A lands + attacks > fails
Tux B lands + attacks
```
Technically the monster will disappear (sprite) and the parameter `out_of_reach` will become `TRUE` (default `FALSE`):
- if the monster lands, then it'll be FALSE again;
- if the monster faints, then it'll be FALSE again;

steps:
- find a technique for fly (or create a new one);
- add the following in the effects, as well as set animation null:
```
	"animation": null,
	"conditions": [
	  	"not status grabbed",
	  	"not status stuck"
	],
	"effects": [
		"fly_off XXX"
	],
```
- `fly_off XXX` XXX is the technique that will trigger when the monster lands;
- add among the effects of XXX technique "land";
- XXX could be [Hawk](https://wiki.tuxemon.org/Hawk)? @Sanglorian 
- when the monster flies off, it shows `"{name} flies up high!"`


bitcraft (dismissed):

Sanglorian:
Thanks, sounds like a cool addition. I think Hawk is a good choice for the fly_off technique.

Could we call the original technique something other than Fly - maybe Swoop or Aerial or Altitude? Just to avoid too much similarities with the other game!

JaskRendix:
@Sanglorian Altitude sounds good, added [Altitude](https://wiki.tuxemon.org/Altitude) and edited [Hawk](https://wiki.tuxemon.org/Hawk), removed Hawk from 4 monsters and replaced it with Assault.

xirsoi (approved):

PR #2052, opened 2023-10-29
enable replacing trans (locale - PO files)

JaskRendix:
PR enables replacing translations inside the PO files. It requires #2045, it uses the same method:

eg Jemuar description:
`msgstr "The gems on its back are the type found when it was born.
There are extensive ROCKITTEN breeding facilities over diamond veins."`
to
`msgstr "The gems on its back are the type found when it was born.
There are extensive ${{trans:rockitten}} breeding facilities over diamond veins."`

This will help in multiple cases, especially in the naming of facility such as tuxecentre, tuxeshop and similar alternatives.


Sanglorian:
Really useful, thanks!

PR #2053, opened 2023-10-30
track damages

JaskRendix:
PR implements tracking of damages (it requires #2037):
- creates `DamageMap` class;
- turns self._damage_map from `MutableMapping[Monster, Set[Monster]]` into `List[DamageMap]`;
- the process is the same (if should_tackle is TRUE, so the techniques hits, it gets added to the list);
- creates def `enqueue_damage` to add the DamageMap to the list;

this will allow to extrapolate numerical values (as before, but with two new entries):

---

_# nr times fainted monster got damaged_ (already existing)
```
hits = len([ele for ele in self._damage_map if ele.defense == monster])
```
---

_# all monsters that hit fainted monster_ (already existing)
```
winners = {ele.attack for ele in self._damage_map if ele.defense == monster}
```

`DamageMap` class:
```
class DamageMap(NamedTuple):
    attack: Monster
    defense: Monster
    damage: int
```


JaskRendix:
**unrelated**, this is why I post this in replies (it'll be moved to Discord too):

At the moment experience gets awarded with:
```
# Award experience
awarded_exp = (enemy.total_experience // (enemy.level * hits)) * enemy.experience_modifier
```
the only issue with this is the following:
```
rockitten attacks (-1 HP) x 9 times (-9 HP)
nut attacks (-85 HP)
total hits = 10

let's say awarded_exp = 35
rockitten 35
nut 35
```
with the values above, in the case there are more monsters, we can understand who has contributed more, but this is just an idea and it doesn't involve this PR. 

the proposal is to move the award experience formula after the `for winner in winners:` and eventually apply a percentage, the results between the damages of the single monster out of the total damages.
```
tot_damage_mon = sum([ele.damage for ele in self._damage_map if ele.defense == monster and ele.attack == winner])

tot_damage_all = sum([ele.damage for ele in self._damage_map if ele.defense == monster])

# Award experience
for winner in winners
     awarded_exp = (enemy.total_experience // (enemy.level * HITS)) * enemy.experience_modifier
     percentage = int(awarded_exp * (tot_damage_mon / tot_damage_all))
     levels = winner.give_experience(percentage )

as per example above:
rockitten 9HP / 94HP > 9.5% * 35 exp = 3 exp
nut 85HP / 94HP >90.5% * 35 exp = 32 exp
```

@Sanglorian I was checking tuxepedia and I found this item: https://wiki.tuxemon.org/XP_Transmitter.

description: _When this item is in your possession, all monsters that did not participate in defeating a monster get half of the XP. The remainder goes to the monsters that did participate._

tell me if what I understood is correct, I'm using the same example as above:
```
party: rockitten, nut, pairagrin, apeoro
participants: rockitten, nut
exp: 35 exp each one (rockitten 35, nut 35)

pairagrin gets 17/18 exp
apeoro gets 17/18 exp

is it correct?
```

JaskRendix:
@Sanglorian I was checking tuxepedia and I found this item: https://wiki.tuxemon.org/XP_Transmitter.

description: _When this item is in your possession, all monsters that did not participate in defeating a monster get half of the XP. The remainder goes to the monsters that did participate._

tell me if what I understood is correct, I'm using the same example as above:
```
party: rockitten, nut, pairagrin, apeoro
participants: rockitten, nut
exp: 35 exp each one (rockitten 35, nut 35)

pairagrin and apeoro get 17/18 exp too?
```

bitcraft (approved):

PR #2062, opened 2023-11-03
split journalstate + swipe (left & right) monsterinfostate

JaskRendix:
PR:
- replaces `Dict` with `dict`, etc.
- split `JournalState` and `MonsterInfoState` (new folder monster_info);
- reorganizes `JournalState`  structure;
- implements the possibility to swipe monsters through left and right buttons (`MonsterInfoState`);
- adds the parameter source (to understand from where it's visited: at the moment `MonsterInfoState` can be accessed from the menu monster, from the monster boxes - PC -, from combat - captured monster - and the action change_state);

structure:
```
journal/__init__.py
journal/journal_choice.py (1/3 page when clicking on tuxepedia)
journal/journal_info.py (3/3 page when clicking on tuxepedia)
journal/journal.py (2/3 page when clicking on tuxepedia)
monster_info/__init__.py
```

it addresses the 3rd point in #1967 : _It's fantastic that the Tuxepedia page about the captured monster pops up after the capture occurs - it really adds something. Is it possible to make Enter/Shift make the page disappear, not just escape?_

requested by @Sanglorian months ago (when I implemented this menu)

I'm referring to the menu accessible through monster menu (submenu info):
Screenshot_2023-11-03_16-11-30

it'll make visible only the pages where there is an actual monster:
Screenshot_2023-11-04_11-09-43


tuxemon/states/journal/journal.py
```diff
@@ -0,0 +1,145 @@
...
+        monsters = sorted(monsters, key=lambda x: x.txmn_id)
+
+        player = local_session.player
+        for mon in monsters:
+            if mon.slug in player.tuxepedia:
+                label = str(mon.txmn_id) + ". " + T.translate(mon.slug).upper()
+                if player.tuxepedia[mon.slug] == SeenStatus.seen:
+                    menu.add.button(
+                        label,
+                        partial(self.client.push_state, JournalInfoState(mon)),
```
bitcraft: Using `partial` and `push_state` with an instance `State` should be avoided.  The reason being is that since the `JournalInfoState` is being created here, *and* is is optional that the play presses the button, we could be wasting CPU time and memory creating the `JournalInfoState`, loading graphics, etc.  In this instance, you should pass the name of the state as a string, and the arguments and kwargs.
Doing it this way ensure that the next state is only created when needed.  In some situations, its fine to `push_state` with an instanced State object if we 100% know that the state needs to change.  But its part of a optional button, so we should avoid creating it if the player might not open it.

tuxemon/states/journal/journal.py
```diff
@@ -0,0 +1,145 @@
...
+                        font_color=(25, 25, 112, 1),
+                        selection_color=(25, 25, 112, 1),
+                        button_id=mon.slug,
+                    ).translate(
+                        fix_width(width, 0.25), fix_height(height, 0.01)
+                    )
+                elif player.tuxepedia[mon.slug] == SeenStatus.caught:
+                    menu.add.button(
+                        label,
+                        partial(self.client.push_state, JournalInfoState(mon)),
```
bitcraft: see other comment about `partial` and `push_state`

bitcraft (changes requested):

tuxemon/states/journal/journal.py
JaskRendix: ok, I'll keep it in mind, fixed

bitcraft (approved):

PR #2063, opened 2023-11-03
fix weight and height

JaskRendix:
PR fixes an issue I bumped into while mass testing on #2062
- changes the weight and height 0 into 0.1;
- adds a validator to check the weight and height;

Reason:
```
  File "/home/julesverne/Tuxemon/tuxemon/formula.py", line 293, in weight_height_diff
    weight = round(((monster.weight - db.weight) / db.weight) * 100, 1)
ZeroDivisionError: float division by zero
```

### DIFF
--- a/tuxemon/states/pc_kennel/__init__.py
+++ b/tuxemon/states/pc_kennel/__init__.py
@@ -192,16 +192,16 @@ class MonsterTakeState(PygameMenuState):
             label = T.translate(monster.name).upper()
             iid = monster.instance_id.hex
             new_image = pygame_menu.BaseImage(
-                transform_resource_filename(monster.menu_sprite_1),
+                transform_resource_filename(monster.front_battle_sprite),
                 drawing_position=POSITION_CENTER,
             )
-            new_image.scale(prepare.SCALE, prepare.SCALE)
+            new_image.scale(prepare.SCALE * 0.5, prepare.SCALE * 0.5)
             menu.add.banner(
                 new_image,
                 partial(kennel_options, iid),
                 selection_effect=HighlightSelection(),
             )
-            diff = round((monster.current_hp / monster.current_hp) * 100, 1)
+            diff = round((monster.current_hp / monster.hp) * 100, 1)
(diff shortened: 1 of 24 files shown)


tuxemon/states/pc_kennel/__init__.py
```diff
@@ -192,16 +192,16 @@ def description(mon: Monster) -> None:
             label = T.translate(monster.name).upper()
             iid = monster.instance_id.hex
             new_image = pygame_menu.BaseImage(
-                transform_resource_filename(monster.menu_sprite_1),
+                transform_resource_filename(monster.front_battle_sprite),
                 drawing_position=POSITION_CENTER,
             )
-            new_image.scale(prepare.SCALE, prepare.SCALE)
+            new_image.scale2x()
```
bitcraft: why changed to scale 2x?

JaskRendix: @bitcraft because the icons need to be smaller, there are 4 columns

bitcraft: we should be using ratios based on scale, because if you change the screen resolution, the SCALE value changes.  therefore, if we hardcode the scale value, the game would have different sized sprites at different resolutions

bitcraft: it might look correct at the resolution you are using, but if it is run at some different resolution, the icons will be the wrong size

JaskRendix: ok

bitcraft: if you want them to be a specific size, you can do something such as `icon_size * (SCALE * .5)`, which would keep the icons the correct size as the screen resolution changes.

JaskRendix: you are right, I'm going to put *.5

without `*.5`
Screenshot_2023-11-04_18-00-00

with` *.5`
Screenshot_2023-11-04_18-02-46

bitcraft (approved):

PR #2069, opened 2023-11-05
scenario based yaml (events)

JaskRendix:
PR implements scenario based YAML (eg spyder.yaml, xero.yaml, etc), the name is based upon this string inside the maps: `<property name="scenario" value="spyder"/>`. 3 new action conditions have been added: **location_type**, **location_inside** and **location_name**, these will allow to define precisely the locations.

This YAML will be active on all the maps with that property (by the way the property has been added to the totality of maps months ago) and it'll manage events such as teleport_faint, evolution, etc.

It was necessary because adding new events to the maps is time wasting as well as not precise (eg evolution was possible only in the maps where there was an event).

All the changes are related to Spyder maps


bitcraft (dismissed):
nice.  conflict though :(

JaskRendix:
@bitcraft fixed

bitcraft (approved):

PR #2070, opened 2023-11-06
fix nodding off

JaskRendix:
PR fixes nodding off (fix 1854)

@Sanglorian I have merged the nodding off and dozing off status. So all the messages and alerts are maintained in one single status, because at the end involves falling asleep and waking up. This will give you the opportunity to decide how long the monster is going to sleep (eg nr min and max of turns, etc). Let me know, so eventually we can remove or transform dozing off in something else.


bitcraft (approved):

Sanglorian:
Thanks! How about:

Sleep lasts for a minimum of one turn.
It has a 50% chance to end after each turn.
If it has gone on for three turns, it ends.

PR #2071, opened 2023-11-06
fix variable is (event condition)

JaskRendix:
PR adds a small fix, it removes the two checking on game_variables (it would block only numbers) and it adds a None return (Optional float) to the function in tools.py. So it's possible to remove the raise valueerror (crash) with a logger error and subsequent False return.


bitcraft:
This is probably ok.  I think that the intention is to accept errors in the map without crashing?

JaskRendix:
@bitcraft you're right, now the task is complete (purged all actions - simple return - and conditions - return False -)

PR #2078, opened 2023-11-08
fix sort

### LINKED ISSUE #2073: Items should always be used at the beginning of the turn.
**Describe the bug**
Currently, everything in a turn depends on the level/speed of the opposing Tuxemon. Items, since they are not moves, should run at the beginning of the turn.

**To Reproduce**
1. use an item
2. Item gets used after opposing Tuxemon moves

**Expected behavior**
Item should be used before the opposing Tuxemon moves

JaskRendix:
fix #2073
removed heal because among the sort there isn't any heal (replaced by quest)

added renaming of **technique** in **method** (EnqueuedAction) since it's carrying multiple methods
```
    technique: Union[Technique, Item, Condition, None]
    target: Monster
```
int
```
    method: Union[Technique, Item, Condition, None]
    target: Monster
```


Grimmys (approved):
Good subtle change 👍

bitcraft (approved):

PR #2079, opened 2023-11-08
text color, background, shadow

JaskRendix:
PR allow to change color to the dialog:
- background. font, shadow colors are moved in **prepare.py**;
- creates a subfolder named **borders** in **gfx**;
- **dialog-borders01.png** has been renamed in **dialog-borders.png**;
- **dialog-borders.png** has been moved in the folder **borders**;
- menu.py have been created self.variable related to background, font and shadow;
- includes 14 different styles (default excluded) from which modders can choose, all the styles have good/excellent results for readability / contrast;

default colors are 248,248,248 (background white), 10,10,10 (font, black) and 192,192,192 (font `shadow)`
fix #671

useful link: https://www.uxmatters.com/mt/archives/2007/01/applying-color-theory-to-digital-displays.php

changed font color: Burnt umber (`#330000`) = 51, 0, 0
Screenshot_2023-11-08_10-02-44
changed font color: Kashmir green (`#003300`) = 0, 51, 0
Screenshot_2023-11-08_10-00-34
changed font color: Midnight blue (`#000033`)  = 0, 0, 51
Screenshot_2023-11-08_09-58-24

PR #2081, opened 2023-11-09
choice_monster, choice_npc and choice_item (event action)

JaskRendix:
PR implements choice_monster (event action), it allows to choose among monsters and it allows to access to the Journal.

By clicking on the name you save the slug of the monster inside a variable, while clicking on the image you access the journal.

`<property name="act082" value="choice_monster budaye:dollfin:grintot:ignibus:memnomnom,myintrochoice"/>`

application before:
Screenshot_2023-11-09_09-37-14
application after:
Screenshot_2023-11-09_09-40-10

items

Screenshot_2023-11-09_12-14-18


tuxemon/event/actions/choice_monster.py
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+
+    name = "choice_monster"
+
+    choices: str
+    variable: str
+
+    def start(self) -> None:
+        def set_variable(var_value: str) -> None:
```
Grimmys: I feel like inner methods often add unnecessary complexity to the code, what would you think about moving this method outside as a (private) instance method?
`player` would become an argument.
Suggested signature:
```suggestion
        def _set_variable(self, value: str, player: Player) -> None:
```

tuxemon/event/actions/choice_monster.py
```diff
@@ -0,0 +1,71 @@
...
+            player.game_variables[self.variable] = var_value
+            self.session.client.pop_state()
+
+        # perform text substitutions
+        choices = replace_text(self.session, self.choices)
+        maybe_player = get_npc(self.session, "player")
+        assert maybe_player
+        player = maybe_player
```
Grimmys: I don't think we need two variables for the player.
Let's just do the check on the fly.
```suggestion
        player = get_npc(self.session, "player")
        assert player
```

tuxemon/event/actions/choice_monster.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # make menu options for each string between the colons
+        var_list = choices.split(":")
+        var_menu = list()
```
Grimmys: Looks like the elements we are putting in this list are quite complex, could we add some typing to this identifier?

tuxemon/states/choice/choice_item.py
```diff
@@ -0,0 +1,81 @@
...
+from tuxemon import prepare
+from tuxemon.db import db
+from tuxemon.menu.menu import PygameMenuState
+from tuxemon.menu.theme import get_theme
+from tuxemon.tools import transform_resource_filename
+
+ChoiceMenuGameObj = Callable[[], None]
+
+
+def fix_width(screen_x: int, pos_x: float) -> int:
```
Grimmys: Could we add more documentation for the parameters of this method?

I'm not sure to understand what's supposed to be `screen_x` for example.

tuxemon/states/choice/choice_item.py
```diff
@@ -0,0 +1,81 @@
...
+    """it returns the correct width based on percentage"""
+    value = round(screen_x * pos_x)
+    return value
```
Grimmys: The statements can be inlined, i.e. we don't need to store the value in a temporary variable that's going to be returned in the line anyway.
```suggestion
    return round(screen_x * pos_x)
```

tuxemon/states/choice/choice_item.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def fix_height(screen_y: int, pos_y: float) -> int:
```
Grimmys: Same comments as for previous method.

tuxemon/states/choice/choice_item.py
```diff
@@ -0,0 +1,81 @@
...
+
+class ChoiceItem(PygameMenuState):
+    """
+    Game state with a graphic box and some text in it.
+
+    Pressing the action button:
+    * if text is being displayed, will cause text speed to go max
+    * when text is displayed completely, then will show the next message
+    * if there are no more messages, then the dialog will close
+    """
```
Grimmys: I think this docstring is a copy/paste from `ChoiceState`. We need to change it, since this class doesn't serve the same purpose as `ChoiceState`.

tuxemon/states/choice/choice_monster.py
```diff
@@ -0,0 +1,112 @@
...
+
+class ChoiceMonster(PygameMenuState):
+    """
+    Game state with a graphic box and some text in it.
+
+    Pressing the action button:
+    * if text is being displayed, will cause text speed to go max
+    * when text is displayed completely, then will show the next message
+    * if there are no more messages, then the dialog will close
+    """
```
Grimmys: Same as for `ChoiceItem`

tuxemon/states/choice/choice_monster.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(
+        self,
+        menu: Sequence[tuple[str, str, Callable[[], None]]] = (),
+        escape_key_exits: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        theme = get_theme()
+        theme.scrollarea_position = POSITION_EAST
+        theme.scrollbar_color = (237, 246, 248)
+        theme.scrollbar_slider_color = (197, 232, 234)
```
Grimmys: Color values should be extracted as constant with proper name into a constants file, so we could see what color is set (because I dare anybody to guess which color `(197, 232, 234)` matches lol) and so we could reuse these colors in other places as well.

Grimmys (commented):
I added a few comments about code smells, but I understand that most of the code here was not added by you, you just copy/pasted (or cut/pasted) things from one place to another.

So please don't feel forced to apply my changes if you don't understand what they are all about.

I just think we should at least update documentation, outdated documentation can be horrible.

tuxemon/states/choice/choice_monster.py
JaskRendix: Planned in another PR, the one regarding the style #2079

tuxemon/event/actions/choice_monster.py
JaskRendix: I agree, it makes sense

tuxemon/event/actions/choice_monster.py
JaskRendix: True, pointless repetition

tuxemon/event/actions/choice_monster.py
JaskRendix: Done, added

tuxemon/states/choice/choice_item.py
JaskRendix: I added fix_width and fix_height months ago (when I coded the Journal), by looking at both now these are identical. So I did something bigger: renamed one in fix_measure, clarified and replaced.

tuxemon/states/choice/choice_item.py
JaskRendix: Yeah, see reply above

tuxemon/states/choice/choice_item.py
JaskRendix: Yeah, it was a big chunk of information, replaced it with a small description

Sanglorian:
Looks really nice! Thanks for this

Grimmys (approved):
Looks better, thanks.

bitcraft (approved):

PR #2082, opened 2023-11-09
update translated dialog (event action)

JaskRendix:
PR updates **translated_dialog** by removing duplicates (passages) as well as by resorting information inside.

Eventually we can consider to remove the field "menu" from open_dialog, since it ends up in opening a ChoiceState inside DialogState, so it's like a never-ending cycle (at the moment there are no translated_dialog with this function, everyone uses translated_dialog_choice) + neither the 54 references of open_dialog in the code

added **menu** commit:
- removes the **menu**
- fixes a typehint (**DialogState(PopUpMenu)** into **DialogState(PopUpMenu[None])**)

add **dialog_box** commit:
- simplifies the formula (calculation **dialog_box**);
- makes clear that **text_area** is the **dialog_box**;


tuxemon/event/actions/translated_dialog.py
```diff
@@ -52,41 +37,25 @@ class TranslatedDialogAction(EventAction):
...
+        key = process_translate_text(self.session, self.raw_parameters, [])
+
         avatar = None
-        for param in self.raw_parameters[1:]:
-            if "=" in param:
-                replace.append(param)
-            else:
-                avatar = get_avatar(self.session, param)
+        if self.avatar:
+            avatar = get_avatar(self.session, self.avatar)
```
Grimmys: Just by curiosity, what's the difference between `avatar` and `self.avatar`?

If there is a difference, I think we should update the name of one of these identifiers to avoid confusion.

Grimmys (changes requested):
Looks like you removed a lot of clunky / unused code, that's nice congrats!
I didn't check if everything is working the same though.

tuxemon/event/actions/translated_dialog.py
JaskRendix: done, it's more clear

Grimmys (approved):

bitcraft:
conflicts

JaskRendix:
fixed

bitcraft (approved):

PR #2085, opened 2023-11-10
npcs from dict to list + iids

JaskRendix:
PR proposes to change **self.npcs** from dict to list; it adds also a method to get a NPC based on his instance_id

### DIFF
--- a/tuxemon/event/__init__.py
+++ b/tuxemon/event/__init__.py
@@ -42,7 +42,14 @@ class EventObject(NamedTuple):
     acts: Sequence[MapAction]
 
 
-__all__ = ["EventObject", "MapAction", "MapCondition", "get_npc"]
+__all__ = [
+    "EventObject",
+    "MapAction",
+    "MapCondition",
+    "get_npc",
+    "get_npc_by_iid",
+    "get_npc_pos",
+]
 
 
 def get_npc(session: Session, slug: str) -> Optional[NPC]:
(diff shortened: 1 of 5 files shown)


tuxemon/event/__init__.py
```diff
@@ -65,10 +71,29 @@ def get_npc(session: Session, slug: str) -> Optional[NPC]:
...
+
+    Parameters:
+        session: The session object.
+        iid: The iid of the NPC that exists on the current map.
+
+    Returns:
+        The NPC object or None if the NPC is not found.
+
+    """
+    from tuxemon.states.world.worldstate import WorldState
```
Grimmys, line 89: Why is the import done there instead of at the top of the file?

tuxemon/event/__init__.py
```diff
@@ -65,10 +71,29 @@ def get_npc(session: Session, slug: str) -> Optional[NPC]:
...
+
+    # Loop through the NPC list and see if the slug matches any in the list
```
Grimmys: Isn't this comment only describing what's happening inside of  `world.get_entity_by_iid`?
If so, I'm not sure if it's useful there.

Grimmys (commented):
Interesting changes, looks like the change from `dict` to `list` is simplifying a lot the implementation!
Thanks.

tuxemon/event/__init__.py
JaskRendix: No idea, I found like this, I try to move it

JaskRendix: (most likely due to a circular import)

tuxemon/event/__init__.py
JaskRendix: good catch

Grimmys (approved):

bitcraft (approved):

PR #2086, opened 2023-11-10
bump pygame-ce to 2.3.2

JaskRendix:
PR proposes to bump pygame-ce, there are no issues, but there are a few new typehints:
- fixes deprecation, from **tostring** to **tobytes**, I remove "utf-8" because it's already defined `encoding: str = "utf-8"`
```
/home/julesverne/Tuxemon/tuxemon/save.py:82: DeprecationWarning: pygame.image.tostring deprecated since 2.3.0
  pygame.image.tostring(screenshot, "RGB")
```
- fixes the following typehints:
```
tuxemon/menu/menu.py:490: note: Following member(s) of "Sequence[float]" have conflicts:
tuxemon/menu/menu.py:490: note:     Expected:
tuxemon/menu/menu.py:490: note:         def __getitem__(self, SupportsIndex, /) -> float
tuxemon/menu/menu.py:490: note:     Got:
tuxemon/menu/menu.py:490: note:         @overload
tuxemon/menu/menu.py:490: note:         def __getitem__(self, int, /) -> float
tuxemon/menu/menu.py:490: note:         @overload
tuxemon/menu/menu.py:490: note:         def __getitem__(self, slice, /) -> Sequence[float]
```

and the following related to this change

### DIFF
--- a/tuxemon/sprite.py
+++ b/tuxemon/sprite.py
@@ -18,7 +18,6 @@ from typing import (
     overload,
 )
 
-import pygame
 from pygame.rect import Rect
 from pygame.sprite import DirtySprite, Group, LayeredUpdates
 from pygame.sprite import Sprite as PySprite
@@ -44,7 +43,7 @@ dummy_image: Final = Surface((0, 0))
 class Sprite(DirtySprite):
     _original_image: Optional[Surface]
     _image: Optional[Surface]
-    _rect: pygame.rect.Rect
+    _rect: Rect
 
     def __init__(
         self,
@@ -73,8 +72,8 @@ class Sprite(DirtySprite):
     def draw(
         self,
         surface: Surface,
-        rect: Optional[pygame.rect.Rect] = None,
-    ) -> pygame.rect.Rect:
+        rect: Optional[Rect] = None,
+    ) -> Rect:
         """
         Draw the sprite to the surface.
 
@@ -97,8 +96,8 @@ class Sprite(DirtySprite):
     def _draw(
         self,
         surface: Surface,
-        rect: pygame.rect.Rect,
-    ) -> pygame.rect.Rect:
+        rect: Rect,
+    ) -> Rect:
         return surface.blit(self.image, rect)
 
     @property
@@ -328,11 +327,11 @@ class SpriteGroup(LayeredUpdates, Generic[_GroupElement]):
         # patch in indexing / slicing support
         return self.sprites()[item]
 
-    def calc_bounding_rect(self) -> pygame.rect.Rect:
+    def calc_bounding_rect(self) -> Rect:
         """A rect object that contains all sprites of this group."""
         sprites = self.sprites()
         if len(sprites) == 1:
-            return pygame.rect.Rect(sprites[0].rect)
+            return Rect(sprites[0].rect)
         else:
             return sprites[0].rect.unionall([s.rect for s in sprites[1:]])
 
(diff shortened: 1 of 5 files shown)


tuxemon/graphics.py
```diff
@@ -175,6 +175,7 @@ def load_sprite(
 
     """
     sprite = Sprite(image=load_and_scale(filename))
+    assert sprite.image
```
Grimmys: I'm don't think this assert is useful.

If there is no image, then it means `filename` was invalid.
Looking at the implementation of `load_and_scale` method and inner calls, if `filename` doesn't exist, it should already  fail at this point because we are not making any null check.

tuxemon/monster.py
```diff
@@ -638,6 +638,7 @@ def get_sprite(self, sprite: str, **kwargs: Any) -> Sprite:
             )
             if flair_path != MISSING_IMAGE:
                 flair_sprite = graphics.load_sprite(flair_path, **kwargs)
+                assert surface.image and flair_sprite.image
```
Grimmys: Again, I don't think this assert can ever fail.
`surface` and `flair_sprite` are in any case loaded using `graphics.load_sprite` or `graphics.load_animated_sprite` methods. These two methods should fail if the provided path is invalid.

tuxemon/sprite.py
```diff
@@ -115,7 +115,7 @@ def rect(self, rect: Optional[Rect]) -> None:
             self._needs_update = True
 
     @property
-    def image(self) -> Surface:
+    def image(self) -> Union[Surface, None]:
```
Grimmys: Let's use `Optional[Surface]` instead

tuxemon/states/combat/combat.py
```diff
@@ -323,6 +323,7 @@ def draw(self, surface: pygame.surface.Surface) -> None:
     def draw_hp_bars(self) -> None:
         """Go through the HP bars and redraw them."""
         for monster, hud in self.hud.items():
+            assert hud.image
```
Grimmys: This method is going to be called at every game iteration, do we really assume `hud.image` can be changed to null at any point?

If there is no external element which can modify the `hud.image`, then I would suggest to do this check at instance creation instead.

tuxemon/states/combat/combat.py
```diff
@@ -334,6 +335,7 @@ def draw_hp_bars(self) -> None:
     def draw_exp_bars(self) -> None:
         """Go through the EXP bars and redraw them."""
         for monster, hud in self.hud.items():
+            assert hud.image
```
Grimmys: Same as for `draw_hp_bars` method

tuxemon/states/combat/combat_animations.py
```diff
@@ -152,6 +152,7 @@ def animate_monster_release(
         feet = (feet_list[0], feet_list[1] + tools.scale(11))
 
         capdev = self.load_sprite(f"gfx/items/{monster.capture_device}.png")
+        assert capdev.image
```
Grimmys: Again, if `capdev.image` is undefined it should already fail earlier (`graphics.load_sprite` would fail)

tuxemon/states/combat/combat_animations.py
```diff
@@ -394,6 +396,7 @@ def build_left_hud() -> Sprite:
                 "gfx/ui/combat/hp_opponent_nohp.png",
                 layer=hud_layer,
             )
+            assert hud.image
```
Grimmys: Same, `hud.image` has to be defined at this point.

tuxemon/states/combat/combat_animations.py
```diff
@@ -406,6 +409,7 @@ def build_right_hud() -> Sprite:
                 "gfx/ui/combat/hp_player_nohp.png",
                 layer=hud_layer,
             )
+            assert hud.image
```
Grimmys: Same.

Grimmys (changes requested):
Most of my comments are related to the over-use of `assert`.
To be honest, I don't like assert pattern outside of tests, it looks archaic for me and not suited to OOP but that's a personal preference.

I just feel like we should at least avoid putting assertions in places where the variable cannot be undefined, because it's adding extra complexity to the code for nothing (since the statements are useless).

JaskRendix:
@Grimmys let's simplify, I put assert because of typehints

Grimmys (approved):

xirsoi (approved):

PR #2089, opened 2023-11-11
fix buy

JaskRendix:
PR fixes a bug. While working on #2088 I noticed it. It was possible to keep buying and going below zero with your wallet.
Another PR will follow as soon as it gets merged.


tuxemon/menu/quantity.py
```diff
@@ -135,8 +136,10 @@ def initialize_items(self) -> Generator[MenuItem[None], None, None]:
             price_tag = T.translate("shop_buy_free")
         else:
             price_tag = str(price)
-            if self.quantity == 0:
+            if self.quantity == 0 and price > wallet:
```
Grimmys, line 139: I don't get why we are adding an extra condition here.

So it means that now "soldout" is going to be displayed only if the item quantity is 0, and the price is bigger than the wallet ?
By extension, it means "soldout" would not be displayed if the quantity is 0, but the player has enough money in their wallet.

JaskRendix: @Grimmys I totally agree with you, but I found really annoying dispay soldout for an item "infinite" like the basic tuxeball. This is sort of a micro "patch", because I would like to add a field inside the class (eg a boolean related to infinite), so we can come up with more precise labels (eg if you are broke, then it would appear "you are broke" instead of "sold out")

but I need to wait #2088 for touching the class

Grimmys: Oh wait does it mean that quantity = 0 is equivalent to quantity = infinite?
So is it really possible that an item is soldout?

JaskRendix: It's possible that an item is soldout, but at the moment isn't enabled, because if you exit and re-enter inside the store, here we go again, the item is returned.

regarding the quantity, I added this `and price > wallet` because of this "issue", let's say you want to buy potions and you play with the quantity by increasing and decreasing it. Well without the  `and price > wallet`  if you select 0 (but you have money in the wallet, so you can afford it), it appears:
Screenshot_2023-11-15_12-52-21
while after  `and price > wallet`  this is what happens:
Screenshot_2023-11-15_12-53-14

Grimmys: Oh ok, it's just that I thought `quantity` meant something like "stock" (how many instances of the given item are still available in the shop), not the "quantity you want to buy".

My bad for not having read the code more carefully.

Still don't get why in your current example it's displaying $20 for the price instead of $0 (not buying anything should be... free, right?).

bitcraft (approved):

PR #2090, opened 2023-11-11
swim to yaml (Spyder)

JaskRendix:
PR continues the task after #2069 (scenario based events) and it moves all the events related to swim and swim encounters in the spyder.yaml.


bitcraft (approved):
great use of the yaml events!

PR #2091, opened 2023-11-11
upgrade current_state (event condition)

JaskRendix:
PR upgrades current_state. It'll allow to check with multiple states at the same time (as usual separated by ":")

eg. "is current_state CombatState:DialogState"

### DIFF
--- a/tuxemon/event/conditions/current_state.py
+++ b/tuxemon/event/conditions/current_state.py
@@ -42,5 +42,5 @@ class CurrentStateCondition(EventCondition):
         current_state = session.client.current_state
         assert current_state
         states = condition.parameters[0].split(":")
-        state = True if current_state.name in states else False
+        return current_state.name in states
         return state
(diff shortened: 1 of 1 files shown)


tuxemon/event/conditions/current_state.py
```diff
@@ -37,4 +41,6 @@ def test(self, session: Session, condition: MapCondition) -> bool:
         """
         current_state = session.client.current_state
         assert current_state
-        return current_state.name == condition.parameters[0]
+        states = condition.parameters[0].split(":")
+        state = True if current_state.name in states else False
```
Grimmys: If you want to return a boolean value, you can just directly return the result of the condition.
Meaning this can be simplified to:
```suggestion
        state = current_state.name in states
```

Grimmys: ... and I see the line below is just about returning this variable, so you can even inline the statements and just have:

```suggestion
        return current_state.name in states
```

Grimmys (changes requested):

JaskRendix:
@Grimmys you're right

bitcraft (approved):

PR #2092, opened 2023-11-11
fix load events from YAML (1/2)

JaskRendix:
PR fixes events from YAML files, since I noticed a bug. The bug was related to the fact that there wasn't order of execution. Traditional actions are done like this one:

```
[EventObject(id=13, name='Teleport to Cotton Town', x=5, y=10, w=2, h=1,
conds=
[
MapCondition(type='player_at', parameters=[], x=5, y=10, width=2, height=1, operator='is', name='cond1'),
MapCondition(type='player_facing', parameters=['down'], x=5, y=10, width=2, height=1, operator='is', name='cond2')],
acts=
[MapAction(type='transition_teleport', parameters=['spyder_cotton_town.tmx', '30', '35', '0.3'], name='act1')
```
There is an id `EventObject(id=13)` as well as `name='act1'`, `name='cond1'` and `name='cond2'`

The **id** is based on **uuid4**, it's a good solution, I tracked with **timeit** and it's faster than an enumerate of the dictionary, moreover we don't need to worry anymore about possible duplicates.

The event object (from the YAML) hasn't the **id** and **name**. It was `name=None`, so default None.
That's pretty bad, because without these components there is no sorting.
The result: events trigger without order.
Note: It doesn't happen with simple ones, but if it gets complicated the probability increases.

I need to admin that I got some suspects months ago, but now I was able to track down this issue.

For instance regarding **act**: `f"act{str(key).zfill(2)}`
so it'll automatically generate **act01**, **act02**, etc.


xirsoi (dismissed):
Nice work!

bitcraft (approved):

PR #2093, opened 2023-11-11
learning technique, removal and overwriting

JaskRendix:
PR centralizes the learning of new techs as well as the removal / overwriting; it removes the hardcoded pieces and it creates a bundle of actions to deal with it. Actions because these can be used and adapted by modders.

actions:
- **get_monster_tech** (it'll open the technique menu, here it'll possible to select the technique)
- **overwrite_tech** (it'll be possible to replace the existing tech with a new one)
- **remove_tech** (it'll be possible to remove the existing tech)

condition:
- **check_max_techs** (it'll be possible to check if one of the party's monster has more than the nr. max of tech, here it'll possible to specify the number, by default it'll be like the constant, so 4)

some uses:
1. `get_player_monster monster_id` (pick the monster, the monster will be saved in the variable monster_id)
2. `get_monster_tech tech_id,monster_id` (pick the monster's technique, technique will be saved in the variable tech_id)
3. `overwrite_tech tech_id,ram` or `remove_tech tech_id` (decide if remove it or overwrite it)

Screenshot_2023-12-21_09-44-08
Screenshot_2023-12-21_09-40-8


tuxemon/event/actions/overwrite_tech.py
```diff
@@ -0,0 +1,59 @@
...
+    eg. "overwrite_tech name_variable,peck"
+
+    """
+
+    name = "overwrite_tech"
+    removed: str
+    added: str
+
+    def overwrite(self, monster: Monster, removed: Technique) -> None:
+        slot = monster.moves.index(removed)
```
bitcraft, line 41: the logic here is fine, but im wondering why you you decided not to use the list indexes
```python
slot = monster.moves.index(removed)
monster.moves[slot] = added
```
not worth changing, but im just curious

bitcraft (approved):

PR #2096, opened 2023-11-12
fix load events from YAML (2/2)

JaskRendix:
PR follows #2092 and it implements default values for **x**, **y**, **width** and **height**, this will allow to write the event without those coordinates. I implemented in this way because of typehints, if there is a short one feel free to share.
```
  Teleport Faint:
    actions:
    - teleport_faint
    conditions:
    - is player_defeated
    - not location_type clinic
    - not location_name wayfarer_inn1:wayfarer_inn2:scoop3:scoop4:paper_town
    height: 1
    width: 1
    x: 0
    y: 0
    type: "event"
```
to
```
  Teleport Faint:
    actions:
    - teleport_faint
    conditions:
    - is player_defeated
    - not location_type clinic
    - not location_name wayfarer_inn1:wayfarer_inn2:scoop3:scoop4:paper_town
    type: "event"
```
so we can remove these:
```
    height: 1
    width: 1
    x: 0
    y: 0
```


tuxemon/map_loader.py
```diff
@@ -66,23 +67,46 @@ def load_events(self, path: str) -> Iterator[EventObject]:
...
+                if event_data.get("x") is None
+                else (event_data.get("x", int))
+            )
+            y = (
+                1
+                if event_data.get("y") is None
+                else (event_data.get("y", int))
+            )
+            w = (
+                0
```
xirsoi: I didn't know an event could have a width or height of 0.
I assume you tested this?

xirsoi (commented):
Mostly looks good, just the one question.

tuxemon/map_loader.py
JaskRendix: right, you're right, I missed it

JaskRendix: it was x and y (0), while w and h (1)

tuxemon/map_loader.py
```diff
@@ -66,23 +67,46 @@ def load_events(self, path: str) -> Iterator[EventObject]:
...
 
         for name, event_data in yaml_data["events"].items():
+            _id = uuid.uuid4().int
             conds = []
             acts = []
-            x = event_data.get("x")
-            y = event_data.get("y")
-            w = event_data.get("width")
-            h = event_data.get("height")
+            x = (
```
bitcraft: seems like you could write this as `event_data.get("x", 0)`?

JaskRendix: you're right!

JaskRendix:
added a fix for change_bg that I noticed while testing right now

bitcraft (approved):

PR #2097, opened 2023-11-12
change call_event (event action) + id (load_events) from maps

JaskRendix:
PR:
- proposes to delete the **call_event** event action, it's unused and calling an event from the id is really tricky, considering the fact the ids in Tiled are progressively assigned;
- proposes to set **uuid** as id as well as for events loaded from maps (as well as I proposed here for #2092 for events from YAML);

### DIFF
--- a/tuxemon/event/actions/call_event.py
+++ b/tuxemon/event/actions/call_event.py
@@ -12,25 +12,25 @@ from tuxemon.event.eventaction import EventAction
 @dataclass
 class CallEventAction(EventAction):
     """
-    Execute the specified event's actions by id.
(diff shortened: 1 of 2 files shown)


xirsoi (dismissed):
Looks good!

I think having a "call event by name" would be a cool/useful feature for the future.
Note to self: make a ticket for a call event by name action.

JaskRendix:
@xirsoi it was my first idea "call event by name", but I wasn't sure someone would be find it useful, so I reverse to event_name

Grimmys (approved):

xirsoi (approved):

PR #2098, opened 2023-11-12
update change bg (event action)

JaskRendix:
PR:
- renames the folder from **bg** to **idle** (as opposite of transition);
- renames **bgstate** into **imagestate** and splits **imagestate** in **imagestate** and **colorstate** (folder organization);
- defines the **background** (inside change_bg as optional, so for ending it it would be necessary put "end");
- implements the possibility to set background as single colors through the rgb;

Removes from the constant ColorLike the str and pygame.Color, the tuple(int, etc) is accepted in pygame.fill.

### DIFF
--- a/tuxemon/graphics.py
+++ b/tuxemon/graphics.py
@@ -32,7 +32,6 @@ logger = logging.getLogger(__name__)
 
 ColorLike = Union[
     pygame.color.Color,
-    str,
     tuple[int, int, int],
     tuple[int, int, int, int],
 ]
--- a/tuxemon/states/idle/color_state.py
+++ b/tuxemon/states/idle/color_state.py
@@ -1,0 +1,28 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+from typing import Optional
+
+from tuxemon import prepare
+from tuxemon.graphics import string_to_colorlike
+from tuxemon.menu.menu import PygameMenuState
+from tuxemon.menu.theme import get_theme
+from tuxemon.platform.events import PlayerInput
+
+
+class ColorState(PygameMenuState):
+    """
+    It imposes a specific color over the world, where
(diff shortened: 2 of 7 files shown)


xirsoi (dismissed):
Good work cleaning this up!

tuxemon/graphics.py
```diff
@@ -31,8 +31,6 @@
 
 
 ColorLike = Union[
-    pygame.color.Color,
```
Grimmys, line 34: Why removing `pygame.color.Color` as valid alias for a color?

I feel like pygame has some nice named constants for color.

tuxemon/states/idle/color_state.py
```diff
@@ -0,0 +1,28 @@
...
+from tuxemon.graphics import string_to_colorlike
+from tuxemon.menu.menu import PygameMenuState
+from tuxemon.menu.theme import get_theme
+from tuxemon.platform.events import PlayerInput
+
+
+class ColorState(PygameMenuState):
+    """
+    Menu for the change of background: color
+    It acts as opposite of transition, it's not limited temporarily.
```
Grimmys: What do you mean by "it's not limited temporarily"?

Grimmys: OK, I see it was already this way before you just copy/pasted it... Hmm... Please if you have any idea about what it could mean, feel free to update the documentation because this sentence doesn't make any sense imo.

Grimmys (changes requested):

tuxemon/states/idle/color_state.py
JaskRendix: because transitions (eg flash, lasts only seconds and then disappear), and yeah, I copy paste, but I was the one who wrote that in the first place

tuxemon/graphics.py
JaskRendix: Because pygame.color.Color accepts tuples too, but I can leave it, no issue

tuxemon/states/idle/color_state.py
JaskRendix: I hope it makes sense now, I don't know what else I can add.

xirsoi (approved):

PR #2099, opened 2023-11-13
remove "interacts" (class TuxemonMap)

JaskRendix:
PR proposes of getting rid of **inits** an **interacts**, because it looks like these are completely useless.

While I was working on #2092 I noticed `event_type == "interact"`, so I decided to track it down, and I realized that both:
```
        interacts: Sequence[EventObject],
```
are related to:
```
            elif obj_type == "interact":
                interacts.append(self.load_event(obj, tile_size))
```
but in maps we have only type="collision" and type="event"

So both lists aren't getting any **EventObject**.

I tried to remove these completely from the game and there are no issues, the game run flawlessly, player can speak, use the PC, etc, I'm not sure, but probably these are from an old version of the game.

I removed also `obj_type = getattr(obj, "class")` because this was added back then #1556 when the main developer of Tiled announced to switch from `type="event"` to `class="event"`, but then he returned on his decision. Back then we acted promptly to fix a potential issue.


xirsoi:
interacts was originally built as a shorthand to avoid needing to say that the player must be adjacent and facing the event and have pressed the interact button.

I think it should stay.

xirsoi:
> interacts was originally built as a shorthand to avoid needing to say that the player must be adjacent and facing the event and have pressed the interact button.
> 
> I think it should stay.

Turns out this feature was never completed, so I'm fine with it going. I'll make a ticket or something to implement some handy shorthand event types, in the vein of what was proposed here.

xirsoi (dismissed):
👍

tuxemon/client.py
```diff
@@ -280,8 +276,6 @@ def _send_event(
             if maybe_game_event is None:
                 break
             game_event = maybe_game_event
```
Grimmys: Since this method logic is now even simpler, I would suggest dropping one of the two variables to make it even clearer.

I don't think it's required to have a temporary variable ; we can just directly assign the returned event to `game_event`.

Grimmys (changes requested):
Nice cleaning!

tuxemon/client.py
```diff
@@ -273,11 +273,8 @@ def _send_event(
         """
         for state in self.active_states:
             maybe_game_event = state.process_event(game_event)
-            if maybe_game_event is None:
-                break
-            game_event = maybe_game_event
-
-        return maybe_game_event
+            return maybe_game_event
```
Grimmys: At this point, we can directly return the result of `process_event` call, no need to store it in a variable.
```suggestion
            return state.process_event(game_event)
```

Grimmys (changes requested):

tuxemon/client.py
Grimmys: Btw why did you remove the break logic & the "bubble out" of the event to the next state logic?
Cf. my comments on Discord.

JaskRendix: @Grimmys I tried to see if without the break logic was working and I didn't notice any weird behavior, opening, closing, saving, loading, teleporting, no issue with any of these.

JaskRendix: fixed

bitcraft:
there are existing events which are the "init" type.  what do you plan to do about them?  you don't think that init-type events are useful?

JaskRendix:
Some numbers:
- type="event" are 3.056 in total among all the maps
- type="collision"  are 4694 in total among all the maps
- type="init" are 9 in total among all the maps (3/9 unused maps) > replaced with event
- type="interact" are 0 in total among all the maps 

> there are existing events which are the "init" type. what do you plan to do about them? you don't think that init-type events are useful?

@bitcraft you're right, but look at these numbers. Among the maps there are only 9 init  out of 7500+

```
  <object id="23" name="Play Music" type="init" x="144" y="16" width="16" height="16">
   <properties>
    <property name="act1" value="play_music music_house_downstairs"/>
    <property name="cond1" value="not music_playing music_house_downstairs"/>
   </properties>
  </object>
```
but this same kind of event is done also with:
```
  <object id="89" name="Route Music" type="event" x="0" y="0" width="16" height="16">
   <properties>
    <property name="act1" value="play_music music_cathedral_theme"/>
    <property name="cond1" value="not music_playing music_cathedral_theme"/>
   </properties>
  </object>
```

bitcraft:
i see you change the `inits` to `event`.  with no conditions, they would just execute every tick?  the inits are executed exactly once.  and it seems now they would execute every tick.  i realize that there is a check that would prevent the music from starting over again, etc, but still seems wasteful.

JaskRendix:
I didn't notice anything weird, the song starts, ends and restarts all over again. The majority of maps plays music with event and it restarts without any issue.

Ok, checked both play_music and play_sound, cleaned up, added loop optional parameter for play_music

bitcraft:
there are more changes that just removing init here, and i can merge it, but i strongly feel that it will be added back some time later.  because there is a use for events which need to fire at the start of a map, and won't run again.  for music, without inits, the music event is going to fire every frame, which is really a small thing considering that the action itself will just run without changing.  the end result is the same with or without `init`, but without `init` the music event is checked every frame needlessly.

JaskRendix:
ok, reversed init

bitcraft (approved):

PR #2102, opened 2023-11-14
npc look (event action)

JaskRendix:
PR implements a new event action called **npc_look**, it'll allow the NPC to look around without wander.

with this it'll look in all directions:
`<property name="act1" value="npc_look spyder_shopassistant"/>`
with this it'll look only left and right
`<property name="act1" value="npc_look spyder_shopassistant,,right:left"/>`

### DIFF
--- a/tuxemon/event/actions/npc_look.py
+++ b/tuxemon/event/actions/npc_look.py
@@ -2,10 +2,9 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-import itertools
 import random
 from dataclasses import dataclass
-from typing import Optional, final
+from typing import Optional, cast, final
 
 from tuxemon.event import get_npc
 from tuxemon.event.eventaction import EventAction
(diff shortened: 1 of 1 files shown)


xirsoi (dismissed):
Looks great! Thanks for the PR 👍

tuxemon/event/actions/npc_look.py
```diff
@@ -0,0 +1,96 @@
...
+                if state.name == "DialogState":
+                    # retrieve NPC talking to
+                    if player.facing == "down":
+                        return
+                    elif player.facing == "up":
+                        return
+                    elif player.facing == "left":
+                        return
+                    elif player.facing == "right":
+                        return
```
Grimmys: I don't get it, why put all these conditions in different blocks if at the end the consequence is always the same?

Change suggestion:
```suggestion
                    if player.facing in ("down", "up", "left", "right"):
                        return
```

Btw, aren't `down`, `up`, `left `and `right `all the possible facing the player can take? If so, it means the condition is always `true`... At this point, better just always `return`.

Grimmys (changes requested):

tuxemon/event/actions/npc_look.py
JaskRendix: you're right

JaskRendix: fixed a couple of more stuff, removing some unused import, setting list(str) and casting at the end as Direction because npc.facing accepts Direction (mypy typehints)

xirsoi (approved):

PR #2103, opened 2023-11-14
menu (event action)

JaskRendix:
PR:
-implements menu (event action);
-adds a series of bool in npc.py;
-edits world_menu.py, so it'll be possible to hide the labels;
-renames **menu_items_map** into **menu** and  **change_state** into **change** (more compact and less_long_text)

**menu** (event action) will be able to enable or disable certain label of the menu

Maybe a modder for a quest needs to disable the monster or bag menu:
`<property name="act1" value="menu disable,menu_monsters"/>`
or
`<property name="act1" value="menu enable,menu_bag"/>`
Screenshot_2023-11-16_21-21-04

### DIFF
--- a/tuxemon/event/actions/menu.py
+++ b/tuxemon/event/actions/menu.py
@@ -1,0 +1,59 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Optional, final
+
+from tuxemon.event.eventaction import EventAction
+
+
+@final
+@dataclass
+class MenuAction(EventAction):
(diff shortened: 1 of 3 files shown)


tuxemon/event/actions/menu.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if self.menu is None:
+            player.menu_bag = result
+            player.menu_load = result
+            player.menu_monsters = result
+            player.menu_save = result
+            player.menu_options = result
+            player.menu_player = result
+
+        if self.menu and self.menu == "menu_bag":
```
xirsoi: i think a switch statement would fit better here.

JaskRendix: you're right, it looks bad

Sanglorian:
Really nice - this will make things simpler and clearer for players in the early game. Thanks!

bitcraft (approved):

PR #2108, opened 2023-11-18
set layer (event action)

JaskRendix:
PR:
- implements **set_layer** (event action) and purges the hardcoded part of the night/day cycle;
- adds the event to Spyder scenario (day night cycle enabled by default since June);

the event action will allow to specify the rgb:
`set_layer 0:0:128:128` this is the night "color", as you can see 128, so transparent
`set_layer` remove the layer

note: it could trigger a conflict with other PR since it adds an event to spyder.yaml, but it's not big deal, easy to fix


bitcraft:
conflicts

JaskRendix:
fixed

tuxemon/states/world/worldstate.py
```diff
@@ -262,6 +263,13 @@ def set_transition_surface(self, color: ColorLike) -> None:
         )
         self.transition_surface.fill(color)
 
+    def set_layer(self) -> None:
+        self.layer = pygame.Surface(
```
bitcraft: it would be better if this Surface were not created/destroyed each frame.  holding a reference to it by something simple like making it an instance attribute would be better.  pygame surface allocations can be slow, since its all software rendering.

bitcraft: this line should be moved to `__init__`.  we shouldn't be creating large Surfaces each frame, just to remove them later.

bitcraft (approved):

PR #2109, opened 2023-11-18
set bubble (event action)

### LINKED ISSUE #667: Add bubbles over npcs
Adding an exclamation point over some npcs could help move the story along a lot. Here is an example: https://github.com/pret/pokered/blob/master/gfx/emotion_bubbles.png

We can't use those as they are ripped directly from pokemon red though. Plus, it would be nice to have them in color

JaskRendix:
follows #2108 and it's a feature, fix #667 

PR creates a new folder in **gfx** called **bubbles** ( bubbles are 16x16)

**set_bubble** will allow to show a bubble above the player's head.

bubbles: https://wiki.tuxemon.org/index.php?title=File:AH_SpriteSheet_Balloon.png

eg:
```
    <property name="act1" value="set_bubble player,exclamation"/> --> adds bubble
    <property name="act2" value="wait 1"/>
    <property name="act2" value="translated_dialog something"/>
    <property name="act3" value="set_bubble player"/> ---> remove bubble
```
screenshot:
Screenshot_2023-11-18_13-51-53

or another example, setting up two event (first makes appear the bubble, the second cleans up):
```
  <object id="32" name="gr" type="event" x="160" y="112" width="16" height="16">
   <properties>
    <property name="act1" value="set_bubble player,note"/>
    <property name="cond1" value="is player_at"/>
    <property name="cond2" value="is player_facing down"/>
   </properties>
  </object>
  <object id="32" name="gr" type="event" x="160" y="112" width="16" height="16">
   <properties>
    <property name="act1" value="set_bubble player"/>
    <property name="cond1" value="is player_at"/>
    <property name="cond2" value="not player_facing down"/>
   </properties>
  </object>
```
screenshot:
Screenshot_2023-11-18_13-31-11


tuxemon/states/world/worldstate.py
```diff
@@ -293,6 +296,13 @@ def set_layer(self) -> None:
         self.layer.fill(self.layer_color)
         self.screen.blit(self.layer, (0, 0))
 
+    def set_bubble(self) -> None:
+        if self.bubble:
+            for npc, surface in self.bubble.items():
+                cx, cy = self.get_pos_from_tilepos(npc.tile_pos)
```
bitcraft: these should be sorted by coordinates value so that they overlap in a predictable way.  if not sorted, they will flicker

bitcraft (approved):

PR #2111, opened 2023-11-19
update PCState (state)

JaskRendix:
PR:
- updates PcState (now PyGameMenu);
- splits PcState and MultiplayerMenu (separate files);
- updates MultiplayerMenu (now PyGameMenu);
- edits some names (too long)
- fix bug (it was possible to move monsters between boxes even above the limit);
- fix bug (it was possible to drop off monsters in full boxes);
- fix bug (it was possible to see and interact with hidden boxes, now the boxes will not appear anymore, neither during picking up or dropping off);

regarding names too long from:
`class MonsterBoxChooseStorageState(MonsterBoxChooseState):`
to
`class MonsterStorageState(MonsterBoxState):`

no monsters/items inside the boxes? then menu storage doesn't appear;
player has only 1 monster or not monster at all? then menu dropoff doesn't appear;

Screenshot_2023-11-19_10-06-28

### DIFF
--- a/tuxemon/states/multiplayer/__init__.py
+++ b/tuxemon/states/multiplayer/__init__.py
@@ -1,0 +1,148 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+from collections.abc import Callable, Generator
+
+import pygame_menu
+
+from tuxemon.animation import Animation
+from tuxemon.locale import T
+from tuxemon.menu.input import InputMenu
+from tuxemon.menu.interface import MenuItem
+from tuxemon.menu.menu import PopUpMenu, PygameMenuState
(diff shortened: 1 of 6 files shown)


tuxemon/states/multiplayer/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+    code salvaged from commit 6fa20da714c7b794cbe1e8a22168fa66cda13a9e
+    """
+
+    shrink_to_items = True
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        menu: list[tuple[str, MenuGameObj]] = []
```
bitcraft, line 40: minor, but when you are making a list, you don't need to use `append` for each item if it doesn't change.
```python
menu: list[tuple[str, MenuGameObj]] = [
   ("multiplayer_host_game", self.host_game),
   ("multiplayer_scan_games", self.scan_for_games),
   ("multiplayer_join_game", self.join_by_ip),
]
```

bitcraft (approved):

PR #2112, opened 2023-11-19
update NuPhone (state)

JaskRendix:
PR:
- splits NuPhone into the components: contacts, map, banking and phone, it'll simplify future operations;
- centralizes RGBs in one single file;


tuxemon/states/pc_locker/__init__.py
```diff
@@ -257,11 +256,7 @@ def __init__(self, box_name: str) -> None:
 
         # menu
         theme.title = True
-        theme.title_background_color = (197, 232, 234)
-        theme.title_font_size = round(0.025 * width)
-        theme.title_font_color = (10, 10, 10)
-        theme.title_close_button = False
-        theme.title_bar_style = MENUBAR_STYLE_ADAPTIVE
+        theme.title_font_size = 30
```
bitcraft: notice that `theme.title_font_size = round(0.025 * width)`.  width is based off screen size, which means that it is scaled to the pixel size of the window.  by setting `30` here, the fotn will only be the correct size at whatever resolution you are using now.  we support multiple resolutions, so this value should be scaled to the screen somehow.  you can use `width`, or prepare.SCALE so the font is always the right size.

tuxemon/states/phone/phone.py
```diff
@@ -0,0 +1,153 @@
...
+                drawing_position=POSITION_CENTER,
+            )
+            new_image.scale(prepare.SCALE, prepare.SCALE)
+            menu.add.banner(
+                new_image,
+                change,
+                selection_effect=HighlightSelection(),
+            )
+            menu.add.button(
+                label, action=partial(uninstall, item), font_size=15
```
bitcraft: font_size needs to be scaled.  same on line 105

tuxemon/states/phone/phone.py
```diff
@@ -0,0 +1,153 @@
...
+            drawing_position=POSITION_CENTER,
+        )
+        theme = get_theme()
+        theme.scrollarea_position = locals.POSITION_EAST
+        theme.background_color = background
+        theme.widget_alignment = locals.ALIGN_CENTER
+
+        # menu
+        theme.title = True
+        theme.title_font_size = 30
```
bitcraft: scaled font size

tuxemon/states/phone/phone_banking.py
```diff
@@ -0,0 +1,177 @@
...
+        menu: pygame_menu.Menu,
+    ) -> None:
+        if "bank_account" not in self.player.money:
+            self.player.money["bank_account"] = 0
+        bank = self.player.money["bank_account"]
+        money = self.player.money["player"]
+        menu.add.label(
+            title=T.translate("wallet") + ": " + str(money),
+            label_id="wallet",
+            font_size=20,
```
bitcraft: scale font size, and line 46

tuxemon/states/phone/phone_banking.py
```diff
@@ -0,0 +1,177 @@
...
+            self.player.money["player"] += amount
+            self.player.money["bank_account"] -= amount
+
+        menu.add.vertical_margin(100)
+        # deposit
+        menu.add.button(
+            title=T.translate("deposit").upper(),
+            action=choice_deposit,
+            button_id="deposit",
+            font_size=20,
```
bitcraft: more scaled font size, and line 137, 159

bitcraft (changes requested):
please make sure all the font size values are scaled to the screen

tuxemon/states/phone/phone_map.py
```diff
@@ -0,0 +1,153 @@
...
+
+MenuGameObj = Callable[[], Any]
+
+
+def fix_measure(measure: int, percentage: float) -> int:
+    """it returns the correct measure based on percentage"""
+    return round(measure * percentage)
+
+
+class NuPhoneMap(PygameMenuState):
```
bitcraft, line 27: i think this is OK as just some experimental code, but because it is hardcoding the map names here, it can't really stay like this forever, since it means the codebase is tied to a specific thing.  i don't have a good solution right now, but i think something along the lines of a game variable which contains the maps would be used to generate this list.  if you don't have a solution, please put in a comment like the following
```
This is experimental code for the NuPhone Map.  Currently these map names are are hardcoded,
but in the future, we will need to use some other data source to get the map names so that this
code is not directly tied to any specific maps.
```

tuxemon/states/phone/phone_map.py
```diff
@@ -0,0 +1,153 @@
...
+    ) -> None:
+        width, height = prepare.SCREEN_SIZE
+        lab1 = menu.add.label(
+            title=T.translate("leather_town"),
+            selectable=True,
+            float=True,
+            selection_effect=HighlightSelection(),
+            selection_color=(232, 48, 48),
+            font_size=round(0.015 * width),
+            background_color=(232, 48, 48),
```
bitcraft: just curious, can this use the theme, or do we want to hardcode the colors/fonts?

tuxemon/states/phone/phone_map.py
JaskRendix: ok

tuxemon/states/phone/phone_map.py
JaskRendix: we can use the theme, no use to hardcode

bitcraft (approved):

PR #2116, opened 2023-11-21
update shop

JaskRendix:
PR updates shop:
- if **no money** in the wallet, then it'll **not be possible to select** certain items;
- the items will be displayed by **price sorting** (from lower to higher);
- **prices** will be displayed in the main window;
- adds **buy** and **sell** inside the PO files, previously both were hardcoded > no translation;
- saves quantities of shop items inside **variables**;
- if a **quantity = 0**, then it'll appear **soldout** label  (see below);
- adds a **field_validator** inside **db.py** for checking if the variable is in the correct format (months ago I added this possibility, so modders can make appear certain items after quests, etc.)
- some edits in **set_economy**, it removes the default shop (it doesn't exist), it removes the raise (since there is the check inside db.py);
- corrects **Mega** Potion name, before it was Imperial, but Imperial exists already;

Screenshot_2023-11-21_12-25-05
Screenshot_2023-11-21_13-48-07
Screenshot_2023-11-22_15-03-57

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -46,7 +46,7 @@ msgid "imperial_potion_description"
 msgstr "Heals a monster by 500 HP."
 
 msgid "mega_potion"
-msgstr "Imperial Potion"
+msgstr "Mega Potion"
 
 msgid "mega_potion_description"
 msgstr "Heals a monster by 200 HP."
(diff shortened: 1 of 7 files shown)


JaskRendix:
This is just a patch to upgrade the shop and to offer a solid mechanism.
The next step will be getting rid of menuitem, change to PyGameMenu and create a new UI to distinguish the shop from the bag.

xirsoi (approved):

PR #2119, opened 2023-11-23
fix libpng warning

JaskRendix:
PR fixes two PNGs with the following warning: **libpng warning: iCCP: known incorrect sRGB profile**


bitcraft (approved):
great!  i wondered where these errors were coming from

PR #2120, opened 2023-11-24
update RegionProperties class

JaskRendix:
PR:
- replaces **RegionProperties** class (see below);
- renames **continue** into **endure** because continue is already used in Python and it can create issues;
- cleans up the **core_outdoor**, hidden lines;
- adds enter/exit/endure to **core_outdoor_nature** (bump, go down, not up, etc);
- adds enter/exit/endure to **other tilesets** (bump, go down, not up, etc);
- updates all the labels into **enter_from**, **exit_from**, **endure** (before there was exit, exit_to, enter, enter_to);
- updates all the value above (enter_from) inside other file tsx;
- removes the collision below from the majority of maps (Xero and Spyder), since it works without;
- creates the **My_tuxemon_sheet_BW**.tsx so the 37707 are updated;
- enables **enter_from = (empty)** (or **exit_from** or **endure**), it works, the tile isn't accessible;
- casting will be removed in #2121;

from 
```
class RegionProperties(RegionPropertiesOptional):
    enter: Sequence[Direction]
    exit: Sequence[Direction]
```
to
```
class RegionProperties(NamedTuple):
    enter_from: Sequence[Direction]
    exit_from: Sequence[Direction]
    endure: Sequence[Direction]
    entity: Optional[NPC]
    key: Optional[str]
```
with this change it would be possible to purge the following collision zone:
```
  <object id="179" type="collision" x="352" y="160" width="112" height="16">
   <properties>
    <property name="endure" value="down"/>
    <property name="enter_from" value="up"/>
    <property name="exit_from" value="down"/>
   </properties>
  </object>
```

----

**endure** means "continue to walk in a precise direction"
if in the tileset.tsx there is **endure=left**, it means that the player continues walking **left**;
endure is a **sequence** (possible multiple values)

```
            tile = self.world.collision_map[self.tile_pos]
            if tile and tile.endure:
                _direction = (
                    self.facing if len(tile.endure) > 1 else tile.endure[0]
                )
                self.move_one_tile(_direction)
            else:
                pass
```
if there are more than 1 direction eg (up and left), then the player will continue in the direction he/she entered the tile, so it takes the direction from **self.facing**


tuxemon/map.py
```diff
@@ -290,36 +295,63 @@ def extract_region_properties(
 
     """
     # this could use a rewrite or re-thinking...
+    label = None
+    endure: list[Direction] = []
     enters: list[Direction] = []
     exits: list[Direction] = []
-    new_props: RegionProperties = {
```
bitcraft, line 295: i agree changing `continue` to a different word is a good idea.  i think `endure` is ok, but after a few months, maybe we will all forget what it means.  please add this comment
```
using word "endure" because continue is already used in Python and it can create issues.
endure means "continue to walk in a precise direction".
if in the tileset.tsx there is endure=left, it means that the player continues walking left.
endure is a sequence (possible multiple values);
if there are more than 1 direction eg (up and left), then the player will continue in the direction he/she entered the tile, so it takes the direction from self.facing
```

tuxemon/map.py
```diff
@@ -280,7 +285,7 @@ def extract_region_properties(
     Given a dictionary from Tiled properties, return a dictionary
     suitable for collision detection.
 
-    Uses `exit_to`, `enter_from`, and `continue` keys.
+    Uses `exit`, `enter`, and `continue` keys.
```
bitcraft: s/continue/endure

also, i see you are standardizing enter/exit.  i think you are going with enter_from and exit_from?  if so, please update this to match

tests/tuxemon/test_map_loader_pytmx.py
```diff
@@ -34,32 +37,26 @@ def test_result_is_point_and_properties_tuple(self):
...
     def test_result_properties_is_not_same_object_as_input(self):
         properties = self.result[0][1]
         self.assertIsNot(properties, self.properties)
 
     def test_result_each_properties_are_not_same_object(self):
-        properties = [i[1] for i in self.result]
+        properties = [i[1] for i in self.result if i[1]]
         comps = combinations(properties, 2)
         self.assertTrue(all(is_not(*i) for i in comps))
-
```
bitcraft, line 52: why is this test being removed?

tuxemon/map.py
```diff
@@ -290,36 +295,63 @@ def extract_region_properties(
...
+    Parameters:
+        direction: str (eg. enter_from = "direction")
+
+    Returns:
+        List with Direction/s
+
+    """
+    try:
+        props = direction.split(",")
+    except:
```
bitcraft: i would prefer that we don't cover up errors with a try/except and just allow the exception to crash the game

bitcraft (commented):
great to see the maps simplified and the region properties used!  really excited about it.  please see the comments

tuxemon/map.py
JaskRendix: yes, I'm going to update

tuxemon/map.py
JaskRendix: ok

tests/tuxemon/test_map_loader_pytmx.py
JaskRendix: mistake

bitcraft (dismissed):

bitcraft:
conflicts

JaskRendix:
solved

xirsoi (approved):

PR #2122, opened 2023-11-25
test_npcs.tmx + test_npcs.yaml + 2 tests + fix 2 bugs

JaskRendix:
PR:
- creates the **test_npcs.json** where there are 500 NPCs called test_npc001, test_npc002, etc.
- creates a new map called **test_npcs** that will host all the 500 NPCs (testing purpose);
- creates the **test_spyder_cotton_town**.yaml, the file that creates 160 NPCs inside the town, if someone wants to test it, then he/she needs to rename it: **spyder_cotton_town.yaml** (spyder_cotton_town has an area of 1600 tiles);
- creates the **test_spyder_cotton_scoop**.yaml, the file that creates 13 NPCs inside the scoop, if someone wants to test it, then he/she needs to rename it: **spyder_cotton_scoop.yaml** (spyder_cotton_scoop has an area of 143 tiles, but only 73 without collisions, it means 13/71 tiles (18.3% occupancy);
- fixes a bug (crash) if someone tries to load a map without indicating a scenario;
- fixes a bug related to the loading of the game; how to reproduce the bug, save in Paper Town, enter the Player house, see mom happily wandering, load the game, reenter the house, mom is stuck
- adds a new parameter in worldstate called **map_area** and it's nothing more than the area of the map (it exists **map_size** but it returns a tuple(40,40), **map_area** will return 1600)

### DIFF
--- a/mods/tuxemon/db/npc/test_npcs.json
+++ b/mods/tuxemon/db/npc/test_npcs.json
@@ -1,0 +1,5002 @@
+[
+  {
+    "slug": "test_npc001",
+    "template": [
+      {
+        "sprite_name": "professor",
+        "combat_front": "professor",
+        "slug": "professor"
+      }
+    ]
+  },
+  {
+    "slug": "test_npc002",
+    "template": [
+      {
+        "sprite_name": "professor",
+        "combat_front": "professor",
+        "slug": "professor"
+      }
+    ]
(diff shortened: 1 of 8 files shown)


bitcraft (approved):
great addition, thanks~

Issue #771, opened 2020-04-28
Pathfinding will sometimes overlap NPCs


bitcraft:
If two NPCs are pathfinding, they may both choose paths that overlap and the sprites will occupy the same tile.  Might be fixed by forcing a tile to be "occupied" when an NPC is moving into it, rather than inside it.  Another solution would be to allow a NPC to wait for a moving NPC to leave the tile before moving into it.  Both would probably be fine if implemented together.

This issue is purely cosmetic.

Related to #662

Issue #662, opened 2020-02-10
Can sometimes walk on top of npcs


kerizane:
If you attempt to walk into a tile at the same time an npc does then you'll both end up on the same tile.

We should be stricter about checking if a tile is blocked or not, maybe "lock" a tile after you decide to move there.

bitcraft:
yeah, i agree on "locking a tile".  i think that if an pc is even a single pixel moving into their next tile, they should be considered in that tile, and the tile they are moving from.

bitcraft:
Ill take this one

kerizane:
It would be nice to still be able to move into a tile if the other npc is moving away from you, and you're not faster than them.

bitcraft:
that was a case i thought of as well, and originally was going to not support, but i do think that it would be nice.

PR #2123, opened 2023-11-26
fix overlap

JaskRendix:
PR proposes a solution for #662 and #771
- adds the def **add_collision** and **remove_collision** in the **entity.py**;
- **add_collision** will add the collision exactly when **set_position** is called, in this way we can keep track of NPCs wandering (this is very important, especially in multiplayer or scenarios with a lot of NPCs, where the position needs to be tracked);
- **remove_collision** will be called at the end of **next_waypoint**;
- **add_collision** and **remove_collision** are really important, because so we can have a true value on how many NPCs there are around (correct value of the elements inside the collision zone), this + **map_area** (#2122) will allow to measure the "walkable" tiles;
- adds `self.player.remove_collision((eo.x, eo.y))` because if not the tile of **player spawn** will not be accessible;
- adds **remove_collision** to **remove_entity**, because if not the tile will stay in the collision zone and it can cause issues;
- updates the logger error for pathfinding as requested by the TODO and adds the name/slug of the NPC involved (name for the player, slug for everything else);

```
    0,0 | 1,0 | 2,0 | 3,0
    0,1 | 1,1 | 2,1 |
    0,2 | 1,2 | 2,2 |
```
by mass testing I noticed that overlapping happens when:
NPC moves from (2,0) to (2,1)
Player moves from (1,1) to (2,1) -> overlap

the issue: **get_exits** scans only (1,0), (0,1), (1,2) and (2,1), but what if there is a NPC in (2,0) ready to move into (2,1)?
the solution: **get_coords_ext** scans all the tiles, if there is a NPC, it looks if (1) it's moving and (2) where it looks, then it follows `self.facing != char.facing:`

the result: NPC (2,0) will move to (3,0) instead of (2,1) because the player is going to move there > no more overlapping
why (3,0)? because of `if player.facing (right towards 2,1) != char.facing (down towards 2,1)`, so the NPC will move right too.


xirsoi (dismissed):

xirsoi:
@JaskRendix I've approved this PR, nice work!

Looks like there's a conflict in the npc.py file though, probably from one of the other PRs I merged today.

JaskRendix:
@xirsoi I was expecting that, solved, I'll test a bit more

Yeah, after testing I noticed there was still a bug
 I forgot to remove the NPC's tile from the collision zone  when it gets removed in game (remove_npc action).

bitcraft (approved):

PR #2126, opened 2023-11-29
class EntityFacing

JaskRendix:
While working on #2121 I noticed: `facing = "front", "back", "left", "right"`
out of curiosity I looked for all the reference and I got only 3, map.py itself as well as npc.py;

PR:
- replaces the **facing** with a class **EntityFacing**, it deletes it from map.py and creates it in db.py;
- upgrades the validator checking in db.py, before it was checking only front, now it's going to check front, back, left and right;

it replaces: 
`anim_types = ["front_walk", "back_walk", "left_walk", "right_walk"]`
with
`anim_types = [f"{facing}_walk" for facing in list(EntityFacing)]`

initially I thought about creating a separate class called **EntityWalking**, but I wasn't sure about, since the "front_walk" references are not many as well as it wasn't much different from **EntityFacing**, only with a **_walk** added.

There is one last piece that needs to be addresses in worldstate.py, but it's connected to #2121 (there are some up, down, left and right), if this gets merged, then it'll be completed there.

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -177,6 +177,13 @@ class MissionStatus(str, Enum):
     failed = "failed"
 
 
+class EntityFacing(str, Enum):
+    front = "front"
+    back = "back"
+    left = "left"
+    right = "right"
+
+
 # TODO: Automatically generate state enum through discovery
 State = Enum(
     "State",
@@ -785,7 +792,10 @@ class NpcTemplateModel(BaseModel):
 
     @field_validator("sprite_name")
     def sprite_exists(cls: NpcTemplateModel, v: str) -> str:
-        sprite: str = f"sprites/{v}_front.png"
+        sprite = f"sprites/{v}_{EntityFacing.front}.png"
+        sprite = f"sprites/{v}_{EntityFacing.back}.png"
+        sprite = f"sprites/{v}_{EntityFacing.right}.png"
+        sprite = f"sprites/{v}_{EntityFacing.left}.png"
         sprite_obj: str = f"sprites_obj/{v}.png"
         if has.file(sprite) or has.file(sprite_obj):
             return v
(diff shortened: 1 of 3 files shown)


tuxemon/db.py
```diff
@@ -785,9 +792,11 @@ def combat_file_exists(cls: NpcTemplateModel, v: str) -> str:
 
     @field_validator("sprite_name")
     def sprite_exists(cls: NpcTemplateModel, v: str) -> str:
-        sprite: str = f"sprites/{v}_front.png"
+        sprite = None
+        for facing in list(EntityFacing):
```
Grimmys: I don't get the logic there.

You loop other the list of all the facing... And for each of them you assign the corresponding sprite path to the `sprite` variable.

But since `sprite` is not an array/list, you are just _**overriding**_ its value at every new iteration. At the end, you only have the last facing sprite path in the variable, what's the point of the loop then?

tuxemon/db.py
```diff
@@ -785,9 +792,11 @@ def combat_file_exists(cls: NpcTemplateModel, v: str) -> str:
...
+            sprite = f"sprites/{v}_{facing}.png"
         sprite_obj: str = f"sprites_obj/{v}.png"
-        if has.file(sprite) or has.file(sprite_obj):
+        if sprite and has.file(sprite) or has.file(sprite_obj):
```
Grimmys: `sprite` should always be defined at this point, since it's something static (the last facing from the Enum).
No need to check it then.

Grimmys (changes requested):
Good improvement idea 👍

tuxemon/db.py
JaskRendix: you're right, I can simplify it, a pointless loop and overriding, I just need to check the 4 parameters and nothing more.

xirsoi (approved):

PR #2127, opened 2023-11-29
get coords (map.py)

JaskRendix:
while working on #2121 I was surprised by the amount of hardcoded "up", "down". Sure #2121 creates a class and centralize, but we can do better.  This is the main reason that drove me to code a def, a def that will allow us to get a list of all the coordinates around the player (at the moment only 4); to explain what I mean, see below:
```
     -  | 1,0 |  -
    0,1 | 1,1 | 2,1
     -  | 1,2 |  -
```
let's say our player is (1,1), there is no way to get a list with (1,0), (0,1), (1,2) and (2,2)

The results are encouraging, especially by seeing how much I was able to simplify certain actions/conditions (both event and item). The casting in **pathfind_to_player** will be removed in #2121, it was needed because of typehints.

### DIFF
--- a/tuxemon/client.py
+++ b/tuxemon/client.py
@@ -156,6 +156,8 @@ class LocalPygameClient:
         self.map_name = map_data.name
         self.map_desc = map_data.description
         self.map_inside = map_data.inside
+        self.map_area = map_data.area
+        self.map_size = map_data.size
 
         # Check if the map type exists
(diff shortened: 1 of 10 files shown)


Grimmys (approved):
Good code quality 👍

xirsoi (approved):

PR #2129, opened 2023-11-30
fix 727 (default name vs custom name) +  mini capture devices issue

### LINKED ISSUE #727: Distinguish between default and custom monster names
Currently a monster's proper name is always saved, regardless of whether or not this was a custom name or just the name of it's species.

This means that when your Cardiling morphs to become Cardinale it will still be called Cardiling. This is not ideal.

Creating a distinction in the code between the proper name and the default name should resolve this issue.

I've also noticed that monsters are now appearing with all lower case names, which leads me to believe we are seeing their untranslated slugs as names. This is a different symptom of the same issue.

JaskRendix:
PR fixes #727 as well as an issue with the small capture devices / tuxeball in the combat.

the issue was:
- someone renamed a Rockitten into Tequila;
- Tequila evolves into Rockat;
- Tequila is now named Rockat

added a check between the translated name (with the slug) vs the actual name, if different (Tequila != Rockitten), then it keeps the name (Tequila), on the contrary gets updated (Rockat).

The issue was no cleaning of the statuses just before fainting. It means that if the monster fainted, then the small tuxeball wasn't black but violet (if the monster fainted had a status). Now it's fixed. I remember there was an issue related to it, but I have never found the problem before.

### DIFF
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -759,6 +759,7 @@ class Monster:
         faint = Condition()
         faint.load("faint")
         self.current_hp = 0
+        self.status.clear()
         self.apply_status(faint)
 
     def end_combat(self) -> None:
--- a/tuxemon/npc.py
+++ b/tuxemon/npc.py
@@ -744,6 +744,11 @@ class NPC(Entity[NPCState]):
         new_monster.taste_cold = old_monster.taste_cold
         new_monster.taste_warm = old_monster.taste_warm
         new_monster.plague = old_monster.plague
+        new_monster.name = (
+            new_monster.name
+            if old_monster.name == T.translate(old_monster.slug)
+            else old_monster.name
+        )
         self.remove_monster(old_monster)
         self.add_monster(new_monster, slot)
 


tuxemon/monster.py
```diff
@@ -384,6 +384,7 @@ def apply_status(self, status: Condition) -> None:
         count_status = len(self.status)
         if count_status == 0:
             self.status.append(status)
+            print("piss", self.name)
```
Grimmys: oh oh 👀

tuxemon/monster.py
```diff
@@ -409,9 +410,11 @@ def apply_status(self, status: Condition) -> None:
                     elif status.repl_pos == ResponseCondition.removed:
                         self.status.clear()
                     else:
+                        print("vomit", self.name)
```
Grimmys: again :eyes:

tuxemon/monster.py
```diff
@@ -409,9 +410,11 @@ def apply_status(self, status: Condition) -> None:
...
                         # chargedup, charging and dozing
                         return
                 else:
+                    print(status.name, "cum", self.name)
```
Grimmys: there as well :eyes:

Grimmys (changes requested):
I'm not sure to have understood your fix.

Your explanation looks like the reverse of what is describe in the issue.

From my understanding, the issue is about Tuxemon with **no nickname** that are **not** renamed to the name of their evolution when evolving.
At least it's what I got from the example stating that "when your Cardiling morphs to become Cardinale it will still be called Cardiling".
Cardiling is the proper name of the species, not a nickname, but still, after morphing, it keep this name, instead of being renamed to Cardinale (the evolved species' name).

tuxemon/monster.py
JaskRendix: yeah, just testing

Grimmys: I know it was for testing, but since you opened your PR for reviews, I assumed you were not testing it anymore (hence that you just forgot to remove the testing statements).

JaskRendix:
There is an action called rename_monster and it allows to change the name of the monster. If you change the name of your monster into Tequila when it evolves it takes the name of the evolution.

As I explained above with Rockitten in Rockat.
If you rename Rockitten into Tequila and it evolves, the name will not be anymore Tequila, but Rockat, this is because in the evolve_monster was missing `new_monster.name = old_monster.name` as it happens to other parameters (eg moves, etc)

xirsoi (approved):

PR #2130, opened 2023-11-30
consolidation conditions

JaskRendix:
PR proposes
- to move def **collide** inside event init like **get_npc** (eventually we can put it in **map.py**); as requested by the TODO;
- to update **npc_at** so it can accept default player without adding the slug (see below), important to say, at the moment **npc_at** is not used at all;
- to update **player_in**, there was a pointless double return and it wasn't able to handle other properties;
- to remove the only **to_use_tile** in all the maps (it'll be implemented as behavior **talk** > **to_talk**);
- to update both **npc_facing_tile** and **player_facing_tile**;
- to update **check_collision_zones**, the return will be a **list of tuples** instead of optional tuple; in this way I was able to implement the checking for different labels (since there will be more obstacles in the tilesets);
- to update **player_facing_npc** by removing the self.session.player and replacing it by a get_npc(player), in this way during the next step we can set up a very useful condition (eg. npc_facing_npc, since at the moment is hardcoded on player);
- to update **player_defeated** by adding the field **character**, in this way it'll not be center on self.session.player anymore; updated the code so it'll be default "player", no issue for successive mass renaming;
- to update **player_sprite** by adding the character parameter as well, in this way we could check if any NPC in the map has a specific sprite;
- to replace the **session.player** in **player_in** and **player_moved** with a **get_npc** hardcoded on **"player**", the next step will be getting rid of the hardcoded part and add the possibility in the condition to accept any npc slug;

`is npc_at` and `is npc_at player` will be identical

collide can be used by **player_at** as well as **npc_at**, I tested and I didn't notice anything wrong.
Both conditions work correctly.

There is no renaming involved in this step.
It reunites the basic changes (before the name change) of #2131, #2133 and #2134

regarding the change of the name, eg **player_defeated**:
```
is player_defeated (original)
is npc_defeated player 
is character_defeated player (longer prefix)
is char_defeated player (short prefix)
is defeated player (we could remove all the prefix)

others final versions could be (this will require separate PR/PRs):

is npc_at / is player_at > is at
is npc_facing_tile / player_facing_tile > is facing_tile npc_slug
is player_facing_npc > is facing_eachother npc_slug1,npc_slug2
is player_moved > is moved npc_slug
is player_in > is in npc_slug,surfable
is player_sprite > is sprite npc_slug,maniac
```


tuxemon/event/conditions/npc_facing_tile.py
```diff
@@ -56,6 +58,17 @@ def test(self, session: Session, condition: MapCondition) -> bool:
         # get all the coordinates around the npc
         client = session.client
         npc_tiles = get_coords(npc.tile_pos, client.map_size)
+
+        # check if the NPC is facing a specific set of tiles
+        world = session.client.get_state_by_name(WorldState)
+        if len(condition.parameters) > 1:
+            value = condition.parameters[1]
+            if value == "surfable":
```
bitcraft, line 66: ok for now, but in general, "surfable" tiles is a feature for the spyder campaign, and we should have a way to get all tiles which have a specific key/value, rather than hardcoding it

bitcraft (approved):

PR #2132, opened 2023-11-30
change npc (npc_defeated, npc_sprite and npc_in)

JaskRendix:
Closed because I was too "fast", there is a better way to handle the change.

### DIFF
--- a/tuxemon/event/actions/play_music.py
+++ b/tuxemon/event/actions/play_music.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from typing import Union, final
+from typing import Optional, final
 
 from tuxemon import prepare
 from tuxemon.db import db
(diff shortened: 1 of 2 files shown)


tuxemon/event/conditions/npc_defeated.py
```diff
@@ -3,43 +3,38 @@
...
-from tuxemon.event import MapCondition
+from tuxemon.event import MapCondition, get_npc
 from tuxemon.event.eventcondition import EventCondition
 from tuxemon.session import Session
 
 
-class PlayerDefeatedCondition(EventCondition):
+class NPCDefeatedCondition(EventCondition):
     """
     Check to see the player has at least one tuxemon, and all tuxemon in their
```
Grimmys: Docstring could be updated to reflect the changes on the class definition ("Check to see if the NPC has at least one Tuxemon [...]")

tuxemon/event/conditions/npc_sprite.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test(self, session: Session, condition: MapCondition) -> bool:
+        npc = get_npc(session, condition.parameters[0])
+        if not npc:
+            return False
+
+        sprite = condition.parameters[1]
+
+        if npc.template:
+            for tmp in npc.template:
```
Grimmys: Name is confusing, `tmp` usually stands for `temporary`, but in this case, looks like it stands for `template`.

I suggest to rename it `template`.

Grimmys (changes requested):

Grimmys (approved):

PR #2135, opened 2023-11-30
char_ facing, at, face

JaskRendix:
closed because I burned it (it was a mess with a merge in between)

### DIFF
--- a/tuxemon/event/actions/play_music.py
+++ b/tuxemon/event/actions/play_music.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from typing import Union, final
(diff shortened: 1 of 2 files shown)


xirsoi:
i'm in favor of this consolidation, but find the naming confusing.

NPC stands for Non-Player Character, so as a modder/scripter i would be confused and think there's no way to check which way the player is facing. I think that char_facing is a much better name for this new version. And the documentation should then be changed back to say "character" instead of "npc".

bitcraft:
good points all around.  it makes good sense to change npc/player actions and conditions to a single term, and "character" is good.  i think that i would prefer "char_" as well, since its less to type.

PR #2137, opened 2023-12-01
check conditions (db.py)

JaskRendix:
PR proposes to add a **check_conditions** inside **db.py**. check_conditions will check the structure of item, technique and status conditions, it'll check the length of the list as well as if the prefix (is or not) is correct.


xirsoi (approved):
Nice work :)

PR #2138, opened 2023-12-02
award experience + award money (combat)

JaskRendix:
PR proposes:
- to allow multiple method of calculation for experience;
- to allow multiple method of calculation for money;
- to add the item [XP Transmitter](https://wiki.tuxemon.org/XP_Transmitter) (Spyder campaign);

Each method will be triggered by a different result in the game variable. In this way a scenario can implement directly  a new formula by using **set_variable**. By default the value will be "default" for both methods.

`<property name="act1" value="set_variable method_experience:METHOD"/>`

The two game variables will be: **method_money** and **method_experience**

at the moment there are 4 formulas (default, proportional, test and xp_transmitter) for experience and 1 for money (default)


xirsoi (approved):
Looks good!

Sanglorian:
Thanks, that's really cool! 

I don't have a strong opinion of how the player should get the XP Transmitter - maybe one of the computers in the Data Centre could provide it? Or left in a bin in the Data Centre?

PR #2148, opened 2023-12-07
fix interact + others

JaskRendix:
PR:
- fixes a bug with interaction (unable to interact with other NPCs)
- adds a small fix for typehints too (wrong parenthesis);
- adds a fix in dialog (`if colors["bg_color"] != "": KeyError: 'bg_color'`);
- updates action_list, since in the last PR the name of the action is changed;
- adds a fix typehint in monster.py (see below);

`tuxemon/monster.py:581: error: Returning Any from function declared to return "int"  [no-any-return]`

### DIFF
--- a/tuxemon/client.py
+++ b/tuxemon/client.py
@@ -279,7 +279,8 @@ class LocalPygameClient:
             _game_event = state.process_event(game_event)
             if _game_event is None:
                 break
-        return _game_event
+            return _game_event
+        return None
 
     def main(self) -> None:
         """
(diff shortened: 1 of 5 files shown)


tuxemon/client.py
```diff
@@ -276,10 +276,8 @@ def _send_event(
 
         """
         for state in self.active_states:
-            _game_event = state.process_event(game_event)
```
bitcraft, line 279: why is being removed?  the purpose is defining the `is None` conditional allows a state to prevent "lower" states from processing an event.  with this change, suppose the player presses "up", then *all* states would process the "up"action.  probably not what we want.

JaskRendix: ok, I left the break

bitcraft (approved):

PR #2151, opened 2023-12-10
start battle (event action) + combat

JaskRendix:
PR proposes:
- to change the **start_battle** action and implement an optional field, so it would be possible to trigger fights among NPCs (see below);
- to move a good part of formulas out of **combat/combat.py** in specific def created inside **combat.py**;
- to update both AI and Combat_Menu, so it can detect who is the enemy, etc. in this way the player isn't locked anymore in the first spot;

from:
`<property name="act30" value="start_battle npc_maple"/>`
to:
`<property name="act30" value="start_battle npc_maple,npc_apple"/>`

inside it'll be like:
```
    character1: str
    character2: Optional[str] = None
```

In this way `start_battle npc_slug` will end up like `player vs npc_slug`, because there is an automatic sort to put the player in the first position (eventually we can integrate this in the mass renaming eg. `start_battle player,`)


tuxemon/event/actions/random_battle.py
```diff
@@ -82,17 +88,17 @@ def start(self) -> None:
...
-        if not check_battle_legal(player):
-            logger.warning("battle is not legal, won't start")
-            return
+        player = self.session.player
 
         # Lookup the environment
         env_slug = player.game_variables.get("environment", "grass")
         env = db.lookup(env_slug, table="environment")
 
+        if not check_battle_legal(player) or not check_battle_legal(npc):
```
bitcraft, line 97: lets avoid removing log lines unless they are no longer useful.  in this case, i think its still a good idea to include 
```
logger.warning("battle is not legal, won't start")
```
when you are debugging why a combat isn't working

JaskRendix: ah ok, I thought the logger inside check_battle_legal was enough.

bitcraft (approved):

PR #2153, opened 2023-12-13
remove key = water (maps)

JaskRendix:
It needs to follow #2164

PR proposes:
- to remove these collision zones from the maps; some maps (with river, sea, etc.) there is:
```
  <object id="17" type="collision" x="192" y="64" width="176" height="32">
   <properties>
    <property name="key" value="water"/>
   </properties>
  </object>
```


bitcraft:
can you elaborate a bit on why "surface=1" was changed to "key=surfable"?

JaskRendix:
> can you elaborate a bit on why "surface=1" was changed to "key=surfable"?

it's changed to key=surfable because before it was a parallel "map". Sea it wasn't a collision, but it was simply another surface (reason why there were a bunch of key=water) blocking the access or sometimes all the water surfaces.

I thought it was better to centralize everything on the existing collision_map (RegionProperties).

bitcraft:
ig when i compare setting (1) "surface=1" to (2) "key=surfable", option 1 seems like it would be easier to manage if a tile has multiple features/constraints.

the key/surfable thing was like this: "key" means the tile/collision has a set of qualities, and we call that set "water".  the game code would then treat water tiles differently, because it has a specific name.  the `surfable` option was supposed to say "this tile/collision has this specific quality", (not just surfable, could be "climbable", "walkable", "swimable", etc) and the map maker could mix and match features on tiles to create things without changing game code, like making surfable sand or something.

it also gets complicated if we want to support "tile features" which has their own options.  suppose we want sand tiles to be walkable and surfable, but to modify the movement speed.  with 1 we could do `{walkable:0.5, surfspeed:0.75}`, or with 2 it would be `{key:[surfable, walkable], walkspeed:0.5, surfspeed:0.75}`.

its fine, i see you did all the work, but im just not sure that the change was needed.  its not really a regression at this point, either, so i don't think it needs to be reverted.

xirsoi (approved):

PR #2155, opened 2023-12-13
update app map (phone)

JaskRendix:
40+ files, but the majority are only lines replacement (from hardcoded to prepare.XXX);
a bunch of files (images) moved and some new images;

PR : 
- updates **app_map**, before all the locations were hardcoded, now all the locations will be obtained from game_variables (see below);
- updates **phone** by implementing the block if there are no trackers in the map, it'll appear the text: *"GPS tracker not updating."*
- updates **control** (options), it was too big, now it opens only the width and height necessary, no more empty space;
- centralizes all the **backgrounds** in **prepare.py**;
- renames the standard background **bg_pcstate** into **gradient_blue** and **new_spyder_map** into **spyder_map**; 
- resize **gradient_blue**, because it was 240x112, while the basic resolution is 240x160, the operation has been done by maintaining the horizontal lines (white);
- moves **gradient_blue** from gfx/ui/**item**/gradient_blue.png to gfx/ui/**background**/gradient_blue.png (same for the other bgs not related with item);
- deletes **spyder_empty.png** since is identical to **gradient_blue** (**spyder_empty** gets replaced in the YAML);
- adds other gradients (same format with white lines, but different colors);
- inside the folder gfx there was a folder called **backgrounds** with only a sub-folder called **test**. Since the battle backgrounds are in another folder, I renamed **backgrounds** into **test**, moved the files and deleted the sub-folder (simply from **gfx/backgrounds/test** to **gfx/test**), mostly to avoid confusion (backgrounds and test - included all the pngs inside test - don't appear anywhere in the code, so eventually we can delete everything, folder included). Note: these images are tests of the cave battle background;
- updates **spyder.yaml** with all the GPS tracker actions;

@Sanglorian in this way we can make appear the labels (locations) progressively, ~but you need to told me~ said:
- **the label will appear by entering the location for the first time** (the Bones had spoken, quote from Willow)
- ~the label will appear after healing the tuxemon inside the center?~
- ~the label will appear.... ?~

this is all, it's possible that other PRs will follow for extending the map functionalities, but at the moment I don't know 

example variable:
`set_variable nu_map_1:leather_town*0.20*0.42`, etc.
leather_town = msgid (PO)
0.20 is x coordinate
0.42 is y coordinate


Sanglorian:
Thanks, this sounds great!

bitcraft (approved):

PR #2157, opened 2023-12-15
set_state and get_state (npc.py)

JaskRendix:
PR proposes to get rid of the session parameter in both set_state and get_state

I noticed that **set_state** in **npc.py** had the session parameter not used; so I removed it, then by checking with mypy I discovered a conflict with entity.py (same def); then I did the same for get_state, but it was necessary to bridge the client through the world. 

I tested by saving and loading, no issues.


bitcraft:
please leave session in where you find it.  there are changes that need to be merged which enable the code base to be used as a multiplayer server, and the `session` object is used to track it.

PR #2158, opened 2023-12-16
update change_bg (event action)

JaskRendix:
PR:
- updates **change_bg**, it's going to add a new parameter image, this can accept the monster_slug as well as the template_slug (all the images inside gfx/sprites/combat/player like ceo, adventurer, etc.);
- introduces a check related to the dimension of the **background**, it must respect the **native_resolution**, if not it'll raise an error;
- introduces a check related to the dimension of the **image** over the background, it must be less than **native_resolution**;
- updates **native_resolution** in prepare.py (tuple instead of list) like **tile_size**;
- updates **icon_size** in prepare.py (tuple instead of list) like **tile_size**;

the issue: some images were "hardcoded", eg:
spyder_gri
before it was `change_bg spyder_gri` after `change_bg gradient_blue,grintot`

it'll follow #2159


mods/tuxemon/gfx/ui/background/spyder_bud.png
bitcraft, line 1: are these files removed?

JaskRendix: yes, because they will be generated by the new function: background blue + monster sprite, not anymore an image already created (I created all the images for the spyder and xero intros back in March)

bitcraft: ok, sounds good, ty

bitcraft (approved):

mods/tuxemon/gfx/ui/background/spyder_bud.png
bitcraft: i see what you mean now by "hardcoded" images

PR #2159, opened 2023-12-16
validator image size (db.py)

JaskRendix:
PR proposes to:
- add validator checks in db.py to verify the dimensions of the images;
- add a small fix inside change_bg;
- set transparent background in some images, so we haven't a "hardcoded" background (maybe in the future we can switch the gradient, with a hardcoded one, it gets tricky);
- set a series of standard dimensions (always tuple format) inside prepare.py;
- resize some images with wrong dimensions (monsters, icons, etc.), the differences aren't big (eg 23x24 instead of 24x24, so it wasn't too tricky);

```
# set sprite size (eg. nurse, bob, etc.)
SPRITE_SIZE: tuple[int, int] = (16, 32)
# set items size (eg. tuxeball, potion, etc.)
ITEM_SIZE: tuple[int, int] = (24, 24)
# set template size (eg. ceo, adventurer, heroine, etc.)
TEMPLATE_SIZE: tuple[int, int] = (64, 64)
# set monster size (eg. rockitten-front, rockitten-back, etc.)
MONSTER_SIZE: tuple[int, int] = (64, 64)
# set monster menu size (eg. rockitten-menu01, etc.)
MONSTER_SIZE_MENU: tuple[int, int] = (24, 24)
# set borders size (eg dialogues black-orange, etc.)
BORDERS_SIZE: tuple[int, int] = (18, 18)
# set element size (earth, metal, etc.)
ELEMENT_SIZE: tuple[int, int] = (24, 24)
etc.
```


bitcraft:
when you resized images, what was the process?   just verifying that you added/removed border pixels without altering the actual content.  it looks that way, but i'd just like to confirm.

JaskRendix:
> when you resized images, what was the process? just verifying that you added/removed border pixels without altering the actual content. it looks that way, but i'd just like to confirm.

@bitcraft Sure, I only added / removed borders pixel by resizing the canvas and maintaining the position, no altering/scaling of the actual content in any way.

bitcraft (approved):

PR #2164, opened 2023-12-19
surfable_map > surface_map

JaskRendix:
#2153 needs to follow this PR.

from: `surfable_map: Sequence[tuple[int, int]],`
to: `surface_map: MutableMapping[tuple[int, int], dict[str, float]],`

It has been upgrade because now it can carry an useful dict, that can be expanded if we want to. The `dict[str, float]` is composed by:
- str = SurfaceKeys = ["surfable", "walkable", "climbable"] (it can be expanded eg. high grass, etc);
- float = moverate (0 area not accessible), others value will modify the `player.moverate *= moverate`, default 1, so no change;

the main difference is that when **get_collision_map** is called, it'll add the tiles with moverate equal to 0 (eg. surfable = 0)

in the case of surfable we can enable swim (so make the tiles accessible) by updating the surface map with **update_tile_properties** event action. The action has two parameters: **label** (always SurfaceKeys) and **value** (always the float).

for Spyder -> `update_tile_properties surfable` since the value is default 1


xirsoi (approved):
excellent work!

PR #2168, opened 2023-12-20
comparison (equals, not_equals, etc.)

JaskRendix:
follows #2167 

PR continues the task with the Comparison class, in this way we can centralize all the sign operators:
- removes all the comparisons from actions and conditions;
- adds a small fix to get_player_monster, to avoid the possibility to close it without selecting anything (if filters haven't been set);


xirsoi:
I kinda liked it better when we could use the operators (">=" instead of "greater_than_equal").

How does @Sanglorian feel about it?

JaskRendix:
@xirsoi I have added the >, <, etc.

Sanglorian:
Thanks both! I don't mind either way, including both sounds like a great compromise!

Grimmys (approved):
Very interesting refactoring, code was quite ugly before, congrats!

bitcraft (approved):

PR #2171, opened 2023-12-21
modify_monster_stats (event action)

JaskRendix:
PR:
- proposes to add **modify_monster_stats** (it'll increase / decrease permanently the monster stats);
- fixes and purges the item effect, the data will be shipped through event action;

This will increase all the stats of all the monsters in the party by 1:
```
    <property name="act2" value="modify_monster_stats"/> 
```
or specific
```
    <property name="act1" value="get_player_monster name_variable"/>
specific monster, dodge, increase
    <property name="act2" value="modify_monster_stats name_variable,dodge,50"/>
specific monster, speed, decrease
    <property name="act2" value="modify_monster_stats name_variable,dodge,-30"/>
specific monster, all stats, decrease
    <property name="act2" value="modify_monster_stats name_variable,,-30"/>
all monsters, all stats, decrease
    <property name="act2" value="modify_monster_stats name_variable,,-30"/>
```


bitcraft:
great idea!  we already have a couple actions like this "modify_player_attribute" and "set_player_attribute", for players.  adding this feature to monsters is good. i think we should try to stick to the existing format of "set" ==, and "modify" +/-, though.  you may notice that modify_player_attribute also allows using `%` for multiplication. https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/event/actions/modify_player_attribute.py

so when i see "set_monster_stats", i think the idea is good, but it looks like it will "modify" the value, instead of "setting" the value.  could you confirm this?

JaskRendix:
@bitcraft I confirm and I have modified the action too, starting with the name, then the parameter value, so it'll accept both float and int, in this way float will be considered a multiplication while int an addition

xirsoi (approved):

PR #2176, opened 2023-12-25
math operations (operator)

JaskRendix:
follows #2167 
the mapping will be moved into **tools.py** to centralize this operation too

PR:
- cleans up **transfer_money**, some lines were pointless;
- moves the operations in **statchange**;
- moves the operations in **variable_math** (the **=** operation in **variable_math** is identical to the action **copy_variable**), since it's not used (anywhere), we could get rid of and add **pow** or other operations;


tuxemon/event/actions/variable_math.py
```diff
@@ -3,14 +3,23 @@
...
-from typing import Union, final
+from operator import add, floordiv, mul, sub
+from typing import Optional, final
 
 from tuxemon.event.eventaction import EventAction
 from tuxemon.tools import number_or_variable
 
 logger = logging.getLogger(__name__)
 
+ops_dict: Mapping[str, Callable[[float, float], float]] = {
```
xirsoi: We should consider moving this dictionary to a central place, instead of defining it everywhere we need it.
Not necessary for this PR, just a thought I had.

xirsoi (dismissed):
looks good

tuxemon/event/actions/variable_math.py
JaskRendix: I agree, I wrote about it above, I would like to put in in tools.py

bitcraft (approved):

PR #2177, opened 2023-12-26
disable forfeit

JaskRendix:
PR disable forfeit (being able to run away from trainer battles), until now was enabled by default and it was possible to block it by adding `"forfeit": false,` inside the npc JSON, now it can be enabled by putting `"forfeit": true,`


xirsoi (dismissed):
looks good!

xirsoi (approved):

PR #2178, opened 2023-12-26
steps monsters

JaskRendix:
PR introduces the parameter "steps" for monster, the temporal parameters (days) isn't enough to qualify how much a monster has been between us, but steps is a pretty good one; this implements only the parameter, something else will follow (where this can be viewed, etc.). Only the ones carried (party) will increase steps (so max 6, nothing more).


xirsoi (approved):
looks good!

PR #2179, opened 2023-12-26
set_monster_health (event action)

JaskRendix:
PR updates set_monster_health:
- replaces **player.find_monster_by_id(monster_id)** with **get_monster_by_iid(self.session, monster_id)**, this point is really important, because at the moment it checks only among 6 (the best case scenario) monsters, the ones in the party, but since the iid is unique, it'll check among all the monsters in the map;
- adds the possibility to add and subtract HP by passing values like 50,-30, etc., it includes a check if it goes below 0 or it overcomes the max;

### DIFF
--- a/tuxemon/event/actions/set_monster_health.py
+++ b/tuxemon/event/actions/set_monster_health.py
@@ -5,8 +5,9 @@ from __future__ import annotations
 import logging
 import uuid
 from dataclasses import dataclass
-from typing import Optional, final
+from typing import Optional, Union, final
 
+from tuxemon.event import get_monster_by_iid
 from tuxemon.event.eventaction import EventAction
 from tuxemon.monster import Monster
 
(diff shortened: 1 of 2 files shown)


tuxemon/event/actions/set_monster_health.py
```diff
@@ -62,8 +67,6 @@ def start(self) -> None:
                 logger.error(f"Game variable {self.variable} not found")
                 return
             monster_id = uuid.UUID(player.game_variables[self.variable])
-            monster = player.find_monster_by_id(monster_id)
-            if monster is None:
-                logger.error("Monster not found in party")
-                return
+            monster = get_monster_by_iid(self.session, monster_id)
+            assert monster
```
xirsoi: another assert

xirsoi (changes requested):
just the one little change here, otherwise looks great

xirsoi (approved):

PR #2181, opened 2023-12-27
enable / disable hp/exp bar (JSON)

JaskRendix:
PR proposes to add 3 parameters in the environment JSONs below the hud section:
- hp_bar_player
- hp_bar_opponent
- exp_bar_player

all the fields are True by default, but this will give to the modder the power to remove the bar in certain environments.


xirsoi (approved):
looks good!

PR #2184, opened 2023-12-29
evolution (path) + evolution for steps

JaskRendix:
PR:
- removes **path**;
- adds **steps** (trigger evolution if the monster has walked a certain amount of steps);

when working on monster I noticed a bottleneck with evolution, we can set:

```
            "path": "standard",
            "at_level": 36,
            "monster_slug": "agnigon"
```
but it doesn't allow to add multiple parameters, like
```
            "path": "standard",
            "at_level": 36,
            "monster_slug": "agnigon"
            "variable": "season:winter"
```
for doing this above, I would be obliged to set path "variable", so I decided to remove the path and doing a multiple check on booleans
```
            "at_level": 36, >true?
            "monster_slug": "agnigon" required parameter
            "variable": "season:winter" > true?
```
in this way we can simplify the code too in **check_evolution**

### DIFF
--- a/tuxemon/event/conditions/check_evolution.py
+++ b/tuxemon/event/conditions/check_evolution.py
@@ -2,12 +2,16 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-from tuxemon.db import EvolutionType, StatType
-from tuxemon.event import MapCondition
+import logging
+
+from tuxemon.db import StatType
+from tuxemon.event import MapCondition, get_npc
 from tuxemon.event.eventcondition import EventCondition
 from tuxemon.monster import Monster
 from tuxemon.session import Session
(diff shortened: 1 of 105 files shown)


tuxemon/event/conditions/check_evolution.py
```diff
@@ -19,118 +23,74 @@ class CheckEvolutionCondition(EventCondition):
...
+                    if evolution.item:
+                        _evolution = False
+                    if _evolution:
+                        evolving.append((monster, evolved))
+
         if evolving:
-            player.pending_evolutions = evolving
-        return _evolution
+            character.pending_evolutions = evolving
+        return bool(evolving)
```
Grimmys: I would instead check whether the list (`evolving`) is empty or not. I feel like it would be more descriptive of what is verified here.

Grimmys (changes requested):

Grimmys (approved):

bitcraft (approved):

PR #2149, opened 2023-12-08
char_at, char_facing, char_face, etc

JaskRendix:
PR proposes:
- to rename **npc_facing** and  **player_facing** into **char_facing**
- to rename **npc_at** and **player_at** into **char_at**
- to rename **npc_face** and **player_face** into **char_face**
- to rename **npc_exists** into **char_exists**
- to rename **player_in** into **char_in**
- to rename **player_sprite** and **has_sprite** into **char_sprite**
- to rename **player_facing_tile** and **npc_facing_tile** into **char_facing_tile**
- to rename **player_facing_npc** into **char_facing_char**
- to rename **player_moved** into **char_moved**
- to rename **player_stop** into **char_stop**
- to rename **npc_look** into **char_look**
- to rename **npc_plague** into **char_plague**
- to rename **npc_walk** into **char_walk**
- to rename **npc_move** into **char_move**
- to rename **npc_run** into **char_run**
- to rename **npc_speed** into **char_speed**
- to rename **npc_wander** into **char_wander**
- to rename **set_player_attribute** and **set_npc_attribute** into **set_char_attribute**
- to rename **modify_player_attribute** and **modify_npc_attribute** into **modify_char_attribute**
- to change **start_battle npc_maple** into **start_battle player,npc_maple**
- to change **has_monster rockitten** into **has_monster player,rockitten** (like has_item player,potion)
- to upgrade all the maps
- moves the **moverate limits** from **char_speed** (old npc_speed) from the action to the **prepare.py**;

some examples `player_` and `npc_` related:
`player_face left` > `char_face player,left`
`is player_facing left` > `is char_facing player,left`
`is player_defeated` > `is char_defeated player`
`is player_moved` > `is char_moved player`
`is npc_at` > `is char_at player`

replaces #2135

PR #2185, opened 2023-12-29
battle + various (small fix)

JaskRendix:
PR:
- updates **battle** class and adds the parameter **fighter**;
- updates **set_battle** and inserts some loggers as well as checks;
- the field **date** has been replaced with **steps** in battle class, this because steps is a way better temporal indication than the ordinal time inside the game and more easier to handle;
- updates **def track_battles** and adds the **session**, in this way we can register the battle through the client as an action;
- adds **def set_var** in combat, so the variables will be created by passing through the action;
- updates class **OutputBattle** and removes both **forfeit** and **run** since unused, so we can maintain the core functions;
- replaces **monster.level** with **npc_monster_details.level** in **npc.py** (def load_party);
- removes **set_party_status** from phase **ran away**, since it already checks in **end_combat**;
- removes **set_party_status** from **track_battles** too (same exact reason point above);
- moves the last **set_party_status** in combat inside **end_combat**;
- removes the width/height/etc from **spyder.yaml** because these are generated by default;
- replaces **monsters_in_play[self.players[0]]** with **monsters_in_play_human** because it returns exactly the previous one;
- replaces **self.players[0]** with **monster.owner** in **combat_animations.py**, since at that point the monster is already captured (player's monster);
- adds check in **def give_money** (npc.py) if **player.isplayer** (temporary, because I have an idea, but #2176 and #2149  need to be merged), this checking was necessary because in a fight among NPCs it could lead to crash, so it's temporary because I would like to ship it through event action and avoid give_money (hardcoded on player);
- fixes only the **background** **caper** **front** and **back**, it was white instead of transparent;
- rewrites some lines in init (monster), it was possible to reduce the number of lines;
- introduces two new game variables to clarify the output of the battle; 
- updates **set_tuxepedia** for two reasons: (1) specific logger info were missing and (2) it was hardcoded on the player;
- updates **set_tuxepedia** in the maps by adding before the **"player"** (it was used only in the Dojo, already checked the new maps);
- moves out from **combat_animation** the **tuxepedia phase**, after capturing the wild monster, because it can be integrated directly in the capture item effect;

regarding the battle class update:
`battle(opponent: npc_maple, outcome: won, etc.)`
lack of who fought against npc_maple
`battle(fighter: player, opponent: npc_maple, outcome:won)`
in this way we can centralize all the battles, even the ones between NPCs.

regarding the game variables:
until now we got **battle_last_result** (win, lost, draw) and **battle_last_trainer** (npc.slug), but these variables are still related to the old model where in **start_battle** there was only 1 parameter, now that we can trigger battles among NPC we need a couple of more generic variables: **battle_last_winner** and **battle_last_loser** 

in this way in a scenario like start_battle npc_maple,apple, there will be the winner immediately associated to the slug; **battle_last_winner:npc_maple** and **battle_last_loser:apple** (if there is a player, it'll be saved the value "player")


tuxemon/combat.py
```diff
@@ -366,66 +384,98 @@ def track_battles(
...
-        loser.game_variables["teleport_clinic"] = OutputBattle.lost
-        message = T.translate("combat_defeat")
+        info = {"name": loser.name.upper()}
+        set_var(session, "teleport_clinic", OutputBattle.lost)
         if trainer_battle:
+            # set variables
+            if loser.isplayer:
+                set_var(session, "battle_last_result", OutputBattle.lost)
+                set_var(session, "battle_last_loser", "player")
+            if not loser.isplayer:
```
Grimmys: Consider replacing this by an else branch, since the condition checked there is just the reverse of the one checked previously.

tuxemon/states/combat/combat.py
```diff
@@ -503,40 +503,42 @@ def transition_phase(self, phase: CombatPhase) -> None:
             pass
 
         elif phase == "ran away":
-            self.players[0].set_party_status()
+            pass
```
Grimmys, line 503: I'm wondering if there would not be a way to get rid of these empty branches... They take some space for nothing.

Maybe we could at least merge them?

Like `elif phase in ("resolve match", "ran away"): pass`, and this could be right before the `else` part to not annoy the reader with it.

Grimmys (changes requested):

tuxemon/states/combat/combat.py
JaskRendix: I changed like in line 442:
```
        elif phase == "resolve match" or phase == "ran away":
            pass
```

Grimmys (approved):

bitcraft (approved):

PR #2194, opened 2024-01-02
random encounter (replace daytime with variable)

JaskRendix:
PR proposes to replace **daytime field (boolean)** with **variable (optional string)**.

Originally daytime was added to guarantee a possible difference between the monster encountered in the night vs day, but daytime is a game variable, so I thought to allow directly to set a variable instead of forcing only 1. In this way it can offer literally infinite possibilities, spawning in seasons, precise hours, days, weekday, after certain quests, etc.

from
```
    "monsters": [
      {
        "encounter_rate": 0.5,
        "daytime": true,
        "exp_req_mod": 1,
```
to
```
    "monsters": [
      {
        "encounter_rate": 0.5,
        "variable": "daytime:true",
        "exp_req_mod": 1,
```

### DIFF
--- a/mods/tuxemon/db/encounter/37707_town.json
+++ b/mods/tuxemon/db/encounter/37707_town.json
@@ -3,7 +3,7 @@
     "monsters": [
       {
         "encounter_rate": 0.5,
-        "daytime": true,
+        "variable": "daytime:true",
         "exp_req_mod": 1,
         "level_range": [
           1,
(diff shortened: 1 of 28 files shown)


Grimmys (approved):
Good changes, although I would advise you to not put / keep that many comments in the code.

Comments are supposed to be for explaining "why is it this way", not "what is it doing".

JaskRendix:
@Grimmys done, removed the comments

bitcraft (approved):

Issue #2193, opened 2024-01-02
Idea: 'No Battle' Mode


ultidonki:
For people who only want to enjoy the story, or who get stuck because it gets too difficult, maybe a 'No Battle' mode would be good?

This would replace each 'start_battle' event with something that just returns a 'True' value or something, without actually starting the battle event. 
(However, there are some variables set only inside the combat/battle event, I think, like 'last_trainer_battled' 'last_battle_win', which might make it more difficult)

Just an idea! :smile_cat:

JaskRendix:
@ultidonki what about if you start with a monster lv100? This will allow you to play without issues any battle and enjoy the story. What do you think?

Davidliang11292:
Good idea, but I think there are a few problems: Tuxemon is a monster hunting RPG, which means there must be more or less " battle" in the game, maybe if the player gets a certain Tuxemon at some point in the story and then leaves all the battle to the game automation?

HippasusTwo:
One could have a less battle oriented campaign at some point, say 'contests' or a mystery simply requiring dialogue choices, or something. But that would probably be something to do after one full campaign is done.

ultidonki:
These are all good ideas - I guess you're right, having a game with no battles is a bit strange, but having a cheat where you can have level 100 tuxemon from the start sounds good - thanks for PR #2195 @JaskRendix , that will probably be fun enough :smiley:

ultidonki:
Very nice - I just tested, and entering the name 'ApexPlayer' now gets me all starting tuxemon at level 100 in the Spyder campaign. Although I tried 'APEXPLAYER' first, and it didn't work, but as it's a cheat code, I guess forcing people only to use the correct case is fine, so I'm happy :smile:

PR #2195, opened 2024-01-02
common checking, remove A (input) and cheat code

JaskRendix:
PR:
- removes the A that appears by default at the end of the name. It adds a simple flag;
- **rename_player** event action instead of setting the name will ship it to **set_player_name** event action;
- adds a logger.info inside **set_player_name** to alert the player about the new name;
- adds **common** (event condition);
- adds **check_char_parameter** (event condition), inspired to the one of similar name (event action), this will simplify the checking for single parameters without the necessity to create an event condition for each one;
- adds a cheat code if the player is named **ApexPlayer**, it'll give to the player all starting monsters (max evolution: eg rockat instead of rockitten, eaglace, etc.) at lv 100; I talked about it with @Sanglorian after showing him this issue #2193  opened by @ultidonki , in this way someone can enjoy the game without bothering too much in levelling up (moreover it can help us to understand if there are bugs later in the game);
- adds the description of skip parameter (it was missing);

to complete after #2149 (add **check_char_parameter** to the conditions list + fix an event after **wayfarer inn**)

Screenshot_2024-01-02_21-12-54

### DIFF
--- a/tuxemon/event/actions/set_player_name.py
+++ b/tuxemon/event/actions/set_player_name.py
@@ -2,12 +2,15 @@
 # Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
+import logging
 import random
 from dataclasses import dataclass
 from typing import final
 
 from tuxemon.event.eventaction import EventAction
 from tuxemon.locale import T
+
+logger = logging.getLogger(__name__)
 
 
 @final
@@ -40,3 +43,4 @@ class SetPlayerNameAction(EventAction):
         else:
             name = self.choice
         self.session.player.name = T.translate(name)
+        logger.info(f"Player name is {T.translate(name)}")
--- a/tuxemon/event/conditions/common.py
+++ b/tuxemon/event/conditions/common.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2024 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+import logging
+
+from tuxemon.npc import NPC
+
+logger = logging.getLogger(__name__)
+
+
+class CommonCondition:
+    name = "Common"
+
+    @staticmethod
+    def check_character_parameter(
+        character: NPC,
+        parameter: str,
+        value: str,
+    ) -> bool:
+        """
+        Check a character's parameter against a given value.
+
+        Parameters:
+            character: The character to check.
+            parameter: The character's parameter (eg. "name", "steps", etc.)
+            value: Given value to check against the parameter's value.
+
+        eg. "player,name,alpha" -> is the player named alpha? true/false
+
+        """
+
+        # check for valid inputs
+        # trigger an AttributeError if the parameter doesn't already exist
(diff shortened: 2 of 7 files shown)


tuxemon/event/actions/set_player_name.py
```diff
@@ -40,3 +43,4 @@ def start(self) -> None:
         else:
             name = self.choice
         self.session.player.name = T.translate(name)
+        logger.info(f"Player name is {T.translate(name)}!")
```
Grimmys: I don't think we need an exclamation mark there, not bringing any value imo.

tuxemon/event/conditions/common.py
```diff
@@ -0,0 +1,49 @@
...
+    name = "Common"
+
+    @staticmethod
+    def check_character_attribute(
+        character: NPC,
+        attribute: str,
+        value: str,
+    ) -> bool:
+        """
+        Check a character's (npc or player) attribute.
```
Grimmys: I feel like this docstring is not explaining anything.

"Check a character's attribute", "The NPC object to check" etc.

But what does it mean "checking" in this context? It's a so generic term that it could be beneficial to elaborate a bit about what is done there.

tuxemon/event/conditions/common.py
```diff
@@ -0,0 +1,49 @@
...
+
+        try:
+            val = type(attr)(value)
+        except TypeError:
+            logger.warning(
+                "The value given cannot be parsed into the correct type for '{0}'",
+                attribute,
+            )
+            return False
+        return bool(attr == val)
```
Grimmys: Bool convertion is unnecessary there.

`==` operation is always going to return a `bool`.

tuxemon/menu/input.py
```diff
@@ -201,11 +202,13 @@ def add_input_char_and_pop(self, char: str) -> None:
...
 
     def add_input_char(self, char: str) -> None:
+        self.counter += 1
         if (
             self.char_limit is None
             or len(self.input_string) <= self.char_limit
         ):
-            self.input_string += char
+            # removes A at the end of the name
+            self.input_string += char if self.counter > 1 else ""
```
Grimmys: I don't really like this implementation to be honest, I don't think we need a counter there in fact, because the only thing we care about is whether this method is called for the first time or not.
So instead, I think we need a **flag**.

Something like `self.is_first_input`, that would be initially `True`, and then assigned to `False` at middle of the first call of this method.

Grimmys (changes requested):
Great PR! I always seen this annoying bug.

tuxemon/event/conditions/common.py
JaskRendix: Without bool it gives me.
`tuxemon/event/conditions/common.py:49: error: Returning Any from function declared to return "bool"  [no-any-return]`
it gives me this typehint because I use mypy --strict

tuxemon/menu/input.py
JaskRendix: true, it looks better

tuxemon/event/conditions/common.py
JaskRendix: I agree, it's confusing. I added an example and renamed some parameters.

Grimmys (approved):

bitcraft (approved):

PR #2200, opened 2024-01-04
fix asserts (journal + phone) + LEFT/RIGHT navigation (journal)

JaskRendix:
PR:
- removes a bunch of `assert not isinstance(lab, list)` , I put it to avoid a typehint (since the output was an annoying Union Any/List), but this approach is way better (journal + phone);
- allows to change screen by using RIGHT and LEFT arrows (screenshot below);
- hides the names of monsters (neither seen or caught);
- hides some info of monsters (seen vs caught);

eg if you have seen monster id 1,45,89 and you click RIGHT, it'll automatically jump from 1 to 45, then 89, then 1, same other direction (LEFT);

before
Screenshot_2024-01-19_09-38-32
after
Screenshot_2024-01-19_09-40-59

monster seen
Screenshot_2024-01-19_09-38-56
monster caught
Screenshot_2024-01-19_09-38-44

split out PlayerState + MonsterInfo in #2202

### DIFF
--- a/tuxemon/states/journal/journal.py
+++ b/tuxemon/states/journal/journal.py
@@ -1,20 +1,22 @@
 # SPDX-License-Identifier: GPL-3.0
-# Copyright (c) 2014-2023 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+# Copyright (c) 2014-2024 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
 from collections.abc import Callable
 from functools import partial
-from typing import Any
+from typing import Any, Optional
 
 import pygame_menu
 from pygame_menu import locals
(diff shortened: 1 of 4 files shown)


Grimmys:
Isn't `menu.add.label()` going to return a `Label` (https://pygame-menu.readthedocs.io/en/4.2.4/_source/widgets_label.html) object in any cases? If so, why not put this as a typehint instead of the way to generic `Any` typing.

JaskRendix:
> Isn't `menu.add.label()` going to return a `Label` (https://pygame-menu.readthedocs.io/en/4.2.4/_source/widgets_label.html) object in any cases? If so, why not put this as a typehint instead of the way to generic `Any` typing.

@Grimmys well yes, but actually no, the typing in the pygame_menu code is Any/list[Any] (event if the docs says otherwise), maybe I'm going to patch it directly on the source.

bitcraft (approved):

PR #2202, opened 2024-01-07
party menu (connected player menu)

JaskRendix:
PR proposes:
- to rename **PlayerState** into **CharacterState**;
- to add the **NPC** parameter when opening **PlayerState**, in this way we can potentially re-utilize it for every NPC and it's not set on session.player;
- to add the **NPC** parameter when opening **MissionState**, in this way we can potentially re-utilize it for every NPC and it's not set on session.player;
- to pass **MonsterInfoMenu** parameters (monster and source) through **kwargs**;
- to update **change_state** event action, since the **MonsterInfoState** function was creating a new monster, now by passing the variable (where is stored the monster_id) it'll possible to show the screen with the enemy monster;
- to add the possibility to open **PlayerState** from **change_state** event action;
- to add a party menu accessible by clicking **right** in the player.menu, this will be accessible only if the player got monsters. It'll show how much the monster in the party walked with the player (sorted by max to min (above to down)) as well as the hidden value generated discussed here #2197 
- enables the possibility to navigate inside the box (from the PC) by clicking LEFT or RIGHT arrows (after opening monster info);

draft because there are others PR waiting

without monsters
Screenshot_2024-01-07_09-09-21
with monsters
Screenshot_2024-01-07_09-00-17
Screenshot_2024-01-07_09-00-05


tuxemon/states/character/__init__.py
```diff
@@ -49,175 +59,161 @@ def add_menu_items(
...
-                "lost": str(lost),
-            },
-        )
+
+        battles = self.char.battles
+        tot = len(battles)
+        _won = OutputBattle.won
+        _draw = OutputBattle.draw
+        _lost = OutputBattle.lost
+        won = sum(1 for battle in battles if battle.outcome == _won)
```
bitcraft: this is fine for now, but if battles becomes very large, then we would want to break this down into something like:
```python
for battle in battles
  if battle.outcome == OutputBattle.won:
     won += 1
  ...
```
because it would loop over the battle list only once.  probably not realistic here, but worth think about it more because iteration of things in python can be slow.
also, this could this be rewritten as `len(battle for battle in battles if battle.outcome == _won)`?  i don't think we need sum here.

JaskRendix: totally agree with you, implemented

bitcraft (approved):

PR #2212, opened 2024-01-16
remove def give_money (npc.py) + update transfer_money / set_money (event action) + create modify_money (event action)

JaskRendix:
I noticed this while working on #2185, but I couldn't act on it because there was #2149 still open.
**give_money** was used only two times.

PR:
- removes **give_money** from **npc.py**;
- removes **give_money** from **combat.py** (it was used to give money to the player - if winner);
- changes the order of some functions inside the combat.py, in this way the message with money prize will appear only to the human player;
- removes **give_money** from an effect, since it gives money instead of damages (battle);
- adds the parameter **player** in front of **transfer_money** (event action);
- creates **modify_money** (event action);
- updates **set_money** (event action);
- simplifies **transfer_money** (event action);
- fixes two maps (Spyder) faulty pay mechanism fee;
- adds a test for money (for checking set_money, money_modify and transfer_money);


tuxemon/event/actions/transfer_money.py
```diff
@@ -15,50 +15,55 @@
...
+    slug2: Optional[str] = None
 
     def start(self) -> None:
         player = self.session.player
 
         # Read the parameters
         transaction = self.transaction
         amount = self.amount
-        wallet_player = player.money.get("player", 0)
+        wallet_player = player.money.get(self.slug1, 0)
```
bitcraft: would we rename the `wallet_` variables here?  i think its great that we are allowing the walllets to change, but if the variable is still named "player", the code becomes difficult to understand

bitcraft (changes requested):

tuxemon/event/actions/transfer_money.py
JaskRendix: I agree with you

JaskRendix: Ok, wallet_character1 and wallet_character2

tuxemon/event/actions/transfer_money.py
```diff
@@ -15,50 +15,55 @@
 @dataclass
 class MoneyMathAction(EventAction):
     """
-    Performs a mathematical transaction on the player's money.
+    Performs a money transaction between entities.
```
bitcraft, line 20: looking through the code it seems like there are a lot of soft-bugs or "gotchas" here.  for example, if you want to move money from the player to another entity, but the entity is not found, the player will still gain or lose the money.  this will be a huge problem since the error isn't handled.  i know the problem already existed, but could we add a comment here warning the developer that this isn't a robust system?  i don't expect you to improve it now, since the pr is just adding a different feature, but we should get a comment in here before somebody forgets.
suggestion:
The logic in handling money transactions is not covered by tests and is known to have issues.

JaskRendix: the logic is faulty, **transfer_money** needs to do only its task (transferring money), this is the reason why I'm going to modify **set_money** by implementing a series of addition and creating **modify_money** (new event action) that will add/remove money from the wallet

since the majority of transactions in the game is the player gaining / reducing the amount of money without someone else gaining / reducing, I put **modify_money** (not many, better do the task now than later); in the other two cases I put **transfer_money**;

JaskRendix: Ok, now **transfer_money** looks clear, because it ships the different functions to precise actions.

bitcraft (approved):
great to see tests added!

PR #2049, opened 2023-10-27
park feature (Spyder)

JaskRendix:
PR implements park feature.

The first step is done. The area where the park is placed isn't accessible at the moment.

- adds new encounters **park** for wild monster -> to complete (only pairaigrin for testing purposes);
- adds **MainParkMenuState** (new state + file in states/combat);
- adds 2 new environments (**park** and **night_park**);
- adds new Tuxeball, [Park Tuxeball](https://wiki.tuxemon.org/Park_Tuxeball);
- adds new maps (**entrance** + **park** + **cave**);
- updates **lion_mountain** map to add the park entrance;
- updates **core_** by adding a bigger bridge (the same one, but bigger)

next steps:
- disabling wild monster attack;
- expanding the cave (?);
- formatting steps variable (it appears like 100.2524284) -> solution here: #2234 (event action: **format_variable**)

PR #2214, opened 2024-01-17
battle menu

### LINKED ISSUE #1952: HP bar behaviour in multi Tuxemon fights

JaskRendix:
PR proposes:
- to add a field (environment JSON) named **menu**, by default **MainCombatMenuState**, it'll allow to use another menu instead of **MainCombatMenuState**;
- to add a field (environment JSON) named **msgid**, by default **combat_monster_choice**, it'll allow the modder to replace the standard sentence **What will Pairagrin do**;
- to add a **validator** in **db.py** for checking **menu** (State);
- to add a **validator** in **db.py** for checking **msgid** (it must be in the PO file);
- to move **build_hud_text** from **combat_animations** to **combat.py** (not the state), **build_hud_text** is the def for generating **Pairaigrin (m) Lv. 3**;
- to simplify def **animate_exp** in **combat_animations.py**;
- to add the parameter **animate** (boolean default **True**) in the def **build_hud** (this is important, so we can trigger the build without the animation, it means more updated and a solution to the issue of XP bar animation, etc);
- to rename monsters_in_play_human into **monsters_in_play_right**;
- to rename monsters_in_play_ai into **monsters_in_play_left**;
- to create def **update_hud** in **combat_animation.py** (avoid the repetition of the same inside the main combat.py);
- to create def **check_hud** in **combat_animation.py** (avoid to build (aka blit) a new hud instead of overwriting the existing one) -> fix #1952 
- to update effects **area** and **give** because in 2vs2 there were issues (the reason was the presence of .isplayer);
- to update loop (whether it appears or not the party hud - minituxeball);

the renaming is for clarity reasons, nowadays human could be left as well as ai can be right (NPC vs NPC), in this way is more clear.

example (night_park) from #2049 :
```
  "slug": "night_park",
  "battle_graphics": {
    "menu": "MainParkMenuState", <---------- by default MainCombatMenuState
    "msgid": "park_character_choice", <---------- by default combat_monster_choice
    "island_back": "gfx/ui/combat/grass_island.png",
```
In this way I could enable the park by replacing **MainCombatMenuState** with **MainParkMenuState**.

Screenshot_2024-01-17_15-17-21


xirsoi (approved):
i dig it!

PR #2216, opened 2024-01-18
remove revive (item effect)

JaskRendix:
PR:
- adds to the **restoration** item the condition that the monster must be alive;
- removes from the **revive** item the effect **revive**, since it's a **duplicate** of **restore**;
- updates item effect **change_stat**, there was a more direct way to ship the value;
- simplifies item effect **heal**;


bitcraft (dismissed):

bitcraft:
conflicts

JaskRendix:
@bitcraft solved

bitcraft (approved):

PR #2217, opened 2024-01-18
check world (event condition) + set_monster_health (event action)

JaskRendix:
PR:
- fixes set_monster_health, if someone sets the hp at 0 (= faint) with `set_monster_health variable,-1.0`, then the monster doesn't faint; the fix changes `monster.current_hp < 0:` into `monster.current_hp <= 0:` since the monster is going to be fainted at 0 HP too; it replaces also the `current_hp = 0` with the `monster.faint()`;
- adds the event condition **check_world**, this will check for specific parameter of the world state.

at the moment there are two:
- **layer** (related to **set_layer** event action);
- **bubble** (related to **set_bubble** event action);

in this way the action stops and it doesn't keep running; it can be handy for modders too.

### DIFF
--- a/tuxemon/event/conditions/check_world.py
+++ b/tuxemon/event/conditions/check_world.py
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2024 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import final
+
+from tuxemon.event import MapCondition, get_npc
+from tuxemon.event.eventcondition import EventCondition
+from tuxemon.graphics import string_to_colorlike
+from tuxemon.session import Session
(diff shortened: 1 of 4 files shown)


tuxemon/event/conditions/check_world.py
```diff
@@ -0,0 +1,51 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@final
+@dataclass
+class CheckWorldCondition(EventCondition):
+    """
+    Check some world's parameter against a given value.
+
```
bitcraft, line 23: it took me a minute to understand what "layer" and "bubble" are used for.  i think we should include some additional information about the parameters.  maybe something like this:

```
layer: color value which is used to overlay the world
bubble: speech bubble of an npc
```
did i get that right?

JaskRendix: you're right, I'm going to add it

JaskRendix: Added below, just above the example

bitcraft (approved):

PR #2219, opened 2024-01-20
bond

JaskRendix:
PR implements https://wiki.tuxemon.org/Bonding

Implemented:
- item (invisible) "friendship_scroll" for triggering all the mechanism (@Sanglorian 's suggestion);
- max and min bond values + default bond value;
- event 1 bond point each 1000 steps (Spyder campaign);
- event 10 bond points when fainting (Spyder campaign);
- event action for changing the bond value;
- bond points as evolution parameter: `"bond": "greater_than:69"` (JSON);
- bond points as evolution mechanism (check_evolution);

Screenshot_2024-01-22_17-20-04


xirsoi (dismissed):
LG;TFTPR!

Enabling the Bond feature via a hidden item is a great idea, well done!

Sanglorian:
Looks great, thank you so much for implementing this!

xirsoi (approved):

PR #2226, opened 2024-01-21
map (cardinals + scenario) + load_map (worldstate.py)

JaskRendix:
PR:
- sets **self.scenario** value as **str / None**, in this way we can remove the **try/except** from **load_map** in **worldstate.py**;
- allows to import **init** from the YAML, before it wasn't possible (only **type=event**) and it merges these with the client.inits (simply added the source to the load_event);
- simplifies **self.cardinals** (north, etc.) by creating a separate def **set_cardinals**, less lines;
- there could be a merging conflict with #2100 ;

the simplification of cardinal will allow modders to set multiple values in this field (tmx):
`  <property name="north" value="route1"/>`
to
`  <property name="north" value="route1,route2"/>`
simply because some maps can have more than one map in one direction, example of the automatic generated sign:
Screenshot_2024-01-21_10-15-08

docs will be updated after merging


tuxemon/states/world/worldstate.py
```diff
@@ -1218,20 +1218,25 @@ def load_map(self, path: str) -> TuxemonMap:
...
         # scenario YAML, try because not all maps have a scenario
-        try:
-            scenario_path = prepare.fetch("maps", txmn_map.scenario + ".yaml")
-            if os.path.exists(scenario_path):
-                new_events = list(txmn_map.events)
-                new_events.extend(YAMLEventLoader().load_events(scenario_path))
-                txmn_map.events = new_events
-        except:
-            pass
+        if txmn_map.scenario:
```
bitcraft, line 1226: this is probably fine for now, but it is a bit wastful as it forces the yaml loader to load and parse the file twice.  would probably be cleaner to return either a dictionary or a struct-type value with the events sorted into "events" and "inits" groups.

bitcraft: you don't need to make this change now if you don't feel like it.  i don't expect it to be an issue since our map logic isn't very complex at the moment

JaskRendix: yeah, I hope this version is better

xirsoi (approved):

PR #2228, opened 2024-01-21
replace action-automatic-releases (build.yaml)

JaskRendix:
PR:
- adds the flatpak part (scripts + yaml) inside the code, in this way it's not necessary to pull; it's also more easier to update the package when we bump requirements (before it was necessary doing the task here, then open a PR in the other repository, then wait that something was merged on top of that);
- replaces **marvinpinto/action-automatic-releases@latest** with https://github.com/softprops/action-gh-release. This because marvinpinto's action isn't updated since Autumn 2022 and it still runs with node 12. Node 12 is deprecated as pointed out [here in github blog](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)


JaskRendix:
@Murmele regarding this, maybe you're using the same actions in other places too.

download-artifact has been updated to v4, but it fails here:
```
      - name: Download artifacts
        uses: actions/download-artifact@v4
        with:
          path: artifacts

      - name: list artifacts folder
        run: |
          echo "Show artifacts folder:"
          ls artifacts
          echo "Show artifacts/tuxemon_windows_installer folder:"
          ls "artifacts/tuxemon_windows_installer" <------------ no issue, it founds the folder
          echo "Show artifacts/Tuxemon-x86_64:"
          ls "artifacts/Tuxemon-x86_64"  <------------ it doesn't found the folder
```

Murmele:
Hi, is the second one the linux version?

JaskRendix:
> Hi, is the second one the linux version?

@Murmele yes, it seems like the second one since it comes after.

xirsoi (dismissed):

Murmele:
> > Hi, is the second one the linux version?
> 
> @Murmele yes, it seems like the second one since it comes after.

But now in the CI it finds it?

The flatpak artifact is here:
https://github.com/Tuxemon/Tuxemon/actions/runs/7686861458?pr=2228

Murmele:
I thought about the workflow with the flatpak manifest and I think it is better to move the manifest into the repository instead of pulling.

1) Manifest in Repository (Easily changeable, also Requirements can be changed easily, maybe with a script so they get changed whenever Requirements.txt changes)
2) The github CI gets much faster, because all those sed commands can be removed

When Publishing, just updating the Flatpak manifest in the Flathub repository. So that means in the Flathub repository there is always the stable version and in the Tuxemon repository there is always the development version.

The current workflow has some issues and is not that comfortable.

JaskRendix:
> > > Hi, is the second one the linux version?
> > 
> > 
> > @Murmele yes, it seems like the second one since it comes after.
> 
> But now in the CI it finds it?
> 
> The flatpak artifact is here: https://github.com/Tuxemon/Tuxemon/actions/runs/7686861458?pr=2228

True, but the publishing is still failing [here](https://github.com/Tuxemon/Tuxemon/actions/runs/7688299541/job/20949421643), probably because we need to change `ls artifacts` into `ls -R artifacts`

JaskRendix:
> I thought about the workflow with the flatpak manifest and I think it is better to move the manifest into the repository instead of pulling.

@Murmele I agree with you about this as well as removing the sed, but I don't understand which manifest.

Murmele:
> > I thought about the workflow with the flatpak manifest and I think it is better to move the manifest into the repository instead of pulling.
> 
> @Murmele I agree with you about this as well as removing the sed, but I don't understand which manifest.

The manifest ( instructions how to build the flapak package) is the yaml file which will be fetched from flathub repo.

JaskRendix:
@Murmele  move `org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml` inside the `build.yaml` or the opposite?

Murmele:
No moving the yaml file somewhere into the repository and then changing the path in the build.yml to that path

https://github.com/Tuxemon/Tuxemon/tree/development/buildconfig/flatpak

Move it in here and remove the comment in the readme :)

JaskRendix:
@Murmele like this?

buildconfig/flatpak/build_flatpak.yaml
```diff
@@ -0,0 +1,51 @@
...
+    post-install:
+      - desktop-file-install --dir=/app/share/applications buildconfig/flatpak/org.tuxemon.Tuxemon.desktop
+      - install -Dm755 buildconfig/flatpak/org.tuxemon.Tuxemon.appdata.xml /app/share/metainfo/org.tuxemon.Tuxemon.appdata.xml
+      - install -Dm755 mods/tuxemon/gfx/icon.png /app/share/icons/hicolor/64x64/apps/icon.png
+      - install -Dm755 mods/tuxemon/gfx/icon_128.png /app/share/icons/hicolor/128x128/apps/icon.png
+      - install -Dm755 mods/tuxemon/gfx/icon_32.png /app/share/icons/hicolor/32x32/apps/icon.png
+
+    sources:
+    - type: git
+      tag: v0.4.34
```
Murmele:
```suggestion
      tag: development
```

Murmele (changes requested):
Rename the build_flatpak.yaml to `org.tuxemon.Tuxemon.yaml`

JaskRendix:
@Murmele done, but it doesn't find the file with cat

.github/workflows/build.yml
```diff
@@ -67,38 +67,14 @@ jobs:
...
-        sed -i 's@config-opts: \["\(.*\)"\]@config-opts: ["\1", "-DDEV_BUILD=${{ github.ref_name }}"]@' org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml
-
-    - name: Replace desktop file name suffix
-      if: github.ref_type != 'tag'
-      run: >
-        sed -i 's@desktop-file-name-suffix: ""@desktop-file-name-suffix: " (Development)"@' org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml
-
     - name: Show Flatpak manifest
-      run: cat org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml
+      run: cat org.tuxemon.Tuxemon.yaml
```
Murmele:
```suggestion
      run: cat buildconfig/flatpak/org.tuxemon.Tuxemon.yaml
```

.github/workflows/build.yml
```diff
@@ -67,38 +67,14 @@ jobs:
...
 
     - name: Build package
       uses: flatpak/flatpak-github-actions/flatpak-builder@v6
       with:
         bundle: Tuxemon.flatpak
-        manifest-path: org.tuxemon.Tuxemon/org.tuxemon.Tuxemon.yaml
+        manifest-path: org.tuxemon.Tuxemon.yaml
```
Murmele:
```suggestion
        manifest-path: buildconfig/flatpak/org.tuxemon.Tuxemon.yaml
```

Murmele: Can you try this?

Murmele (changes requested):

.github/workflows/build.yml
JaskRendix: @Murmele done, but same output

JaskRendix:
do we need to update this?
```
    - name: Checkout development branch
      run: |
        cd org.tuxemon.Tuxemon <<---------
        git checkout development
```

Murmele:
> do we need to update this?
> 
> ```
>     - name: Checkout development branch
>       run: |
>         cd org.tuxemon.Tuxemon <<---------
>         git checkout development
> ```

This is not needed anymore. Also the comment above can be removed.

Remove this: 
```
# Turn this on once the requirements (requirements.txt) changed and the build is not working anymore
    # Create development branch at https://github.com/flathub/org.tuxemon.Tuxemon and
    # update the requirements as described in the Readme. Turn it off again when the requirements in the
    # master branch on flathub are aligned again with the Tuxemon requirements
    - name: Checkout development branch
      run: |
        cd org.tuxemon.Tuxemon
        git checkout development
```

Remove also:
```
 - name: Clone flatpak manifest from Flathub
      uses: GuillaumeFalourd/clone-github-repo-action@v2.3
      with:
        branch: 'development'
        owner: 'flathub'
        repository: 'org.tuxemon.Tuxemon'
```

JaskRendix:
@Murmele done

Murmele:
Can you add an 'ls' to see in which folder the ci is before the cat command?

Murmele:
- name: Checkout repository
      uses: actions/checkout@v4

This is missing before the stackoverflow comment.
Otherwise the repo is not available

JaskRendix:
@Murmele good! done! now it stops here:
```
** (flatpak-builder:187): ERROR **: 18:10:26.133: Failed to load included manifest (/__w/Tuxemon/Tuxemon/buildconfig/flatpak/portmidi.json): Failed to open file ?/__w/Tuxemon/Tuxemon/buildconfig/flatpak/portmidi.json?: No such file or directory
/usr/bin/xvfb-run: line 181:   187 Trace/breakpoint trap   (core dumped) DISPLAY=:$SERVERNUM XAUTHORITY=$AUTHFILE "$@" 2>&1
Error: Build failed: Error: The process '/usr/bin/xvfb-run' failed with exit code 133
```

JaskRendix:
@Murmele done, I have added the portmidi, libshake, etc. at the moment

Murmele:
> @Murmele done, I have added the portmidi, libshake, etc. at the moment

Great. Can you add also the readme and the scripts from there? So that everything is together

JaskRendix:
@Murmele everything is added

Murmele:
Great work 🙂

bitcraft (approved):

JaskRendix:
@Murmele  the PR has been merged, everything is fine, but it doesn't publish (last step) as it was before:
```
tuxemon-installer.exe
Show artifacts/tuxemon_windows_installer folder:
artifacts/tuxemon_windows_installer:
tuxemon-installer.exe
Show artifacts/Tuxemon-x86_64:
ls: cannot access 'artifacts/Tuxemon-x86_64': No such file or directory
Error: Process completed with exit code 2.
```
this is the piece of code:
```
      - name: list artifacts folder
        run: |
          echo "Show artifacts folder:"
          ls -R artifacts
          echo "Show artifacts/tuxemon_windows_installer folder:"
          ls -R "artifacts/tuxemon_windows_installer"
          echo "Show artifacts/Tuxemon-x86_64:"
          ls -R "artifacts/Tuxemon-x86_64"
```
what could be the issue? because windows installer is fine, but tuxemon-86 no. Do we need to put **ls -R artifacts** again before **echo "Show artifacts/Tuxemon-x86_64:"**?

Murmele:
@JaskRendix the list artifacts folder was just to check if the folders exist, this is not required. For some reason it does not find the Bundle.

I checked a few pullrequest actions: https://github.com/Tuxemon/Tuxemon/actions/runs/7947846121
There the flatpak package is available. Next step is to find out why it will not be found by this part: https://github.com/Tuxemon/Tuxemon/blob/development/.github/workflows/build.yml#L177-L187

Do you have a link where you found this error message?

JaskRendix:
@Murmele sure, here: https://github.com/Tuxemon/Tuxemon/actions/runs/7945212993/job/21691724375

this appear only when a PR is merged and I think this is the cause of this:
> Next step is to find out why it will not be found by this part: https://github.com/Tuxemon/Tuxemon/blob/development/.github/workflows/build.yml#L177-L187

Murmele:
Hmm looking into the flatpak build:
```
Uploaded /__w/Tuxemon/Tuxemon/Tuxemon.flatpak (99.8%) bytes 201326592:209715199
Uploaded /__w/Tuxemon/Tuxemon/Tuxemon.flatpak (100.0%) bytes 209715200:209970981
Total size of all the files uploaded is 209970982 bytes
File upload process has finished. Finalizing the artifact upload
Artifact has been finalized. All files have been successfully uploaded!

The raw size of all the files that were specified for upload is 209977872 bytes
The size of all the files that were uploaded is 209970982 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage
```

```
Show artifacts folder:
artifacts:  <--- No Tuxemon-x86_64 is listed. But also any other name is listed
tuxemon_windows_installer

artifacts/tuxemon_windows_installer:
tuxemon-installer.exe
Show artifacts/tuxemon_windows_installer folder:
artifacts/tuxemon_windows_installer:
tuxemon-installer.exe
Show artifacts/Tuxemon-x86_64:
ls: cannot access 'artifacts/Tuxemon-x86_64': No such file or directory
```
grafik

JaskRendix:
@Murmele  I think the issue is here:
```
      - name: list artifacts folder
        run: |
          echo "Show artifacts folder:"
          ls -R artifacts
          echo "Show artifacts/tuxemon_windows_installer folder:"
          ls -R "artifacts/tuxemon_windows_installer"
          echo "Show artifacts/Tuxemon-x86_64:"
          ls -R "artifacts/Tuxemon-x86_64"   <-------------
```
because I think if we switch the names
```
          echo "Show artifacts/Tuxemon-x86_64:"
          ls -R "artifacts/Tuxemon-x86_64"
          echo "Show artifacts/tuxemon_windows_installer folder:"
          ls -R "artifacts/tuxemon_windows_installer"
```
we would have the same issue with windows, it's like the ls -R is unable to find it

maybe we need only to do like this
```
          ls -R artifacts
          echo "Show artifacts/tuxemon_windows_installer folder:"
          ls -R "artifacts/tuxemon_windows_installer"
          ls -R artifacts
          echo "Show artifacts/Tuxemon-x86_64:"
          ls -R "artifacts/Tuxemon-x86_64"
```

Murmele:
> artifacts

But why` ls -R artifacts` did not list anything? It should list the Tuxemon-x86_64 folder if it exists.

Can you extend the flatpak run? (`verbose: true`) is the first line I added
```
    - name: Build package
      uses: flatpak/flatpak-github-actions/flatpak-builder@v6
      with:
        bundle: Tuxemon.flatpak
        manifest-path: buildconfig/flatpak/org.tuxemon.Tuxemon.yaml
        cache: false
        branch: ${{ steps.flatpak_release_branch.outputs.value }}
        verbose: true

    - name: Download artifacts (just for testing!)
        uses: actions/download-artifact@v4
        with:
          path: artifacts

    - name: list artifacts folder (just for testing)
      run: |
        echo "Show artifacts folder:"
        ls artifacts
        echo "Show artifacts/Tuxemon-x86_64:"
        ls "artifacts/Tuxemon-x86_64"
```

PR #2235, opened 2024-01-27
home (battle)

JaskRendix:
PR:
- sets new parameters in **prepare.py** corresponding to the position of the lines inside the hud (precisely where is blit the "name monster, lv 3, etc.";
- renames correctly the labels in **prepare.py**, before player? opponent? no, there is right and there is left;
- fixes a difference between wild encounters and trainer battles. In the trainer battles Boxali (the opponent) has the head outside the screen, so I decided to lower the sprite. By doing this (as first image), I discovered that in wild encounters the image is always lower, so I edited the value (scale).

this is boxali in a trainer battle
Screenshot_2024-01-27_20-05-39
this is boxali in a wild encounter
Screenshot_2024-01-27_20-30-30
this is boxali in a wild encounter after the PR
Screenshot_2024-01-27_21-35-45


xirsoi (dismissed):

xirsoi:
@JaskRendix i'm not sure exactly what it's pissed about, but the tests are failing you on a style guide issue.

xirsoi (approved):

PR #2241, opened 2024-02-01
test_actions + various

JaskRendix:
PR:
- adds some new tests (from 121 to 198);
- replaces some **return** with **raise**;
- fixes some weird things I bumped into (lack of checks, etc.);

anticipated from #2253  and added tests:
- modify_money (values stored in variables) ;
- money_is (checks against values stored in variables);
- format_variable (added -int and -float, removed rounding since I want to keep it simple, I'll add it separately);


Grimmys (changes requested):
That's not how tests are supposed to be done.

One test = one def.

Because else, you can introduce side effect from one test to another one, and if it's failing, it's harder to see what's the issue exactly, because it would not be a specific method that would be failing, but the only single test defined in the file.

EDIT: some simple example from the official documentation of the testing framework we are using: https://docs.python.org/3/library/unittest.html#basic-example
We should aim for a similar structure.

Grimmys (approved):

bitcraft (approved):
great to see tests added, thank you!

PR #1304, opened 2022-09-10
add money dict

JaskRendix:
This was intended for fixing the #1089 

@ultidonki I think the idea of the dictionary can simplify a lot the troubles with "buying and selling" (recent updates in npc.py).

Now we can associate the money to a specific slug, it could be the slug of the scoop as well as of NPCs.

**Following updates:**

1. inserted a new translation called "player_wallet";

```
msgstr "${{name}} Geldbörse: ${{money}} ${{currency}}" -> DE
msgstr "${{name}}'s wallet: ${{money}} ${{currency}}" -> EN
msgstr "La cartera de ${{name}}: ${{money}} ${{currency}}" -> ES
msgstr "La cartera de ${{name}}: ${{money}} ${{currency}}" -> ES/MX
msgstr "Portefeuille de ${{name}}: ${{money}} ${{currency}}" -> FR
msgstr "Il portafoglio di ${{name}}: ${{money}} ${{currency}}" -> IT
msgstr "Carteira ${{name}}: ${{money}} ${{currency}}" -> PR
```
So we are already covered on the majority of languages.

2. inserted in local.py the ${{money}} related to the player's wallet;
3. cleaned up artshop (there was a duplicate of a translation in three .po files);
4. save_upgrader, defined a function to set an amount of money for all the old savegames;
5. fixed a bunch of logger.error;

Now buy works perfectly, it reduces player's money.

PR #2242, opened 2024-02-02
fix locale + control

JaskRendix:
closed because old


tuxemon/locale.py
```diff
@@ -216,7 +216,8 @@ def replace_text(session: Session, text: str) -> str:
     text = text.replace("${{name}}", player.name)
     text = text.replace("${{currency}}", "$")
     text = text.replace(r"\n", "\n")
-    text = text.replace("${{money}}", str(player.money["player"]))
+    money = player.money.get("player", 0)
```
xirsoi: why do we need to specify the player when we already have the player?

xirsoi: this way seems more complicated.
admittedly, it's a marginal increase in complexity, but it's still an increase.

Is there a benefit to this method?

JaskRendix: because the money is a dict[str, int], where str is "player"

tuxemon/locale.py
```diff
@@ -247,29 +244,23 @@ def replace_text(session: Session, text: str) -> str:
...
-        )
-        text = text.replace(
-            "${{monster_" + str(i) + "_hp_max}}",
-            str(monster.hp),
-        )
-        text = text.replace(
-            "${{monster_" + str(i) + "_level}}",
-            str(monster.level),
-        )
+        _name = "${{monster_" + str(i) + "_name}}"
```
xirsoi: I'm not sure this is an improvement. Saving the new strings to a var doesn't really help anything, since they're only used once. It's maybe *slightly* more aesthetic, but I think that's just a matter of opinion.

I'd like to get other people's opinion though. @vnmabus @bitcraft or @kerizane.

JaskRendix: I don't know, potentially we can remove everything in that loop (included the loop itself), because it's used only once in an unused map ("You watch some TV with your beloved Lv ${{monster_0_level}} ${{monster_0_name}}.,0")

tuxemon/locale.py
xirsoi: why is it that way? i must have not noticed the change in a previous PR. is this to support multiple wallets or something?

tuxemon/locale.py
xirsoi: no, i mean that i don't think it makes sense to save the constructed strings to a local variable (_name, _desc, etc.) when they are used only once and on the very next line.

the previous implementation has the strings built in-line with the string replace

tuxemon/locale.py
JaskRendix: exactly, it supports multiple wallets, here we go #1304, back in September 2022

tuxemon/locale.py
JaskRendix: ah ok, no problem

JaskRendix: I went back to the original structure, since I have an idea.

tuxemon/locale.py
```diff
@@ -214,9 +214,17 @@ def replace_text(session: Session, text: str) -> str:
     player = session.player
     client = session.client
     text = text.replace("${{name}}", player.name)
+    text = text.replace("${{NAME}}", player.name.upper())
     text = text.replace("${{currency}}", "$")
     text = text.replace(r"\n", "\n")
-    text = text.replace("${{money}}", str(player.money["player"]))
+    money = player.money.get("player", 0)
+    text = text.replace("${{money}}", str(money))
+    # tuxepedia
```
Grimmys: Unnecessary comment. It explains nothing. Let's remove it.

tuxemon/locale.py
```diff
@@ -247,29 +251,82 @@ def replace_text(session: Session, text: str) -> str:
...
-        )
-        text = text.replace(
-            "${{monster_" + str(i) + "_hp_max}}",
-            str(monster.hp),
-        )
-        text = text.replace(
-            "${{monster_" + str(i) + "_level}}",
-            str(monster.level),
-        )
+        # name
```
Grimmys: In my opinion, all the comments below are useless.
They are just repetitions of what can already been deduced from reading the name of the variables.

tuxemon/locale.py
```diff
@@ -247,29 +251,62 @@ def replace_text(session: Session, text: str) -> str:
...
+        _name = monster.name
+        text = text.replace("${{monster_" + str(i) + "_name}}", _name)
+        _description = monster.description
+        text = text.replace("${{monster_" + str(i) + "_desc}}", _description)
+        _types = ""
+        for _type in monster.types:
+            _types += T.translate(_type.name) + " - "
+        text = text.replace("${{monster_" + str(i) + "_types}}", _types[:-3])
+        _category = monster.category
+        text = text.replace("${{monster_" + str(i) + "_category}}", _category)
```
xirsoi: I think i wasn't being clear, sorry about that.
This line should just be

`text = text.replace("${{monster_" + str(i) + "_category}}", monster.category)`

Same with all the other lines in this section. There's no point in saving the value, except in the case of types, since there's multiple.

xirsoi (changes requested):

tuxemon/locale.py
JaskRendix: I understood, you were crystal clear in the previous comment, I simply judged the code too long (verbose), reason why I saved the value. Everything has been fixed.

bitcraft:
@xirsoi do the changes look good to you now?

PR #2244, opened 2024-02-03
fix teleport

JaskRendix:
PR fixes teleport.

if someone uses **transition_teleport** in the same map:
```
player in map_cotton (1,1)
transition_teleport map_cotton,1,5,0.3)
player in map_cotton (1,5)
```
then the tile (1,1) isn't accessible, because it's in the **collision_map** (since occupied by the player); so it was necessary to add a **remove_collision**.

### DIFF
--- a/tuxemon/event/actions/delayed_teleport.py
+++ b/tuxemon/event/actions/delayed_teleport.py
@@ -45,7 +45,8 @@ class DelayedTeleportAction(EventAction):
         if world.delayed_teleport:
             logger.error("Stop, there is a teleport in progress")
             return
-
+        player = self.session.player
+        player.remove_collision(player.tile_pos)
         world.delayed_teleport = True
         world.delayed_mapname = self.map_name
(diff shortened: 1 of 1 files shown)


Grimmys:
Shouldn't collision be re-enable later on?

JaskRendix:
No, because the tile (1,1) isn't occupied anymore.
player.tile_pos is the old position (1,1) while self.position_x, self.position_y will be the new one (1,5)
the collision for (1,5) is automatically enabled

Grimmys:
No it's not what I meant.

After tp, you added a line for disabling collision checks, so if I got it correctly it means the player is now free to go wherever they want.

However, you never enable it back. So does it mean the player can move freely accross walls after the teleportation?

JaskRendix:
> So does it mean the player can move freely accross walls after the teleportation?

No

Grimmys (approved):

bitcraft (approved):

PR #2247, opened 2024-02-04
regionproperties + collision zone

JaskRendix:
PR fixes wrong behavior of collision zone:
- acts on **add_collision**;
- acts on **remove_collision**;
- updates the **get_collision_map**;
- ameliorates swim (Spyder) by adding a new event (Spyder.yaml);

let's say there is a tree and the trunk is both **enter_from** and **exit_from** as **up**, **left** and **right**

right now, this doesn't work, because when walking "behind" the tree (as supposed), these limitations get lost, specifically these RegionProperties are overwritten and replaced with (up, down, left and right).
`enter_from = ["up", "right", "left"] -->  ["up", "right", "left", "down"]`

after the PR the system checks if there is an existing collision zone, it checks if there is a region property; if there is a regionproperties is going to updated only the specific field (eg NPC when the player or another NPC is walking), in this way there is no more overwriting.

### DIFF
--- a/tuxemon/states/world/worldstate.py
+++ b/tuxemon/states/world/worldstate.py
@@ -798,21 +798,44 @@ class WorldState(state.State):
 
         # Get all the NPCs' tile positions
         for npc in self.get_all_entities():
-            prop = RegionProperties(
-                enter_from=[], exit_from=[], endure=[], entity=npc, key=None
-            )
-            pos = npc.tile_pos
-            collision_dict[pos] = prop
-
-        for coords, surface in self.surface_map.items():
-            for label, value in surface.items():
-                _prop = RegionProperties(
+            region = self.collision_map.get(npc.tile_pos)
+            if region:
+                prop = RegionProperties(
+                    region.enter_from,
+                    region.exit_from,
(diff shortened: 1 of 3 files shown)


tuxemon/states/world/worldstate.py
```diff
@@ -941,13 +964,16 @@ def get_explicit_tile_exits(
...
+                for direction in tile.exit_from:
+                    exit_tile = tuple(dirs2[direction] + position)
+                    if skip_nodes and exit_tile in skip_nodes:
+                        continue
 
-                adjacent_tiles.append(exit_tile)
-            return adjacent_tiles
+                    adjacent_tiles.append(exit_tile)
+                return adjacent_tiles
+            else:
```
Grimmys: No need for this else, since nothing is done inside.

Grimmys (changes requested):

bitcraft:
@Grimmys do the changes look good to you now?

adrienave (approved):

Grimmys (approved):

bitcraft (approved):

PR #2238, opened 2024-01-29
update maps (Spyder)

JaskRendix:
@Sanglorian regarding the expansion of the Spyder map

PR:
- updates **spyder_bedroom**, the y and x aren't necessary since both generated;
- renames **menu_park_encountered** into **menu_park_seen** (verbosity);
- adds **format_variable** inside the **park**;
- fixes dialogues (Cotton Cafe), it replaces **man** with **buddy** (fix #2249);
- adds question Y/S to confirm the choice, beginning of Spyder, as requested here: #2262 

maps:
- fixes **spyder_greenwash.tmx** collision zone (outside the map);
- fixes map **spyder_lion_mountain.tmx**;
- fixes map **spyder_park.tmx**;
- adds map **spyder_cave.tmx** to the **spyder.world**;
- fixes map **spyder_route3.tmx** collision zone;
- fixes map **spyder_leather_town.tmx** collision zone;
- adds map **spyder_park_south.tmx** and **spyder_park_cabin.tmx**, it'll follow another PR since these maps are empty at the moment;
- adds map **spyder_brideswood** (east of Paper Town), **spyder_diamond_hill** (west of Candy Town, south of Tunnel) and **spyder_candy_port** (south of Candy Town);
- adds map **routef**, **routeg**, **routeh** and **routei** (temporary names); these maps will complete the scenario map; the maps are above **Citypark**, it'll follow a PNG showing the entire world;
- renames **sphalian_** maps into **crystal_** so it follows to the anniversary gift rule as well as to avoid  confusion with the old version (there is no risk of incompatibility with old savegame because the area is not accessible, the access from route7 is still blocked);
- moves **rivertogo** variable from **Candy Town** to **Candy Port** and it updates all the transition teleports, because when the player chooses Candy it'll be teleported to Candy Port; 
- fixes imports (tilesets), the right one is .tsk, not .png;
- split out the events (yaml) for certain maps (eg all the centers in Spyder, looking forward at #2253 );
- adds **the door is locked** to all the locked doors in Spyder maps (9 buildings);
- adds inn in **Candy Town**;
- adds bank in **Crystal Town**;

PR #2250, opened 2024-02-05
fix player stuck bedroom (Xero)

JaskRendix:
PR split out from #2238 
- updates **player_bedroom** yaml, the y and x aren't necessary since both generated; in player_bedroom was missing **type:event** (so the action wasn't loaded and the player was stuck in the bedroom);


JaskRendix:
@HippasusTwo this is the reason why is stuck, there is the solution too.

bitcraft (approved):

PR #1520, opened 2022-12-28
taba town locations

JaskRendix:
PR addresses the new locations in Taba Town:
- creation of new NPCs (JSONs + inside the locations);
- creation of new dialogues (PO file);
- creation of the "signals" outside the new locations (the additions in the Taba Town tmx);

@Qiangong2  tagged for informing you and waiting for your approval.

Addendum:
In the maple bedroom (wall) there is a calendar with the 27th of April circled, while in the house of Mr and Ms we have two circles on the 30th September and 12th February. Random dates, no particular reason.

PR #2251, opened 2024-02-05
remove maps (Xero)

JaskRendix:
PR:
- removes all my additions to the Xero campaign because these don't fit into the plot. All the stuff removed have been added in these two PRs: #1508 + #1520; 
- the locations are the 4 houses in Taba and Cotton Misa in Cotton Town;
- all the maps have been deleted;
- all the dialogues have been deleted;
- all the NPCs have been deleted;
- all the buildings have the door and if the player interacts with the door appears "The door is locked";


Qiangong2:
I believe there was some misunderstanding when we were talking over discord. I have no issue with the additions, but I'm working on other parts of the story at the moment, so cannot address them right now. I'll get to them eventually, but since it doesn't interfere or mess up any part of the current story, I see no reason to remove it.

PR #2252, opened 2024-02-06
modify_monster_bond & modify_monster_stats (event action)

JaskRendix:
PR:
- updates **modify_monster_bond** (event action) and adds the possibility of giving a random value.
eg `modify_monster_bond name_variable,,1,10`
eg `modify_monster_bond name_variable,,-11,-5`

- updates **modify_monster_stats** (event action) and adds the possibility of giving a random value.
eg `modify_monster_stats name_variable,,,1,5` 

- adds tests for this new function;


tuxemon/event/actions/modify_monster_stats.py
```diff
@@ -88,6 +92,8 @@ def start(self) -> None:
 
         monster_stats = [StatType(self.stat)] if self.stat else list(StatType)
         amount_stat = 1 if self.amount is None else self.amount
+        if amount_stat == 1 and self.rand1 and self.rand2:
```
Grimmys: Shouldn't we check somewhere if `rand1` and  `rand2` are in the bounds? (i.e., 1 <= `rand1` <= 5)

Grimmys (changes requested):

tuxemon/event/actions/modify_monster_stats.py
JaskRendix: what's the point? amount_stat is overwritten, literally the second line:
`amount_stat = random.randint(self.rand1, self.rand2)`
both values are put by the modder in `modify_monster_stats name_variable,,,1,5`

adrienave: Ah right, these are the bounds ok.
Then I guess we could find better names for these identifiers, because at first I thought it was something else.

Maybe `min_stat_increase` / `max_stat_increase`.
There should at least be a `min`/`max` opposition in the names.

Grimmys: Right, I misunderstood the point of these identifiers.

Then they should be rename, to emphase they are in a min/max opposition.
Maybe `min_stat_change`/`max_stat_change`.

JaskRendix: I used simply lower_ and upper_bond, since there is another event action using these labels

Grimmys: Ok, names look better, but I don't get how the values are going to be provided...
They are never updated as far as I can see.
I guess they should be provided in the init, right?

Grimmys (approved):
Still don't get how `lower_bound` and `upper_bound` attributes are working (they are read in some places, but never updated...), but apparently tests are passing. I approve it.

bitcraft (approved):

PR #2253, opened 2024-02-06
cathedral events (Spyder)

JaskRendix:
PR implements Cathedral events:
- creates the following game variables **cathedral_share** (the percentage of money detracted after each win, by default 50%), **cathedral_interest** (the interest on the total amount of debt, by default 10%), the **cathedral_fee** (the fee if the player faints, by default 0, to be decided);
- creates a wallet (**bill_cathedral**) to manage the debt;
- the bill will be visible in the App Bank (phone);
- adds **battle_last_prize** to the battles variables, which will be used inside the YAML Spyder to detract if the trainer wins the battle, it'll trigger only if there is a debt for cathedral (check cathedral's wallet);

locations:
```
spyder_candy_center.tmx--> done
spyder_flower_center.tmx--> done
spyder_healing_center.tmx --> done
spyder_leather_center.tmx--> done
spyder_crystal_center.tmx--> done
spyder_timber_center.tmx--> done
```

```
msgstr "Would you like to heal your monsters?"

Question: Y/N

msgstr "All done. That will be ${{currency}}${{var:cathedral_hp}}."

If the player doesn't have the money:

msgstr "Oh, you can't afford to pay? No problem. We'll add it to your bill, plus ${{var:cathedral_interest}}% interest."
msgstr "Plus a ${{currency}}${{var:cathedral_fee}} call-out fee since we had to send the ambulance to collect you."
msgstr "When you win money, we'll take a share [${{var:cathedral_share}}%] until you've paid us back. Easy peasy!"

If the player has the money:

msgstr "Thank you for choosing Cathedral! Have a wonderful day!"
```
shady guy
```
msgstr "Would you like to sell one of your monsters?"

Question: Y/N

if yes
msgstr "Which one?"

pick monster
msgstr "I can give you ${{currency}}${{var:scoop_price}}, is it ok?"

if no
msgstr "Let me know if you change your mind!"

if yes, after selling
msgstr "What, you thought you'd get cash back? Tuxemon are pets, not things to be bought and sold. I'm just helping you out with your debt."

note: the player doesn't receive money, the money will lower automatically the debt/bill
```

PR #2255, opened 2024-02-09
update phone_banking (app_bank spyder)

JaskRendix:
split out from #2253

PR rewrites some parts of phone_banking, it generalizes some loop and def and it enables the possibility to list the "bills", bills are nothing else than wallet starting with "bill_", in this case a modder can create a bill by using set_money.

eg `set_money bill_electricity,100`, then it'll appear in the list **bill_electricity**, eventually the modder can add the msgid to the PO **bill_electricity** with the true label (eg: bill_electricity > "Energy Bill")

if someone uses `set_money spyder_bill_mom,100`, it looks like this (because there is no corresponding msgid in the PO):
Screenshot_2024-02-09_15-49-08

### DIFF
--- a/tuxemon/states/phone/phone_banking.py
+++ b/tuxemon/states/phone/phone_banking.py
@@ -33,99 +33,133 @@ class NuPhoneBanking(PygameMenuState):
     ) -> None:
         if "bank_account" not in self.player.money:
             self.player.money["bank_account"] = 0
-        bank = self.player.money["bank_account"]
-        money = self.player.money["player"]
+        bank_account = self.player.money["bank_account"]
+        wallet_player = self.player.money["player"]
+
+        _wallet = f"{T.translate('wallet')}: {wallet_player}"
         menu.add.label(
-            title=T.translate("wallet") + ": " + str(money),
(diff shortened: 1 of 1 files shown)


tuxemon/states/phone/phone_banking.py
```diff
@@ -33,99 +33,135 @@ def add_menu_items(
     ) -> None:
         if "bank_account" not in self.player.money:
             self.player.money["bank_account"] = 0
-        bank = self.player.money["bank_account"]
-        money = self.player.money["player"]
+        if "bill_cathedral" not in self.player.money:
```
bitcraft: this kind of mod-specific code should be avoided.  i think its ok for now since it is labeled "cathedral", but if we expand on the banking system in the future, we should plan out a way to do it with plugins.  no need to change it now, but if we make changes again, lets work out a plug instead.

JaskRendix: I'm open to plugin, but I don't know if you have some "rule" or method to do it.

bitcraft (approved):

PR #2254, opened 2024-02-08
dryad's grove (Spyder), lima pie (new item) and wolffsky (monster)

JaskRendix:
PR:
- adds new exit in the Cotton Tunnel map towards Dryad's Grove;
- edits the Dryad's Grove map: removes some trees, adds a new building (not accessible, no plan for it), enlarges the "lake", adds a new high grass zone, adds two NPCs and adds the event RUFF (before it was in **Fondent_region** - Tuxepedia, now it can be found in https://wiki.tuxemon.org/Dryad%27s_Grove);
- adds new accesses and a mini maze in the sea part of Dryad's Grove, which it'll be accessible by swimming from Sea Route C;
- adds the 279th monster (Wolffsky);
- adds the item [Lima Pie](https://wiki.tuxemon.org/Lima_Pie), the item will trigger the evolution for a new monster; 

there are some small additions to do when #2282 will be merged, but I'll open a separate PR (kinda of minor fixes, since it'll be just a couple of lines);

before
spyder_dryadsgrove_old
after
spyder_dryadsgrove

PR #2256, opened 2024-02-09
update set_party_status (event action)

JaskRendix:
split out from #2253 and #2254

PR updates set_party_status (event action):
- it shorts some words (verbosity);
- adds two new variables related to the party: **party_lost_hp** (for #2253) and **party_healthy** (for #2254)
- **party_lost_hp** is nothing than the sum of the lost HP (eg 40/100 --> 60);
- **party_healthy** is nothing more than a yes/no variable, yes when all the monsters in the party are healty (current_hp = max_hp);

### DIFF
--- a/tuxemon/event/actions/set_party_status.py
+++ b/tuxemon/event/actions/set_party_status.py
@@ -33,30 +33,37 @@ class SetPartyStatusAction(EventAction):
     character: str
 
     def start(self) -> None:
-        character = get_npc(self.session, self.character)
-        if character is None:
+        char = get_npc(self.session, self.character)
+        if char is None:
             logger.error(f"{self.character} not found")
             return
-        if not character.monsters:
-            logger.error(f"{character.name} has no monsters!")
+        if not char.monsters:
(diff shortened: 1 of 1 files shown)


tuxemon/event/actions/set_party_status.py
```diff
@@ -33,39 +33,50 @@ class SetPartyStatusAction(EventAction):
...
+            if monster.level < level_lowest:
+                level_lowest = monster.level
+            if monster.level > level_highest:
+                level_highest = monster.level
+            if monster.current_hp == monster.hp:
+                _healthy += 1
+            _lost_hp += monster.hp - monster.current_hp
+            level_average += monster.level
+
+        level_average = int(level_average / len(char.monsters))
```
bitcraft: does it make more sense to round the number instead of truncating it?

JaskRendix: True, I'm going to add the round

bitcraft (approved):

PR #2258, opened 2024-02-10
daycare (Spyder)

### LINKED ISSUE #700: "Daycare" option for tuxemon to gain XP and learn techniques
Spun out from #541 

Not an urgent issue, but at some point it would be good to have a Daycare Centre, where players can leave one or two monsters. While at the Daycare Centre, the monsters gain 1 XP per step the player takes. When the player returns, if they have taken at least 1,000 steps, each monster has learned one technique at random from another monster in the Daycare Centre. When a player takes their monster(s) back, they have to pay $1 for every XP gained. 

That would give an incentive to catch and evolve monsters that you never plan to use in battle, to unlock their techniques to teach to others. It also ties in with the theme of commercialism being replaced by cooperation.

JaskRendix:
fix #700 

PR adds the daycare in Paper Town:
- **format_variable**, **money_is** and **modify_money** will be merged in other PRs);
- adds the checking inside the boxes in **info** (event action);
- creates **give_experience** (event action);
- changes the name of the location from **daycare** to **paper_daycare** (since it's in paper town);
- resizes a collision zone;
- adds new tests (**give_experience**);

spyder_daycare


Sanglorian:
> draft because, it needs:
> 
>     * a dialogue where granny paper notices that the player has only 1 monster in the party (and that is not possible);
> 
>     * a dialogue for the book over the table (since the book will have a text when opened, I don't know something like: "it looks like a business plan");
> 
>     * other descriptions?
> 
>     * adding the tuxemon in the fenced area;

For granny noticing, how about: "Dearie, I can't take your only tuxemon! Come back when you have another."

For the book, I like "It looks like a business plan."

Sanglorian:
Looks great to me, thanks @JaskRendix !

bitcraft (approved):

PR #2261, opened 2024-02-13
music and sound (db.py)

JaskRendix:
PR:
- adds field validators to **music** and **sound** (to check if the file exists);
- adds more validators for checking the music in the battle, the sound of techniques and conditions as well as the sound of monsters, at least know everything is checked and if some file is missing we can discover it immediately;
- by checking I noticed that **sfx_faint** was missing, but it was wrongly named in the sound.json as sfx_feint; to the feint technique, since there is not feint sound, I put the default sfx_blaster;

regarding music:
- removes duplicate music **music_house_downstairs**
- renames **music_house_downstairs** into **music_home** since it's the same file: **All of Us.ogg**;
- (the renaming for Spyder maps has been handled in #2238 )
- adds to the JSONs all the missing files (the ones inside the folder, but not in the JSON);


mods/tuxemon/db/sounds/techniques.json
```diff
@@ -36,7 +36,7 @@
         "file" : "technique/drill.ogg"
     },
     {
-        "slug" : "sfx_feint",
+        "slug" : "sfx_faint",
```
xirsoi, line 39: feint is the correct spelling.
In English, a "feint" is to trick someone into opening themselves up for attack (in basketball a player might 'feint' to trick their opponent into going one way, but actually go the other).
To "Faint" is to pass out.

tuxemon/db.py
```diff
@@ -487,6 +487,18 @@ class MonsterSoundsModel(BaseModel):
         ..., description="The sound used when the monster faints"
     )
 
+    @field_validator("combat_call")
+    def combat_call_exists(cls: MonsterSoundsModel, v: str) -> str:
+        if has.db_entry("sounds", v):
+            return v
+        raise ValueError(f"the sound {v} doesn't exist in the db")
+
+    @field_validator("faint_call")
```
xirsoi, line 496: this is the correct spelling for the monster passing out

mods/tuxemon/db/sounds/techniques.json
xirsoi: i suspect this is for a monster technique that lowers the target's defense temporarily

xirsoi (dismissed):
minor change needed, just switch that one file name and slug to feint and it's good.

mods/tuxemon/db/sounds/techniques.json
JaskRendix: you're right, but if you look at the name of the file ogg, then you see faint.ogg. So I guess it must be the faint sound, while there is no feint sound (replaced it the with default blaster)

bitcraft:
@xirsoi do the changes look good?

bitcraft (approved):
it loks like the changes by x were addresed. so i will merge this.  thank you!

PR #2262, opened 2024-02-13
Updated README with TOC, Project Improvements

willsonr8:
Added a table of contents to the README file. Also added a new section titled "Project Improvements" that includes non-bugs that could be added to improve gameplay and give direction to new developers. Expanded program description. Combined "Ubuntu/Debian" section with "Ubuntu/Debian with Virtual Environment" section for readability and conciseness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,10 +333,15 @@ Proposed Improvements
 ---------------------
 - Add WASD movement compatibility to the base instruction set.
 - Add confirm button when user chooses original Tuxemon during introduction. Currently, Tuxemon selection is confirmed by user selecting the Tuxemon name. Selection of Tuxemon image sprite opens up menu of Tuxemon bio and stats. Addition of "confirm" button to this page view would increase user experience during introduction.
+- Add diagonal walking capability.
+- Add autosave feature.
+- Add save and exit feature.
+- Allow exit button to take user back to menu rather than closing program. 
 
 Known Bugs
 -----------
-- Gameplay may freeze after first Tuxemon battle encounter is completed. 
+- Gameplay may freeze after first Tuxemon battle encounter is completed.
+- Input key modulation not saving new key selections. 
 
 License
 -------
--- a/README.md
+++ b/README.md
@@ -338,11 +338,6 @@ Proposed Improvements
 - Add save and exit feature.
 - Allow exit button to take user back to menu rather than closing program. 
 
-Known Bugs
------------
-- Gameplay may freeze after first Tuxemon battle encounter is completed.
-- Input key modulation not saving new key selections. 
-
 License
 -------
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ level are welcome to join.
     - [CLI Examples](#cli-examples)
 - [Building](#building)
 - [Proposed Improvements](#proposed-improvements)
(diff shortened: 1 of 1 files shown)


JaskRendix:
I don't know about project improvements, because for instance the modder now can set the autosave, but it depends on the modder side, not the user; same for the confirmation. I suggest to open specific issues instead of listing these in the readme, because some can be easily fixed.

README.md
```diff
@@ -309,6 +328,15 @@ They will add and remove packages and could leave your OS in a bad
 state.  You should not use them on your personal computer.  Use in a vm
 or container.
 
+Proposed Improvements
```
xirsoi, line 331: Your proposals are welcome, but they don't belong in the readme file.
Please create issues for these ideas.

xirsoi (changes requested):

JaskRendix:
> Add confirm button when user chooses original Tuxemon during introduction. Currently, Tuxemon selection is confirmed by user selecting the Tuxemon name. Selection of Tuxemon image sprite opens up menu of Tuxemon bio and stats. Addition of "confirm" button to this page view would increase user experience during introduction.

done in #2238

README.md
willsonr8: Okay, I understand. Thank you for the feedback.

Qiangong2:
Any objections to closing this PR? It seems like everything it tries to add could just be split into issues.

README.md
```diff
@@ -1,15 +1,38 @@
 Tuxemon
 ==============
 
-Tuxemon is a free, open source monster-fighting RPG.  It's in constant
-development and improving all the time!  Contributors of all skill and
-level are welcome to join.
+Tuxemon is a free, open source monster-fighting role player game (RPG). It hosts a turn-based combat style along with monster acquisition features.
```
bitcraft, line 4: please don't change the formatting.  line length should be limited to 79 characters, unless it is part of markup (long hyperlinks, markdown, etc)

README.md
bitcraft: i agree.  please move these to issues

bitcraft (changes requested):

ultidonki:
I've just looked at this PR, and I actually quite like the Table of Contents, and the fact that you can click on it to jump straight to certain sections. 

However, I think the secondary levels of nesting is probably unnecessary for 'Controls' and 'CLI Interface', as those sections are very short, so I would personally just delete those 6 lines (Game Controls, Debugging, Map Editor, Setting Up, Commands, CLI Examples). 

I would keep the secondary level underneath 'Installation', as people just want to know what you can install it on, and jump straight there, so I'd keep that because it's useful.

For the rest of this PR, I think if you remove all of the other changes, and just have the changes I recommend above, to show the Table of Contents, I would personally think that's a good change. 

All of the changes to add 'Note:' and stuff, and re-word phrases, I don't think are an improvement - to me, the new phrases seem harder to read than the original. 

(For context, I'm not against changing the text and making it neater, or easier to read, I just don't think this PR does that)

And as suggested, the 'Project Improvements' section should be removed too - many of these features are already added or will never be added.

So, part of this PR is good, if you make those changes then in my opinion it's a good PR. :+1:

PR #2270, opened 2024-02-17
modify_monster_health (event action)

JaskRendix:
PR:
- creates a new event action called **modify_monster_health**, it's nothing more than a part from **set_monster_health**, but it's necessary to maintain clear distinction between the two functions;
- adds some missing event actions to the action_list;
- inverts the **float, int** into **int, float**, this because with float before I noticed the annoying behavior to transform a 2 into a 2.0, in this way there will be no more confusion;
- fixes **tests** and adds 5 new tests for **modify_monster_health** (as usual standard, positive / negative int, positive / negative float), now we are over 200+ tests);
- adds **tuxemon.event.actions.info.InfoAction** to the action_list (before it would have triggered a conflict since there was input_variable)


bitcraft (dismissed):

bitcraft:
looks good, conflicts

JaskRendix:
fixed

bitcraft (dismissed):

bitcraft:
sorry, conflicts again

bitcraft (approved):

PR #2274, opened 2024-02-19
AnimationModel (db.py)

JaskRendix:
waiting #2261 (because it'll trigger a conflict)

PR:
- adds **AnimationModel** (db.py) and includes a validator to check the file;
- uniforms animation **name** (before item and technique `_00.png` while tileset `.000.png`, now all `_00.png` because the majority was _00 (>1400 (techs + items) vs 70 (> tilesets));
- creates class **AnimationEntity** (since Animation is already present);
- updates **load_method_animation** in **combat.py** (duplicate method, it can be handled by the new class);
- sets **frame_time** as constant in **prepare.py** (0.09);
- gets rid of **load_animation_from_frames**, by moving the frames inside the new animation class, this was surperflous; 
- removes all the `self.images: Sequence[str] = []` from item, technique and condition;
- **AnimationEntity** will accept **slug**, **duration** (float, default **frame_time**) and **loop** (bool default **False**);
-with the parameters above it'll be created the animation (**self.play**) in this way **create_animation** will be found only in **AnimationEntity**:

if someone has a better proposal (instead of AnimationEntity), then feel free to suggest.
this follows the PR on **play_map_animation**

NOTE: I created a new file called **animation_entity**, but eventually we can delete it and move the class **AnimationEntity** inside animation.py, no issues in moving it


bitcraft:
#2261 has been merged

bitcraft (approved):

PR #2280, opened 2024-02-22
play_sound (event_action) + audio.py

JaskRendix:
PR:
- replaces all the **.play()** around with the action shipped through client (client.execute...);
- adds **if player** so we can use the action event outside the game (when the game starts, the sound, etc.)
- deletes **audio.py**, this because **audio.py** is serving only **play_sound** (event action), while **play_music** (event action) is using a mixer from an inside module (`platform/__init__`) that ends up in pygame.mixer; this is the reason why for the following point;
- moves parts of **audio.py** inside the **play_sound** event action (the class and some exceptions);


JaskRendix:
I'm not so sure about getting rid of audio.py

PR #2281, opened 2024-02-24
Add Black adventurers of both genders and nonbinary gender options

HippasusTwo:
Increases early game customization options, removes an errant pixel from the penguin walk animation, and changes the intro background to reflect the extra characterization options. While it is true you can just edit to have whatever sprite as your player, there weren't any Black adventurer sprites before (unless you count the firefighter) and some people would like to be lazy / or don't have the know-how on how to change their sprite, so this will appeal to casuals.
I asked on the forum and one other person was fine with the silliness of having a penguin suit character as an option, but if needed that can be removed.
I play-tested and encountered no issues; it's a pretty simple set of edits.


JaskRendix:
@Qiangong2 and @Sanglorian do you want this kind of customisation? Eventually I can split the choice for scenarios, so each one can independently decide the choices. Let me know.

Qiangong2:
Hi @HippasusTwo! I like the pr for the most part! I think it's important that we have diversity/customization in the game, it just hasn't been a priority since there's been so much else to do.

However, I actually use that warlock sprite in my campaign as the Omnichannel Acolytes. I'm fine substituting the sprite that's currently set if you want to keep it in the pr, but I'll need help finding a substitute (to be fair, it should probably be changed anyways).

Normally it would be fine, but the Omnichannel Acolytes are kinda the bad guys in the Xero campaign, so it probably wouldn't give the best impression if your character was one of the evil guys.

As for the choices though, I don't see an issue with them. As long as the Penguin sprite is different from the Tux Tuxemon sprite, it's fine by me.

I think we should consider having clothes and skin tone customization at the beginning of the game in the future. Even if there are only three or four options in the end, the prompt could be "What do you look like?" instead of "Are you a boy or a girl?"

Another thing to tackle later on is campaigns. I've been trying to leave pronouns related to the player out as much as possible, but there are some instances where a stretch of dialog may sound more "normal"/"easy to read" if there's a he/her in it.  For that, we can always establish a male/female variable that's set at the beginning, but it's something to consider later on in the dev process.

Also, maybe an option could be given to campaigns to limit player sprites to humanoid characters? It'd kinda be weird for Allie to be hitting on a Penguin. Not judging, but it makes some parts... more questionable... than others if the dialog is about an animal and not a human.

bitcraft:
I think this is fine.  I think that the penguin and warlock are odd choices, but its fair to leave them in for now.

bitcraft (approved):

PR #2286, opened 2024-02-28
minor fixes

JaskRendix:
Small fixes:
- completes #2254 (set the name is RUFF, gender and traded);
- corrects a description (give_experience);
- adds 3 miners (Spyder Route 3) to block the access to the map (west), it'll be opened later;
- str(prize) because set_var accepts str (mypy typehint);

### DIFF
--- a/mods/tuxemon/db/npc/spyder_route3_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_route3_npcs.json
@@ -108,5 +108,35 @@
         "slug": "spyder_rookie"
       }
     ]
+  },
+  {
+    "slug": "spyder_route3_miner1",
+    "template": [
+      {
+        "sprite_name": "miner_blue",
+        "combat_front": "miner",
+        "slug": "miner"
+      }
+    ]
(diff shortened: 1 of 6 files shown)


JaskRendix:
unrelated:

@Sanglorian do you want to define the gender of Billie's monsters? because at the moment it can happen that in the first fight the starter is male, in the second is female, etc. let me know (just point out which is male and which is female); the same if you want to define specific names for Billies monsters

bitcraft (approved):

Sanglorian:
> unrelated:
> 
> @Sanglorian do you want to define the gender of Billie's monsters? because at the moment it can happen that in the first fight the starter is male, in the second is female, etc. let me know (just point out which is male and which is female); the same if you want to define specific names for Billies monsters

Thanks for flagging this, @JaskRendix ! I've added genders to the monsters on Billie's wiki page:

    STARTER (max evolution) - Male
    Cardinale - Female
    Eyesore - Female
    Viviphyta - Male

PR #2287, opened 2024-03-01
skinsprite (class enum)

JaskRendix:
PR adds **skinsprite** inside the **db.py**. It'll be used to define each sprite, because the goal is to split the gender choice and the "race" (aka sprite skin color) in two separate choices. This will not be handled anymore by game variables.

Terms from, but I don't care about the names.
skin


bitcraft (approved):
good to have a foundation.  lgtm

Issue #2289, opened 2024-03-03
When you lose a battle, the game doesn't wait until you've exited the battle to continue events


Qiangong2:
When you lose a battle, the events will immediately start after the "defeated" screen pops up. Due to this, you can have text boxes overlapping. This issue is not present when you win a battle. When you win a battle, events wait to continue until you're back on the map.

**EDIT**: The game also tends to crash when that happens with the following error:
```
Traceback (most recent call last):
  File "C:\Users\Kurt\Downloads\Tuxemon\run_tuxemon.py", line 49, in <module>
    main.main(load_slot=args.slot)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\main.py", line 90, in main
    client.main()
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\client.py", line 301, in main
    update(clock_tick)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\client.py", line 346, in update
    self.event_engine.update(time_delta)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\event\eventengine.py", line 359, in update
    self.update_running_events(dt)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\event\eventengine.py", line 450, in update_running_events
    action.start()
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\event\actions\unlock_controls.py", line 30, in start
    sink_state = self.session.client.get_state_by_name(SinkState)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\client.py", line 539, in get_state_by_name
    return self.state_manager.get_state_by_name(state_name)
  File "C:\Users\Kurt\Downloads\Tuxemon\tuxemon\state.py", line 668, in get_state_by_name
    raise ValueError(f"Missing state {state_name}")
ValueError: Missing state <class 'tuxemon.states.sink.SinkState'>
```

JaskRendix:
Sinkstate means that there is a lock_control pending, where happens the issue? Taba battle arena 4?

JaskRendix:
> EDIT: The game also tends to crash when that happens with the following error:

@Qiangong2 I have no idea where you experimented this. Ah yeah, so I was forced to test the #2290 before it was merged.

[...]

PR #2290, opened 2024-03-03
Taba battle area 4th acolyte

Qiangong2:
Adds the 4th acolyte to the story and ensures the player's Tuxemon is at least level 16 for the upcoming boss


JaskRendix (approved):

JaskRendix:
Approved before noticing #2289 

First solution: wasted some time to set up a proposal for a chain of events with player vs acolyte with 3 monsters, here the file: 
[taba_ba_br_4.txt](https://github.com/Tuxemon/Tuxemon/files/14482078/taba_ba_br_4.txt)
in this case these msgid aren't necessary anymore (if you accept):
```
msgid "tabaacolyte4talk3"
msgstr "Heh. You've got spirit, I'll give you that.\n"
"But that's not even half of my power!\n"
"FOR THE GLORY OF OMNICHANNEL... YOU SHALL NOT PASS!!"

msgid "tabaacolyte4talk4"
msgstr "Hahaha! Is that really all you got?\n"
"You'll have no chance of beating the Battle Area Master if this is your limit. Let's test it more!\n"
"FOR THE GLORY OF OMNICHANNEL... YOU SHALL NOT PASS!!"
```

if you want to maintain your method 1 vs 1 vs 1 (separated monsters), it'll take me to waste some time to fix all the chain of events

JaskRendix:
here you go buddy, this is the original one (separated monsters), I wasted some time to fix all the variables, as you can notice you used too many variables, this was the cause of #2289 (crash related to the wrong position of lock/unlock as well as multiple boxes): [taba_ba_br_4.txt](https://github.com/Tuxemon/Tuxemon/files/14483363/taba_ba_br_4.txt)

Qiangong2:
@JaskRendix Thank you for the fixes. Just some clarification: why do we need to take out the `last_battle: check` variable? It's there to reset it. Otherwise, if you lost the previous battle, it will immediately run teleport_faint, right?

Also, shouldn't the unlock_controls be after the battle, not before? Otherwise there's a chance that the player could move and be out of position.

JaskRendix:
@Qiangong2 

>  Just some clarification: why do we need to take out the last_battle: check variable? It's there to reset it. Otherwise, if you lost the previous battle, it will immediately run teleport_faint, right?

it could happen, but it depends on the event you set up as well as on the complexity of the events (this one was pretty tough). You put three fights against the same dude, one after another. Look at what I did here, this is the reason why I decided to use **is char_defeated player** instead of last_battle. Because **is char_defeated player** will kick in after the battle -> checks the player's party -> confirm everyone is fainted -> true -> then the event starts
```
  <object id="18" name="check second battle-loss" type="event" x="192" y="16" width="16" height="16">
   <properties>
    <property name="act1" value="clear_variable 2ndbattle"/>
    <property name="act2" value="char_face loch,left"/>
    <property name="act4" value="translated_dialog tabaacolyte4talklost"/>
    <property name="act7" value="teleport_faint 1.5"/>
    <property name="cond1" value="is variable_set 2ndbattle:done"/>
    <property name="cond2" value="is char_defeated player"/> 
   </properties>
  </object>
```

> Also, shouldn't the unlock_controls be after the battle, not before? Otherwise there's a chance that the player could move and be out of position.

I'm accustomed to unlock before the battle because during the battle kicks in the **CombatState** (**start_battle**, during this state the player cannot move). Immediately the battle kicks in the **DialogState** (eg. translated_dialog) and the player cannot move (eg read a sign and/or speak with an NPC, and the player cannot move until there is the dialogue open).

But if you think someone can keep pressed any arrow (up, down, right or left), then feel free to add lock and unlock again, but lock and unlock in the same event, eg here:
```
    <property name="act1" value="clear_variable 1stbattle"/>
    <property name="act2" value="char_face loch,left"/>
    <property name="act3" value="translated_dialog tabaacolyte4talklost"/>
    <property name="act7" value="teleport_faint 1.5"/>
    <property name="cond1" value="is variable_set 1stbattle:done"/>
    <property name="cond2" value="is char_defeated player"/>
```
to 
```
    <property name="act1" value="lock_controls"/>
    <property name="act2" value="clear_variable 1stbattle"/>
    <property name="act3" value="char_face loch,left"/>
    <property name="act4" value="translated_dialog tabaacolyte4talklost"/>
    <property name="act6" value="unlock_controls"/>
Note: unlock always before any teleport_
because when the teleport_ kicks in, it's another state and it gets locked / unlocked by default
    <property name="act7" value="teleport_faint 1.5"/>
    <property name="cond1" value="is variable_set 1stbattle:done"/>
    <property name="cond2" value="is char_defeated player"/>
```

Qiangong2:
PR has been updated with requested changes

Qiangong2:
@JaskRendix Could you please take a look and verify everything was changed correctly?

JaskRendix:
Everything looks fine

bitcraft (approved):

PR #2293, opened 2024-03-09
Replace villain disciple/knight enby customization with Asian character

HippasusTwo:
Addresses multiple complaints / concerns from multiple people. In future hopefully we'll have more options / have race and gender as separate questions, but this will do for now.

Also edits the penguin to make it clear it's a person in a suit and not a real penguin so dialogue makes... a tiny bit more sense. Ish.


bitcraft (approved):
thank you for this.  with implementing more customization options, we should consider adding a disclaimer somewhere that we are working towards being more inclusive but there are technical limitations at the moment that are preventing from full customization.  ill try to add something next week unless someone else does first.

PR #2295, opened 2024-03-16
load_yaml (event action)

JaskRendix:
PR adds a new event action called load_yaml, in this way we can load events from file.yaml not related to the single map or scenario, it'll simplify the centers (Tuxemon centers) in the Spyder campaign by centralizing the duplicate functions. It'll come handy in adding new content, quests, etc.

**load_yaml file_yaml** (the yaml file must be in the maps folder, it can be easily changed and it must be used in the map)

eg. Tuxemon center map, spyder_healing_center, and event with act1 load_yaml cathedral, it'll load the cathedral.yaml

note: we can do this shit on Spyder because all the healing centers have the exact size (dimension) - only Candy is bigger, but "horizontally" so the main coordinates are identical -


JaskRendix:
I just noticed today the complaint regarding #2253 but without being on Discord I wasn't able to see it

bitcraft (approved):

PR #2298, opened 2024-03-19
Add upstairs to Xeno Taba House

HippasusTwo:
For some reason the map didn't seem to exist even though many Xeno maps further ahead seem to? Also gives a working script, if rather plain, but someone could always improve upon it with better dialogue later. Unlocks first badge after a battle and lets you wander past the pet rock. Plus boost the wild rate of some nearby mons because by the time you reach them you are well past level 2-5 and should be in the 6-15s.
The reason the map has wild grass on it is to prevent an accidental no-win game scenario where you can't grind anywhere to get stronger. (Didn't do this, but might be a good idea to unlock route 1 mons when the trial begins instead of after first badge considering how low level they are?)
The master mons should probably be adjusted at some point after the starter mons are better balanced. Right now Cardaling is absurdly powerful due to Give All, but also early mons all tend to be able to heal and a lot can one-shot mons of near the same level. The extreme randomness makes it hard to strategize much too.
Both winning and losing, and losing twice, work. The latter took three hours straight to get working...

### DIFF
--- a/mods/tuxemon/db/encounter/default_encounter.json
+++ b/mods/tuxemon/db/encounter/default_encounter.json
@@ -6,8 +6,8 @@
        "variable": "daytime:true",
        "exp_req_mod": 1,
        "level_range": [
-         2,
-         4
+         5,
+         10
        ],
        "monster": "eyenemy"
      },
@@ -16,8 +16,8 @@
        "variable": "daytime:true",
        "exp_req_mod": 1,
        "level_range": [
-         2,
-         4
+         5,
+         9
        ],
        "monster": "dandicub"
      },
@@ -26,8 +26,8 @@
        "variable": "daytime:true",
        "exp_req_mod": 1,
        "level_range": [
-         2,
-         4
+         5,
+         9
        ],
        "monster": "cardiling"
      },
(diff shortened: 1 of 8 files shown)


Qiangong2:
Hi @HippasusTwo, thanks for the pr! "Jim Master" made me laugh a lot 😆 I've actually been working on finishing up the Taba Battle Area on my fork which I'll submit a pr for when it's ready.

The reason that route 1 tuxemons aren't enabled yet is because Omnichannel hasn't made an official announcement that they're now found in the wild. The populace still has to be notified first, which will happen after you defeat the battle area master.

As for the comment about wild Tuxemon, that was done because, frankly, the stats portion of Tuxemon battles are kind of broken. You've already encountered the random choices that the opposing Tuxemon choose, but if you're Rockitten for instance, you need to be at least 8-10 levels above the opposing Tuxemon to survive most battles. There's an xp multiplier on each trainer battle, so you should be able to level up your Tuxemon relatively quickly.

HippasusTwo:
@Qiangong2  Ah, so I should just close this then?
Well, feel free to use any part of it.

PR #2299, opened 2024-03-20
Fix game breaking has 2 potions bug

HippasusTwo:
Very tiny change.
Check for has potion, not 'has exactly two potions', because players can buy from the shop, which would cause the old script to never trigger and then the player would be unable to progress in the game unless they guessed they needed to toss away all but two potions. Not my idea of fun and actually how my first playthrough went! I was very confused and thought I was doing something wrong.


mods/tuxemon/maps/taba_town.tmx
```diff
@@ -670,7 +670,7 @@
     <property name="cond2" value="is char_facing player,down"/>
     <property name="cond3" value="not variable_set moveon:yes"/>
     <property name="cond4" value="not variable_set timetomove:finite"/>
-    <property name="cond5" value="is has_item player,potion, equals,2"/>
+    <property name="cond5" value="is has_item player,potion"/>
```
Qiangong2: You shouldn't have any credits to purchase items yet at this point in time, but regardless, `greater_or_equal` would probably be a better fit since as it stands currently, a single potion would fulfill the requirements: https://wiki.tuxemon.org/Event_Reference#has_item

HippasusTwo: I could def. buy more potions, this happened to be me in a playthrough.

You're right. I just wasn't very familiar with the code. I'll update.

ultidonki:
I can confirm this change works, and I can also confirm that you start the game with 500 tuxedollars, so it's possible to buy more potions at the mart now.  

One small thing to ask, is can you update your Tiled version to 1.10 or higher? I think if you use the default one from the distro repository, it can sometimes lag behind (I used to use a really old version, too). The problem is just that this change should only be 1 line, but there's loads of other changes appearing too, as I think the compression is different too. Hopefully if you use version 1.10, that compression change will disappear, so it's more obvious there's only one line being changed.

I don't have the ability to merge Pull Requests, but I think this change would be nice. :smile:

HippasusTwo:
I updated, but reverting to a different compression will make a commit that is just a bunch of unintelligible changes again... and I'm not sure it actually gives the same compression anyway. So I'm not sure it helps.

kerizane (approved):

Issue #2291, opened 2024-03-04
how do you advance past the part with the first med centre?


RustoMCSpit:
im at the omnichannel hq place and im totally lost, there was some tunnel but i couldnt get past as it "was still being drilled". ive been confused for the past hour of where im supposed to go

ultidonki:
Sometimes there are bugs that will stop you from going any further, but I've just tested and it should be OK:

Follow the guide on this map - first go to '1', and you might see a cutscene before you go into the cafe, then go inside the cafe and talk to the guy below - After that you can leave town by going to '2'. 

(As you say, the tunnel opens much later in the game, you can't use it yet)

![magician_grey_front](https://github.com/Tuxemon/Tuxemon/assets/89098769/0b0fc48f-7d5a-4784-83ea-6404c7476ba8)
![spyder_cotton_town_walkthrough](https://github.com/Tuxemon/Tuxemon/assets/89098769/25d55a10-4fd9-46d5-b1c4-20f8cb4b9c44)

RustoMCSpit:
> Sometimes there are bugs that will stop you from going any further, but I've just tested and it should be OK:
> 
> Follow the guide on this map - first go to '1', and you might see a cutscene before you go into the cafe, then go inside the cafe and talk to the guy below - After that you can leave town by going to '2'.
> 
> (As you say, the tunnel opens much later in the game, you can't use it yet)
> 
[...]

Issue #1946, opened 2023-06-11
NPCs don't move when entering a cafe in Cotton City


pincher2012:
**Describe the bug**
NPCs don't move when entering a cafe in Cotton City. I think that the cut scene should play

**To Reproduce**
1. Enter the cafe in Cotton City
2. See error

**Expected behavior**
I can interact with NPC

**Screenshots**
![tuxemon_cotton_city_cafe](https://github.com/Tuxemon/Tuxemon/assets/1856521/c85c22c6-ab5c-4f22-bebc-55e13829b966)

**Desktop (please complete the following information):**
- OS: Manjaro 23.0.0
- Python Version: 3.11.3

**Additional context**
[slot2.save.zip](https://github.com/Tuxemon/Tuxemon/files/11714610/slot2.save.zip)

JaskRendix:
Yeah, this is "normal".
At the moment the player needs to move towards the guy behind (bearded one) to trigger the sequence.
I'm not going to touch it because @Sanglorian is going to update the Cotton Cafe soon.

freshreplicant:
Ah yeah this isn't a bug, but good to know it's being updated. It is pretty jarring having them all stand there like mannequins.

kerizane:
Are we happy with this scene now?

PR #2300, opened 2024-03-21
Add some movement to cafe, trainer

HippasusTwo:
Confused Guy helps lost players figure things out, and get some exp.

(Like this confused player:
https://github.com/Tuxemon/Tuxemon/issues/2291)

Some of the npcs will now occasionally change their direction they are facing. One wanders around inside and is a trainer (so grinding isn't as painful - I always found it weird that you fought your rival and then immediately again with no action in-between.)
Cafe issue;  I didn't touch the cutscene part as I think that's only supposed to trigger when you actually talk to the hacker guy:
https://github.com/Tuxemon/Tuxemon/issues/1946

### DIFF
--- a/mods/tuxemon/db/npc/confusedperson.json
+++ b/mods/tuxemon/db/npc/confusedperson.json
@@ -1,0 +1,10 @@
+{
+  "slug": "confusedperson",
+  "template": [
+    {
+      "sprite_name": "childactor_fiery",
+      "combat_front": "tennisplayer_fiery",
+      "slug": "noclass"
+    }
+  ]
+}
--- a/mods/tuxemon/db/npc/spyder_cotton_town_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_cotton_town_npcs.json
@@ -64,7 +64,7 @@
     "template": [
       {
         "sprite_name": "goth",
-        "combat_front": "adventurer",
+        "combat_front": "goth",
         "slug": "goth"
       }
     ]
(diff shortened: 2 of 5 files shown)


ultidonki:
I haven't tested it but this looks fine. If there's any problems they can be tested when doing a full playthrough. So thanks!

ultidonki:
Actually, after thinking about it - I don't like the idea of a trainer ambushing you if you're in the friendly base/cafe. During my playthrough I'm often walking around the village with low HP tuxemon, because I want to explore before going to the Health center (which is a bit boring). 

I think there was a suggestion to have a new map, a 'training area' underneath the cafe, where you could level up or catch tuxemon. If someone created that, it might be better to have a trainer or two in there.

If you still want an NPC to give advice on where to go, I think they should give advice without fighting you first. :grinning:

Sanglorian:
> Actually, after thinking about it - I don't like the idea of a trainer ambushing you if you're in the friendly base/cafe. During my playthrough I'm often walking around the village with low HP tuxemon, because I want to explore before going to the Health center (which is a bit boring).
> 
> I think there was a suggestion to have a new map, a 'training area' underneath the cafe, where you could level up or catch tuxemon. If someone created that, it might be better to have a trainer or two in there.
> 
> If you still want an NPC to give advice on where to go, I think they should give advice without fighting you first. 😀

I agree.

At the very least, maybe the NPC can ask *if* the player wants a battle, and helps them out regardless?

HippasusTwo:
Ah, it is true getting ambushed can be unfun. I'll add choices. If someone at some point wants to put trainers elsewhere, that'd be fine, but for now we really kinda need some before the 2nd rival battle as going straight from rival battle to rival battle feels kinda bad and incredibly grindy.

If you leave the town rightward you immediately get attacked by the rival. Someone who can't beat two level two mons is not going to be able to leave regardless, but with an optional lower-level battle right in front of the blocked cave they have a sense of a path toward progression which is what really matters.

kerizane (approved):

PR #2301, opened 2024-03-21
Major speed bug breaking battle orders fix, balance changes

HippasusTwo:
A comment about about the rock kitten's stats being 'bad' got me digging. This actually wasn't true: Rockitten has great speed and strength, so why was it so bad? Faster monsters were actually LESS likely to attack first, which absolutely broke Hunter body shape monsters that were frail speedsters, and made strong tanky but slow monsters into strong tanky AND FAST monsters. The opposite of game balance. If absolutely nothing of my code is taken except for the fix (simply subtracting from 0 the speed when feeding into turn order so that it would be negative/smaller for higher speeds and thus rank lower in turn order.) that would still be great.

(I first tested by setting the fast priority moves to return 0. This confirmed pretty well that lower numbers were getting prioritized, and I switched back to the randomness after.)

However, there are other changes too, mostly moving the heal spam moves up a few levels so that one is not murdered by a level 2 caterpillar before one manages to get to the tuxemon center, which obviously isn't much fun, and boosting catch rates so that one isn't chucking 5 balls in a row trying to catch a caterpillar at low health (which shouldn't seem like it should be able to escape so easily, right?). 

With those patches + my other pull request for a confused character who battles you and asks for directions out of town + fix for the game breaking 'have exactly 2 potions in Xeno' bug, the early game should be MUCH less frustrating and actually, y'know, fun. Or at least I had fun, where before I was just having raw frustration running into all of these issues. With the stat and combat system more balanced, an obvious next step would be more appropriately scaling encounters and trainer exp, but I didn't do too much of this. Feel free to tweak as you like as long as there is some kind of balance change, because the game really really needs it.

Another thing that should probably be done is lower the miss rate / randomness of hitting because that's... honestly pretty annoying to have early game combat that goes 'Combatant one missed! Combatant two missed!' like a comedy of errors. But simply having priority working correctly, and having weaker moves on level two mons, is going to do a lot to address move imbalance concerns as discussed here:
https://github.com/Tuxemon/Tuxemon/issues/1849


Sanglorian:
Thanks for this @HippasusTwo , and sorry I haven't replied until now.

Those changes sound good to me. Can you outline here how you've changed the moves and catch formula and so on, so I can make the same changes on the wiki?

ultidonki:
1. Thank you for investigating the speed problem - I was also wondering why my tuxemon, the stronger it got, the slower it appeared to get against weak tuxemon! 

However, when I tested your change, I was kind of confused why the speed values needed to be negative numbers (0 - randrange(0, speed)) for the calculation to work. Actually, I checked and it turned out the list was being sorted correctly before your change, so bigger numbers were at the beginning, and smaller ones came afterwards. 

So the problem was, why wasn't the action appearing in the order in the list? It turns out that in tuxemon/states/combat.py, handle_action_queue() - it's using self._action_queue.pop() - I think this is popping the last element in the list, rather than the first one. I think it should be pop(0) instead. I'm fairly sure this is where the actual problem is, but I haven't tested it yet.

2. It looks like your text editor has indentation for .json files set to 2 spaces, unfortunately in Tuxemon it looks like 3 is the norm. When I'm looking at your diff, I just see a bunch of lines that haven't been modified, only the indentation has changed, so it's hard to see what you've changed. 

I think you're just changing the levels certain moves are learned at, which is good - I also noticed Aardorn was kind of useless with its original moves, as Tonguespear seemed quite strong but needed a cooldown, so your only other option was Hibernate which puts you to sleep for several turns. So I think these changes are good, even though it's hard to read with the indentation.

3. I think changing the catch rate is good, to make tuxemon early on easier to catch, but I'm not sure if changing that Tuxemon's catch rate is the best way to do that - I think Catch Rate is a percentage, so bumping it up to 160% or higher seems strange? Maybe I'm wrong though as I haven't looked at the catch calculation, but I feel like changing it somewhere else might be better. Still, I do think tuxemon should be easier to catch earlier, so maybe this change is fine for now (in my opinion). 

4. The only other change I see is making the rival stronger, to make the Cardiwing level 6 instead of level 3... Honestly, I'm struggling with this battle as it is currently. (In fact, I'll move this comment below:)

These comments are not part of this pull request, but just general balance if anyone wants to look at them:
As an aside, I also agree that improving the miss rate is a good idea, as I'm also noticing a LOT of missed attacks (maybe more than 50% right now?) and a pull request to fix that somehow would probably be appreciated. 

As another aside, levelling up your Tuxemon in the first area (route 1) in the Spyder campaign is kind of miserable, as you need approx 50-100 EXP to level up, but one battle against a level 2 aardorn only gave me 2 EXP, I think? I'll keep battling and see if it gets any easier, but currently I'm struggling to level up my tuxemon for the Rival battle. But some sort of experience boost to the calculation might be appreciated too, because right now I feel it's way too low at the beginning. 
Or, lowering the levels of the 2nd Rival battle might be helpful too, as most of the Tuxemon I've caught so far are only level 2 and 3, but the rival has level 6 tuxemon already.

ultidonki:
Another thought here (nothing to do with your PR), but if the config `items_consumed_on_failure` could be set to False by default, then it doesn't matter if the tuxemon are hard to catch - as you won't waste all your capture devices on them. 

After setting that config to False myself, I find myself much less frustrated when trying to catch tuxemon, especially as it's so hard to earn money to buy capture devices in the first place in the early game.

(I might start a Discussion later to gather all these game-balance ideas, sorry for putting them all in your PRs!)

Sanglorian:
> Another thought here (nothing to do with your PR), but if the config `items_consumed_on_failure` could be set to False by default, then it doesn't matter if the tuxemon are hard to catch - as you won't waste all your capture devices on them.
> 
> After setting that config to False myself, I find myself much less frustrated when trying to catch tuxemon, especially as it's so hard to earn money to buy capture devices in the first place in the early game.
> 
> (I might start a Discussion later to gather all these game-balance ideas, sorry for putting them all in your PRs!)

Thanks @ultidonki - a Discussion sounds good. I wonder if it's not a problem with the formulas, but with the implementation of the formulas? There shouldn't be anyone missing 50% of the time, or even close to that amount.

ultidonki:
> Thanks @ultidonki - a Discussion sounds good. I wonder if it's not a problem with the formulas, but with the implementation of the formulas? There shouldn't be anyone missing 50% of the time, or even close to that amount.

I think you're right - I've played a bit longer, and I think I was overreacting - most of the techniques only have 80% accuracy, and they seem to miss about 1/5 times, so I think that's right. 
There was a time when 4 or 5 attacks missed one after another, which was a bit frustrating, but I really think it's just random numbers and not actually bad. When I use techniques with higher accuracy they hit more often, so I think it's probably not a problem.

ultidonki:
> It turns out that in tuxemon/states/combat.py, handle_action_queue() - it's using self._action_queue.pop() - I think this is popping the last element in the list, rather than the first one. I think it should be pop(0) instead. I'm fairly sure this is where the actual problem is, but I haven't tested it yet.

After testing this, I was wrong about this - changing it to `pop(0)` meant several other things broke, eg. potions were used after attacks, and switching tuxemon took place after attacks too. I've raised an issue to discuss it here: #2315 
Although the change in your PR actually fixed it better than my pop(0) idea, so I've left that as an option to whoever looks at the issue.

HippasusTwo:
@Sanglorian 
old if technique.is_fast:
            return int(random.randrange(0, int(self.speed)) * 1.5)
vs the new one:
            return 0 - int(random.randrange(int(self.speed/2), int(self.speed)) * 2.5) 
 
and then this part:
old       else:
            return random.randrange(0, int(self.speed))
vs new:
            return 0 - random.randrange(0, int(self.speed))
So what I did was actually turn high speeds negative, as @ultidonki pointed out, but only for these checks: that ensured nothing else would break since I was unsure why exactly it was evaluating like that. I also slightly boosted the 'fast technique' thing because I felt it would compensate if you sent a low level mon against a fast higher level mon better, as it doesn't guarantee hitting first.

I mostly did not add moves but I did add a couple of headbutts - it is a very weak move, but as previously noted it makes Aardorn far less useless since before you were basically stuck getting asleep half the time. In compensation for the healspam I added negation to Tweesher, but I think I'll revert that since you nerfed Wallow. I also gave Rockitten Stone Rot which is a poisoning move, which might be over-compensating for people previously saying Rockitten is bad: a status move lets it do something useful and then switch out even if it's frail. I'll go ahead and revert that too, and people can playtest and add that back on if they feel it needs it later.

I was under the impression the catch rate was miming pokemon where it is out of 255, which was why I boosted them to what I did, mainly because if set to 100 it doesn't actually catch nearly 100 percent of the time! (I actually playtested this, that's why I upped the rate.) It does occur to me making it out of 255 might be a bit confusing for people. I do think early mons should be easier to catch than the others regardless - route 1 mons are supposed to be the easiest catches in the whole game, so I stand by my changes there in spirit if nothing else.

I'm not sure why my editor sometimes mucked up the indentation and other times left it alone: possibly because on some I only changed values of existing parts. The mucked indentation ones are largely just headbutt getting added on, so that moving a move upward wouldn't do something like leave a tuxemon with only one move, and also to fix some annoyance in move selection (ex: Aardorn mainly).

HippasusTwo:
I just can't seem to fix the styles issue. x_x I might try creating a new branch and slowly re-adding until I see what finally causes it to break or hopefully avoid it entirely, and close this pull. That would have the side effect of making the changes much cleaner for everyone else.
edit: nope, even the code snippet alone, where I could have sworn I hadn't altered the spaces at all, is making the styles checker unhappy.

HippasusTwo:
I'm gonna close this for now. A different pull request is being written up by someone else trying to update speed test.

PR #2321, opened 2024-06-17
润色中文

cooollawf:


Davidliang11292 (changes requested):
Chinese:

感谢你对Tuxemon中文翻译的润色，这些修改的大部分看起来都很好，但有一些润色存在问题，我想如果把「否」改成「不对」的话看起来会很奇怪，与此相似的还有把「但是没打中。」改成「没打到！」另外，Tuxemon目前对全角符号的支持有些问题，我很抱歉，但在目前，请使用半角符号。

English:

Thank you for your efforts in refining the Chinese translation of Tuxemon. Most of the revisions seem well done, but there are a few instances where the changes might not sit right. For example, changing "否" to "不对" could appear odd in context, and similarly, altering "但是没打中。" to "没打到！" might not convey the intended nuance. Moreover, Tuxemon is currently encountering issues with supporting full-width symbols, and for that, I apologize. Until this is resolved, please make use of half-width symbols in your translations.

PR #2331, opened 2024-08-06
class ActionQueue and speed_test function

### LINKED ISSUE #2315: If two tuxemon attack, the one with the lower speed attacks first
Problem: When two tuxemon perform an action with the same ranking (ie. both attack), the one with the smallest, not the highest, speed moves first.

Expected: The tuxemon with the highest speed should attack first.

Why it happens:
tuxemon/states/combat/combat.py

rank_action takes an action, calculates a speed, and returns a tuple(Int, Int) giving a ranking, and a speed of the tuxemon (slightly randomised).
The first Int is the ranking of the action, which is defined in 
tuxemon/prepare.py:
```
# what comes first, second, etc. during a battle
SORT_ORDER: list[str] = [
    "potion",
    "utility",
    "food",
    "quest",
    "meta",
    "damage",
]
```
Lower integers are completed first. Most battle actions are 'damage', the last one in the list, index 5.

The 2nd Int in the tuple is the calculated speed for this action, based on either the Tuxemon's or NPC's speed, and with some randomness added.

`rank_action` is used to sort a list of actions, usually only 2 - the 2 actions that need to be performed this turn.

tuxemon/states/combat/combat.py:
```
        self._action_queue.sort(key=rank_action, reverse=True)
```
It returns the list in reverse order, so higher elements are at the start, and lower elements at the end.

Finally, actions are pop()-ed in handle_action_queue: 
tuxemon/states/combat/combat.py - handle_action_queue
```
            action = self._action_queue.pop()
```
It uses pop(), so elements at the end of the list are popped first, ie. ones with smaller rankings. 

This is a problem, because if 2 actions both have the same ranking (The first Int in the tuple returned by rank_action), the order is then determined by the speed - however, in this case, you don't want the smallest speed, but the largest, which is the opposite way the speed should work (tuxemon with higher levels have higher speeds)

Here is an example of some sorted tuples returned by rank_action:

[(5, 200), (0, 10)]
Above, 1 tuxemon is using an attack, with a speed of 200, and the other tuxemon is using a potion, with a speed of 10. 
Potions should be used before attacks, so the 2nd action should be executed first, even though it has a lower speed.
This works in the current codebase.

[(5,200), (5,10)]
Above, both tuxemon are using attacks, but the first has a speed of 200, and the other a speed of 10. In this case, (only when both rankings are the same), the one with the highest speed should be chosen. Instead, the 2nd action is executed first, as it always pop()s from the end of the list.

Possible solutions:
1. Switch the rankings in tuxemon/prepare.py, so rankings and speed both have the property higher=faster.

2. Change the speed calculation so speeds are always negative from rank_action - this will sort speeds in the correct order in the list. (This already exists in an open PR - EDIT: #2301)

3. Have some extra code somewhere that sorts the ranking order from higher to lower, then the speeds from lower to higher.

Things I tried that won't work:
Changing pop() into pop(0)  to get the first element of the list means all the rankings were in reverse priority (see the 1st of the 2 examples, after making this change potions would always execute after attacks, which is wrong).

Changing reverse=True into reverse=False - same problem as pop(0), this changes the order of both the rankings and the speed, so lower rankings were performed later.

JaskRendix:
PR: 
- monster.py (enhances def speed_test);
- combat.py (changes sort > action_sort_type, rank > get_action_sort_key and simplifies the "if") > clarity reasons;
- adds "Custom sorting to handle same primary order" -> fix #2315 ;
- moves the classes out of **combat.py** in another file called **combat_classes**;
- fixes some import related to the moving (point above);
- moves the magic numbers in **prepare.py**;
- renames **DamageMap** in **DamageReport**;

original version:
- simpler: only considers the monster's base speed and a potential speed bonus based on the technique's is_fast attribute;
- limited: doesn't account for other monster attributes that might influence speed;
- potential issues: might lead to inconsistent result due to the random nature of the speed calculation and the lack of a minimum speed (@ultidonki and @HippasusTwo);

improved version:
- more complex: incorporates multiple factors influencing speed: base speed, techniques is_fast, dodge penalty and armor penalty;
- robust: ensures a minimum speed of 1, preventing issues with random number generator;
- flexible: allows for customization of attribute modifiers to balance gameplay;
- fairness: includes a basic tiebreaker mechanism to handle equal speed modifiers;

pointless words without a massive testing, here we go (GO TO THE REPLIES TO SEE THE DATA)

tested, no new typehints

unrelated: @Sanglorian  what about terrain modifiers? Look at this:

plain, grass, interior = 1.0 for everyone (neutral)
mountain: metal 1.2 (faster) > ??? 0.8 (slower)
forest: wood 1.2 (faster) -> ??? 0.8 (slower)
desert: fire 1.2 (faster) > ??? 0.8 (slower)
ocean: water 1.2 (faster) -> ??? 0.8 (slower)
cave: earth 1.2 (faster) -> ??? 0.8 (slower)

let me know


Sanglorian:
Thanks @JaskRendix!

Do I understand these options correctly? 

Old formula: The creature's maximum Speed is equal to its Speed (* 1.5 if the technique is fast). Choose a random number between 0 and the creature's maximum Speed - that is its Speed for this round.

New formula: Creature's Speed + 50 (if fast) - (Dodge*0.2 + Armor*0.05) + random number between 1 and 100 = Speed for the round.

Have I understood that correctly? I'm not very familiar with programming code unfortunately. 

If so, I don't like the idea of other stats imposing a penalty on Speed. I think speed should be based on just the Speed stat. But we could change the Speed stat itself if it needed to be more varied or something like that. 

Because Speed increases by level, I think is_fast needs to be a multiplier not a static bonus. Otherwise is_fast will become less important over time. 

I don't mind the idea of a random number being added though. That could be a good way of randomising things. (Won't you still need a tiebreaker though, in case one monster's Speed + random integer happens to equal another monster's Speed + random integer?)

---

I'm also not opposed to different terrains improving speed of certain types - although personally I hope to implement the extra elements one day, so I don't think it has to be the five types exactly matching up to five terrains. Maybe these could be special terrain or weather effects that exist in special places or after special techniques are used, rather than necessarily applying to all terrain.

JaskRendix:
@Sanglorian your points are valid and raise interesting considerations:
- dependency on other stats: you're right that linking speed to dodge and armor can be counterintuitive, it might be better to isolate speed as a core attribute;
- is_fast as a multiplier: making it a multiplier would ensure its impact remains relevant as speed increases;
- random tiebreaker: event with a random number, ties are still possible. A more robust tiebreaker might be necessary;

Output **improved**:
```
def speed_test(self, action: EnqueuedAction) -> int:
    """
    Calculate the speed modifier for the given action.
    """
    assert isinstance(action.method, Technique)
    technique = action.method
    base_speed = float(self.speed)
    base_speed_bonus = 1.5 if technique.is_fast else 1.0
    speed_modifier = base_speed * base_speed_bonus 

    # Add a controlled random element
    random_offset = random.uniform(-10, 10) 
    speed_modifier += random_offset

    # Use dodge as a tiebreaker
    speed_modifier += float(self.dodge) * 0.01

    # Ensure the speed modifier is not negative
    speed_modifier = max(speed_modifier, 1)

    return int(speed_modifier)
```
or do you want to keep the original one?
or are you thinking about something else?
Let me know, so eventually I can close it

HippasusTwo:
The speed test function definitely needed some kind of tweak due to the weird bugginess. There was something kind of off with mine (formatting issue that I don't have the energy to try to deal with right now) so I'll close my pull - that way we don't have conflicting pulls both trying to mod speed test at the same time.

I think it would be interesting to have certain campaign options to change how battles work, such as having terrains influencing or not, critical hits by speed or not: it would give a 'different generations having different mechanics feel' and make modders feel like they don't have to actively override the base game code.

JaskRendix:
As suggested by the TODO, I have implemented an ActionQueue class. This class provides a dedicated interface for operations (adding, sorting, removing, etc.), improving code clarity and reducing the potential for mistakes. 

Translate to no-coders: imagine you're running a restaurant , you have a list of orders coming in. You could just scribble them down on a greased piece of paper (like using a simple list), but that can get messy and smells and hard to keep track of. So instead you use a proper system (like an action queue), this can help you to organize orders, easily add new orders, quickly find the next order to prepare and keep track of what's going on.

Sanglorian:
> @Sanglorian your points are valid and raise interesting considerations:
> 
>     * dependency on other stats: you're right that linking speed to dodge and armor can be counterintuitive, it might be better to isolate speed as a core attribute;
> 
>     * is_fast as a multiplier: making it a multiplier would ensure its impact remains relevant as speed increases;
> 
>     * random tiebreaker: event with a random number, ties are still possible. A more robust tiebreaker might be necessary;
> 
> 
> Output:
> 
> ```
> def speed_test(self, action: EnqueuedAction) -> int:
>     """
>     Calculate the speed modifier for the given action.
>     """
>     assert isinstance(action.method, Technique)
>     technique = action.method
>     base_speed = float(self.speed)
>     base_speed_bonus = 1.5 if technique.is_fast else 1.0
>     speed_modifier = base_speed * base_speed_bonus 
> 
>     # Add a controlled random element
>     random_offset = random.uniform(-10, 10) 
>     speed_modifier += random_offset
> 
>     # Use dodge as a tiebreaker
>     speed_modifier += float(self.dodge) * 0.01
> 
>     # Ensure the speed modifier is not negative
>     speed_modifier = max(speed_modifier, 1)
> 
>     return int(speed_modifier)
> ```
> 
> or do you want to keep the original one? or are you thinking about something else? Let me know, so eventually I can close it

This sounds great, thank you! (As in, the new formula you describe sounds great).

JaskRendix:
@Sanglorian I simulated the improved one (Rockitten lv 5 vs Cataspike lv 5) and I don't like the output, it gives 100 to 0 in some conditions. I'm going to test something else. Here the data until now.
```
Rockitten lv5 vs Cataspike lv 5
- Rockitten multiplier 7 speed (lv 5 + 7 coeff) * 5 = 12 *7 = 84
- Cataspike multiplier 5 speed (lv 5 + 7 coeff) * 5 = 12 *5 = 60
by default Rockitten is faster than Cataspike
```
data with the old formula (10k simulations):
```
both technique.is_fast TRUE: Rockitten 63.12%, Cataspike 35.72%, Equal 1.16%
both technique.is_fast FALSE: Rockitten 62.84%, Cataspike 36.0%, Equal 1.16%
Rockitten TRUE and Cataspike FALSE: Rockitten 75.3%, Cataspike 23.94%, Equal 0.76%
Rockitten FALSE and Cataspike TRUE: Rockitten 46.71%, Cataspike 52.04%, Equal 1.25%
```
data with the ditched formula (10k simulations) (the one with **50.0 if technique.is_fast else 0.0**):
```
both technique.is_fast TRUE: Rockitten 66.48%, Cataspike 32.87%, Equal 0.65%
both technique.is_fast FALSE: Rockitten 66.58%, Cataspike 32.55%, Equal 0.87%
Rockitten TRUE and Cataspike FALSE: Rockitten 94.97%, Cataspike 4.75%, Equal 0.28%
Rockitten FALSE and Cataspike TRUE: Rockitten 22.99%, Cataspike 76.4%, Equal 0.61%
```
data with the "improved" ditched formula (10k simulations) (the one with **base_speed_bonus = 1.5 if technique.is_fast else 1.0**):
```
both technique.is_fast TRUE: Rockitten 100.0%, Cataspike 0.0%, Equal 0.0%
both technique.is_fast FALSE: Rockitten 100.0%, Cataspike 0.0%, Equal 0.0%
Rockitten TRUE and Cataspike FALSE: Rockitten 100.0%, Cataspike 0.0%, Equal 0.0%
Rockitten FALSE and Cataspike TRUE: Rockitten 24.53%, Cataspike 71.95%, Equal 3.52%
```

eg. Rockitten 24.53% > Rockitten attacks first on the 24.53% of times.

Sanglorian:
Thanks for those calculations. What if the random offset were -50 to +50 instead of -10 to +10?

JaskRendix:
@Sanglorian , it looks better, the equals are even lower:
```
both technique.is_fast TRUE: Rockitten 79.86%, Cataspike 19.43%, Equal 0.71%
both technique.is_fast FALSE: Rockitten 70.44%, Cataspike 28.7%, Equal 0.86%
Rockitten TRUE and Cataspike FALSE: Rockitten 94.43%, Cataspike 5.11%, Equal 0.46%
Rockitten FALSE and Cataspike TRUE: Rockitten 43.74%, Cataspike 55.29%, Equal 0.97%
```

Sanglorian:
Nice, thanks for running those numbers. That formula looks good to me then.

HippasusTwo:
That's definitely a lot better than before.

tuxemon/monster.py
```diff
@@ -761,12 +761,28 @@ def end_combat(self) -> None:
...
+        base_speed = float(self.speed)
+        base_speed_bonus = multiplier_speed if technique.is_fast else 1.0
+        speed_modifier = base_speed * base_speed_bonus
+
+        # Add a controlled random element
+        speed_offset = prepare.SPEED_OFFSET
+        random_offset = random.uniform(-speed_offset, speed_offset)
+        speed_modifier += random_offset
+
+        # Use dodge as a tiebreaker
```
kerizane: Optional: swap these two lines to ensure tiebreaker still works.

`speed_modifier = max(speed_modifier, 1)`
`speed_modifier += float(self.dodge) * 0.01`

tuxemon/states/combat/combat_classes.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class EnqueuedAction(NamedTuple):
+    user: Union[Monster, NPC, None]
+    method: Union[Technique, Item, Condition, None]
+    target: Monster
+
+
+class ActionQueue:
+    _action_queue: list[EnqueuedAction] = []
```
kerizane: Not keen _action_queue being a class attribute.

Can you make it an instance attribute and use the same ActionQueue object, rather than creating it each time?

kerizane (changes requested):

tuxemon/monster.py
JaskRendix: ok

tuxemon/states/combat/combat_classes.py
JaskRendix: ok

kerizane (approved):

tuxemon/states/combat/combat_classes.py
```diff
@@ -0,0 +1,164 @@
...
+                    else:
+                        return primary_order, action.user.speed_test(action)
+
+        self._action_queue.sort(key=get_action_sort_key, reverse=True)
+
+        # Custom sorting to handle same primary order
+        for i in range(len(self._action_queue) - 1):
+            if self._action_queue[i][0] == self._action_queue[i + 1][0]:
+                self._action_queue[i : i + 2] = sorted(
+                    self._action_queue[i : i + 2], key=lambda x: (x[0], -x[1])
```
kerizane: Can't we do this in get_action_sort_key?

tuxemon/states/combat/combat_classes.py
```diff
@@ -0,0 +1,164 @@
...
+        """Clears the entire queue."""
+        self._action_queue.clear()
+
+    def sort(self) -> None:
+        """Sorts the queue based on the action's sort key (game rules).
+        * Techniques that damage are sorted by monster speed
+        * Items are sorted by trainer speed
+        """
+
+        def get_action_sort_key(action: EnqueuedAction) -> tuple[int, int]:
```
kerizane: Some unittests for the sorting logic would be good.

kerizane (changes requested):

tuxemon/states/combat/combat_classes.py
```diff
@@ -0,0 +1,164 @@
...
+        return len(self._action_queue) == 0
+
+    def clear_queue(self) -> None:
+        """Clears the entire queue."""
+        self._action_queue.clear()
+
+    def sort(self) -> None:
+        """Sorts the queue based on the action's sort key (game rules).
+        * Techniques that damage are sorted by monster speed
+        * Items are sorted by trainer speed
```
kerizane: Explain the highest priority actions are at the end of the queue.

tuxemon/states/combat/combat_classes.py
kerizane: You're not using "reverse=True" here, so surely this reproduces the old issue where slower attacks happen first?

tuxemon/states/combat/combat_classes.py
```diff
@@ -0,0 +1,164 @@
...
+                action_sort_type = action.method.sort
+                primary_order = prepare.SORT_ORDER.index(action_sort_type)
+
+                if action_sort_type in ["meta", "potion"]:
+                    return primary_order, 0
+                else:
+                    if action.user is None:
+                        return 0, 0
+                    else:
+                        return primary_order, action.user.speed_test(action)
```
kerizane: `                        return primary_order, -action.user.speed_test(action)`

I think if you do this you don't need the next section

kerizane (changes requested):

tuxemon/states/combat/combat_classes.py
kerizane: I think it would be clearer if you applied the negative to the primary_order, and left the reverse=True out.

JaskRendix: I'm explaining it, even if it's worthless and a waste of time, but you're right, because by negating the result (-), we are sorting the action with the same primary order in descending order based on the speed test result, so we can remove it

tuxemon/states/combat/combat_classes.py
JaskRendix: wasted some time to implement your "would be good", do you want some more tests?

JaskRendix: by the way, next time "would be good" not as requested change

kerizane (approved):

PR #2332, opened 2024-08-09
fix CATCH_RATE_RANGE constant

JaskRendix:
After multiple complaints about the difficulty in catching monsters, I decided to dig deeper to find the root of the issue. It wasn't until I wasted some time investigating that I realized the problem wasn't with the monster's stats, but rather with the catch rate calculation itself.

I'd like to propose a modification to adjust the **CATCH_RATE_RANGE** constant from (0,255) to (0,100).

The reason for this adjustment is that the majority of monsters have catch rates equal to or less than 100. By scaling down the range, we can make the catch rate values more representative of the actual capture chances. 

To illustrate the improvement, let's take a look at the graphic (results after 10k simulations):

Figure_1

Before the catch rate calculation, the probability to capture a wild monster, no matter the level, without modifiers, using a standard Tuxeball, at 5% of its lifespan (e.g., 5/100) was around 0.38 (38%). This is a relatively low capture rate.

Alternatively, we could keep the actual value of 255 and focus on making targeted adjustments to address specific complaints. For example, we could focus on adjusting the catch rates for the infamous monsters found on Route 1 and Route 2, which are notoriously difficult to catch.

Let me know

### DIFF
--- a/tuxemon/monster.py
+++ b/tuxemon/monster.py
@@ -154,11 +154,12 @@ class Monster:
         self.height = 0.0
         self.weight = 0.0
 
-        # The multiplier for checks when a monster ball is thrown this should be a value between 0-255 meaning that
-        # 0 is 0% capture rate and 255 has a very good chance of capture. This numbers are based on the capture system
-        # calculations. This is inspired by the calculations which can be found at:
-        # https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_by_catch_rate
-        self.catch_rate = 125.0
+        # The multiplier for checks when a monster ball is thrown this should be a value between 0-100 meaning that
+        # 0 is 0% capture rate and 100 has a very good chance of capture. This numbers are based on the capture system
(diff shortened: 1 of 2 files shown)


Sanglorian:
Thanks so much for figuring that out, @JaskRendix . I think using 100 instead of 255 would work great, and the odds shown in your graph (for 100) look sensible to me.

Thanks for your other comments and pull requests too - will try to look at them in the next few days.

HippasusTwo:
I had suspected something like that was going on but didn't have the energy to dig in to see if it was using 0 to 255 like I suspected. Thank you.
I think it would be easier to reduce to 0 to 100 rather than go through every monster and adjust them upward: it's pretty clear that the people who were setting the catch rates and the original person who wrote the code had very different expectations for how catch rates would actually work.
The only thing I have to add is that 'this is inspired by Pokemon Catch Rates' comment is no longer true, so perhaps should be removed or modified to 'was originally inspired by pokemon catch rates, but has been modified to fit with most people's intuitions'

kerizane (approved):

PR #2333, opened 2024-08-09
fix combat skip & fix 2340

### LINKED ISSUE #2340: If one tuxemon misses, the other tuxemon usually misses the same turn
**Describe the bug**
If one of the tuxemon's attack misses, there's a 90% chance the other tuxemon's attack will miss this turn too.

I believe it's because the value for _random_tech_hit in tuxemon/states/combat/combat.py is used for BOTH tuxemon in this turn. 

For example, if an attack with 80% accuracy misses, the other attack of 80% or lower will also miss. 

Therefore, when attacks miss, they usually come in 2's, 4's, or 6's... etc. which is frustrating for the player.

Recently, I said that in the 'Game Balance' discussion, that the attacks miss too much, but this might be the real issue - when one misses, the other one usually misses too, so you notice misses more and it feels more frustrating.

**To Reproduce**
1. Go into any tuxemon battle - random or trainer.
2. When a player/enemy attack hits, see if the other enemy/player attack hits. (It usually does).
3. When a player/enemy attack misses, see if the other enemy/player attack misses. (It usually does).

**Expected behavior**
If an attack hits or misses, it should not have any effect on whether the other tuxemon's attack hits or misses.

JaskRendix:
PR:
- simplifies the **loop** in **combat.py**;
- moves below the **AI()** after the **recharge**;
- turns **_random_tech_hit** into a **dict** and moved inside the **loop**;
- updates all the effects that use **_random_tech_hit** and simplifies some;

I stumbled upon a minor issue while testing a different scenario (Tuxemon battles with only one technique).

During a fight against a wild Pairagrin, I noticed that it couldn't repeat the same technique (Peck) consecutively, instead following a pattern of "Peck" -> "Skip" -> "Peck", and so on.

Curious, I investigated further and discovered that the recharge mechanism was occurring after the AI module, preventing the technique from recharging as expected.

To resolve this issue, I'm going to implement a micro fix that moves the recharge mechanism to occur before the AI module, ensuring that techniques recharge properly and can be used consecutively as intended.

fix #2340 as reported by @ultidonki it confirms the issue that the random was similar for both monsters, so I have implemented a fix, even if I'm thinking about something else, but for now this fix the issue (after #2331 because I don't want to trigger conflicts).

before, same turn:
```
0.15208814455599795 Peck Pairagrin Cataspike
0.15208814455599795 Bullet Cataspike Pairagrin
```

after
```
0.454057320050562 Peck Pairagrin Cataspike
0.11096621957467656 Bullet Cataspike Pairagrin
```

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -464,10 +464,10 @@ class CombatState(CombatAnimations):
                 # tracks ai players who need to choose an action
                 for trainer in self.ai_players:
                     for monster in self.monsters_in_play[trainer]:
-                        AI(self, monster, trainer)
                         # recharge opponent moves
                         for tech in monster.moves:
                             tech.recharge()
+                        AI(self, monster, trainer)
 
         elif phase == "action phase":
             self.sort_action_queue()
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -227,7 +227,7 @@ class CombatState(CombatAnimations):
         self._run: bool = False
         self._post_animation_task: Optional[Task] = None
         self._xp_message: Optional[str] = None
-        self._random_tech_hit: float = 0.0
+        self._random_tech_hit: dict[Monster, float] = {}
 
         super().__init__(players, graphics)
         self.is_trainer_battle = combat_type == "trainer"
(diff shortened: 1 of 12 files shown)


Sanglorian:
Thank you for catching and fixing this! This bug would have effectively given all NPC monsters slower recharges, which wouldn't have been very fair or balanced.

JaskRendix:
add a fix #2340

JaskRendix:
add a fix #2340 @ultidonki

kerizane (approved):

PR #2338, opened 2024-08-13
monster JSON - tags field

JaskRendix:
@Sanglorian as per discussion #2337 

This PR introduces a new 'tags' field to each monster's JSON file, a straightforward addition that primarily affects the JSON files. In `db.py`, I've added a simple validation check to ensure the 'tags' field exists and contains at least one tag. Since each monster has a body shape that counts as a tag, this check guarantees that every monster has a minimum of one tag. Additionally, I've incorporated all available terrain data from Tuxepedia, allowing us to centralize this information. This is the first step in a larger effort, and I'm excited to build upon this foundation.


kerizane:
Will merge after @Sanglorian approves.

Sanglorian:
That looks great, thanks @kerizane and @JaskRendix

kerizane (approved):

PR #2344, opened 2024-08-16
fix evolution

### LINKED ISSUE #2339: If one tuxemon evolves, sometimes others evolve when their level is too low
**Describe the bug**
When one of my tuxemon evolved, some of my other monsters in my team tried to evolve too, even though their levels were WAY below what they need to be to evolve.

I had the following team (in the save file, attached below):
Puparmor - level 11
Bolt - level 10
Aardorn - level 5
Pairagrin - level 5
Eyenemy - level 5
Axylightl - level 8

When Puparmor reaches level 12, or Bolt reaches level 12, they try to evolve (correctly).
However, sometimes other tuxemon try to evolve at the same time - eg. Aardorn, Pairagrin, and Eyenemy try to evolve too. (Axylightly has never tried to evolve, correctly)

I have a save file from just before they reached these levels, but I'm struggling to consistently reproduce this: Sometimes, just the Puparmor or Bolt evolve when they reach level 12, and other times the other tuxemon try to at the same time.

It happened 3 times:
- In one battle, all 6 tuxemon fought, and all 5 tried to evolve afterwards, but only puparmor was level 12, bolt was level 11 (the rest lower)
- In one battle, only 3 tuxemon fought, and all 3 tried to evolve afterwards after Bolt reached level 12 (the others were level 5), the others did not try to evolve
- In another battle, only 1 tuxemon fought (I think) - puparmor, but another 2 tried to evolve at the same time even though they hadn't fought (I think).

Now, I'm trying to reproduce this issue, and sometimes only puparmor or Bolt evolves - correctly. So I'm not sure exactly what's happening, but it's happened 3 times, so I'm sure it will happen again. 

Here is zip file, with the save file - the tuxemon have the levels above. If you just fight the guy standing next to me (Curie), it happened once that they all evolved, but trying it from this save point, it seems to work correctly which is frustrating - but if I fight several trainers, I can usually get the bug to happen (but it takes longer to test).
[slot2.zip](https://github.com/user-attachments/files/16629408/slot2.zip)

(You might need to be on JaskRendix's pull request #2317 to fix some bugs on Route 3 if you try to use this save file, or else the game might crash when battling trainers on Route 3 (you start this save file on route 3))

**To Reproduce**
I'm struggling to reproduce this exactly, so hopefully the steps above can help track it down.
If I work out how to reproduce it better I'll add a comment or new save file below.

**Expected behavior**
Only the tuxemon that have the correct level should evolve after a tuxemon battle.

JaskRendix:
PR:
- fix #2339 ;
- rewrites the condition and the action;

I have wasted some time in testing the savegame shared by @ultidonki and I have noticed that Puparmor wasn't updating after reaching lv 12.

### DIFF
--- a/tuxemon/event/conditions/check_evolution.py
+++ b/tuxemon/event/conditions/check_evolution.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 
-from tuxemon.db import StatType
+from tuxemon.db import MonsterEvolutionItemModel, StatType
 from tuxemon.event import MapCondition, get_npc
 from tuxemon.event.eventcondition import EventCondition
 from tuxemon.monster import Monster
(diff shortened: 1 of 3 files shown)


tuxemon/event/conditions/check_evolution.py
```diff
@@ -41,60 +41,75 @@ def test(self, session: Session, condition: MapCondition) -> bool:
...
+    ) -> bool:
+        # Check if the evolution is actually possible
+        if evo.monster_slug == monster.slug:
+            return False
+
+        conditions = [
+            evo.at_level
+            and compare("greater_or_equal", monster.level, evo.at_level)
+            or False,
+            evo.gender and evo.gender == monster.gender or False,
```
kerizane: Why do we have `or False` for all of these?

JaskRendix: Because if any of the attributes (e.g. evo.gender, evo.element, etc.) are missing (i.e., None), the entire condition will evaluate to None, not False.

Do you prefer this?
```
conditions = [
    bool(evo.at_level and compare("greater_or_equal", monster.level, evo.at_level)),
    bool(evo.gender and evo.gender == monster.gender),
    # ...
]
```

kerizane: Yeah, I think that's better.

Ensuring it's a bool isn't strictly necessary, as None is Falsey, but feel free.

tuxemon/event/conditions/check_evolution.py
```diff
@@ -41,60 +41,75 @@ def test(self, session: Session, condition: MapCondition) -> bool:
...
+        if evo.bond:
+            parts = evo.bond.split(":")
+            operator, value = parts[:2]
+            _bond = monster.bond
+            conditions.append(compare(operator, _bond, int(value)))
+
+        # If the evolution requires an item, do not evolve
+        if evo.item:
+            return False
+        # The monster can evolve if all conditions are met
```
kerizane: "all conditions are met".

"all" in comment contradicts `any(conditions)`

kerizane (changes requested):

tuxemon/event/conditions/check_evolution.py
JaskRendix: ok

tuxemon/event/conditions/check_evolution.py
JaskRendix: ok

kerizane (approved):

PR #2345, opened 2024-08-16
fix pending_queue (combat)

JaskRendix:
after mass testing for #2333 
the issue: monster flies off -> it faints (eg poison) -> it lands -> damage
the issue: monster flies off -> it faints (eg poison) -> nothing else

PR:
- fix **pending_queue** by improving the **loop**;
- removes the **fainted** user or target from the **pending_queue**;
- simplifies both effects related to the pending_queue (fly_off and land);

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -473,18 +473,22 @@ class CombatState(CombatAnimations):
             self.sort_action_queue()
 
         elif phase == "post action phase":
+            # remove actions from fainted users from the pending queue
+            self._pending_queue = [
+                pend
+                for pend in self._pending_queue
+                if pend.user
+                and isinstance(pend.user, Monster)
+                and not (fainted(pend.user) or fainted(pend.target))
+            ]
+
             # apply condition effects to the monsters
             for monster in self.active_monsters:
-                # check if there are pending actions (eg. counterattacks)
-                if self._pending_queue:
-                    pend = None
-                    for pending in self._pending_queue:
-                        pend = pending
-                    if pend:
-                        self.enqueue_action(
-                            pend.user, pend.method, pend.target
-                        )
-                        self._pending_queue.remove(pend)
+                # Check if there are pending actions (e.g. counterattacks)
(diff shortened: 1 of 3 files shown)


tuxemon/states/combat/combat.py
```diff
@@ -473,18 +473,23 @@ def transition_phase(self, phase: CombatPhase) -> None:
...
 
         elif phase == "post action phase":
+            # remove actions from fainted users from the pending queue
+            self._pending_queue = [
+                pend
+                for pend in self._pending_queue
+                if pend.user
+                and isinstance(pend.user, Monster)
+                and not fainted(pend.user)
+                or fainted(pend.target)
```
kerizane: need to think about this

kerizane: please explain the filtering here

kerizane: Is this meant to be `not (fainted(pend.user) or fainted(pend.target))` ?

I think you might want one of the following, don't mind which one.

```
 self._pending_queue = [
                pend
                for pend in self._pending_queue
                if pend.user
                and isinstance(pend.user, Monster)
                and not fainted(pend.user)
                and not fainted(pend.target)
]
```

```
 self._pending_queue = [
                pend
                for pend in self._pending_queue
                if pend.user
                and isinstance(pend.user, Monster)
                and not (fainted(pend.user) or fainted(pend.target))
]
```

Do we also want to check if pend.target is a Monster?

kerizane (changes requested):

JaskRendix:
No, pend.target is fine, it was only the pend.user the issue, thank I forgot a "not"
```
    method: Union[Technique, Item, Condition, None]
    target: Monster
```

tuxemon/states/combat/combat.py
JaskRendix: No, pend.target is fine, it was only the pend.user the issue, thank I forgot a "not"
```
    method: Union[Technique, Item, Condition, None]
    target: Monster
```

kerizane (approved):

PR #2346, opened 2024-08-16
caching icons (combat)

JaskRendix:
PR:
- **caches icons** inside a dict[str, Sprite] (as requested by the TODO);
- simplifies the loop and mechanism by splitting it in two defs;

pending:
- waiting #2333


tuxemon/states/combat/combat.py
```diff
@@ -713,47 +714,54 @@ def add_monster_into_play(
                 (partial(self.alert, message), 0)
             )
 
+    def get_status_ico(
```
kerizane: Maybe call it `get_status_icon_position`?

kerizane (changes requested):

tuxemon/states/combat/combat.py
JaskRendix: ok

kerizane (approved):

PR #2348, opened 2024-08-17
fix forfeit & run

JaskRendix:
PR:
- fixes forfeit and run, modifying the structure to accommodate multiple formulas for escaping. This change relocates the formula (run) to formula.py, enabling the use of multiple formulas that can be triggered by adjusting the game variable;
- add 3 new methods;

3 new methods:
- **always_method**, escape probability 100%

`    <property name="act1" value="set_variable method_escape:always_method"/>
`
- **never_method**, escape probability 0%

`    <property name="act1" value="set_variable method_escape:never_method"/>
`
- **relative_method**, the formula takes into account the relative strength of the user and the target, as well as their level and speed differences. The formula is designed to give a higher escape chance to users (trainer monsters) who are higher level than the monster (target, the wild one), have higher speed, and are facing a weaker monster. The max and min functions are used to ensure that the escape chance is within the valid range of 0 to 1.

`    <property name="act1" value="set_variable method_escape:relative_method"/>`

### DIFF
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -358,3 +358,57 @@ def capture(shake_check: float) -> tuple[bool, int]:
         if random_num > int(shake_check):
             return (False, i + 1)
     return (True, total_shakes)
+
+
+def attempt_escape(
+    method: str, user: Monster, target: Monster, attempts: int
+) -> bool:
+    """
+    Attempt to escape from a target monster.
+
+    Parameters:
+    - method: The escape method to use.
+    - user: The monster attempting to escape.
+    - target: The monster from which the user is attempting to escape.
+    - attempts: The number of attempts the user has made to escape so far.
+
+    Returns:
+    - bool: True if the escape is successful, False otherwise.
+
+    Raises:
+    - ValueError: If the specified method is not supported.
+    """
+
+    def relative_method() -> bool:
+        monster_strength = (target.melee + target.ranged + target.dodge) / 3
+        level_advantage = user.level - target.level
+        escape_chance = (
+            0.2
+            + (0.1 * level_advantage)
+            - (0.05 * monster_strength / 10)
+            + (0.05 * user.speed / 10)
+        )
+        escape_chance = max(0, min(escape_chance, 1))
+        return random.random() <= escape_chance
+
+    def always_method() -> bool:
+        return True
+
+    def never_method() -> bool:
+        return False
+
+    def default_method() -> bool:
+        escape_chance = 0.4 + (0.15 * (attempts + user.level - target.level))
+        return random.random() <= escape_chance
+
+    methods = {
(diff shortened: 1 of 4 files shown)


Sanglorian:
Great, thank you!

tuxemon/formula.py
```diff
@@ -358,3 +358,31 @@ def capture(shake_check: float) -> tuple[bool, int]:
...
+
+def attempt_escape(
+    method: str, user: Monster, target: Monster, attempts: int
+) -> bool:
+    """
+    Attempt to escape.
+
+    Parameters:
+    - method: The method used.
+    - level_user: The level of the user.
```
kerizane: Docstring is inconsistent with actual arguments.

E.g. level_user Vs user.

tuxemon/technique/effects/run.py
```diff
@@ -37,26 +37,36 @@ def apply(
...
+            }
+
+        # Attempt to escape
+        if formula.attempt_escape(
+            escape_method,
+            user,
+            target,
+            attempts,
+        ):
+            game_variables["run_attempts"] = attempts + 1
```
kerizane, line 66: Doesn't this mean that every time we successfully run away from a battle, running away becomes permanently easier? Seems a bit odd. Is this intentional?

kerizane (changes requested):

tuxemon/technique/effects/run.py
JaskRendix: what if it wasn't my choice? https://github.com/Tuxemon/Tuxemon/pull/1395#issuecomment-1279959092

JaskRendix: come on, let's tag the formula's creator @Sanglorian

JaskRendix: @kerizane the final decision is determined by the formula in `formula.py`, which returns a boolean value. This means that someone could develop a formula without considering the attempts counter.

JaskRendix: added other methods (no use of attempts)

JaskRendix:
@Sanglorian in this way we can set certain location where is always possible or impossible to escape from a wild monster.

kerizane (approved):

kerizane:
I think the default run method logic was slightly wrong when you first wrote it, but I will merge this and raise an issue to discuss it further.

JaskRendix:
@kerizane could you stop assuming it was my idea?

kerizane:
> @kerizane could you stop assuming it was my idea?

Yes, I understand. The behaviour was defined by Sanglorian, but I think you misunderstood what he meant slightly.

Sanglorian:
Sorry for the confusion, @kerizane and @JaskRendix . Have clarified in the other thread.

PR #2358, opened 2024-08-20
fix run attempts

### LINKED ISSUE #2355: Running will get easier on each request
See discussion in https://github.com/Tuxemon/Tuxemon/pull/2348

The formula includes "attempts", so after each attempt the user has a higher chance to escape.

```
    def default_method() -> bool:
        escape_chance = 0.4 + (0.15 * (attempts + user.level - target.level))
        return random.random() <= escape_chance
```

I believe this should be incremented after each attempt within a battle, and set to zero at the start of a new battle. It seems that this value is incremented after successfully running from battle, which means that there are two issues:
1. Running does not get easier within a single battle.
2. Running gets easier after each battle that you run from.

Initially implemented in https://github.com/Tuxemon/Tuxemon/pull/1395#issuecomment-1279959092

@Sanglorian, thoughts?

JaskRendix:
PR fixes #2355 
hoping this time I understood correctly, if the PC escape a  battle with a wild monster, then the counter resets, otherwise the counter keep increasing each failed attempt

waiting reply from @Sanglorian if renaming it

```
    def default_method() -> bool:
        if target.level > user.level:
            return False
        elif target.level < user.level:
            return True
        else:
            return random.random() >= 0.5
```

### DIFF
--- a/tuxemon/technique/effects/run.py
+++ b/tuxemon/technique/effects/run.py
@@ -63,7 +63,7 @@ class RunEffect(TechEffect):
             target,
             attempts,
         ):
-            game_variables["run_attempts"] = attempts + 1
+            game_variables["run_attempts"] = 0
             ran = True
 
         # Trigger the run effect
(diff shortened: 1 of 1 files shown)


JaskRendix:
it looks like the spyder formula will be imposed to all campaigns, boh

kerizane (approved):

PR #2363, opened 2024-08-20
post talk (Spyder)

### LINKED ISSUE #2341: Small Route 3 mapping bugs
**Describe the bug**
Some small issues with the map in Spyder Route 3 campaign:

1. 
* Go to spyder_leather_town.tmx 
* Go up to Route 3 in the top-left corner
* Go down to Leather town again
* Now, click 'Up' to try and go up to route 3 again. You can't, so you have to go down 1 square, then up again.
Expected: Clicking 'up' after the first 3 steps, you should be able to go to route 3 again without having to go down 1 square first.

2. The following trainers on Route 3 won't speak to you if you try to talk to them after they've won/been defeated.
Wanda, Spyder Knight (Weavil), the miner in the bottom-left. 
The boy in the bottom-left of the screen, near Leather Town, if you beat him then go away and come back, you can walk through him and he also won't talk to you. 
Expected: Standing in front of an NPC and clicking 'talk' should make them talk to you, even if they've already won/been defeated.

Not sure if this is a bug - None of the trainers seem to fight you again if you've already lost against them - (did that happen on the earlier routes or did I imagine that?...)

JaskRendix:
PR:
- introduces events that enable the player to engage in conversation with the NPC after a fight and fixes #2341
- adds `return value == str(int(value))`, because this will now return False for values like 1.0 or any other values that contain decimal characters, ensuring that only integer are accepted;

@Sanglorian , here #2341  @ultidonki says that he can't talk with some NPCs after fighting them, this could be solved by adding an extra event and allow the NPC to repeat the sentence that says after the combat.

right now:
NPC says XYZ -> fight > says QWE > interact again > no reply
after:
NPC says XYZ -> fight > says QWE > interact again > repeats QWE

is it ok for you or do you want to keep NPC that cannot "talk"? Let me know, because this PR solves the cases in different maps. Of course it the NPC doesn't disappear (eg. Omnichannel, etc.)

as usual I bumped into weird things, like events over collision zone, some duplicates, wrong position, etc.

### DIFF
--- a/mods/tuxemon/maps/cotton_underground.tmx
+++ b/mods/tuxemon/maps/cotton_underground.tmx
@@ -69,7 +69,7 @@
   <object id="291" type="collision" x="352" y="208" width="32" height="48"/>
   <object id="292" type="collision" x="160" y="224" width="16" height="16"/>
   <object id="293" type="collision" x="144" y="240" width="16" height="16"/>
-  <object id="294" type="collision" x="400" y="16.3333" width="48" height="16"/>
+  <object id="294" type="collision" x="400" y="16" width="48" height="16"/>
   <object id="295" type="collision" x="384" y="0" width="16" height="16"/>
  </objectgroup>
  <objectgroup color="#ffff00" id="5" name="Events">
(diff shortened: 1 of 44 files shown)


Sanglorian:
Thanks for this, good idea! Will feel more interactive than just getting no response from NPCs.

kerizane (approved):

Issue #2366, opened 2024-08-22
Game crashes in Route 4/Flower City


ultidonki:
I have some non-game crashing bugs I'll post in a day or two, but I'll put these 2 game-crashing bugs here seperately as they're more important:

(UPDATE: I've just noticed point 1 below is related to issue #2289 - it seems to be the same issue.)

1. In Route 4, if you fight Billie in the top-left hand corner (she appears if you walk around near the exit),  and lose, the game crashes with the following error. No dialogue appears, the game just immediately crashes after showing the map for about 1 second. (If you win against Billie, the game doesn't crash):
The map is spyder_route4.tmx
> Traceback (most recent call last):
  File "run_tuxemon.py", line 49, in <module>
    main.main(load_slot=args.slot)
  File "tuxemon/main.py", line 90, in main
    client.main()
  File "tuxemon/client.py", line 301, in main
    update(clock_tick)
  File "tuxemon/client.py", line 346, in update
    self.event_engine.update(time_delta)
  File "tuxemon/event/eventengine.py", line 359, in update
    self.update_running_events(dt)
  File "tuxemon/event/eventengine.py", line 450, in update_running_events
    action.start()
  File "tuxemon/event/actions/pathfind.py", line 37, in start
    assert self.npc
AssertionError

2. If you go into this house in Flower city (spyder_flower_city.tmx), and then leave through the exit on the left, into this garden:
There's no door back into the house again, and you can't walk over the fence, so you're softlocked and need to load a save file from before going through the door.
Expected behaviour: You should be able to go inside the house again somehow.

![spyder_flower_city_softlock](https://github.com/user-attachments/assets/575a826d-bd5b-4bae-ba70-363031bed3dd)

JaskRendix:
2 is already solved, regarding the first I have an idea, but I don't think it's related to #2289

JaskRendix:
I wasn't able to reproduce it (teleport to Route4 and losing against Billie on the first turn with a lv3 monster), but I got teleport back to the health center, if you can share the savegame it would be great

ultidonki:
> I wasn't able to reproduce it (teleport to Route4 and losing against Billie on the first turn with a lv3 monster), but I got teleport back to the health center, if you can share the savegame it would be great

That's strange - here's the save game I'm using: I've removed 5 tuxemon from my party so there's just 1, the game crashes every time for me. 
[...]

PR #2367, opened 2024-08-22
fix 2366 (collision zone in Flower City)

JaskRendix:
related to #2366 by @ultidonki 

> If you go into this house in Flower city (spyder_flower_city.tmx), and then leave through the exit on the left, into this garden: There's no door back into the house again, and you can't walk over the fence, so you're softlocked and need to load a save file from before going through the door. Expected behaviour: You should be able to go inside the house again somehow.

fix issue with collision zone, there was the event teleport, but it was over a collision zone, so no way to reenter inside

the other one needs more investigation


ultidonki:
That was fast :open_mouth:  Thank you for the fix!!

kerizane (approved):

PR #2385, opened 2024-08-29
menu item

JaskRendix:
I've made some changes to the MenuItem class that I wanted to walk you through. First, I added an 'enable' option to the inits, which will allow us to disable certain functions by default without those annoying popups.

As a result, I removed the code that triggered the popup, as well as the conditions in the forfeit.json file. I also got rid of the various messages in different languages that were related to the forfeit being impossible due to certain conditions, since that mechanism is no longer needed.

On a related note, I added a `__repr__` method to the MenuItem class, which helped me out with some debugging. I'm also considering changing the game_object to a callable, but I'll test that out in a separate PR.

In the menu.py file, I removed the hardcoded colors and replaced them with the centralized colors defined in prepare.py. I also renamed some variables and added some descriptions to make the code easier to understand. The biggest changes were in the `search_items()` and `on_menu_selection()` functions - I updated the return values and removed an assert statement in favor of raising a ValueError with more informative error messages. Finally, I cleaned up the combat_menu.py file by removing some unnecessary code related to the forfeit logic.

before (selectable + popup)
Screenshot_2024-08-29_15-07-53
after (not selectable)
Screenshot_2024-08-29_15-07-06


JaskRendix:
@Sanglorian forfeit needs to be blocked if there is a condition (like run)? I think we can remove that function, because forfeit needs to be intended like an action outside the scope of battle, like during a boxing match.

Moreover, eventually we could do the same for techniques, do you want it or do you prefer the popup?

https://github.com/user-attachments/assets/a76dd736-f194-474d-b6ff-359e4ed6877a

JaskRendix:
@Sanglorian forfeit needs to be blocked if there is a condition (like run)? I think we can remove that function, because forfeit needs to be intended like an action outside the scope of battle, like during a boxing match.

Moreover, eventually we could do the same for techniques, do you want it or do you prefer the popup?

this is without pop, not selectable at all
https://github.com/user-attachments/assets/a76dd736-f194-474d-b6ff-359e4ed6877a

Sanglorian:
Thanks @JaskRendix, I think we can allow Forfeit even if a condition would stop the monster from running. Exactly as you say, like a boxing match - the boxer might be trapped but they can still forfeit. 

I like greying it out for techniques too - just making it not selectable. Much cleaner than the pop up.

JaskRendix:
Sounds like we're on the same page. I think this approach is a big win, not just for the user experience, but also for our codebase. By allowing forfeit even when a condition would prevent the monster from running, we can actually simplify the implementation. I was duplicating the run mechanism when implementing forfeit, but now it looks like we can avoid that. Plus, greying out techniques that aren't selectable is a much cleaner solution than popping up a message. And as a bonus, we can get rid of those pointless messages that were just cluttering things up.

kerizane (approved):

PR #2390, opened 2024-09-01
animation evolution

JaskRendix:
PR finally added the evolution transition! It’s basically just an animation that shows the monster leveling up. Some people were really bummed it wasn't there before, but now it's finally in the game.


JaskRendix:
@ultidonki @Sanglorian

Sanglorian:
Brilliant, thanks!

HippasusTwo:
Having 7 seconds of the critter just fading back and forth seems a bit unpolished. Can we have the mid-sequence of it change color to white? Like 'baby form, white baby form, white adult form, normal colored adult form' sequence in addition to the fading in and out. I'm not experienced with Python but this code looks like it may work, or perhaps the project already has a colorize function somewhere I'm unaware of.
If we can't make it fancier, I'd just make it slightly shorter tbh.

`def colorize(image, newColor):
    """
    Create a "colorized" copy of a surface (replaces RGB values with the given color, preserving the per-pixel alphas of
    original).
    :param image: Surface to create a colorized copy of
    :param newColor: RGB color to use (original alpha values are preserved)
    :return: New colorized Surface instance
    """
    image = image.copy()

    # zero out RGB values
    image.fill((0, 0, 0, 255), None, pygame.BLEND_RGBA_MULT)
    # add in new RGB values
    image.fill(newColor[0:3] + (0,), None, pygame.BLEND_RGBA_ADD)

    return image
`
uhhh that did not display the way I was hoping to, sorry. Original source I looked at was here: https://gamedev.stackexchange.com/questions/26550/how-can-a-pygame-image-be-colored

edit:
looking around some more, it looks like it may be possible to do the coloring in just one line if all you want to do is preserve the transparent as transparent. But I'd still make a copy since we want both the white version and the original.
background.fill((255, 255, 255), special_flags=pygame.BLEND_MULT)

JaskRendix:
@HippasusTwo to make sure I get it just right, can you give me some specific feedback on what you're looking for? (precisely when and how, I don't know if you are talking about black and white, if not there is the risk we keep wasting time in back and forth, like if it was an order at the drive-thru, so I can deliver the result).

Something like: original, evolve, original, evolve b/w, etc.

>  If we can't make it fancier, I'd just make it slightly shorter tbh.

Does the fast-forward button at the end really matter to you, or is it not a big deal? Even if it's included, are you still looking for a shorter overall length? Tell me your requests.

I'm going to split out everything else in another PR.

HippasusTwo:
Sorry for lack of clarity! I'm good with either full white or 'black and white'.
If you look here:
https://www.youtube.com/watch?v=mJn5LJlX1jQ
You can see different pokemon animations and how they do timing. The early ones do black and white, but the later ones like Gen 3 have the image turn completely white, then after transitioning to the new mon, the color returns. Either is totally fine with me.

I'm not finicky about the fast forward. It just seems strange to me to have 7 seconds of just flickering back and forth between two images without any kind of transition effect like the Pokemon games had, be it turning fully white or going black and white or something else (some of the later ones got much fancier).

JaskRendix:
@HippasusTwo I wasted some time to implement some stuffs, does it meet your expectations or still not enough? It doesn't anticipate the new sprite (evolved monster), so the player can discover it at the end. I need to revisit a couple of things because it can be simpler, maybe add something else.


Remember: item evolution and sound

edit: item thing is progressing in other PRs while the sound I'll look into soon

https://github.com/user-attachments/assets/6bc3a3c0-1a17-41c1-9ddc-f2fd588b24db

JaskRendix:
@ultidonki @Sanglorian (updated video, the previous didn't meet @HippasusTwo standard)

HippasusTwo:
That's exactly what I hoped for! Thank you.

kerizane:
Looks good, well done.

kerizane (approved):

PR #2402, opened 2024-09-05
MusicPlayerState()

### LINKED ISSUE #502: Global variable to fade music in and out
I've noticed (probably like a lot of people) that the music change is rather jarring between areas. If there was a global variable to set a fade out and fade in between the different music, it would be a lot easier to handle.

For instance, in my latest edit on route2, I decided to change the music to more fit the mood of the conversation (low, darkish, kind of sad). But when the conversation finishes, you just get the loud, blaring route theme.

JaskRendix:
PR replaces the current music setup with a shiny new `MusicPlayerState` class. Gone are the days of manually managing music status, song, and previous song - now it's all neatly wrapped up in a single object. But that's not all - this PR also:
* cleans up imports in `tuxemon/platform/__init__.py` to get rid of that TODO
* creates a new `MusicPlayerState` class in `audio.py` to keep everything organized
* moves the `mixer` import from `platform` into the new class, renaming it to `mixer2` to avoid confusion
* defines clear methods for playing, pausing, stopping, fading out, and more
* adds a `__repr__` method for easy debugging
* introduces a new `MusicStatus` enum in `db.py` to keep track of music states
* updates event actions like `play_music`, `fadeout_music`, and `pause_music` to work with the new class
* adds an optional `duration` parameter to `pause_music` for more flexibility
* creates a new `unpause_music` event action to balance things out
* updates the `music_playing` event condition to work with the new setup
* improves the control system, so increasing volume happens immediately - no more exiting and reentering the map
* simplifies music actions in non-map scenarios, like starting a battle, by using `self.client.current_music` directly
* fixes #502 and introduces a `fade_ms` parameter in `play_music` to enable smooth volume transitions - set to 1 second (1000ms) by default"
* updates the `fadeout_music` and introduces the constant FADEOUT - set to 1 second (1000 ms) by default 
* updates to the 'teleport' event action, as well as 'transition_teleport'. Now, when a player teleports, the current music will automatically stop playing. This change addresses an issue reported by @ultidonki in #2404 , where the music from the previous map would continue playing even after transitioning to a new map without music. This could be pretty annoying, especially if a modder has created a map with no music, but the previous map's soundtrack keeps playing and disrupts the experience. By stopping the music on teleport, we can ensure a smoother transition and avoid any potential audio clashes. Plus, this change also makes it easier to identify which maps are intentionally designed to be silent, rather than having to wonder if the event action 'play_music' isn't present


Sanglorian:
So good!

kerizane (approved):

PR #2417, opened 2024-09-08
icon cache (fix #2412)

### LINKED ISSUE #2412: If both Tuxemon have the same status effect, only 1 icon appears
**Describe the bug**
If both tuxemon have the same status effect, only the player's icon appears.
However, in the text at the bottom of the screen, it will correctly say things like 'X has a status effect', 'Y has a status effect'.
If the status effects are different, both the enemy's icon and the player's icon appear.

See screenshot, both Aardart and Aardorn are asleep right now, but only the player's icon appears: I've also seen the same problem happen with 'Constrict', which seems to constrict both tuxemon, but only the player's icon appears.
both_asleep

**To Reproduce**
Use an Aardorn on Route 1, eventually you'll find another Aardorn.
If you use Hibernate, you'll fall asleep. The other Aardorn only has 2 or 3 abilities, so it will also use Hibernate soon. 
When they're both asleep, only the player will have the sleeping 'Z' status effect icon showing.

Or:
In the Scoop HQ, several tuxemon use the technique 'Constrict', which gives both tuxemon the 'grab' status effect. The same thing happens.

JaskRendix:
PR fixes #2412

@ultidonki I'm going to write the cause of it, even if probably I'll never get a reply (excluding maybe an emoticon or a thumbs up).

The main issue with the first version of the code what that it was using a cache to store the icon instances, but it was using the icon name as the key. This meant that if two monsters had the same status icon, they would both be using the same icon instance from the cache.

The problem with this is that the icon instance is being added to the self.sprites list multiple times, but it's the same instance each time. This means that when the icon is updated or removed, it'll affect all the monsters that are using that icon instance.

Now, I have changed the cache to use a tuple of the icon name and its position as the key. This means that each monster gets its own unique icon instance from the cache, even if they have the same status icon.

Screenshot_2024-09-08_21-30-18

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -168,7 +168,9 @@ class CombatState(CombatAnimations):
         self._run: bool = False
         self._post_animation_task: Optional[Task] = None
         self._xp_message: Optional[str] = None
-        self._status_icon_cache: dict[str, Sprite] = {}
+        self._status_icon_cache: dict[
+            tuple[str, tuple[float, float]], Sprite
+        ] = {}
         self._random_tech_hit: dict[Monster, float] = {}
 
         super().__init__(players, graphics)
@@ -657,6 +659,7 @@ class CombatState(CombatAnimations):
 
         # add status icons
         for monster in self.active_monsters:
+            self._status_icons[monster] = []
             for status in monster.status:
                 if status.icon:
                     icon_position = (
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thanks for fixing that, @JaskRendix - and for the very clear explanation of what was going on. I do find it interesting to get an insight into how the programming works and what's going on behind the scenes.

kerizane (approved):

ultidonki:
> @ultidonki I'm going to write the cause of it, even if probably I'll never get a reply 

Thank you for explaining it, I always appreciate how quickly you create pull requests and fix the issues I report :smile:

Issue #2410, opened 2024-09-08
Timber Town/Scoop HQ mapping bugs


ultidonki:
- When I lost a battle in spyder_scoop3.tmx, it didn't transport me to the most previous healer. I couldn't move, open the menu, etc. so I had to close the game and reload a save.
It worked OK when I was defeated on spyder_scoop1.tmx
- In spyder_scoop4.tmx , there's a trainer called 'Arachne' (see screenshot). If you fight this trainer, and complete the map - leaving using the stairs in the top-right - your mom will give you a surfboard.
However, I managed to complete spyder_scoop4.tmx without defeating this trainer, and the surfboard event never happened. 
After completing spyder_scoop4.tmx, Arachne leaves the map so you can never go back to get the surfboard.
In the screenshot, Arachne is in the white circle, the red line shows how I escaped without fighting Arachne.
![arachne](https://github.com/user-attachments/assets/41e47bb2-2926-4a6f-a112-36cca27017e4)

- Scoop HQ spyder_scoop2.tmx - See Save file/screenshot. The only way to access the rest of Scoop HQ is to go 'up' on the very right of the map, as if there should be a door there. (I only knew there was something missing because I'd played the dungeon before)
![spyder_scoop_2](https://github.com/user-attachments/assets/880cae54-1581-48a7-9e69-9e9957445e90)
[slot3_scoop_hq_2.zip](https://github.com/user-attachments/files/16922752/slot3_scoop_hq_2.zip)

- Same map: Maybe have a warning before you go up the stairs - "Are you sure you want to leave Scoop HQ?" As I was surprised the stairs took me outside, I thought I'd completed the dungeon already.
- After rescuing the Sailor/Captain from the Mansion, he appears in Timber Town and offers to take me to Paper, Cotton etc. towns. However, when I select an option, nothing happens - I always stay in Timber Town. The same thing happens in Flower Town too. (You can use the above save file, slot3_scoop_hq_2.zip to check it out)

JaskRendix:
> When I lost a battle in spyder_scoop3.tmx, it didn't transport me to the most previous healer. I couldn't move, open the menu, etc. so I had to close the game and reload a save. It worked OK when I was defeated on spyder_scoop1.tmx

it has been fixed in #2401 (it was a crashing issue, because the teleport back to scoop2 was directed to a point outside the map).

[...]

PR #2418, opened 2024-09-08
fix scoop2

JaskRendix:
PR fixes from #2410 

> Scoop HQ spyder_scoop2.tmx - See Save file/screenshot. The only way to access the rest of Scoop HQ is to go 'up' on the very right of the map, as if there should be a door there. (I only knew there was something missing because I'd played the dungeon before)

adds the door (missing both in spyder and xero version)

spyder_scoop2

and

> Same map: Maybe have a warning before you go up the stairs - "Are you sure you want to leave Scoop HQ?" As I was surprised the stairs took me outside, I thought I'd completed the dungeon already.

if the player moves in the tile close to the NPC, the NPC turns right and says:
"Are you sure you want to leave Scoop HQ?"
then 
Yes / No
if NO -> it goes in the tile below the stair
if YES -> it goes out

@Sanglorian


Sanglorian:
Brilliant, thank you - sorry I made it so confusing in the first place!

kerizane (approved):

PR #2427, opened 2024-09-12
various

JaskRendix:
This PR simplifies our database model by removing the redundant `moveset` element from `db.py`. Additionally, it introduces a minimum requirement of 1 tech per monster, which should prevent issues like #2328 from arising in the future.

The PR also refactors the condition for checking if monsters are available for breeding, making the logic more clear and concise.

### DIFF
--- a/tests/tuxemon/test_monster_actions.py
+++ b/tests/tuxemon/test_monster_actions.py
@@ -40,7 +40,7 @@ class TestMonsterActions(unittest.TestCase):
     _agnite = MonsterModel(
         slug="agnite",
         category="false_dragon",
-        moveset=[],
+        moveset=[{"level_learned": 1, "technique": "ram"}],
         evolutions=[],
         history=[],
         tags=["dragon", "coastal", "desert", "mountains"],
@@ -58,7 +58,7 @@ class TestMonsterActions(unittest.TestCase):
     _nut = MonsterModel(
         slug="nut",
         category="hardware",
-        moveset=[],
+        moveset=[{"level_learned": 1, "technique": "ram"}],
         evolutions=[],
         history=[],
         tags=["blob"],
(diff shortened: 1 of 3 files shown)


JaskRendix:
unrelated @Sanglorian what do you think about these new fields in the monster JSON (moveset)

I'm asking it because of this:

> I have developed a potion that allows a tuxemon to re-learn a technique that it has forgotten. Shall I give the Re-Learner Potion to one of your tuxemon?

in this way the can avoid to forget certain techniques (eg reserved techniques) as well as avoid that some techniques can be learned again once forgotten. It could simplify all the process depending on some flags.
```
    relearnable: bool = Field(
        ..., description="Whether the monster can relearn the technique if it forgets it"
    )
    forgettable: bool = Field(
        ..., description="Whether the monster can forget the technique"
    )
```
it'll appear:
```
    "moveset": [
        {
            "level_learned": 2,
            "technique": "ram",
            "relearnable": True,
            "forgettable": True
        }
    ],
```

kerizane (approved):

Sanglorian:
> unrelated @Sanglorian what do you think about these new fields in the monster JSON (moveset)
> 
> I'm asking it because of this:
> 
> > I have developed a potion that allows a tuxemon to re-learn a technique that it has forgotten. Shall I give the Re-Learner Potion to one of your tuxemon?
> 
> in this way the can avoid to forget certain techniques (eg reserved techniques) as well as avoid that some techniques can be learned again once forgotten. It could simplify all the process depending on some flags.
> 
> ```
>     relearnable: bool = Field(
>         ..., description="Whether the monster can relearn the technique if it forgets it"
>     )
>     forgettable: bool = Field(
>         ..., description="Whether the monster can forget the technique"
>     )
> ```
> 
> it'll appear:
> 
> ```
>     "moveset": [
>         {
>             "level_learned": 2,
>             "technique": "ram",
>             "relearnable": True,
>             "forgettable": True
>         }
>     ],
> ```

That sounds like a good idea. Although I'm happy for reserved techniques to be forgettable. Sometimes, a reserved technique won't be the right fit for a particular player's strategy. 

But yes, there might be other techniques we don't want to be forgettable. 

I think all techniques would be re-learnable.

PR #2437, opened 2024-09-14
tech tags

JaskRendix:
PR for @Sanglorian as requested  [here](https://github.com/Tuxemon/Tuxemon/discussions/2268#discussioncomment-10644249) First, it refines some translations and tweaks the accuracy and potency parameters of techniques to make them more balanced. But that's not all - it also introduces two new fields that will be game-changers for monster movesets. The first field is for categorizing techniques, think 'animal', 'exotic', and so on. The second field is for tagging techniques with specific keywords like 'bird', 'megafauna', or whatever fits the bill. These new fields will be super helpful in automating the assignment of techniques to monsters, ensuring that we don't end up with a water monster wielding fire techniques or a monster with an overpowered moveset. Finally, this PR transforms the 'cavity' technique into a reserved technique and replaces it with a new one in the monsters that previously had it in their moveset.

---

@Sanglorian the data you asked for, remember that these results are without the 20 technique above. If you have specific questions, feel free to ask.

What do we need?
- we need a range for min and max number of techniques (moveset), at the moment min 4 and max 15 (see below)
- we need a range of % for moves - for instance **simple** techniques must be between 40-60%, etc. depending on xxx (see monster examples at the bottom)

```
Average number of moves per monster: 10.104026845637584
Minimum number of moves: 4
Maximum number of moves: 15
```
Here: https://wiki.tuxemon.org/Technique_Distribution

Data per specific monster, some examples:
```
hydrone -> standalone
Simple 40.0 %
Basic 20.0 %
Powerful 30.0 %
Animal 10.0 %

eyenemy shares the name techniques with eyesore
Exotic 25.0 %
Simple 50.0 %
Animal 12.5 %
Condition Imposer 12.5 %
```

PR #2438, opened 2024-09-15
new monster + tuxepedia sync

JaskRendix:
PR adds the new monsters and sprites.

I noticed that some of the sprites weren't the standard 24x24 (6) or 64x64 (only 1) sizes, so I've corrected and replaced them on Tuxepedia. Additionally, I've completed all the translations and descriptions for the species.

As for the technique, I'll be tackling that in a separate PR. I'm waiting for PR #2437 to be resolved first, and then I'll move forward with the technique updates. In the meantime, I'd appreciate it if you could hold off on adding any new techniques (with the exception of evolving monsters that already have a defined moveset) until I've had a chance to get that sorted out.

2 commits: 1st sprites, 2nd json


JaskRendix:
Hey @Sanglorian, just a heads up, I've made some updates to Tuxepedia. I've fixed the healing power to match what we have in the game, and I've also added proportional healing and capped the value at 3. I've also added some descriptions to the technique form to make it easier to input the right type of value. Oh, and I've added Solar Synthesis too.

I took a quick look at the monsters and I think we need to revisit some of their techniques. I mean, take Banling for example - it's a tree monster, but it has Peck? That doesn't really fit. I think we can do better. What do you think? Want to start reassigning some techniques to the monsters that need them?

Could you give me the ranges for each move family? Like, what percentage of moves are exotic techniques, reserved, and so on? That way, I can run a check and see which monsters need their movesets updated. We can focus on the numbers for now and worry about the levels later. Just give me a rough idea of what the distribution looks like, and we can go from there.

reserved from 0% to 5%
exotic from 10 to 15%
etc.

We need a range for min and max number of techniques (moveset), at the moment min 4 and max 15, is it ok 4 and 15? or do you prefer leave it uncapped?

Sanglorian:
> Hey @Sanglorian, just a heads up, I've made some updates to Tuxepedia. I've fixed the healing power to match what we have in the game, and I've also added proportional healing and capped the value at 3. I've also added some descriptions to the technique form to make it easier to input the right type of value. Oh, and I've added Solar Synthesis too.

Thank you!

> I took a quick look at the monsters and I think we need to revisit some of their techniques. I mean, take Banling for example - it's a tree monster, but it has Peck? That doesn't really fit. I think we can do better. What do you think? Want to start reassigning some techniques to the monsters that need them?

I think this is a great idea. Would you like to set up a new Discussions page for them? We could have one post per Tuxemon family. (Just starting with one or two, I mean, and adding more as we sort them out.)
I won't have a lot of time in the next month or so I'm afraid, but we might be able to crowdsource some ideas from the community?

> Could you give me the ranges for each move family? Like, what percentage of moves are exotic techniques, reserved, and so on? That way, I can run a check and see which monsters need their movesets updated. We can focus on the numbers for now and worry about the levels later. Just give me a rough idea of what the distribution looks like, and we can go from there.
> 
> reserved from 0% to 5% exotic from 10 to 15% etc.
> 

Hmm, how about
Reserved technique 0 to 10%
Animal + basic technique 10 to 25%
Condition imposers 10 to 25%
Simple techniques 10 to 60%
Exotic techniques 0 to 30%
Powerful techniques 0 to 40%

At least this could be a flag, for us to investigate further. Maybe there'll be some mons we want to break these rules.

> We need a range for min and max number of techniques (moveset), at the moment min 4 and max 15, is it ok 4 and 15? or do you prefer leave it uncapped?

15 sounds like a reasonable maximum. We might want to break the rule in special cases but I think it's a good rule of thumb.

JaskRendix:
> At least this could be a flag, for us to investigate further. Maybe there'll be some mons we want to break these rules.

https://wiki.tuxemon.org/Technique_Distribution

JaskRendix:
@Sanglorian 

> At least this could be a flag, for us to investigate further. Maybe there'll be some mons we want to break these rules.

https://wiki.tuxemon.org/Technique_Distribution

It looks like there was a bit of a mix-up with how properties were used in Tuxepedia. Instead of using the correct format for the `Technique Allocation` template, it seems the entries were made like this:
```
{{Technique Allocation
|Has Technique=Biting Winds
|Technique Level=52
}}
```
But the proper way to format it should have been:
```
{{Technique_Allocation
|Has_Technique=Biting Winds
|Technique_Level=52
}}
```
This discrepancy caused some issues because the data for `Has_Technique` and `Technique_Level` wasn't saved correctly. As a result, it became impossible to display the same information on other pages, like the technique page that lists monsters using that technique.

JaskRendix:
@Sanglorian so the main structure is all set, but ome of the new monsters feel a bit bare. They only have a couple of moves, especially the ones that don't evolve from others monsters (already added). I'm thinking about generating some new moves based on their tags, but maybe it's better to leave them as is and fill them later.  Waiting for changes on Tuxepedia and then sync.

> At least this could be a flag, for us to investigate further. Maybe there'll be some mons we want to break these rules.

https://wiki.tuxemon.org/Technique_Distribution

Sanglorian:
> @Sanglorian so the main structure is all set, but ome of the new monsters feel a bit bare. They only have a couple of moves, especially the ones that don't evolve from others monsters (already added). I'm thinking about generating some new moves based on their tags, but maybe it's better to leave them as is and fill them later. Waiting for changes on Tuxepedia and then sync.
> 
> > At least this could be a flag, for us to investigate further. Maybe there'll be some mons we want to break these rules.
> 
> https://wiki.tuxemon.org/Technique_Distribution

Thank you! Yes, I think okay to leave them for now and fill in later - unless you're feeling particularly inspired.

JaskRendix:
I'm glad I could surprise you with a bit of inspiration at the end. I was able to whip up a simple association script that used the existing tags to assign moves, even if they're not the best ones yet. At least it gives us some discrete placeholders for techniques, so I'm almost half satisfied with the result. I think with some further refinement of the tags and maybe increasing the number of moves, we can do a solid job.

I also took the opportunity to apply the Solar Synthesis techniques more widely, specifically to monsters with the "plant" tag. It actually helped to weed out some weird techniques, like Peck for Banling (that monster tree) and Peregrine, as well as a couple of fire techniques that didn't quite fit.

Sanglorian:
> I'm glad I could surprise you with a bit of inspiration at the end. I was able to whip up a simple association script that used the existing tags to assign moves, even if they're not the best ones yet. At least it gives us some discrete placeholders for techniques, so I'm almost half satisfied with the result. I think with some further refinement of the tags and maybe increasing the number of moves, we can do a solid job.
> 
> I also took the opportunity to apply the Solar Synthesis techniques more widely, specifically to monsters with the "plant" tag. It actually helped to weed out some weird techniques, like Peck for Banling (that monster tree) and Peregrine, as well as a couple of fire techniques that didn't quite fit.

Thank you!

kerizane (approved):

Issue #2434, opened 2024-09-14
Route 6/Candy Town mapping bugs


ultidonki:
- Billie appears and disappears from nowhere in spyder_route6.tmx
Come onto spyder_route6.tmx by coming from the map on the left. Billie will appear in the middle of the screen, then walk towards the player. After the battle, she will walk to the middle of the screen and disappear.

- The game crashes when you try to battle this guy on spyder_candy_hospital1.tmx (he's in the top-right corner):
![game_crash1](https://github.com/user-attachments/assets/8139c6c9-1d9d-4a70-8f1a-6e33edf7b38b)

With the error message:
> Traceback (most recent call last):
  File "tuxemon_dev/run_tuxemon.py", line 49, in <module>
    main.main(load_slot=args.slot)
  File "tuxemon_dev/tuxemon/main.py", line 91, in main
    client.main()
  File "tuxemon_dev/tuxemon/client.py", line 299, in main
    update(clock_tick)
  File "tuxemon_dev/tuxemon/client.py", line 344, in update
    self.event_engine.update(time_delta)
  File "tuxemon_dev/tuxemon/event/eventengine.py", line 366, in update
    self.update_running_events(dt)
  File "tuxemon_dev/tuxemon/event/eventengine.py", line 457, in update_running_events
    action.start()
  File "tuxemon_dev/tuxemon/event/actions/add_monster.py", line 55, in start
    raise ValueError(
ValueError: Anu doesn't exist (monster or variable)

However, you CAN fight an Anu, if you fight the guy in the bottom-left corner

- spyder_candy_hospital1.tmx
"This should help!" - but, what did he give me? An item?
![this_might_help](https://github.com/user-attachments/assets/d863590d-b86e-4e21-946a-6d3717f0d36a)

- Can't walk out of spyder_candy_cafe.tmx - the door doesn't work.

- In spyder_candy_center.tmx, you use the elevator by examining the switches by their side - nothing happens if you try to open the doors.
In the 1st and 2nd floors, spyder_candy_hospital1.tmx, spyder_candy_hospital2.tmx, it's the opposite - you change floors by examining the doors, not the switches.
(Either way is fine, but it should be consistent on all 3 floors)
These screenshots may or may not help:
![elevators2](https://github.com/user-attachments/assets/0404bcc6-619a-4854-b21b-63ad7a9f838a)

[...]

PR #2439, opened 2024-09-15
update change_bg (event action)

JaskRendix:
With this PR, we've expanded the capabilities of the `change_bg` feature, making it possible to display not just backgrounds, but also item sprites. Requested [here](https://github.com/Tuxemon/Tuxemon/issues/2434#issuecomment-2350847646) by @Sanglorian


kerizane (approved):

Sanglorian:
Thank you! I think this will make it a lot more obvious to players when they've gained items, which will get them thinking about how to use them.

PR #2444, opened 2024-09-16
check evolutions

JaskRendix:
PR updates the event condition for checking evolution. While I wasn't entirely happy with the previous implementation, the real driving force behind this change was the need to support some new evolution methods for our upcoming monster.

After putting the new code through its paces, I discovered an issue with chained parameters that was causing some unexpected behavior (see below). The old code would allow evolution to occur even if only one of the conditions was met, which wasn't ideal.

To fix this, I've updated the code to check if each parameter is not None and then use the 'all' function to ensure all conditions are met. This should prevent any issues with chained parameters going forward. As a bonus, I've also made the 'at_level' parameter Optional[int] to match the other parameters, and removed any JSON entries with a value of 0, which are now redundant.

---

It was getting a bit overwhelming to keep track of everything, so I figured it would be better to break it out into a separate class. That way, we can manage it more easily and keep things organized.

So, what's new since the last version? I've made some changes to the code. I've moved the `evolve_monster` method out of `npc.py` because, let's be honest, it never really belonged there. Now, all the evolution-related stuff will be handled by a new class called `Evolution`, which makes a lot more sense. I've also tweaked the parameters for `evolve_monsters` because now that monsters can trigger their own evolution, the `old_monster` parameter was redundant. This simplifies things and makes the code a bit more streamlined.

I took the opportunity to give the event action a bit of a refresh. I added two new fields - `variable` and `evolution` - which makes the event action way more useful. Now, if someone specifies the variable (which contains the monster instance ID and the name of the evolution), we can actually evolve a monster.  I also added some checks inside to make sure the evolution string is legit - i.e., it's actually part of the monster's evolution or history (which includes previous forms). If it's not, we stop everything and log a helpful error message.

---

On a related note, I've also updated the 'level' condition to 'can_evolve', which will allow us to use the 'item' parameter in a similar way.

```
    "evolutions": [
        {
            "at_level": 2,
            "monster_slug": "pairagrim",
            "gender": "male",
            "inside": true,
            "party": ["pharfan", "rockitten"], <--- min 1, max (MAX_PARTY-1 = 5)
            "variables": ["var1:var", "var2:var"], <--- min 1
            "moves": ["peck"], <--- min 1, max (MAX_MOVES=4)
            "taste_warm": "refined",
            "taste_cold": "sweet"
        }
    ],
```


tuxemon/event/conditions/check_evolution.py
```diff
@@ -41,14 +41,14 @@ def test(self, session: Session, condition: MapCondition) -> bool:
...
-        self.client = session.client
+        context = {"map_inside": session.client.map_inside, "use_item": False}
 
         evolving_monsters = []
         for monster in character.monsters:
             if monster.evolutions:
                 for evolution in monster.evolutions:
-                    if self.can_evolve(monster, evolution):
+                    if can_evolve(monster, evolution, context):
+                        print(monster.name)
```
kerizane: remove print statement?

JaskRendix: you're right, I forgot it, I'm going to remove it

mods/tuxemon/db/monster/chromeye.json
```diff
@@ -27,22 +27,22 @@
         {
             "at_level": 18,
             "monster_slug": "angrito",
-            "variable": "season:winter"
+            "variable": ["season:winter"]
```
kerizane: can you rename this to variables?

kerizane (changes requested):

mods/tuxemon/db/monster/chromeye.json
JaskRendix: ok

JaskRendix: done

JaskRendix:
@Sanglorian so if a monster is ready to evolve because it's daytime (only daytime as condition, no level), but the player doesn't want it to, what happens next? Does it never get another chance, or does it have to level up again before it can try to evolve again?

Sanglorian:
> @Sanglorian so if a monster is ready to evolve because it's daytime (only daytime as condition, no level), but the player doesn't want it to, what happens next? Does it never get another chance, or does it have to level up again before it can try to evolve again?

Have to level up again, I would say. But I think in most cases we would want to require both daytime and minimum level.

JaskRendix:
> Have to level up again, I would say. But I think in most cases we would want to require both daytime and minimum level.

Yeah, that's usually the case, but then there are those weird exceptions that come out of nowhere, like Vivipere evolutions and a few others.

kerizane (approved):

PR #2446, opened 2024-09-17
open_technique_menu (combat_menu.py)

JaskRendix:
PR updates combat_menu.py.

I made some tweaks to the **open_technique_menu** in **combat_menu.py**. Nothing too groundbreaking, just some minor improvements to make the code a bit more readable and streamlined. I got rid of some unnecessary if/elif/else statements, which should make things easier to follow. The biggest change is actually with the 'skip' technique - instead of adding it to the monster's moves, it now just creates a button in the menu. This simplifies things nicely, and we can even remove the code that was previously used to remove the technique.

@Sanglorian unrelated: now that we've made these changes, it's technically possible to make recharging moves unselectable. That would actually eliminate the need for the popup altogether. I'm curious, do you think the popup was originally added just because there wasn't a way to prevent the move from being selected in the first place? It's possible that it was a workaround at the time, but now that we have this new functionality, we might be able to simplify things even further by getting rid of the popup altogether. Let me know.

Screenshot_2024-09-17_14-23-22

### DIFF
--- a/tuxemon/states/combat/combat_menus.py
+++ b/tuxemon/states/combat/combat_menus.py
@@ -205,27 +205,33 @@ class MainCombatMenuState(PopUpMenu[MenuGameObj]):
         """Open menus to choose a Technique to use."""
 
         def choose_technique() -> None:
+            available_techniques = [
+                tech
+                for tech in self.monster.moves
+                if not combat.recharging(tech)
+            ]
+
             # open menu to choose technique
             menu = self.client.push_state(Menu())
             menu.shrink_to_items = True
 
-            # add techniques to the menu
-            filter_moves = []
+            if not available_techniques:
+                skip = Technique()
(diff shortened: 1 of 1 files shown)


kerizane (approved):

Sanglorian:
> @Sanglorian unrelated: now that we've made these changes, it's technically possible to make recharging moves unselectable. That would actually eliminate the need for the popup altogether. I'm curious, do you think the popup was originally added just because there wasn't a way to prevent the move from being selected in the first place? It's possible that it was a workaround at the time, but now that we have this new functionality, we might be able to simplify things even further by getting rid of the popup altogether. Let me know.
> 
> Screenshot_2024-09-17_14-23-22

Brilliant, yes I think that was just a work around and it's better to make them undetectable.

PR #2448, opened 2024-09-18
class encounter + random_encounter + encounter JSON files

JaskRendix:
While working on **check_evolutions** #2444, I stumbled upon an idea and decided to run with it. I dug into the 'variable' fields in db.py and found the EncounterModel, which handles the **random_encounter** event action. I thought it could be improved, so I refactored the code to move the **Encounter** class to a separate file, and broke down the methods into a more organized structure. This change makes the event action cleaner and easier to manage.

As a bonus, I added the ability to set multiple variables (totally optional, but super useful for modders). Now, you can create more complex conditions for wild monster encounters, like only triggering during nighttime in the summer, or on leap years.

```
    {
      "monster": "furnursus",
      "encounter_rate": 3.5,
      "variables": ["daytime:false", "season:summer"],
      "exp_req_mod": 1,
      "level_range": [30, 35]
    },
```

I also took the chance to tidy up most of the encounter JSONs, cleaning up the non Spyder from the variable field (added for the Spyder campaign), and clearly separated the spyder and xero ones to avoid any confusion.


JaskRendix:
unrelated @Sanglorian do you have any idea the max of **encounter_rate**? 
```
     {
       "encounter_rate": 0.5,   <------
       "exp_req_mod": 1,
       "level_range": [
         2,
         4
       ],
       "monster": "eyenemy"
     },
```
all the files JSON have a range from 0.5 to 10, so the max is 100?

Sanglorian:
Thank you, it's great to have some more customisation for encounters.

From memory, I don't think there is a maximum. I think the idea is that the odds of any monster are divided by all the eligible monsters. So if Rockat has a rate of 5, Tweesher 3 and Katapill 4, the odds of a Rockat encounter would be 5/(5+4+4)

kerizane (approved):

PR #2450, opened 2024-09-19
healing overhaul (technique effect)

JaskRendix:
I've been tracking this for weeks, and the time has finally come to implement some much-needed improvements to our healing effect. This PR refactors the formula to make it more efficient and moves it to its new home in `formula.py`.

One of the key changes is the introduction of an `additional_factors` field (it mimics damage calculation structure), which allows other effects to tap into the healing calculation and modify the amount healed based on their own logic. This should make it easier to add new effects and interactions in the future.

I've also replaced the old multiplication code with `math.prod`, which simplifies the formula and makes it easier to read. Overall, this PR should make our healing effect more flexible, maintainable, and efficient.

### DIFF
--- a/tuxemon/formula.py
+++ b/tuxemon/formula.py
@@ -64,10 +64,8 @@ def simple_damage_multiplier(
     Parameters:
         attack_types: The types of the technique.
         target_types: The types of the target.
-        additional_factors: Optional dictionary of additional factors
-        that affect the damage multiplier. The keys of the dictionary
-        should be the names of the factors, and the values should be
-        the corresponding multipliers.
+        additional_factors: A dictionary of additional factors to apply to
+        the damage multiplier (default None)
 
     Returns:
         The attack multiplier.
@@ -93,8 +91,8 @@ def simple_damage_multiplier(
                 )
     # Apply additional factors
     if additional_factors:
-        for _, multiplier_value in additional_factors.items():
-            multiplier *= multiplier_value
+        factor_multiplier = math.prod(additional_factors.values())
+        multiplier *= factor_multiplier
     return multiplier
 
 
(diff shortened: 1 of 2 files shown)


JaskRendix:
unrelated @Sanglorian I think we need to revisit the **healing_power** values. I've been testing some scenarios and I think we need to lower the cap from **10** to at a maximum of **3.0**.

For example, the **healing_power** of hibernate is currently **8**, which is way too high. I ran some tests and a level 3 monster (shape humanoid, lowest multiplier hp) with hibernate would regain a whopping 80 HP - that's double its base HP of 40! If we cap it at 3, the same monster would only regain 30 HP, which feels more balanced.

kerizane (approved):

Sanglorian:
> unrelated @Sanglorian I think we need to revisit the **healing_power** values. I've been testing some scenarios and I think we need to lower the cap from **10** to at a maximum of **3.0**.
> 
> For example, the **healing_power** of hibernate is currently **8**, which is way too high. I ran some tests and a level 3 monster (shape humanoid, lowest multiplier hp) with hibernate would regain a whopping 80 HP - that's double its base HP of 40! If we cap it at 3, the same monster would only regain 30 HP, which feels more balanced.

Thanks for raising, @JaskRendix. Hibernate does have quite a high cost, putting the user to sleep for a while. Giving up one or more turns of damage. But I can see for low HP monsters like humanoids, it is very substantial healing, especially low HP high defence monsters.

What if we changed healing power to be as a share of maximum HP, say 66% for Hibernate? So a monster with 100 max HP, down to 50 HP from damage, would go back to full health, but the same monster on 1 HP would go to 67 HP.

PR #2452, opened 2024-09-20
solar synthesis (plant) + shooting_star (3 tags) (techniques)

JaskRendix:
This PR  brings two new techniques to the table, the 189th and 190th, along with an effect that triggers healing based on a unique 'utility' curve. Think of it like this: if modders want a move to reach its peak healing power at a specific hour, they'll need to set both a start and end time for that effect. This way, anything outside that time frame won't count, and the technique will fail to heal.

The magic behind this is a new formula called **calculate_time_based_multiplier**, which calculates a useful multiplier to make it all work. For now, we're focusing on healing techniques, but this concept can easily be applied to damaging techniques as well. One thing to note is that these techniques don't have a monster assigned to them yet, but that's something we'll tackle in a future update, likely in PR #2438.

(moved below)

@Sanglorian I think it would be an interesting starting technique (lower levels), what do you think? Animal? Simple? Exotic? I'm going to create it in Tuxepedia only when merged.


Sanglorian:
Thanks, good idea! I would call it an exotic technique because it's only useful in certain circumstances. 

Maybe it should have a default level of healing (half its maximum potential, maybe?) so it's still useful in other circumstances?

I think it makes sense for any of those monsters. And if we give it the Plant tag, then even plant monsters who don't have it in their moveset could learn it.

JaskRendix:
As you can see in these graphs, it's strongest at midday and gets weaker as the day goes on. If the battle's indoors or in dark place, it won't work at all. It's a bit of a limitation, but it's still a really powerful tool, especially for specific monsters (wood, etc.). The red line is the max HP value of the monster.
Figure_1
this demonstrate how the formula can calculate the same during the night, but still I haven't figure out which kind of technique. I'm planning a damage version too (eg Solar Flare), because at the end it calculates a multiplier and we can use it in the same way. The technique will damage regularly, but during the daytime it'll damage more.
Figure_2


> Maybe it should have a default level of healing (half its maximum potential, maybe?) so it's still useful in other circumstances? I think it makes sense for any of those monsters. And if we give it the Plant tag, then even plant monsters who don't have it in their moveset could learn it.

I don't know, because all this work would result in a complete waste of time.

Sanglorian:
> > Maybe it should have a default level of healing (half its maximum potential, maybe?) so it's still useful in other circumstances? I think it makes sense for any of those monsters. And if we give it the Plant tag, then even plant monsters who don't have it in their moveset could learn it.
> 
> I don't know, because all this work would result in a complete waste of time.

I was still thinking it would increase during the sunny hours of the day (when outdoors), just that the minimum wouldn't be zero.

JaskRendix:
Yeah, I think I might have chosen the wrong graph style for this data. Bars would have been a way better choice, it would have made the comparison way clearer. No worries though, it's already handled.

Sanglorian:
> Yeah, I think I might have chosen the wrong graph style for this data. Bars would have been a way better choice, it would have made the comparison way clearer. No worries though, it's already handled.

Oh great! I like the idea.

And a night attack technique using the same calculation in reverse could be Sneak Attack or Falling Star or something else darkness themed? Falling Star would make sense for it to also be limited to outside only.

JaskRendix:
> And a night attack technique using the same calculation in reverse could be Sneak Attack or Falling Star or something else darkness themed? Falling Star would make sense for it to also be limited to outside only.

Falling Star sounds good, I guess exotic as well, for which kind of monsters?

Sanglorian:
> > And a night attack technique using the same calculation in reverse could be Sneak Attack or Falling Star or something else darkness themed? Falling Star would make sense for it to also be limited to outside only.
> 
> Falling Star sounds good, I guess exotic as well, for which kind of monsters?

I'm not very creative, turns out we already have Starfall which is basically the same name in reverse 😂.

Let's call it Shooting Star, and it could have the Celestial, Light and Darkness tags.
Potential mons: Yiinaang, Snarlon, Lunight, Solight

JaskRendix:
> Let's call it Shooting Star, and it could have the Celestial, Light and Darkness tags.

Done, while "plant" is solar_synthesis's tag. 

in another PR, it'll be assigned:

potential monster moveset: [Banling](https://wiki.tuxemon.org/Banling) (+ evolutions), [Baoby](https://wiki.tuxemon.org/Baoby) (+ evolutions), [Burrlock](https://wiki.tuxemon.org/Burrlock) (+ evolution), [Conifrost](https://wiki.tuxemon.org/Conifrost), [Conileaf](https://wiki.tuxemon.org/Conileaf), [Dandylion](https://wiki.tuxemon.org/Dandylion), [Dankush](https://wiki.tuxemon.org/Dankush), [Cohldrabi](https://wiki.tuxemon.org/Cohldrabi) (+ evolutions), [Cowpignon](https://wiki.tuxemon.org/Cowpignon) I'm not sure about the mushroom, [Fruitera](https://wiki.tuxemon.org/Fruitera) (+ evolutions), [Rosarin](https://wiki.tuxemon.org/Rosarin) (+ evolutions), [Pearamanca](https://wiki.tuxemon.org/Pearamanca), [Helipi](https://wiki.tuxemon.org/Helipi) (+ evolutions), [Trapsnap](https://wiki.tuxemon.org/Trapsnap) (+ evolution), [Shybulb](https://wiki.tuxemon.org/Shybulb) (+ evolution), [Sprightly](https://wiki.tuxemon.org/Sprightly) (+ evolution), [Weedsea](https://wiki.tuxemon.org/Weedsea) (+ evolution), [Turnipper](https://wiki.tuxemon.org/Turnipper) (+ evolution) I'm not sure, [Oxifeuille](https://wiki.tuxemon.org/Oxifeuille) and probably others too

kerizane (approved):

PR #2456, opened 2024-09-22
transitions

JaskRendix:
PR made some tweaks to the **FlashTransition** to make it more flexible and modder-friendly. I've moved some hardcoded parameters to injected ones, so they can be easily changed later on. I've also added descriptions for the fade and flash init transitions, which should make things clearer. On top of that, I've finally implemented the new transitions I've been working on over the past few weeks. @Sanglorian might recall me asking about the background color for flashbacks - well, these new transitions are the part result of that work, and they'll allow us to introduce some specific events in a more interesting way. Some can be changed even more, it's a starting point.

I've been thinking about how we can trigger the different transitions, and I realized that the screen_transition event action is currently hardcoded to fade. One possible solution could be to use the **change_state** event action, which is what we use for the PCState, but that would just add more clutter to it. A better approach might be to introduce a new event action, maybe something like 'transition', that would allow us to trigger different transitions in a more elegant way.

Some examples (other are too heavy to be uploaded and the negative color doesn't work when I try to record the screen):
this below can be right, left, up or down

https://github.com/user-attachments/assets/cb49a4da-894e-4722-8daa-edb6d1545395

https://github.com/user-attachments/assets/689cefca-2157-4b68-a2ef-ce94d1daf937

https://github.com/user-attachments/assets/9708c0ce-7077-447a-ae70-7bd426d15bae

https://github.com/user-attachments/assets/1ab8f47e-825c-42f9-b477-188071b3bc09

### DIFF
--- a/tuxemon/states/transition/__init__.py
+++ b/tuxemon/states/transition/__init__.py
@@ -4,5 +4,12 @@ import logging
 
 from .fade import FadeInTransition, FadeOutTransition
 from .flash import FlashTransition
+from .mosaic import MosaicTransition
+from .negative import NegativeTransition
+from .pixelation import PixelationTransition
+from .static import StaticTransition
+from .swirl import SwirlTransition
+from .wipe import WipeTransition
+from .zoom import ZoomInTransition, ZoomOutTransition
 
 logger = logging.getLogger(__name__)
--- a/tuxemon/states/transition/fade.py
+++ b/tuxemon/states/transition/fade.py
@@ -30,6 +30,16 @@ class FadeTransitionBase(State):
         caller: Optional[State] = None,
         color: ColorLike = prepare.BLACK_COLOR,
     ) -> None:
+        """
+        Parameters:
+            state_duration: The duration of the transition state in seconds.
+                If not provided, a default value will be used.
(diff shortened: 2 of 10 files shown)


JaskRendix:
@Sanglorian what do you think?

Sanglorian:
Thanks for these, I can different transitions being useful in different circumstances. 

I liked the pixel effect, I think that would be good when the player is using a magical teleporter, or if we had more out-there adventures like shrinking into a computer or something like that. 

I think screen wipe is a good one but it moved a bit slowly for my liking. 

And I think flickering is a nice versatile transition but the colours were too strong and the flickering too quick for me - maybe a little slower and just black and white?

JaskRendix:
Sure, I'm going to allow people define the speed, colors, etc.

kerizane:
I'm worried that the flickering one could be bad for people with epilepsy. Please remove it. Others looks good.

JaskRendix:
> I'm worried that the flickering one could be bad for people with epilepsy. Please remove it. Others looks good.

I'm going to remove it

JaskRendix:
removed

kerizane (approved):

PR #2457, opened 2024-09-22
prop_healing (tech effect) + healing_power cap

JaskRendix:
PR made some changes to **healing_power**. @Sanglorian it's set to **3.0** the **healing_power** cap. So anything above 3 is rounded down to 3. Also, I changed **hibernate** from 8 to 0. This is because we needed a new effect for healing, similar to the one we have for damage. I created a new effect called **prop_healing** which heals proportionally based on the user or target's max HP. This was, hibernate will heal 66% of the max HP as you requested [here](https://github.com/Tuxemon/Tuxemon/pull/2450#issuecomment-2365690221). Let me know if this is ok.

### DIFF
--- a/mods/tuxemon/db/technique/battery_discharge.json
+++ b/mods/tuxemon/db/technique/battery_discharge.json
@@ -7,7 +7,7 @@
     "healing user"
   ],
   "flip_axes": "",
-  "healing_power": 4,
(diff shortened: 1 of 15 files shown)


Sanglorian:
Brilliant, thank you!

kerizane (approved):

PR #2458, opened 2024-09-22
change name (Axylightl to Axolightl)

JaskRendix:
waiting for #2448 

PR corrects a mistake. @Sanglorian caught a mistake he made with a monster's name, so we had to update it. This was a good chance to clean up some old stuff I added a while back. My old save loaded fine, and I even created a way to automatically change name using a dictionary. We can add more to this dictionary if we need to change a name in the future, maybe because someone gets offended or there is a copyright problem.


Sanglorian:
Thanks! Once I learned how to properly spell axolotl, the old name started bugging me.

kerizane (approved):

PR #2459, opened 2024-09-23
transfer (tech effect)

JaskRendix:
PR implements a new tech effect called **transfer**. A new effect inspired by @Sanglorian's idea for the 'suck_poison' technique. Here's how it works: if the user has a poisoned condition, they'll lose it and the target will gain it. I actually had a similar idea myself, so it wasn't too hard to implement - I just needed to tweak a few details. The best part is that this effect can transfer any condition: eg `transfer poison` or `transfer burn` or `transfer whatever`. You just need to specify it in the technique's JSON. One important thing to note is that the condition will only be transferred if the technique actually hits - if it misses, nothing will happen. I'm planning to implement this new effect in `suck_poison` in PR #2437 .


Sanglorian:
Thank you! I'd wanted to realise the Suck Poison effect for a while but wasn't sure if it was possible. Now the Nudiflots get a unique reserved technique!

kerizane (approved):

PR #2465, opened 2024-09-23
Give stronger Nudiflot if no tuxemon in Candy town

ultidonki:
In Candy Town, all your current Tuxemon get taken away, and if you don't have any in your Kennel, you get a level 10 Nudiflot.

However, all the tuxemon in the nearby dungeons and villages are level 25-ish, and I think level 20-ish in the previous route (B?). So I've made the Nudiflot you receive stronger (level 25). 

This will make it easier to catch other tuxemon and build up a team without getting instantly defeated by the level 25 tuxemon you'll probably meet. I considered making it higher but the important thing here is catching more tuxemon, not defeating them, so having a similar level to them is probably better. :smile:


Sanglorian:
Good idea!

kerizane (approved):

PR #2466, opened 2024-09-24
fix 2464 & fix 2461

### LINKED ISSUE #2461: Can take Tuxemon out of Quarantine before cure is found
**Describe the bug**
After entering Candy Town, my tuxemon are taken away to be quarantined, and I need to rescue them.
However, currently I can just take my tuxemon out of the 'Quarantine' box, there's nothing stopping me.

**To Reproduce**
See the following save file: I've just had my tuxemon taken off me, to be quarantined, but I haven't got the cure yet.
[slot2_after_quarantine.zip](https://github.com/user-attachments/files/17103154/slot2_after_quarantine.zip)

1. In the PC, select 'Pick up monster'
2. You can see 2 kennels, 'Shelter' (the tuxemon I've caught), and 'Quarantine' (a new kennel created, where my 6 infected party tuxemon have been placed)
3. You can take tuxemon from 'Shelter' or 'Quarantine'.
4. You should only be able to access 'Shelter'. You shouldn't be able to access 'Quarantine' until after the cure is found.

### LINKED ISSUE #2464: If you put all your surviving monsters in a kennel, the game thinks you've been defeated
**Describe the bug**
If you have 2 defeated tuxemon, and 1 tuxemon still alive.
If you put the tuxemon that's still alive in the kennel, the game thinks you've been defeated, and calls teleport_faint to your nearest health center. 

**To Reproduce**
Here's a save file: 2 tuxemon are defeated, and Botbot still has a few HP.
[slot1_candy_health_center.zip](https://github.com/user-attachments/files/17105020/slot1_candy_health_center.zip)
Put 'Botbot' in the kennel, and you will be teleported to the previous health center.

Well, it's not the biggest problem, just a bit strange. When your tuxemon are in quarantine, you have to take some tuxemon out of the kennel. When I took out one of my tuxemon (which was defeated when I put it in the kennel) I was surprised to be teleported to the health center of the previous town. You can come back using the Shipmaster (if you've saved him) but otherwise you'd need to walk through the route b/6 again.

JaskRendix:
PR fixes #2464 and fixes #2461 .

Related to #2461 

First, I decided to rename `set_kennel` to `create_kennel`. I thought this name was more fitting, and it also gave me the opportunity to simplify the function by removing the option to set the kennel as hidden. We already have a separate event action for that, so it made sense to break it out.

Next, I created a new condition called `kennel` that checks three different states: visible, hidden, or exist. I think this will make the code easier to read and understand, and it'll be more consistent with our other conditions.

To test out these changes, I set up a few event actions in the main YAML file. I created a kennel quarantine at the beginning, and then I made it invisible if the main variable of the cure wasn't set. Once the variable is set (the cure is found, etc.), the kennel will be visible again (@Sanglorian so the player can take out everything else).

While I was in there, I also noticed that the `has_box` condition was checking the exact number of monsters inside the kennel. I thought it would be better to rename it to `has_kennel`, just to keep our naming conventions consistent and make it clearer what the condition is checking.

Finally, I found a more efficient way to check for fainted monsters in the `char_defeated` event action. I think this will help streamline the code and make it run a bit smoother. 

Related to #2464 

this new change prevents the selection of the last monster still standing, see video below (note: this video was recorded before the fix for #2461):
https://github.com/user-attachments/assets/74dc402b-c10d-41c6-b083-43ae344d1752


JaskRendix:
@Sanglorian I'm thinking we could make the infected monsters unselectable instead of having the pop-up hardcoded.

msgid "menu_storage_infected_monster"
msgstr "Virus detected. For the safety of all tuxemon in our system, you cannot store an infected tuxemon."

Sanglorian:
> @Sanglorian I'm thinking we could make the infected monsters unselectable instead of having the pop-up hardcoded.
> 

My only concern with this is whether players will understand why the monsters are unselectable - they might be confused.

kerizane (approved):

Issue #1909, opened 2023-06-05
Teleportation bug at beginning of game


kerizane:
After the opening cutscene I was left in the players bedroom with no player character. I was able to open and close the menu, but wasn't able to move. I may have been mashing some keys at the time. 
My character was female (probably not relevant).

The broken save file is attached.

[slot3.zip](https://github.com/Tuxemon/Tuxemon/files/11657250/slot3.zip)

JaskRendix:
> I may have been mashing some keys at the time.

It doesn't help. Details missing in the bug reporting:
- the scenario is Spyder
- the person reporting has chosen a female sprite (gender choice done)
- the person reporting has chosen a budeye (monster choice done) 

So probably it (the person) didn't interrupt anything, but pressed somehow, somewhere, escape.

kerizane:
There used to be a bug where walking whilst teleporting would cause a similar issue, where your sprite would be out of bounds. The coordinates in the save file look correct, though.

JaskRendix:
the coordinates in the savegame are wrong (bedroom is smaller):
```
    "tile_pos": [
        26,
        11
    ],
```
if you replace it with 3,4, everything works fine. I have no idea how the player could have ended up in 26,11.

26,11 doesn't pop up anywhere in the code

kerizane:
I think 26, 11 is the value I manually entered after the glitch happened. I thought I took that from a working file, but must have down something wrong.

Then this is probably the same error as a while back.

JaskRendix:
I don't know, because it would have been popped up before. This is where the player is headed after the Mart scene:
`<property name="act122" value="transition_teleport spyder_bedroom.tmx,3,4,0.3"/>` 
I tracked down the teleport, but I didn't notice anything weird.
I tried to click other keys while teleporting, but nothing happened.

JaskRendix:
I don't know, each time I read the title of this issue, it doesn't sound right. The game doesn't freeze if you can open the menu. Although there are no clues it's the "old" not reproducible glitch, then it can be corrected with "teleport glitch".

Sanglorian:
Given there's some confusion with the save file, shall we close this one and reopen if the problem re-occurs?

PR #2473, opened 2024-09-29
class Teleporter

JaskRendix:
So I dug into the code (after #2472) and noticed that the `change_map` method was popping up in two places - the teleport event action and `handle_delayed_teleport`. I took a closer look at both and found that `handle_delayed_teleport` was missing a couple of important things, like the locking system and `cancel_path`. That reminded me of issue #1909, which @kerizane opened, and it seems like a pretty plausible explanation for the problem. 

To fix it, I moved the `teleport_character` code outside of the World class and into its own method. Now, it checks if the destination coordinates are within the map boundaries. If they're not, it raises a `ValueError` right away.

To get this working, I also moved the delayed teleport logic, which had the added bonus of completely eliminating all 'delayed_' attributes from the World class. This means we've successfully consolidated all teleportation-related code into a single, dedicated class - the Teleporter class.

It's a much cleaner and more organized approach, and it's great to see everything related to teleporting in one place.

### DIFF
--- a/tuxemon/event/actions/char_face.py
+++ b/tuxemon/event/actions/char_face.py
@@ -58,7 +58,7 @@ class CharFaceAction(EventAction):
         if character.isplayer:
             world_state = self.session.client.get_state_by_name(WorldState)
             if world_state.in_transition:
-                world_state.delayed_facing = direction
+                world_state.teleporter.delayed_facing = direction
             else:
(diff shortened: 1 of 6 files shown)


kerizane (approved):

kerizane:
Good work!

PR #2479, opened 2024-09-30
class Bar (HP + EXP) + fix exp bar (level up)

### LINKED ISSUE #577: Level number and exp bar don't update to reflect changes after leveling up
If a monster levels up during battle, the player's HUD is not updated to reflect this. That includes both the level number as well as the newly added exp bar, they each stay the same.

I already investigated and have no idea what might be causing this: The exp bar at least does increase after the monster faints when you don't level up, but if you do level up then the bar isn't modified either. The strange thing is that if I go to [animate_exp in combat_animations.py](https://github.com/Tuxemon/Tuxemon/blob/development/tuxemon/core/states/combat/combat_animations.py#L257) and add `print(value)` at the end of this function, a lower value is printed when the enemy monster faints as you level up, meaning it's seeing the new level! eg: When the battle starts it may print 0.9, and once you defeat the enemy monster and level up it correctly shows 0.1.

So why doesn't even the bar move any more considering its animate function is being called right there? The bar does move if you gain exp without leveling up, so it's likely not HUD updates being blocked entirely once the enemy monster faints. For the level text a separate update may need to be added, one may have not been put in place after leveling up was implemented.

JaskRendix:
This PR fixes #577 an old issue with the exp bar when leveling up.

https://github.com/user-attachments/assets/e9fcd5f8-7cfe-4d87-ab61-0f35058e1066

### DIFF
--- a/tuxemon/menu/interface.py
+++ b/tuxemon/menu/interface.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2014-2024 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-from typing import ClassVar, Generic, Optional, TypeVar
+from typing import Generic, Optional, TypeVar
 
 import pygame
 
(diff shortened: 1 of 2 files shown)


JaskRendix:
unrelated: I'd love for you to take a look, @Sanglorian  - can you check out the video and let me know if everything looks good now? Also, I was thinking, should we change the color of the bar to make it more visually appealing? If yes, feel free to drop a RGB color. Ah, the video is old, now the update of the number (level) happens after the exp bar.

kerizane (approved):

Sanglorian:
> unrelated: I'd love for you to take a look, @Sanglorian - can you check out the video and let me know if everything looks good now? Also, I was thinking, should we change the color of the bar to make it more visually appealing? If yes, feel free to drop a RGB color. Ah, the video is old, now the update of the number (level) happens after the exp bar.

Thanks! The bar looks good to me. I'm happy to change the colour but don't have any strong preference.

PR #2494, opened 2024-10-06
fix 2491

### LINKED ISSUE #2489: I cannot catch Drokoro or Volcori
I beat the game, but I wasn't able to catch 2 bonus monsters, Drokoro and Volcoli.

My team was amazing but the best monsters were two level 48 tuxemon, Drokoro was level 50.
After getting Drokoro's HP down to this much, I spent 12 minutes (I counted!) throwing tuxeballs at Drokoro, but couldn't catch it. Sometimes it came close, with 3 or 4 wiggles before it was caught, but I never actually managed to catch it.

drokoro_cannot_catch

Is that right or wrong? I don't know what the catch rate is supposed to be, or how to make it more likely to be caught, but that seems a bit harsh.

It happened again with the level 50 Volcoli - I had a level 60 tuxemon this time, spent 5 minutes throwing tuxeballs at it after reducing its health, and gave up.

### LINKED ISSUE #2491: Shops only sell Potions throughout the game
When I go to the Mart (the shop) in Candy Town, the final town in the (spyder) campaign, there's still only potions, tuxeballs, and revives in the shop. (They don't add more useful items as the game progresses)

A potion only heals 20HP, but at this point my tuxemon all have over 200-400 HP.

If I'm in one of the final dungeons, my options are:
1. Walk back to the nearest healer.
2. Use 10-20 potions per Tuxemon to bring them back to full health.
Neither are fun.

A super potion restores 50HP, which would be better than a normal potion, but it would still take 8 or 9 to get my strongest Tuxemon back to full health. 

Please either add to the store:
- Cureall (this restores all health and status effects)
- A potion that restores 100 or 200 HP or something
- Super Potions would be useful around about Route 3/Route 4 too, I think.

JaskRendix:
PR fixes #2491, additionally, I've centralized all the spyder shops into a single JSON file, making it easier to manage.

@Sanglorian + @ultidonki 

info about healing [items](https://wiki.tuxemon.org/Category:Item#Medicine):  Super 50HP, Mega 200HP, Imperial 500HP

- Paper Scoop: no modifications (always closed)
- Cotton Scoop: no modifications
- Leather Scoop: Super Potion (instead of Potion) + Cure All
- Flower Scoop: Super + Mega Potion (instead of Potion) + Cure All
- Timber Scoop: Super + Mega Potion (instead of Potion) + Cure All
- Candy Scoop: Mega + Imperial Potion (instead of Potion) + Cure All

Let's talk about the prices. We've got Potion priced at 20, which recovers 20HP. Following that logic, we've got Super at 50, Mega at 200, and Imperial at 500.

Tuxeball #2501


JaskRendix:
Hey @Sanglorian and @ultidonki, this pull request is now focused solely on the potions. If that works for you, I can go ahead and mark it as ready to merge. Just let me know if there are any other items that need to be added or adjusted!

Sanglorian:
Thank you, that sounds good to me - as do the prices. 

Made me realise that the price of items is a pretty obvious thing I never thought to add to the wiki! I've created a field for it now.

kerizane (approved):

PR #2500, opened 2024-10-08
self.plague (monster.py)

JaskRendix:
PR follows #2497

I've refactored the plague model to make it more flexible and less hardcoded. Initially, I had set the `self.plague` field to a specific value, `PlagueType.healthy`, but I've now changed it to a dictionary, `self.plague: dict[str, PlagueType]`. This change allows me to easily determine if a monster is healthy by checking if the dictionary is empty or if a specific key is missing.

As a result of this change, I was able to remove the `healthy` value from the `PlagueType` enum class in `db.py`. I also created a custom method in `save_upgrade` to upgrade old savegames to the new system.

Additionally, I removed the `plague` attribute from the `NPC` class, as the infection is now handled differently. This led to updates in `char_plague`, `quarantine`, and the `party_infected` condition to accommodate the new format.

One interesting aspect of this refactor is the renaming of the `spyderbite` effect to a more generic `plague` effect.

```
  "effects": [
    "spyderbite" --> "plague spyderbite,0.125"
  ],
```

This new effect now carries two attributes: the name of the plague (e.g., `spyderbite`) and the spreadness value (e.g., 0.125 for `spyderbite` (1/8 = 0.125)). This change allowed me to remove the hardcoded party logic in `combat.py`, both in states and non-states, as well as the `plague` method, which is no longer needed.

I've added a new event action called `set_monster_plague` that allows modder to target a specific monster with the plague + condition, rather than affecting the entire party. I works similarly to the other actions, where you'll need to specify the variable with the instance ID of the monster you want to target (by using `get_player_monster` before). This gives more control and flexibility, especially when compared to the `char_plague` action which affects all party members.

Now let's talk about how contagion works. If a monster only has one plague, like `spyderbite`, nothing changes. But if a monster is infected with multiple plagues, say `spyderbite` and `bubonic plague`, things get interesting. Once we get all the plagues from the monster, it randomly selects one of them and applies it, using the corresponding technique (spyderbite -> technique spyderbite, bubonic -> technique bubonic). The best parts that modders have control over the rate of spreading, which can be set as percentage (eg. 0.125 spyderbite -> 12.5%). This adds a whole new layer of depth to the game, and allows for some really cool and complex scenario.

```
    infected_slugs = [
        slug
        for slug, plague in monster.plague.items()
        if plague == PlagueType.infected
    ]

---> it gets all the plague slugs with condition infected

slug = random.choice(infected_slugs)

---> it randomly selects one of them and applies it
```

@Sanglorian just a heads up - I wanted to keep you in the loop, but don't worry, nothing's changing in terms of the basic mechanics, so no need to take action.


Sanglorian:
> @Sanglorian just a heads up - I wanted to keep you in the loop, but don't worry, nothing's changing in terms of the basic mechanics, so no need to take action.

Thanks, appreciate it!

kerizane (approved):

PR #2502, opened 2024-10-10
refactor technique menu (the one accessible from monster_menu)

JaskRendix:
https://github.com/user-attachments/assets/328d2867-6c81-4e2d-92a0-c4d43501065c

PR refactors the technique menu and follows #2454 . I've made some changes to the technique menu in the game. Now, instead of having to switch monsters to see their techniques, you can just click left or right to choose a different monster and see their techniques right away. You can also use the up and down arrows to change techniques. The rest of the menu is the same, you can still go back and do everything else as before.

Note: the old technique menu will still appear (eg when using the get_monster_technique event action). We'll need to make some more changes to the pygame menu before we can fully refactor that part as well.

@Sanglorian

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -736,6 +736,18 @@ msgstr "{name} cannot be used on any of your Tuxemon right now."
 msgid "technique_description"
 msgstr "ID.{id} ({types}) - Accuracy {acc}%, Potency {pot}%, Power {pow}/3, Recharge {rec} turns"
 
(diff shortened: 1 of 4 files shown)


Sanglorian:
Yes, a very nice quality of life improvement - thank you. 

I just noticed with those images, some of the monster sprites look blurry? Is that something to do with the file type or the dimensions?

JaskRendix:
> I just noticed with those images, some of the monster sprites look blurry? Is that something to do with the file type or the dimensions?

Hmm, Aardon does look a bit blurry, doesn't it? I'm going to take a closer look at the file, see if there's something going on with the image that's causing it to appear that way.

kerizane (approved):

PR #2503, opened 2024-10-10
museum leather town (Spyder)

JaskRendix:
PR updates the museum (leather town - Spyder campaign) and fixes a bug with the fee. @Sanglorian I added 3 NPCs and a plaque. Do the new dialogues sound good? I'm trying to reinforce the plot around Nimrod and his activities at the museum.

https://github.com/user-attachments/assets/6bc18ece-ae83-4519-a597-95d7842302c1

I've added some trash cans outside the museum and inside them, I've placed some of the notes and journals that were previously on display (in the museum). It's like they've been thrown away, discarded. Players can read and learn more about the world's history. It's like they're uncovering secrets that Nimrod wants to keep hidden. Let me know if this works for you, or if there is anything that doesn't feel right (maybe only the first name without last name?).

https://github.com/user-attachments/assets/c2a4e6ec-7329-4e08-b5bf-17bce2da85f4

### DIFF
--- a/mods/tuxemon/db/npc/spyder_leather_town_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_leather_town_npcs.json
@@ -1,242 +1,218 @@
 [
   {
     "slug": "spyder_leathergym_chadine",
-    "template": 
-      {
-        "sprite_name": "barmaid_blonde",
-        "combat_front": "barmaid",
-        "slug": "barmaid"
-      }
-  
+    "template": {
+      "sprite_name": "barmaid_blonde",
+      "combat_front": "barmaid",
+      "slug": "barmaid"
+    }
   },
   {
     "slug": "spyder_leathergym_chadfort",
-    "template": 
-      {
-        "sprite_name": "soldier",
-        "combat_front": "soldier",
-        "slug": "soldier"
-      }
-  
+    "template": {
+      "sprite_name": "soldier",
+      "combat_front": "soldier",
+      "slug": "soldier"
+    }
   },
   {
     "slug": "spyder_leathergym_chad",
-    "template": 
-      {
-        "sprite_name": "magician_grey",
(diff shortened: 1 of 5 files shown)


Sanglorian:
These are all great, and very welcome additions. Thanks for honouring the Spyder campaign plot by developing it further. 

And it's a good idea to put things in the trashcans because we've already trained the player that good things can be there, since that's where they got their first tuxemon. 

I agree there's something a bit odd about firstname lastname - I think just firstname would be worth a try (Dr Maria, etc).

My only other thought is - should there be an introductory sentence? Like "*rummage* *rummage* What's this here?" or "Huh, there's a notebook under some plastic wrappers" so it doesn't go straight into the writings?

JaskRendix:
@Sanglorian 

> I agree there's something a bit odd about firstname lastname - I think just firstname would be worth a try (Dr Maria, etc).

I agree.

> My only other thought is - should there be an introductory sentence? Like "rummage rummage What's this here?" or "Huh, there's a notebook under some plastic wrappers" so it doesn't go straight into the writings?

Sure, something like this

*rummages through the trash can, pulling out a banana peel with a few days' worth of mold on it* Ugh, wait, what's this? A leather notebook... Letter from Dr Henry, Former Nimrod Researcher

*digs deeper into the trash can, pushing aside banana peels and broken pens* Gross. Okay, what's this? Another notebook... Field Notes of Dr Sophia, Conservation Biologist

*pulls out a moldy pizza box and a stack of old newspapers* This trash can is a treasure trove of grossness. Oh, a journal... Journal of Dr John, Historian

*reaches to the bottom of the trash can, pulling out a soggy notebook* Ugh, I hope this one isn't ruined. *wipes off the cover* Ah, it's still legible... Notebook of Dr Maria, Tuxemon Biologist

kerizane (approved):

PR #2482, opened 2024-10-03
Emails and notes hospital (Spyder)

JaskRendix:
PR introduces a series of emails in the hospital, shedding light on Dr. Lee. Lee had been secretly working to find a cure for the deadly Spyderbite virus. The emails reveal the shocking events that unfolded as Dr. Lee tried to save lives and the Cathedral tried to silence them.

---

Hospital: no flashbacks, simple emails, because a flashback triggered by a single line of dialogue might not be as impactful or clear. I'm having trouble figuring it out, and it's not like I can just hop on a computer in the hospital - the hospital doesn't exactly have a lot of them lying around. The only one I can think of is the one on the ground level, and that's the one I can access to check its boxes. I could add portables around.

**Email 1: Researcher 1 to Dr. Lee**
msgid "spyder_hospital_pc_message1" msgstr "E-mail from Researcher 1: 'Dr. Lee, I've been analyzing the virus samples, and I think we should try a different strain. It might trigger a new reaction in the antibodies.'"

**Note: Researcher 1 and Researcher 2 have been working together on the project, but they have different approaches and personalities. Researcher 1 is more cautious and methodical, while Researcher 2 is more impulsive and willing to take risks.**

**Email 2: Researcher 2 to Dr. Lee**
msgid "spyder_hospital_pc_message2" msgstr "Note from Researcher 2: 'Dr. Lee, I've found a potential combination of enzymes and antibodies that could slow down the mutation rate. I'm running simulations now.'"

**Email 3: Researcher 2 to Dr. Lee**
msgid "spyder_hospital_pc_message3" msgstr "Simulation results from Researcher 2: 'The enzyme-antibody combination shows promising results. We might have a cure on our hands.'"

**Email 4: Researcher 1 to Researcher 2**
msgid "spyder_hospital_pc_message_rivalry1" msgstr "E-mail from Researcher 1: 'Researcher 2, I'm not sure I agree with your approach. We need to be more careful and methodical in our research. Your simulations are promising, but we can't rush into conclusions.'"

**Email 5: Researcher 2 to Researcher 1**
msgid "spyder_hospital_pc_message_rivalry2" msgstr "E-mail from Researcher 2: 'Researcher 1, you're always so cautious and slow. We need to take risks and push the boundaries of what's possible. Our research is too important to be held back by your fear of failure.'"

**Email 6: Enforcer to Dr. Lee**
[...]

PR #2506, opened 2024-10-13
flashback Captain (Spyder)

JaskRendix:
split out from #2482 

So the trigger for the flashback of the drunk captain is actually inside the Flower City shop - the shopkeeper notices the player approaching and asks if they want to know what happened the previous night. @Sanglorian does that setup feel right, or is there something that needs adjusting?

Trigger + Flashback: https://github.com/user-attachments/assets/2f0430fb-f9bd-4aed-b20a-1927bfe76200


JaskRendix:
Same here: @Sanglorian I've removed the location, 'days ago', and title from the start.

Sanglorian:
Thanks! Looks good to me.

kerizane (approved):

PR #2508, opened 2024-10-13
flashback Greenwash (Spyder)

JaskRendix:
split out from #2482 

After @Sanglorian  pointed out the issue with the florists (same for the other Greenwash related flashbacks), I went back and reworked the scene. I wanted to get into the florists' heads and make their conversation feel more personal and relatable. I ditched the scientific jargon and tried to make them sound like, well, florists.

The scene's now set in Timber Cafe, which gives the player a chance to get their feet wet, so to speak. The trigger for the flashback is a woman who overheard the florists' discussion - it's a nice way to get the player invested. 

note: The video now asks if they want to hear the discussion, with 'no' at the top and 'yes' below, so it's totally up to them.

trigger + flashback: https://github.com/user-attachments/assets/f2a0fd78-f6d2-4a1a-a468-af0625b15944

---

I've reworked the incident flashback at Greenwash to make it more concise and impactful. Instead of having multiple flashbacks, we're now focusing on the key moments that really drive the story forward. The memories of the security guard at the Candy Cafe will still trigger the flashback, but now it's more streamlined and efficient. This way, we can really get into what happened and give the player a better sense of Looten's character. And, of course, this all sets the stage for the player's eventual tumble into Looten's office through that mysterious hole in the ground.

trigger + flashback: https://github.com/user-attachments/assets/3bcae3e6-e8e3-459a-9c85-b186a0b810a8

---

I've made some changes to the flashbacks involving Looten. I've cut one of the three original flashbacks, but the two that remain are still really important to the story. One of them is the scene with the Fossilator, and the other is the conversation between Looten and Dempsey. I've actually moved the conversation to a different location within Greenwash, on the third level.

I've been working to make the story more cohesive, and that's involved combining some similar themes and ideas. For example, I've merged the introduction of the fusion technology with the De-Fossilator into a single flashback. This actually meant I could cut the first flashback altogether. Now, when you access the PC at Greenwash, you'll trigger one of these flashbacks. I've kept it simple, with one flashback per PC. This should help to make the story feel more streamlined and easier to follow.

trigger: https://github.com/user-attachments/assets/96367fcd-68d8-4799-9854-40a7d1f67c09

Conversation between Looten and Dempsey:
original: https://github.com/user-attachments/assets/12d1bf60-6b9a-4c22-9f8a-10434071dc09

Fossilator:
original: https://github.com/user-attachments/assets/dbd408c4-5e2b-4517-a72a-417f9b716767


Sanglorian:
Thank you! I appreciate you taking the feedback on board and re-working these. I think they're tighter, will fit in seamlessly to the gameplay, and add a nice storytelling element.

JaskRendix:
@Sanglorian thank you. I enjoy filling in gaps like these as it allows me to gain new skills and insights. If you'd like to address anything else (Billie, Greenwash, etc), feel free to check #2482, which is still pending, and tell me.

kerizane (approved):

PR #2512, opened 2024-10-16
conditions (target field + icon)

JaskRendix:
I started this PR by cleaning up some unnecessary code - specifically, removing the `target` field from the condition JSON. It was a leftover from when conditions and techniques were still merged, and it was just taking up space.

While I was at it, I also updated some of the icons that were way too heavy (we're talking 7kb vs 200-400 bytes for the others). It was a good opportunity to tidy things up.


Sanglorian:
Thanks @JaskRendix , I've proposed a bit of an alternative system over at #2523 - keen to hear your thoughts.

JaskRendix:
I ended up trashing all the proposals, so now the PR only addresses the icons and removing the target field from the JSON. I'm curious to know if the new system will get rid of some of the old conditions with stats modifiers, but something tells me they'll stick around.

kerizane (approved):

PR #2516, opened 2024-10-17
refactor effects (switch, give, reverse and sacrifice)

JaskRendix:
This PR gives a bunch of effects a much-needed overhaul. Let's be real, the original code was written back when I was still figuring things out, and it shows. But, as part of my ongoing quest for coding, I've revisited these effects and given them a serious refactor. They're now more readable, maintainable, and just plain better. It's a reminder that we're all constantly learning and improving - and that's what makes this so awesome.


tuxemon/technique/effects/reverse.py
```diff
@@ -32,19 +32,18 @@ class ReverseEffect(TechEffect):
...
     ) -> ReverseEffectResult:
-        done: bool = False
-        if self.objective == "user":
-            user.reset_types()
-            done = True
-        elif self.objective == "target":
-            target.reset_types()
-            done = True
-        elif self.objective == "both":
+        if self.objective not in ["user", "target", "both"]:
```
kerizane: Minor thing, but it's better to use tuples rather than lists for checks like this.

tuxemon/technique/effects/switch.py
```diff
@@ -39,61 +39,73 @@ class SwitchEffect(TechEffect):
...
-                    done = True
-                elif self.objective == "both":
-                    user.types = [ele]
-                    target.types = [ele]
-                    done = True
+        monster_map = {"user": user, "target": target}
+
+        if self.element == "random":
+            element_map = {
+                "user": Element(random.choice(elements)),
```
kerizane: I think a dict is overkill. Could just be user_element, target_element.

Won't make you change this.

kerizane (changes requested):

tuxemon/technique/effects/switch.py
JaskRendix: Yeah, it was overkill, I changed.

kerizane (approved):

kerizane (approved):

PR #2519, opened 2024-10-18
update technique effects ('fly_off' and 'land')

JaskRendix:
PR updates and changes name to 'fly_off' and 'land' effects (technique)t.

The main goal was to make the technique effects more generic and less hardcoded. To achieve this, I've renamed the 'fly_off' effect to '**disappear**' and the 'land' effect to '**appear**'. This change makes the logic more flexible and opens up possibilities for future moves that use similar mechanisms, such as '**burrow**' or '**dive**'.

While testing the changes, I noticed a few opportunities to simplify the code. For instance, I realized that the extra code in 'fly_off' was unnecessary, as we can use the existing '**use_success**' field instead. This led me to take a closer look at the combat.py file, where I was able to remove some redundant code related to displaying messages. This cleanup helps to make the code more efficient and easier to maintain.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -835,9 +835,6 @@ msgstr "Phone keeps ringing but no answer!"
 
 ## NOTIFICATION TRANSLATIONS ##
 # Combat notifications
-msgid "combat_fly"
-msgstr "{name} flies up high!"
-
 msgid "combat_scope"
 msgstr "AR:{AR} DE:{DE} ME:{ME} RD:{RD} SD:{SD}"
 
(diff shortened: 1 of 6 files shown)


mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
```diff
@@ -835,9 +835,6 @@ msgstr "Phone keeps ringing but no answer!"
 
 ## NOTIFICATION TRANSLATIONS ##
 # Combat notifications
-msgid "combat_fly"
```
kerizane, line 838: Is this translated in other locales? If so, please rename those too.

kerizane (dismissed):

mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
JaskRendix: Nope, and I checked.

kerizane (approved):

PR #2520, opened 2024-10-18
ItemEffectResult (from typedict to dataclass)

JaskRendix:
follows #2517

PR converts `ItemEffectResult` from a `TypedDict` to a `dataclass`, which I believe is a better approach. By using a `dataclass`, we can take advantage of its built-in features, such as automatic generation of `__init__` and `__repr__` methods, and improved performance. And, we can access the fields directly as attributes, rather than having to use dictionary keys.

```
@dataclass
class ItemEffectResult:
    name: str
    success: bool
    num_shakes: int
    extra: list[str]
```


kerizane (approved):

kerizane:
Agreed, I think dataclass is generally better than typeddict. Good work.

PR #2521, opened 2024-10-18
CondEffectResult (from typedict to dataclass)

JaskRendix:
follows #2520

PR converts `CondEffectResult` from a `TypedDict` to a `dataclass`.

```
@dataclass
class CondEffectResult:
    name: str
    success: bool
    condition: list[Condition]
    technique: list[Technique]
    extra: list[str]
```


tuxemon/condition/condeffect.py
```diff
@@ -14,11 +14,13 @@
...
 
-class CondEffectResult(TypedDict):
+@dataclass
+class CondEffectResult:
+    name: str
     success: bool
-    condition: Union[Condition, None]
-    technique: Union[Technique, None]
-    extra: Union[str, None]
+    condition: list[Condition]
```
kerizane: As these are now lists you should rename to `conditions`, `techniques`, `extras`.

tuxemon/combat.py
```diff
@@ -81,8 +81,11 @@ def pre_checking(
         monster.status[0].combat_state = combat
         monster.status[0].phase = "pre_checking"
         result_status = monster.status[0].use(target)
-        if result_status["technique"]:
-            technique = result_status["technique"]
+        if result_status.technique:
+            if len(result_status.technique) > 1:
```
kerizane: You could drop the branches and always use random.choice; it would be more concise. Up to you.

kerizane (changes requested):

tuxemon/combat.py
JaskRendix: it makes sense

kerizane (approved):

PR #2527, opened 2024-10-20
DamageEffect (technique)

JaskRendix:
Here's what's changed:
* gotten rid of the 'area' effect and instead implemented a new mechanism in damage.py that allows for targeting multiple monsters in 2vs2 battles. This is now handled automatically by checking the targets in the technique's target field (new methods in stats/combat.py as well as combat.py).
* renamed the 'enhance' effect to 'empty' and removed it from most techniques, as it's become redundant. I originally added this effect back in January 2023 because techniques were failing without a 'damage' component, due to issues with checking if a target was hit or missed. However, thanks to several updates since then, this is no longer a problem. We're keeping the 'empty' effect for now, though, as it allows a placeholder technique (for status/condition) to function without failing.
* updated a few other effect techniques to make sure everything is working smoothly.

This effect "area" was born out of a discussion with @Sanglorian months ago, where we were brainstorming ways to target multiple monsters on the battlefield, especially in 2v2 scenarios (lava, earthquake, avalanche, etc).

I'm talking about this field:
```
  "target": {
    "enemy_monster": false,
    "enemy_team": false,
    "enemy_trainer": false,
    "own_monster": false,
    "own_team": false,
    "own_trainer": false
  },
```
example  1 vs 1:
- if "**enemy_monster**" or  "**own_monster**" are True -> nothing changes
- if "**enemy_team**" or  "**own_team**" are True -> nothing changes
- if "**enemy_trainer**" or  "**own_trainer**" are True -> the attack is going to damage all the monsters in the party

example  2 vs 2 or 1 vs 2 or 2 vs 1:
trainer 1: nut and rockitten 
trainer 2: pairagrin and botbot
- if "**enemy_monster**" or  "**own_monster**" are True -> the monster (target) selected
- if "**enemy_team**" or  "**own_team**" are True -> the monster (target) selected + all the others in the field (same side): **rockitten** targets **botbot**, but the technique has only **enemy_team** (True), so it'll damage both **botbot** and **pairagrin**
- if "**enemy_trainer**" or  "**own_trainer**" are True -> the attack is going to damage all the monsters in the party

Note: if a technique doesn't have any True values in its target field, it won't inflict any damage (there is a logger.error).

regarding "**enemy_trainer**" or  "**own_trainer**" , look here, technique **Peck** with only **enemy_trainer** **True**, everything else **False**.

https://github.com/user-attachments/assets/c8cf4d75-dcc6-44a7-a1f0-ee086d1ea976


JaskRendix:
@Sanglorian the ''give'' effect from:
`"give exhausted,target"` to `"give exhausted,enemy_monster"`
`"give exhausted,user"` to `"give exhausted,own_monster"`
allowing the possibility to use also **enemy_team**, etc. so we can handle the effects more easily for double battles

what do you want to do with the "both":
- amnesia `"give exhausted,both"`
- blood_bond `"give lifeleech,both"`
- bubble_trap `"give grabbed,both"` (this is the only technique that's managed to grab both, but if we consider the bubble trap as `own_monster:enemy_monster`, then that opens up the possibility of creating an additional category that specifically targets both team. Something like: Spider Web or Gravity Crush or EMP (Electromagnetic Pulse))
- lava `"give burn,both"` (if we apply the 'floor is lava' logic, then it's `own_team:enemy_team`)
- levitate `"give sniping,both"`
- muddle `"give focused,both"`
- sleep_bomb `"give noddingoff,both"`

I was surprised to notice that there aren't many techniques with "both", however remember to think always in `2vs2` because the target **_team** is going to work even in 1 vs 1. You have the following option:
- `give xxx,own_monster:enemy_monster`?
- `give xxx,own_team:enemy_team`?
- `give xxx,own_trainer:enemy_trainer`?

same as above, but for other effects:
- cloud_aether `"switch both,random"` (switch the element/types randomly)
- neutralize `"reverse both"` (reverse do the opposite of switch, so the element/types are back to normal)

Sanglorian:
> @Sanglorian the ''give'' effect from: `"give exhausted,target"` to `"give exhausted,enemy_monster"` `"give exhausted,user"` to `"give exhausted,own_monster"` allowing the possibility to use also **enemy_team**, etc. so we can handle the effects more easily for double battles

Oh nice! Does own_monster mean either monster in a double battle, or just the user? I think in general with these "both" ones, we want to give it to the user and to a single enemy monster of the player's choice. I e. Even in a double battle if Vamporm uses Blood Bond it should be the one who gets lifeleech, it shouldn't be able to choose its ally to get lifeleech.

> 
> what do you want to do with the "both":
> 
>     * amnesia `"give exhausted,both"`

Target and user 
> 
>     * blood_bond `"give lifeleech,both"`
> 
>     * bubble_trap `"give grabbed,both"` (this is the only technique that's managed to grab both, but if we consider the bubble trap as `own_monster:enemy_monster`, then that opens up the possibility of creating an additional category that specifically targets both team. Something like: Spider Web or Gravity Crush or EMP (Electromagnetic Pulse))

I like Gravity Crush as a "bigger" version of Bubble Trap!

> 
>     * lava `"give burn,both"` (if we apply the 'floor is lava' logic, then it's `own_team:enemy_team`)
> 

Yeah I think this is a good candidate for own+enemy teams.

>     * levitate `"give sniping,both"`
> 
>     * muddle `"give focused,both"`
> 
>     * sleep_bomb `"give noddingoff,both"`

Another good one to affect both teams. We could have "Dream Encounter" or something as a user+one enemy monster variant.
> 
> 
> I was surprised to notice that there aren't many techniques with "both", however remember to think always in `2vs2` because the target **_team** is going to work even in 1 vs 1. You have the following option:
> 
>     * `give xxx,own_monster:enemy_monster`?
> 
>     * `give xxx,own_team:enemy_team`?
> 
>     * `give xxx,own_trainer:enemy_trainer`?
> 
> 
> same as above, but for other effects:
> 
>     * cloud_aether `"switch both,random"` (switch the element/types randomly)

Own team and enemy team could be fun here - lots of changes at once!
> 
>     * neutralize `"reverse both"` (reverse do the opposite of switch, so the element/types are back to normal)

I think this should neutralise for everyone. It's already quite niche so I think it's nice for it to have broad effect.

JaskRendix:
> Oh nice! Does own_monster mean either monster in a double battle, or just the user? I think in general with these "both" ones, we want to give it to the user and to a single enemy monster of the player's choice. I e. Even in a double battle if Vamporm uses Blood Bond it should be the one who gets lifeleech, it shouldn't be able to choose its ally to get lifeleech.

`own_monster` is actually the replacement for "user" in the majority of technique effects, which helps to standardize the language and eliminate differences with the target field in techniques (such as enemy_team, own_team, etc.). The mechanism doesn't change if everything gets replaced as proposed, the only issue was for "both", but you have already addressed it.

---

I think we really need is a separate effect that's specifically for healing, without any extra attributes attached to it. The thing is, we can't have both '**damage**' and '**healing**' in the effects at the same time. We can get around this by using '**prop_healing**' instead. There are only 2 techniques with "healing" and "damage" and since both are ranged, we can use prop_healing, but you need to set a percentage.

Blossom 
```
  "effects": [
    "damage",
    "prop_healing own_monster,xxx" 0.33? (33% of user HP? something else?)
  ],
```
Battery Discharge
```
  "effects": [
    "damage",
    ""prop_healing own_monster,xxx" 0.33? (33% of user HP? something else?)
  ],
```
Sleeping Powder, in its original form, this technique had an 'area' effect, but since it can now target multiple monsters, we could consider changing it to `enemy_team:own_team`, what's your opinion?
```
  "effects": [
    "give noddingoff,enemy_team:own_team"
  ],
```

Sanglorian:
> I think we really need is a separate effect that's specifically for healing, without any extra attributes attached to it. The thing is, we can't have both '**damage**' and '**healing**' in the effects at the same time. We can get around this by using '**prop_healing**' instead. There are only 2 techniques with "healing" and "damage" and since both are ranged, we can use prop_healing, but you need to set a percentage.
> 
> Blossom
> 
> ```
>   "effects": [
>     "damage",
>     "prop_healing own_monster,xxx" 0.33? (33% of user HP? something else?)
>   ],
> ```
> 
> Battery Discharge
> 
> ```
>   "effects": [
>     "damage",
>     ""prop_healing own_monster,xxx" 0.33? (33% of user HP? something else?)
>   ],
> ```

Thanks! Longer term a separate effect for healing sounds good, but in the meantime prop_healing is a good idea. I'd say 20% is enough as it also does damage. 

> 
> Sleeping Powder, in its original form, this technique had an 'area' effect, but since it can now target multiple monsters, we could consider changing it to `enemy_team:own_team`, what's your opinion?
> 
> ```
>   "effects": [
>     "give noddingoff,enemy_team:own_team"
>   ],
> ```

Maybe just "enemy_team" for this one, but lowering Potency to 80%.

JaskRendix:
@Sanglorian where should we document the fact that [Sleeping Powder](https://wiki.tuxemon.org/Sleeping_Powder) specifically targets the opposing team (enemy_team)? We could use Effects and put `give noddingoff,enemy_team`? Consider renaming 'User Condition' to 'Own Monster Condition' and 'Target Condition' to 'Enemy Monster Condition'. Instead of introducing separate fields for 'Enemy Team' and 'Own Team', we could potentially handle team-specific effects within the existing effect field. Thoughts?

Sanglorian:
> @Sanglorian where should we document the fact that [Sleeping Powder](https://wiki.tuxemon.org/Sleeping_Powder) specifically targets the opposing team (enemy_team)? We could use Effects and put `give noddingoff,enemy_team`? Consider renaming 'User Condition' to 'Own Monster Condition' and 'Target Condition' to 'Enemy Monster Condition'. Instead of introducing separate fields for 'Enemy Team' and 'Own Team', we could potentially handle team-specific effects within the existing effect field. Thoughts?

Hmm, am I right that in-game it's all handled through effects? Maybe the wiki should have one field that just shows the effects code, and then separately a field where the effects can be written in simple prose? Something like:

Potency 80%
Effects give nodding_off,enemy_team
Explanation 80% chance for enemy team to nod off.

Edit: Sorry I closed it when I meant to post comment instead.

JaskRendix:
no problem, ready to merge

kerizane (approved):

PR #2528, opened 2024-10-22
fix levels (combat.py)

JaskRendix:
I found a bug while testing for #2527 and this pull request fixes it. The issue was that when two monsters leveled up after a battle, only the last one would get the technique. This was happening because the levels were being overwritten. To fix this, I've changed the code to pass the list of learned techniques instead of the levels. This should resolve the bug, but I've noticed that the message still doesn't display entirely. However, that's being addressed in a separate PR related to the callback, so we can tackle that issue next.

I've tweaked the string concatenation code and it's now fixed the issue where an extra value would cause an empty line to appear.

unrelated: @Sanglorian what if we made it so that monsters learn techniques at uneven levels and evolve at even levels (or vice versa)? Since there are fewer evolutions to modify than techniques in the moveset, it might be easier to implement evolving at uneven levels. This could be a nice quality of life improvement. I could also add a simple test to make sure everything works as expected. What do you think?

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -876,7 +876,11 @@ class CombatState(CombatAnimations):
             m = None
 
         if result_tech["extra"]:
-            m = (m or "") + "\n" + T.translate(result_tech["extra"])
+            m = (
+                (m or "")
+                + ("\n" if m else "")
+                + T.translate(result_tech["extra"])
+            )
 
         if m:
             message += "\n" + m
@@ -1072,6 +1076,7 @@ class CombatState(CombatAnimations):
         """
         winners = get_winners(monster, self._damage_map)
         if winners:
+            new_techniques = []
             for winner in winners:
                 # Award money and experience
                 awarded_mon = award_money(monster, winner)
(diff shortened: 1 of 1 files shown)


kerizane (approved):

Sanglorian:
> unrelated: @Sanglorian what if we made it so that monsters learn techniques at uneven levels and evolve at even levels (or vice versa)? Since there are fewer evolutions to modify than techniques in the moveset, it might be easier to implement evolving at uneven levels. This could be a nice quality of life improvement. I could also add a simple test to make sure everything works as expected. What do you think?

I like the idea of a simple rule for avoiding having an evolution and a new technique on the same level. My only concern with even versus odd is that it limits the technique progressions we can have to once every two levels, once every four levels and once every six levels (1, 3, 5 or 1, 5, 9 or 1, 7, 13, etc).

What if we sorted out a few standard progressions (once every three levels, once every four levels and once every five levels, maybe), and then for each of those progressions we could recommend evolution levels that don't clash?

(E.g. if you pick a once every three levels technique progression - 1, 4, 7, 10, 13, 16, etc, then standard level up options could be 9 and 17 for a fast evolver that evolves twice, 15 and 30 for a medium evolver and 21 and 42 for a slow evolver.)

A monster wouldn't have to learn a technique every three levels or anything like that, just that when it does learn a technique it appears in one of those "slots".

PR #2529, opened 2024-10-23
test + at_level (evolutions) + level_learned (moveset)

JaskRendix:
From @Sanglorian [reply](https://github.com/Tuxemon/Tuxemon/pull/2528#issuecomment-2430392707)

PR updates the 'level_learned' and 'at_level' fields. It also adds new tests to 'test_jsons_monster' to catch potential issues. The 'level_learned' field in the moveset will follow a specific sequence (1, 3, 4, 7, 10, ..., 100), while 'at_level' will have more flexibility without a predefined sequence. Additionally, a new test has been included to flag 'double' levels, where a level appears in both the moveset and evolution, to ensure consistency and accuracy. Moveset only exception is 1 (allowing 2 techniques at level 1).

The only evolution levels changed are: Jelillow (from 40 to 41), Stonifly (from 10 to 9)

Is it ok @Sanglorian ? Do you want some additional check?

### DIFF
--- a/mods/tuxemon/db/monster/aardart.json
+++ b/mods/tuxemon/db/monster/aardart.json
@@ -3,23 +3,23 @@
     "category": "anteater",
     "moveset": [
         {
-            "level_learned": 2,
+            "level_learned": 1,
             "technique": "tonguespear"
         },
         {
-            "level_learned": 2,
+            "level_learned": 1,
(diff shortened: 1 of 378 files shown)


Sanglorian:
Nice! That sounds good to me, thanks.

kerizane (approved):

PR #2531, opened 2024-10-24
test jsons technique

JaskRendix:
split out from #2527 

PR adds test for technique files JSON. I've included standard tests to catch duplicates and missing tech_ids, as well as a test to verify the 'range' technique has the correct effects (damage and other using simple_damage_calculate). This was previously in db.py, but I've moved it here so we can catch any issues before merging. I've also added tests to check for valid combinations (which will be necessary later when 'healing' loses its attributes) and a couple of tests to ensure moves with simple_damage_calculate have a power greater than 0 and those with simple_heal have a healing_power greater than 0.

### DIFF
--- a/mods/tuxemon/db/technique/pseudopod.json
+++ b/mods/tuxemon/db/technique/pseudopod.json
@@ -8,7 +8,7 @@
   ],
   "flip_axes": "x",
   "is_fast": false,
-  "potency": 0.0,
+  "potency": 0.7,
   "power": 1.25,
   "range": "reach",
   "recharge": 1,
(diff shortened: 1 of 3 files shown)


JaskRendix:
@Sanglorian I've added a test to ensure all techniques with the status-adding effect "give" have a potency greater than 0 (no potency = no condition). [Pseudopod](https://wiki.tuxemon.org/Pseudopod) technique has a potency of 0. Could you tell me the value?

Sanglorian:
> @Sanglorian I've added a test to ensure all techniques with the status-adding effect "give" have a potency greater than 0 (no potency = no condition). [Pseudopod](https://wiki.tuxemon.org/Pseudopod) technique has a potency of 0. Could you tell me the value?

Good catch, thank you. I've gone with 70%.

kerizane (approved):

PR #2534, opened 2024-10-25
leather flashback (Spyder)

JaskRendix:
PR adds flashback. @Sanglorian  I wanted to follow up on our previous conversation about Zoolander.

I asked about this because I've added a flashback scene in Leather Town. The scene introduces Sarah, Zoolander's girlfriend. It also shows Zoolander as a complex and introspective person, which seems to be struggling with the consequences of her actions, particularly with regards to the mine and the decision to divert the river and drain the lake.

Do you think this flashback can be helpful? Let me know.


Sanglorian:
Thanks @JaskRendix , it looks good to me!

kerizane (approved):

PR #2535, opened 2024-10-26
update choice states (monster, item, state, etc.)

JaskRendix:
PR updates choice states.

First, I've removed all the magic numbers from the code and replaced them with constants at the top of the page, which makes everything much cleaner and easier to understand. I've also extracted a new method to declutter the init code, making it more modular and manageable. 

Additionally, I've fixed a bug that was causing issues with the monster info and journal info. Specifically, if a monster had the same evolution twice (like when it can evolve using two different items), it would show up twice, but that's now been addressed. 

Lastly, I've made some improvements to the `translated_dialog_choice`. If there are more than 13 elements, a scroll bar will now appear, which prevents the screen from crashing. The scroll bar is also capped at 80% of the screen height, so everything should be much more stable.

### DIFF
--- a/tuxemon/states/choice/choice_item.py
+++ b/tuxemon/states/choice/choice_item.py
@@ -14,6 +14,14 @@ from tuxemon.menu.menu import PygameMenuState
 from tuxemon.menu.theme import get_theme
 
 ChoiceMenuGameObj = Callable[[], None]
+LENGTH_NAME_ITEM = 10
+MAX_MENU_ELEMENTS = 7
+MAX_MENU_HEIGHT_PERCENTAGE = 0.8
+SCALE_SPRITE = 0.5
+WINDOW_WIDTH_PERCENTAGE_LONG = 0.4
+WINDOW_WIDTH_PERCENTAGE_SHORT = 0.3
+TRANSLATE_PERCENTAGE_LONG = 0.4
+TRANSLATE_PERCENTAGE_SHORT = 0.3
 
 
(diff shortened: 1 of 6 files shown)


tuxemon/states/choice/choice_item.py
```diff
@@ -14,6 +14,14 @@
 from tuxemon.menu.theme import get_theme
 
 ChoiceMenuGameObj = Callable[[], None]
+LENGHT_NAME_ITEM = 10
```
kerizane: typo

kerizane (changes requested):

tuxemon/states/choice/choice_item.py
```diff
@@ -14,6 +14,14 @@
 from tuxemon.menu.theme import get_theme
 
 ChoiceMenuGameObj = Callable[[], None]
+LENGTH_NAME_ITEM = 10
```
kerizane, line 17: This const seems to be unused.

kerizane: Same for the other files.

kerizane: Haven't checked if the other consts are used.

kerizane (changes requested):

tuxemon/states/choice/choice_item.py
JaskRendix: it was the only one unused

kerizane (approved):

PR #2536, opened 2024-10-26
updates and refactors TranslatorPo

JaskRendix:
PR updates and refactors TranslatorPo.

I've taken the original `TranslatorPo` class and broken it down into three separate classes, each with its own specific responsibility.

Now we have `LocaleFinder`, which is in charge of searching for locales and returning `LocaleInfo` objects. It's got a single method, `search_locales`, that yields `LocaleInfo` objects as it finds them.

Next up is `GettextCompiler`, which handles compiling PO files into MO files. Its main method, `compile_gettext`, takes a PO file path and an MO file path as input and does the rest.

Finally, we've got `TranslatorPo`, which is the main class that ties everything together. It uses `LocaleFinder` and `GettextCompiler` to collect languages, build translations, and load a selected locale for translation.

Each class has a single job to do, making it way easier to understand and modify the code in the future (like loading and merging file.po from multiple mods or something similar)

output:
`{'ja', 'it_IT', 'fi', 'cs_CZ', 'nb_NO', 'zh_CN', 'de_DE', 'eo', 'README.md', 'es_ES', 'pt_BR', 'pl', 'fr_FR', 'en_US', 'es_MX'}`

do we need to standardize? jp_JP instead of ja, etc?

### DIFF
--- a/tuxemon/locale.py
+++ b/tuxemon/locale.py
@@ -8,6 +8,7 @@ import logging
 import os
 import os.path
 from collections.abc import Callable, Generator, Iterable, Mapping, Sequence
+from gettext import GNUTranslations
 from typing import Any, Optional
 
 from babel.messages.mofile import write_mo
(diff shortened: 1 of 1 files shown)


kerizane:
> do we need to standardize? jp_JP instead of ja, etc?

Don't mind.

kerizane (approved):

PR #2537, opened 2024-10-27
purge class Mission

JaskRendix:
PR purge the Mission class, the attribute missions (NPC) as well as all the related methods in the class (NPC), the event action set_mission and the event condition check_mission, the str enum MissionStatus in db.py and the MissionModel in db.py. 

I've been taking a closer look at the Mission class lately, and I have to admit, I'm not convinced it's really adding much value. I tried to revamp it, but the more I worked on it, the more I realized it was just too generic and might not have been the best idea to implement as a standalone feature. That being said, I think I've found a better home for the mission concept - integrating it with the phone app, specifically the contacts feature. This way, it can be more naturally woven into the game's existing mechanics (another PR).

By the way there is only 1 mission and it's in Spyder.


kerizane:
Maybe we should leave this in until you can reimplement it?

PR #2553, opened 2024-11-23
shop

JaskRendix:
PR follows #2544 and introduces the ability to buy Tuxemon by defining it in the economy JSON. It also enables the following features:

• The economy can be given a custom name, which will be displayed in the window.
• The background of the shop can be changed, allowing different shops to have distinct appearances. The default background remains the blue gradient.
• Items and monsters will be assigned to NPCs, but instead of using traditional items and monsters, they will be placed in their respective boxes. This change helps reduce duplicate code and improves overall efficiency.

```
    "slug": "shop",
    "monsters": [
      {
        "name": "rockitten",
        "price": 9999,
        "inventory": 1, (default 1, impossible to set infinite value -1)
        "level": 5 (default 1, impossible negative numbers or 0)
      },
    "items": [
      {
        "name": "potion",
        "price": 20,
        "cost": 5
      }
    ],
```

https://github.com/user-attachments/assets/d0b2514a-b9bd-45da-8c4d-d16b88d9f77f


Sanglorian:
Nice, thank you! 

I think displaying the descriptions might be too much detail, at least on the sale screen - is it possible to leave those out? That would allow more items to be shown (I'm imagining shops might buy back most items, even if they only sell a handful).

JaskRendix:
> I think displaying the descriptions might be too much detail, at least on the sale screen - is it possible to leave those out? That would allow more items to be shown (I'm imagining shops might buy back most items, even if they only sell a handful).

@Sanglorian there's no problem with leaving out the descriptions on the sale screen, that's definitely doable. As for the shop buyback, the only issue is that we'd need to add most items to the JSONs, which could be a bit of work. This is because at the moment the shop only buys back the items that it sells, and since the value is missing for other items, it would result in a crash. Alternatively, we could add a default value to the item class to simplify things, that way we wouldn't have to explicitly define every single item in the JSONs.

JaskRendix:
@Sanglorian could you please complete the cost for each item on the wiki (https://wiki.tuxemon.org/Category:Item) (bottom table)? That way, I can update the shop to buy back the majority of items, rather than just the ones it sells.

Sanglorian:
> @Sanglorian could you please complete the cost for each item on the wiki (https://wiki.tuxemon.org/Category:Item) (bottom table)? That way, I can update the shop to buy back the majority of items, rather than just the ones it sells.

Oh thanks, that makes a lot of sense. I've made a start today, will do some more later and let you know when I've done them all (except ones that shouldn't be sell-able, which I'll leave at no cost entered).

Sanglorian:
Okay I think I've added prices for all the ones that should be sell-able. Thanks again for doing this!

I changed a few of the prices - I thought they were a bit low. But let me know if that doesn't make sense and I can change them back.

P.S. I noticed the Blocker Ball mentions protection moves. Do we have any of these currently? I like the idea of them.

JaskRendix:
@Sanglorian 

> Okay I think I've added prices for all the ones that should be sell-able. Thanks again for doing this! I changed a few of the prices - I thought they were a bit low. But let me know if that doesn't make sense and I can change them back.

Thank, I'm going to check.

> P.S. I noticed the Blocker Ball mentions protection moves. Do we have any of these currently? I like the idea of them.

I think you're talking about the Crusher Ball, not the Blocker Ball. As for protection moves, I took a closer look at the mechanism and from what I can see, it's not a specific move per se, but rather the presence of a positive status on the wild monster.

```
        if item.slug == "tuxeball_crusher":
            crusher = ((target.armour / 5) * 0.01) + 1
            if crusher >= 1.4:
                crusher = 1.4
            if (
                self._calculate_status_modifier(target)
                == prepare.STATUS_POSITIVE
            ):
                crusher = 0.01  <----------
            tuxeball_modifier = crusher
```

Sanglorian:
> > P.S. I noticed the Blocker Ball mentions protection moves. Do we have any of these currently? I like the idea of them.
> 
> I think you're talking about the Crusher Ball, not the Blocker Ball. As for protection moves, I took a closer look at the mechanism and from what I can see, it's not a specific move per se, but rather the presence of a positive status on the wild monster.
> 
> ```
>         if item.slug == "tuxeball_crusher":
>             crusher = ((target.armour / 5) * 0.01) + 1
>             if crusher >= 1.4:
>                 crusher = 1.4
>             if (
>                 self._calculate_status_modifier(target)
>                 == prepare.STATUS_POSITIVE
>             ):
>                 crusher = 0.01  <----------
>             tuxeball_modifier = crusher
> ```

Ah thank you, yes! I've updated the description to explain it's about the condition used.

JaskRendix:
@Sanglorian I'm a bit confused about the cost and price of items. From what I understand, the cost refers to the amount of money a player would receive if they sold the item to a shop, whereas the price can vary from shop to shop (eg in Cotton you pay 500 while in Candy only 250, etc).

However, I noticed a discrepancy between the price of the Nocturnal Ball (as well as other items) in the game (300) and its cost on Tuxepedia (350).

This would imply that the player purchases the item for 300, but if they were to resell it to any shop, they would receive 350, effectively making a profit of 50.

Sanglorian:
Ah, sorry for the confusion, @JaskRendix !

What I put down on the wiki was what I had in mind for the standard sale price, rather than the cost/resale price the player can get for selling it. For resale price, I would go with 50% of the standard sale price (rounded up) - so if a Tuxeball usually sells in shops for $100, its cost would be $50.

And then I imagined that most shops - at least in the Spyder campaign - would sell at the standard sale price. With maybe some exceptions for special events or opportunities (a trainer who when defeated sells you a potion cheap, or someone deep in the woods who sells a Cureall at discount once per day, etc.).

JaskRendix:
> And then I imagined that most shops - at least in the Spyder campaign - would sell at the standard sale price. With maybe some exceptions for special events or opportunities (a trainer who when defeated sells you a potion cheap, or someone deep in the woods who sells a Cureall at discount once per day, etc.).

That's an interesting twist, but it would indeed require a significant overhaul of the system.

PR #2538, opened 2024-10-29
Adding Tech Shops to Candy and Cotton Scoops

Sanglorian:
I hope I've done this correctly, I'm afraid I can't run Tuxemon on my system at the moment to confirm that it's working. 

If these two work, I'll replicate them for the other towns. 

I wanted to make a place for TMs specifically to be sold, as we've got a lot of techniques that could become TMs and I didn't want them blocking the more basic items in the Scoop stores. 

I also wanted to make the evolution items for the "new" starters available. 

**Yet to do**
I'd also like these shops to sell the "new" starters. Is there a way to add tuxemon for sale in an economy, or does that need to be done through a separate process?

PR #2557, opened 2024-12-01
scoops maps (Spyder)

JaskRendix:
PR incorporates changes from #2538, affecting the scoop and all related areas. Key updates include:
- Blocking access to the behind-counter area, allowing interaction with the shopkeeper only from the counter.
- Adding an additional payment terminal.
- Removing the event action with behavior, as the path is now blocked.
- Simplifying and removing some collision lines.


Sanglorian:
Thank you!

kerizane (approved):

PR #2559, opened 2024-12-07
standard cost (JSON)

JaskRendix:
PR adds a 'cost' field to the items JSON, which represents the standard cost of an item. This cost will be displayed when the item is not available in the shop, specifically when the player intends to resell the item to the shop. This implementation lays the groundwork for the changes proposed in #2553.

@Sanglorian, I noticed that some new tuxeballs (e.g. Gambler, Noble) were missing from Tuxepedia, so I've added them. Please feel free to fill in the standard cost for these items.


Sanglorian:
Thank you, I think this will make shops much more usable for players - and it's a useful guide for anyone setting up their own shop prices.

kerizane (approved):

PR #2563, opened 2024-12-25
TechEffectResult (from typedict to dataclass)

JaskRendix:
follows #2520

PR converts `TechEffectResult` from a `TypedDict` to a `dataclass`. By using a `dataclass`, we can take advantage of its built-in features, such as automatic generation of `__init__` and `__repr__` methods, and improved performance. And, we can access the fields directly as attributes, rather than having to use dictionary keys.

```
@dataclass
class TechEffectResult:
    name: str
    success: bool
    damage: int
    element_multiplier: float
    should_tackle: bool
    extras: list[str]
```


JaskRendix:
Many draft PRs are stuck because of this one, which is waiting on #2520. I'm not sure what the issue is with #2520, but if we could merge this one first, I can handle any conflicts that come up with #2520 later.

kerizane (approved):

kerizane:
#2520 was merged months ago.

JaskRendix:
@kerizane wrong number, it was #2530

PR #2566, opened 2024-12-29
Soldier Dialogue (Spyder)

JaskRendix:
PR expands the Soldier's dialogue to include additional lines and details.

What do you think, @Sanglorian? Do you have any suggestions for enhancing the dialogue? I was thinking of an ending where the soldier gifts the player a specific tuxemon, but we need to decide which one, at what level, and how it could be utilized again later in the game.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -9262,40 +9262,78 @@ msgstr "You got it to morph? \n ... \n And you did it in that particular way? \n
 
 msgid "spyder_route1_beachcomber"
 msgstr "To run, just hold down the SHIFT key. \n Look! I'm free as an ELOFLY!"
+
 msgid "spyder_cottonhouse1_homemaker"
 msgstr "I feel so safe seeing all the Enforcers around. You can never be too careful. \n The newspapers are always reporting terrible crimes in this very region!"
+
 msgid "spyder_cottonhouse1_firefighter"
 msgstr "Did you know it's possible for tuxemon to have two elements? \n Even a technique can belong to two elements at one."
+
 msgid "spyder_cottonhouse2_magician"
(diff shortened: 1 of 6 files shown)


Sanglorian:
Thanks @JaskRendix, sounds good to me. 

For the papers in Nimrod HQ, it seems strange that they are so similar but not identical. "The soldier has shown a significant decline in mental health, frequently expressing guilt and remorse over past missions." vs "The overseer has shown significant signs of mental distress, often expressing feelings of guilt and conflict regarding recent decisions at the mine."

I think these should be identical, like the doctors are filling out the same form, or more distinct to reflect different writing styles for the two doctors. 

I like the idea of the soldier gifting the player a new monster - maybe Rosarin as it doesn't have a place in the plot yet, and its weapons make it quite soldiery?

JaskRendix:
@Sanglorian you're right, the language used in the two messages is quite similar. What do you think about this?

msgid "spyder_nimrod_papers2"
msgstr "Subject - Overseer #### by Dr. Orion - Mental Health Assessment\n 'The overseer is exhibiting signs of significant emotional distress, grappling with ethical dilemmas and the weight of recent operational decisions. \n She reports a growing sense of isolation and anxiety, particularly concerning the evolving demands from investors that have necessitated a shift in project priorities. \n Her fixation on the environmental impact of the river diversion project, including the draining of the lake, suggests a profound sense of responsibility and potential moral conflict.' "

> I like the idea of the soldier gifting the player a new monster - maybe Rosarin as it doesn't have a place in the plot yet, and its weapons make it quite soldiery?

Good choice, Rosarin does have a fitting design for a soldier's gift, and its abilities would make for an interesting addition to the player's team. I'm actually considering bringing the soldier back into the story at a later point, perhaps in a different context or location, to add more depth to their character and create a sense of continuity.

Sanglorian:
That change in the assessment looks good, thanks!

JaskRendix:
Updated soldier dialogue, removed Route 5 instance (to be added later), ready for merge.

kerizane (approved):

PR #2568, opened 2024-12-31
refactor calc_dialog_rect + positon parameter (dialog)

JaskRendix:
PR refactors the `calc_dialog_rect` function to include a `position` parameter, allowing for more flexibility in positioning dialog boxes on the screen. Specifically:
* Added a `position` parameter to the `calc_dialog_rect` function, which can take one of the following values: "top", "bottom", "center", "topleft", "topright", "bottomleft", "bottomright", "left", "right"
* Updated the function to position the dialog box according to the specified position
* Removed hardcoded positioning logic, making the function more reusable and flexible

The main reason: enables the possibility to use different positioning of dialog boxes to deliver a better experience to the players when there is a dialogue among two or more NPCs (eg flashbacks), allowing for a clearer understanding of who says what and enhancing the overall narrative flow

https://github.com/user-attachments/assets/37a23f58-0e1e-492f-b82b-4812fa5b5bc2

video above generated by: 
```
    <property name="act11" value="translated_dialog spyder_papertown_sunnyside"/>
    <property name="act12" value="translated_dialog spyder_papertown_sunnyside,,bottomright"/>
    <property name="act13" value="translated_dialog spyder_papertown_sunnyside,,right"/>
    <property name="act14" value="translated_dialog spyder_papertown_sunnyside,,topright"/>
    <property name="act15" value="translated_dialog spyder_papertown_sunnyside,,top"/>
    <property name="act16" value="translated_dialog spyder_papertown_sunnyside,,topleft"/>
    <property name="act17" value="translated_dialog spyder_papertown_sunnyside,,left"/>
    <property name="act18" value="translated_dialog spyder_papertown_sunnyside,,bottomleft"/>
    <property name="act19" value="translated_dialog spyder_papertown_sunnyside,,center"/>
    <property name="act20" value="translated_dialog spyder_papertown_sunnyside"/>
    <property name="cond1" value="is char_facing_tile player"/>
    <property name="cond2" value="is button_pressed K_RETURN"/>
```

PR #2569, opened 2024-12-31
Billie's House (Spyder)

JaskRendix:
Waiting #2568 

PR adds interactions for Billie's room, including her bed, bookshelf, and weights (still missing). It also includes a flashback sequence that plays when the player interacts with the TV that reveals more about Billie's personality and background.

descriptions:
- Bed: "Bed is messy, but it looks like she's been sleeping well. There's a tuxemon plushie on her pillow."
- Bookshelf: "The bookshelf is filled with tuxemon strategy guides and training manuals."
- Weights: "There's a diagram on the wall nearby that shows how to use the weight to train a tuxemon's strength."

flashback: 

https://github.com/user-attachments/assets/4fa5cf43-58af-4c7f-832f-2cf3ff4a0533

@Sanglorian I've also tested the new dialog position and I think it's a big improvement. It helps the player to understand better who is speaking and adds a nice touch of realism to the game. I'm currently using a top-down approach. Do you think this works well?

Additionally, the fight scene. Do you think it's too short, or does it feel like a good length to you? Are there any other suggestions you have for improving the scene or adding more depth to the story?

Let me know your thoughts.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -11344,6 +11344,15 @@ msgstr "Home sweet home!"
 msgid "spyder_papertown_rival"
 msgstr "Billie's House"
 
+msgid "spyder_rivalbedroom_bed"
+msgstr "Bed is messy, but it looks like she's been sleeping well. There's a tuxemon plushie on her pillow."
+
+msgid "spyder_rivalbedroom_bookshelf"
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks, that looks good to me! I think the fight is a good length - especially since there's Billie's commentary after it. 

And the new dialogue position works well - was immediately sensible to me who was speaking. 

I like the nod towards a popular video game franchise in the names of the battlers, too! 

Sorry I've been incommunicado for a while - will go through other outstanding pull requests over the next couple of days.

kerizane (approved):

PR #2580, opened 2025-02-06
Modifier(BaseModel)

JaskRendix:
PR:
- introduces a new data model `Modifier` to represent damage modifiers with attributes, values, and multipliers
- implements the `weakest_link` and other formulas to calculate damage
- implements the unittest suite for all the formulas

weakest_link -> Returns the smallest damage multiplier that applies to the given monster.
strongest_link -> Returns the largest damage multiplier that applies to the given monster.
cumulative_damage -> Returns the cumulative product of all applicable damage multipliers for the given monster.
average_damage -> Returns the average of all applicable damage multipliers for the given monster.
first_applicable_damage -> Returns the first applicable damage multiplier for the given monster.

weakest_link will be used for the burn and poison condition

### DIFF
--- a/mods/tuxemon/db/condition/blinded.json
+++ b/mods/tuxemon/db/condition/blinded.json
@@ -1,5 +1,6 @@
 {
   "animation": null,
+  "damage_modifiers": [],
   "category": "negative",
   "effects": [
     "statchange"
--- a/mods/tuxemon/db/condition/blinded.json
+++ b/mods/tuxemon/db/condition/blinded.json
@@ -1,6 +1,6 @@
 {
   "animation": null,
-  "damage_modifiers": [],
+  "modifiers": [],
   "category": "negative",
   "effects": [
     "statchange"
--- a/mods/tuxemon/db/condition/burn.json
+++ b/mods/tuxemon/db/condition/burn.json
@@ -1,6 +1,6 @@
 {
   "animation": null,
-  "damage_modifiers": [
+  "modifiers": [
     {
       "attribute": "type",
       "values": ["fire"],
--- a/mods/tuxemon/db/condition/burn.json
+++ b/mods/tuxemon/db/condition/burn.json
@@ -1,5 +1,17 @@
 {
   "animation": null,
+  "damage_modifiers": [
+    {
+      "attribute": "type",
+      "values": ["fire"],
+      "multiplier": 0.0
+    },
+    {
+      "attribute": "type",
+      "values": ["frost"],
+      "multiplier": 2.0
+    }
+  ],
   "category": "negative",
   "effects": [
     "burnt 8"
--- a/mods/tuxemon/db/condition/chargedup.json
+++ b/mods/tuxemon/db/condition/chargedup.json
@@ -1,6 +1,6 @@
 {
   "animation": null,
-  "damage_modifiers": [],
+  "modifiers": [],
   "category": "positive",
   "effects": [
     "statchange",
--- a/mods/tuxemon/db/condition/chargedup.json
+++ b/mods/tuxemon/db/condition/chargedup.json
@@ -1,5 +1,6 @@
 {
   "animation": null,
+  "damage_modifiers": [],
   "category": "positive",
   "effects": [
     "statchange",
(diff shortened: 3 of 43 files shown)


Sanglorian:
Thanks @JaskRendix 

We could assign each condition that does damage an "element" - Fire for burn, Venom for poison, etc? Then use that damage multiplier for the damage from the condition? 

Otherwise I like the idea of some elements being immune to certain conditions.

JaskRendix:
@Sanglorian probably I wrote too much above, but shortly I have added these to the burn condition.

(blabla)

example message:
"XXX is immune to YYY"

Sanglorian:
> @Sanglorian probably I wrote too much above, but shortly I have added these to the burn condition.
> 
> burn
> 
> ```
> { "attribute": "type", "values": ["fire"], "multiplier": 0.0 } eg damage 20 * 0.0 -> 0 HP damage
> { "attribute": "type", "values": ["water"], "multiplier": 0.5 }
> { "attribute": "type", "values": ["metal", "earth"], "multiplier": 0.75 }
> { "attribute": "type", "values": ["wood"], "multiplier": 1.5 }
> { "attribute": "type", "values": ["frost"], "multiplier": 2.0 } -> for the 13 types
> ```
> 
> For example, if a fire monster takes burn damage, it gets 0 damage (immune). But if a frost monster takes burn damage, it gets double damage.
> 
> The questions::
> 
>     * do you agree with the damage multipliers I set?

I think 0.75 and 1.5 might be a bit fiddly to keep track of for players - they're pretty close to 1 or 2 respectively. 

How about making Fire and Water immune, and Wood and Frost take double? 

>     * should we show a message when a monster is immune (0.0x damage)? We could do similar messages for poison and lifesteal. Examples below.
> 
> 
> example message for burn: "XXX is immune to YYY, it loses the condition" "XXX's fiery nature makes it immune to burn damage, the condition is removed."
> 
> example message for poison: "XXX is immune to poison, it loses the condition." "XXX's natural defenses protect it from poison, the condition is removed."
> 
> example message for lifeleech: "XXX is immune to lifeleech, it loses the condition." "XXX's resilient spirit makes it immune to lifesteal, the condition is removed."

Yeah, I think that's a good idea - show the message and remove the condition. I prefer the simpler "XXX is immune to YYY" version, but don't have a strong preference either way.

JaskRendix:
> I think 0.75 and 1.5 might be a bit fiddly to keep track of for players - they're pretty close to 1 or 2 respectively. How about making Fire and Water immune, and Wood and Frost take double?

"Fiddly". You're saying that using 0.75 and 1.5 is too complicated for players to keep track of? Seriously? We've got a range from 0 to 2, but we can only use 0, 1, or 2? Doesn't make sense to me, but ok, I'll use it for new things. Look, I'll set:
- 0 for Fire - immunity (burn)
- 2 for Frost (burn)
- 0 for Venom - immunity (poison)

JaskRendix:
@Sanglorian  just let me know if you want to apply to other conditions, like positive or negative multipliers (for things like healing - or damaging or making immune - certain types or targets or something else). If not, I can go ahead and set it to ready for review (we can always add those values later if needed).

Sanglorian:
> @Sanglorian just let me know if you want to apply to other conditions, like positive or negative multipliers (for things like healing - or damaging or making immune - certain types or targets or something else). If not, I can go ahead and set it to ready for review (we can always add those values later if needed).

Thanks! I'd recommend setting it ready for review - as you say, can add other values later.

HippasusTwo:
I don't think half values like adding or removing .5 is too fiddly. If someone is digging around in the code to mod it, they probably grok 'slightly more or less than one' imo. '1.5' is STAB same-type bonus in Pokemon, and four year olds play Pokemon.
But if you've already made the changes, or if that's what everyone wants, then that's fine.

JaskRendix:
> But if you've already made the changes, or if that's what everyone wants, then that's fine.

@HippasusTwo that's just the first step. We can start a new conversation to talk about how we might adjust conditions related to elements, tags, or anything else.

HippasusTwo:
It might not be a bad idea to have a special tag that gives immunity to a certain element's side effects. There are many cases in nature of animals having poison/venom resistance, for instance, badgers and snake eagles famously go after snakes, and some animals eat poisonous organisms (Monarch butterflies eat toxic milkweed).

kerizane (approved):

PR #2584, opened 2025-02-07
removal ElementType enum

JaskRendix:
This PR makes the element system more flexible and moddable by removing the hardcoded `ElementType` enum and instead using a string to represent the element type. This allows modders to easily add new elements to the game without having to modify the underlying code. Additionally, a couple of new unit tests have been added to `test_evolution.py` to cover the Element class, which was previously missing, and will now ensure that all elements are properly tested.


kerizane (dismissed):

JaskRendix:
fixed the conflict (import)

kerizane (approved):

PR #2588, opened 2025-02-09
add LifeShareEffect (technique) 

JaskRendix:
PR adds a `LifeShareEffect` to the game, which allows two monsters to share their current HP. The effect can be configured to share HP in one of three ways: 
- Weighted average: The weighted average takes into account the maximum HP of each monster.
- Geometric mean: The geometric mean gives more weight to the monster with lower HP.
- Simple average: This method simply takes the average of the two HP values.

First, let's define the HP values of the monsters:
- Monster A: 50/100
- Monster B: 180/200

Here are the results:
* Simple average: 115 -> Monster A 100/100 + Monster B 115/200
* Weighted average: 136.67 -> Monster A 100/100 + Monster B 137/200
* Geometric mean: 94.87 -> Monster A 95/100 + Monster B 95/200

The direction of the sharing can also be specified, either from the user to the target or from the target to the user.
This effect can be used to create interesting mechanics, such as HP-sharing abilities or symbiotic relationships between monsters.

@Sanglorian @HippasusTwo


Sanglorian:
Thanks @JaskRendix - I think simple average is the best way to go for this, to keep it easily understandable to players.

JaskRendix:
@Sanglorian the modder can choose the formula.

HippasusTwo:
Very neat!

kerizane (approved):

PR #2589, opened 2025-02-09
add LifeSwapEffect (technique) 

JaskRendix:
PR adds a `life_swap` effect to the game, which swaps the current HP amounts of the two monsters involved in a technique. The direction of the swap is determined by the `direction` attribute, which can be either `user_to_target` or `target_to_user`. The effect ensures that the HP values do not exceed the maximum HP of each monster after the swap.

Example:
 * Monster A has 42/69 HP.
 * Monster B has 80/150 HP.

After HP Swap:
 * Monster A has 69/69 HP.
 * Monster B has 42/150 HP.
 
@Sanglorian @HippasusTwo


Sanglorian:
Cool! This could be a fun effect for a few techniques. 

Are you thinking it would be used on one ally with lots of HP, to effectively heal another ally with low HP? Or as a kind of attack, with a creature using it on their opponent? (Or I guess you could have different techniques using it in different ways.)

HippasusTwo:
How does changing directions matter? This swaps so monster A has Monster's B HP, and Monster B has monster A's HP, yes?

JaskRendix:
> How does changing directions matter? This swaps so monster A has Monster's B HP, and Monster B has monster A's HP, yes?

@HippasusTwo you're right, I'm going to update

kerizane (approved):

PR #2591, opened 2025-02-09
add LifeGiftEffect(CondEffect) (condition)

JaskRendix:
This PR adds a new status effect called 'LifeGiftEffect' which allows a monster to transfer some of its HP to its target, acting differently than a traditional lifeleech effect. Unlike lifeleech, LifeGift heals the opponent while damaging the user, introducing a new layer of strategy and complexity to combat interactions. I have reused the same icon, just recolored.

To support this new effect, two new combat state messages have been added:
msgid "combat_state_lifegift_get"
msgstr "{target}'s energy is being restored."
msgid "combat_state_lifegift_end"
msgstr "{target} has been infused with lifeforce."

The introduction of LifeGiftEffect opens up new possibilities for developing unique mechanics and moves, such as sacrificial healing moves or abilities that reward selfless behavior. This can lead to the creation of new and interesting moves, such as a "Healing Sacrifice" or "Lifeforce Transfer".

@Sanglorian


JaskRendix:
@Sanglorian what do you think of this?

Sanglorian:
Sounds good! Is there an option to choose an ally as the target, in two v two matches?

kerizane (approved):

PR #2594, opened 2025-02-13
add StepDamageEffect & StepHealingEffect (technique) 

JaskRendix:
requires #2563 

PR adds two new effects that allow techniques to scale their `power` and `healing_power` based on the number of steps taken by the monster, making them grow stronger as the game progresses. The effects, one for damaging and one for healing, utilize a logarithmic formula to calculate the scaled power.

The formula used is:
new_power = healing_factor * \log(1 + user.steps / scaling_constant)

where the `healing_factor` and `scaling_constant` parameters can be adjusted to fine-tune the effect's behavior.

Below, you can find an image illustrating the growth of the power value based on the number of steps taken. 
https://github.com/user-attachments/assets/b164c557-a1b6-4578-aea4-931d329027b9

The graph shows nine possible combinations of `healing_factor` and `scaling_constant` values, with the central combination being the recommended one (factor 1, constant 1500 steps).

The red points on the graph represent the calculated power values at different step counts, demonstrating how the power grows as the monster walks with the trainer. As shown, the power value can exceed 3.0 at high step counts, but considering the amount of walking required to reach these values, this may not be an issue. However, we can discuss capping the power value at 3.0 if necessary.

From analysis of various savegames, the highest step count observed in Route 3 is approximately 3.5k / 4k steps.

@Sanglorian @HippasusTwo

### DIFF
--- a/tuxemon/technique/effects/step_damage.py
+++ b/tuxemon/technique/effects/step_damage.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from tuxemon.combat import get_target_monsters
+from tuxemon.formula import simple_damage_calculate
+from tuxemon.technique.techeffect import TechEffect, TechEffectResult
+
+if TYPE_CHECKING:
+    from tuxemon.monster import Monster
+    from tuxemon.technique.technique import Technique
+
+
+@dataclass
+class StepDamageEffect(TechEffect):
+    """
+    This effect calculates damage to the target based on the number of steps
+    taken by the monster. The damage is scaled using a specified formula, which
+    is logarithmic.
+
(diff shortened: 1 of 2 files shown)


HippasusTwo:
I guess a some good move names could be something like 'Growing Power' and 'Affection'.
edit:
Also, I am strangely relieved to see it is not just me that is having issues with Styles. I was half-convinced Python just hated me.

JaskRendix:
@Sanglorian is it ok for you? Any suggestions on this effect?

Sanglorian:
Thanks - it's an interesting idea! My thoughts go to: 
* It could reflect friendship (since the Tuxemon and owner have travelled together a long time)
* It could be fun for a reserved technique, since it's a bit of an unusual effect 
* There's a few travel based mons that could use this effect

Fellow Traveller (heal one target based on step count)
Fernweh (damage one target based on step count) - reserved tech for Chenipode and Exapode?

HippasusTwo:
I like the idea of Chenipode and Exapode having some travel related techniques. Another 'mon that could use such a thing is maybe one of the dolphin (Dolfin) evos, because dolphins/whales tend to be highly migratory, and also have strong relationships, so 'Fellow Traveler' might make sense.

kerizane (approved):

PR #2601, opened 2025-02-18
Added the options menu to the starting menu

Kent985:
The purpose of this PR is to add the options menu to the start menu.

I removed (only for the start menu) the options that are tied to the music, the temperature and the hemisphere, since these can't be modified if there is not a player character yet. 

I also had to remove `pop_state()` in the function `process_event` from the control state, because this would lead to a black screen if the options menu was closed in the main menu. This now makes it that if the options menu is closed in game, the player is led back to the menu right before it instead of bringing him back to the game.

### DIFF
--- a/tuxemon/states/control/__init__.py
+++ b/tuxemon/states/control/__init__.py
@@ -156,6 +156,8 @@ class ControlState(PygameMenuState):
         theme = get_theme()
         theme.scrollarea_position = locals.POSITION_EAST
         theme.widget_alignment = locals.ALIGN_CENTER
+        self.main_menu = "main_menu" in kwargs and kwargs["main_menu"]
+        kwargs.pop("main_menu", None)
         super().__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


JaskRendix:
Instead of getting rid of the "pop_state" entirely, let's just add a condition to it. This way, it will only activate when we're not on the main menu. Think of it like adding a safety net to make sure it only triggers when it's supposed to.
```
    def process_event(self, event: PlayerInput) -> Optional[PlayerInput]:
        if event.button == buttons.BACK:
            self.reload_controls()
            if not self.main_menu:
                self.client.pop_state()
```
It looks like the style test is having some issues. No worries, you can run the command `tox -eftm` . This should get things back on track.

Kent985:
Thank you for your feedback. I've added the condition and ran the command for the style.

JaskRendix (approved):

kerizane (approved):

PR #2607, opened 2025-02-19
add ForesightEffect (technique) 

JaskRendix:
PR introduces a new effect called "ForesightEffect" that allows players to plan ahead and execute a predetermined attack after a specified number of turns. The attack can be a specific technique (`"foresight 2"`).

For example, let's say it's turn 3 and the user (monster A) uses a move with the `foresight` effect, specified as `"foresight 2"`. This means that the user will use the "foresight" attack at turn 5 (3 + 2) with a power of 2.0 (equal to the turns)

Here's how it would play out:
* At turn 5:
	+ A attacks B (as part of the normal turn sequence)
	+ B attacks A (as part of the normal turn sequence)
	+ A then executes the predetermined "earthquake" attack (at the end of the turn)

To implement this effect, I had to make some changes to the way pending actions are handled. Specifically, I decided to move the logic for managing pending actions inside the `ActionQueue` class, which now includes a method for adding pending actions to the queue.

@Sanglorian @HippasusTwo


Sanglorian:
This sounds like a lot of fun. 
Possible techniques: Storm Warning (Lightning), Bloody Mary (Shadow).
I think they should be quite powerful since the player is giving up a technique in the short term

JaskRendix:
@Sanglorian Thanks! I'm glad you think it sounds fun! I agree that techniques like Storm Warning and Bloody Mary should pack a punch since players are planning ahead. What about a Boomerang technique?

ideas:
- **Boomerang**; The user throws a boomerang that travels out to hit the target and then returns to the user, allowing for a second attack.
- **Storm Warning**: The skies darken as the user channels the power of a brewing storm. With a foreboding crack of thunder, the user prepares to unleash a devastating lightning attack.
- **Bloody Mary**: A chilling aura envelops the battlefield as the user invokes the dark spirit of Bloody Mary. With a whisper of dread, the user prepares to unleash a haunting attack that strikes fear into the hearts of their enemies.
- **Self Destruct Sequence (SSS)**: The user initiates a countdown sequence, preparing for a powerful self-detonation that will unleash a massive explosion. (only available to mechanical or robot monsters)

example storm_warning.json
```
  "accuracy": 1,
  "effects": [
    "foresight 5"
    "damage"
  ],
  "power": 0.0
  "recharge": 5,,
```
Monster A uses Storm Warning no damage gets inflicted.
... after 5 turns ....
Monster A uses Storm Warning and the damage is calculated using the power of 5.0 (foresight 5 = 5 power)

HippasusTwo:
I think there should be a psychic move, maybe 'Seer', that uses this.

kerizane (approved):

PR #2619, opened 2025-02-25
Add Cancellation Mechanism to Event Actions and Engine

JaskRendix:
PR introduces a way to cancel both individual actions and entire events, making the event system much more flexible and robust.

Event Actions can now be cancelled:
   * added a "cancelled" flag to each action
   * we can now call a "cancel()" method on an action to stop it
   * the action's code now checks this flag before doing anything, so cancelled actions don't run
   * added better error handling and logging, so if something goes wrong, we know about it

Events can now be cancelled:
   * added a "cancelled" flag to each running event
   * we can now cancel a specific event or all events at once
   * the engine now checks if an event is cancelled before running its actions, so cancelled events are skipped
   * the engine also checks if the action is cancelled before updating it or executing it

We can now stop actions and events whenever we need to, like when a player moves away from something or a game state changes. Cancelling events and actions stops unnecessary processing, which can improve performance. This lets us create more complex and dynamic game behaviors, like interrupting cutscenes or cancelling tasks. Checking for cancelled actions and events prevents errors that would occur from attempting to execute a cancelled object. The added logging makes it much easier to track what's happening and find problems.

Basically, this update makes the event system more reliable and gives us more tools to create interesting and responsive gameplay.

### DIFF
--- a/tuxemon/event/eventaction.py
+++ b/tuxemon/event/eventaction.py
@@ -84,6 +84,7 @@ class EventAction(ABC):
     session: Session = field(init=False, repr=False)
     _done: bool = field(default=False, init=False)
     _skip: bool = field(default=False, init=False)
+    cancelled: bool = field(default=False, init=False)
 
     def __post_init__(self) -> None:
         self.session = local_session
@@ -96,6 +97,9 @@ class EventAction(ABC):
         Context Protocol.
 
         """
+        if self.cancelled:
+            logger.warning("Event is cancelled, not starting")
+            return
         self.start()
 
     def __exit__(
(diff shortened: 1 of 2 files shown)


HippasusTwo:
I'd really like for path-finding or other things like moves to automatically stop crashing the entire game if something goes even the tiniest bit wrong (like an npc tries to pathfind to you but can't because another npc wandered into that spot - that was an issue I had awhile back, or running into an unhandled move combo case like with Riposte).

It sounds like this can do that, but looking over the code this merely adds the possibility, it doesn't actually implement in it any auto-cancelation on errors, correct?

JaskRendix:
@HippasusTwo yeah buddy, "merely" adding things. I'm not going to bother with the complaint about the pathfinding crash. If it was still a problem, some issue would have come up by now. But thanks for remembering about it.

kerizane (approved):

PR #2626, opened 2025-02-27
Improving Spyder Intro

JaskRendix:
PR improves the Spyder intro by making the characters more relatable. Here's what's changed:
* the conversations between trainers and Dante help show their personalities, making them feel more real
* gives each trainer a unique personality makes the world feel more alive
* makes feel Dante more like a person and less like a robot

@Sanglorian is it ok? Suggestions?

https://github.com/user-attachments/assets/74ee6294-980e-4d2d-9b4b-9da40e770022

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -8317,6 +8317,66 @@ msgstr "Please, now return the form to the Shop Assistant, so we can give you th
 msgid "spyder_intro_shopkeeper4"
 msgstr "I'm sorry, you are not a Gold member. This offer is for Gold members only."
 
+msgid "spyder_trainer1_dante1"
+msgstr "I expect only the best. I've heard these tuxemon are quite exceptional."
+
+msgid "spyder_trainer1_dante2"
+msgstr "Indeed, these are exceptional tuxemon. I'm sure you'll find it to your liking."
+
+msgid "spyder_trainer1_dante3"
+msgstr "Yes, yes, very good. Now, let's see if it lives up to the hype."
+
+msgid "spyder_trainer1_dante4"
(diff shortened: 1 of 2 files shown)


HippasusTwo:
I'd make it shorter: when I'm doing bug testing I ended up restarting the game SO MANY times, which makes long cut unskippable scenes more agonizing (I largely avoid playing Xeno for this reason, it takes forever to get to a monster which is amusing the first time but not the tenth). Giving Billie some personality is good, because she shows up all the time! But you don't even fight any of these side characters in the first route (though perhaps you should - even after I added two trainers who want to fight you in town, the time between the first and second Billie fight feels super short and training that doesn't involve just throwing yourself at wild enemies would be welcome).

Or, alternatively instead of shortening it to just improving Billie, you could implement a skip, like 'zone out while other people talk?' question would be pretty cool imo.

JaskRendix:
@HippasusTwo every modder does things their own way, so if players are impatient, that's on them. I'm not going to change it to fit someone else's preferences - it's either good enough or it's not.

Sanglorian:
> @Sanglorian is it ok? Suggestions?
> test.mp4

Thanks @JaskRendix , I like the idea of giving the others more personality, but I think it should be up to the player if they want to engage or just get straight into the game. 

What if after the intro scene, the player could wander around the room - then they could choose to talk to some or all of the trainers, or go straight to Dante. Whenever they go to Dante, we could have Billie push in, and then everyone else could follow Billie, still leaving the player to be the last. 

Just thinking about this approach, then we could also have capture devices out on display for the five starters, and the player could approach each one in turn to see the preview for them again. 

If we do add more personality to the trainers, it'd be fun to have them show up later in the game. They wouldn't have to appear multiple times like Billie, but each one could appear as a regular trainer somewhere in the world.

JaskRendix:
@Sanglorian ok

PR #2639, opened 2025-03-06
Refactor Taste

JaskRendix:
PR refactors the **Taste** feature into a JSON format, similar to how `Shape` and `Element` are structured. This change allows modders to easily add or remove tastes as needed, rather than relying on a hardcoded model that was specific to a single campaign. The previous approach, suggested by @Sanglorian, was too rigid and did not accommodate the diverse requirements of different campaigns. 

Here’s an example of the new JSON structure:
```json
{
  "slug": "flakey",
  "name": "taste_flakey",
  "taste_type": "cold",
  "modifiers": [
    {
      "attribute": "stat",
      "values": ["ranged"],
      "multiplier": 0.9
    }
  ]
}
```
This new format allows for multiple modifiers and provides a broader range for the multiplier, which now spans from 0 to 2. This change enables more intuitive adjustments, allowing for the complete reduction or doubling of specific stats. Additionally, I have updated a series of unit tests to ensure they align with these changes, and I have also modified a couple of methods to support the new structure.

### DIFF
--- a/mods/tuxemon/db/taste/taste.json
+++ b/mods/tuxemon/db/taste/taste.json
@@ -1,0 +1,122 @@
+[
+  {
+    "slug": "mild",
+    "name": "taste_mild",
+    "taste_type": "cold",
+    "modifiers": [
+      {
+        "attribute": "stat",
+        "values": ["speed"],
+        "multiplier": 0.9
+      }
+    ]
+  },
+  {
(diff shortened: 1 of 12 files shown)


kerizane (approved):

Sanglorian:
Thanks @JaskRendix !

PR #2643, opened 2025-03-09
bump requirements

JaskRendix:
PR updates the requirements by bumping the version.


JaskRendix:
Hey @Murmele, I was trying to set up testing for version 3.13, but I'm running into issues. It keeps flagging me that I need additional packages like maturin (flagged by pydantic) or  mesonpy (meson-python flagged by pygame-ce), etc. Do you have any idea how can we handle this?

Murmele:
did you update the requirements.json with the script (https://github.com/flathub/org.tuxemon.Tuxemon/blob/master/updateRequirements.sh)?

JaskRendix:
@Murmele yes, I have used the last version of flatpak-pip-generator (update: check flatpak-pip-generator, there is a PR to update the script)

https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator

pygame-ce toml
```
[build-system]
requires = [
  "meson-python<=0.17.1",
  "meson<=1.7.0",
  "ninja<=1.12.1",
  "cython<=3.0.11",
  "sphinx<=8.1.3",
  "sphinx-autoapi<=3.3.2",
]
build-backend = 'mesonpy'
```

Murmele:
Maybe you have to add mesonpy to the flatpak manifest manually?

https://github.com/flathub/org.sugarlabs.MusicKeyboard/pull/16

They added it as dep here:
https://github.com/flathub/flathub/pull/5542/files

JaskRendix:
@Murmele ok, I'm going to add it separately

JaskRendix:
@Murmele I have tried to add meson-python as well as maturin (required by pydantic), but maturin requires setuptools-rust, but then it fails because there is no wheel

JaskRendix:
@Murmele I got that `maturin `problem sorted. Turns out, that fancy script for updating the requirements? Totally wrong. I had to ditch it, go back to the old way, and fix each link by hand. Now, I've just got this one last little issue to fix.

Murmele:
> @Murmele I got that `maturin `problem sorted. Turns out, that fancy script for updating the requirements? Totally wrong. I had to ditch it, go back to the old way, and fix each link by hand. Now, I've just got this one last little issue to fix.

Packaging python applications into flatpak is an anyoing job. It is so unclear what is required ...

JaskRendix:
@Murmele  we did it

The problem was with version p310; I needed to switch it to p312 as well as to specify a particular version, which was pydantic-core 2.27. It would have been impossible to do it using just the tools available. I'm going to test a few more things just out of curiosity, maybe we can avoid python-meson.

Edit: Done! By using `only-arches` (x86_64 and aarch64) for pygame-ce, we can avoid python-meson. Tomorrow, I’ll remove the files and get everything ready for review.

bitcraft:
maybe a minority opinion here, but python 3.9 is still supported until the end of this year.  im glad we got some testing for 3.13.  is there any way to keep python 3.9 support and also support 3.13?  im also fine with just using 3.12 as our "newest supported version", since 3.13 is still only a few months old and most distributions are still shipping 3.11 and 3.12

JaskRendix:
@bitcraft no problem, removed the 3.13, we can easily add it whenever we want it

kerizane (approved):

PR #2644, opened 2025-03-09
fix various

JaskRendix:
PR anticipates that the fixes for type hints already made in #2643 will be necessary due to the new dependencies.

### DIFF
--- a/mods/tuxemon/db/technique/spyderbite.json
+++ b/mods/tuxemon/db/technique/spyderbite.json
@@ -8,7 +8,6 @@
     }
   ],
   "flip_axes": "",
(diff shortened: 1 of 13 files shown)


mods/tuxemon/db/technique/spyderbite.json
```diff
@@ -8,7 +8,6 @@
     }
   ],
   "flip_axes": "",
-  "icon": "gfx/ui/icons/status/icon_spyderbite.png",
```
kerizane, line 11: Why?

kerizane (approved):

PR #2647, opened 2025-03-11
New Linting Workflow and Tox Command for Flake8

JaskRendix:
PR creates a new workflow specifically for linting. This new workflow can be triggered manually whenever we want to check our code for linting issues without affecting the main build process.

Additionally, I've integrated the linting step into our `tox.ini` file. This means that you can now run the lint checks as part of your local development process by simply using the `tox -elint` command.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,29 @@
+name: Lint
+
+on:
+  workflow_dispatch: # Allows manual triggering of the workflow
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


bitcraft:
i feel like the flake8 steps are a good way to check for errors prior to building.  why do we need to remove it?

JaskRendix:
@bitcraft I'm not removing it,  I'm creating a separate yaml that we can use whenever we need to

la-xirsoi:
@JaskRendix i think it'd be more useful to have the linting be automatically triggered on a PR than to have it be manual.

I really like the ability to run the linting locally though, that's super nice

JaskRendix:
@la-xirsoi sounds good, done

la-xirsoi (approved):

PR #2657, opened 2025-03-14
JSONDatabase: Configuration and Mod Handling

JaskRendix:
follows #2656 

PR updates the `JSONDatabase` with configuration management, mod support, error handling, and file format versatility:
- renames `JSONDatabase` into `ModData` (since it loads also YAML)
- breaks `JSONDatabase`  into `ModData`  and `DataLoader`
- introduced a `DatabaseConfig` dataclass for structured database settings, including active mods, mod versions, table exclusions, and mod activation
- added a `load_config` function to load the configuration from a JSON file
- control over mod loading, allowing specification of tables per mod and table exclusions
- added support for loading data from YAML files in addition to JSON
- implemented a custom `EntryNotFoundError` exception for better lookup error management
- added a `reload` method to reload specific database tables
- the lookup function now has an optional table parameter, and defaults to monster
- the path information is now removed from `preloaded` data before models are loaded.

regarding the mods, for now there is only "tuxemon", but in the future it'll be "spyder", "xero", etc.


Sanglorian:
This is very exciting! Mod support would be a huge leap forward so I'm glad we're getting closer.

bitcraft (approved):

PR #2659, opened 2025-03-16
Improve Platform and Storage Path Handling

JaskRendix:
PR is connected to #2658 (no conflict):
- replaced os.path with pathlib.Path
- improved the logic for finding the user storage directory, especially on Android, and added support for XDG_DATA_DIRS on linux
- added a helper function `is_android` to check if it's android
- added a helper function `_get_android_storage_paths` to handle android storage paths
- added a try...except block to the init() function to catch and log any errors during Pygame mixer initialization
- the `get_system_storage_dirs` function has been refactored to use the new `_get_android_storage_paths` function

### DIFF
--- a/tuxemon/platform/__init__.py
+++ b/tuxemon/platform/__init__.py
@@ -5,8 +5,9 @@ Put platform specific fixes here
 from __future__ import annotations
 
 import logging
-import os.path
+import os
 from collections.abc import Sequence
+from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Davidliang11292:
Does this mean an Android version is coming soon?

JaskRendix:
@Davidliang11292 no

Davidliang11292:
> @Davidliang11292 no

Thanks

kerizane (approved):

PR #2662, opened 2025-03-17
Change language from options + TuxemonConfig + ControlState

JaskRendix:
PR implements change language from options:
- added the msgids
- added the option font_file, hemisphere, unit_measure, sound_volume and music_volume in the tuxemon.cfg
- simplified ControlState
- added method to audio.py for getting the volume
- the button `mute music` will disappear if the volume is 0

https://github.com/user-attachments/assets/021042fb-96c2-4d95-96a1-d6fbf80bc965

Screenshot_2025-03-18_15-30-53


kerizane:
Good, but I think it would be better to show each language name in it's own language. e.g. Deutsch, Francais.

Not sure how difficult that is though, so I will merge this.

kerizane (approved):

JaskRendix:
@kerizane I'm going to test it

PR #2670, opened 2025-03-20
Held Item Mechanics Added for Monsters

JaskRendix:
PR is all about giving monsters the ability to hold items. Here's what's changed:
- monsters can now carry **one** item at a time, just like they do in the wild.
- when you look at a monster's menu, you'll see what item it's holding, along with some other useful stats. This means you don't have to click on the monster's info to see what it's carrying.
- the monster menu has been updated
- added two new event actions: one that lets you give an item to a monster, and another that lets you take an item away from a monster
- for an item to be held by a monster, it needs to have a special setting called "`holdable`" set to True. Right now, there aren't any items like this, so we'll need to add some and figure out which ones should be added (there are some in Tuxepedia)

If you want to see how it all works, there's a video that explains it in more detail. Encounter mode, new attribute `held_items`, by default empty sequence.
```
    {
      "monster": "cataspike",
      "held_items": ["tea", "mystery_tea", "imperial_tea"], <----if more than 1, then randomly picked
    },
```

https://github.com/user-attachments/assets/11a2f9b9-c1a7-4630-b962-d94d8ba3ab24


bitcraft:
great PR.  thanks for the video and extra information.  very cool feature~

bitcraft (approved):

Sanglorian:
This is a really exciting expansion to the rules, thank you!

Issue #2001, opened 2023-08-30
I want the dialog box to show the name of the NPC


DavidLiang1129:
**Is your feature request related to a problem? Please describe.**
A member of my QQ group asked me to report this to Tuxemon, and I thought it was a good idea to refine the game and familiarize players with NPCS

**Describe the solution you'd like**
I would like to see the name of the protagonist /NPC in the upper left/upper right corner of the dialog box when talking to an important NPC. I think this feature can also enable the top left corner of the dialog box to say "got the item" and the dialog box to say "you found a Tuxeball" or the top left corner of the dialog box to say "Tips" and the dialog box to say "to run, Please press the SHIFT key "and so on

**Describe alternatives you've considered**
We can still label the NPC's name directly in the original dialog.

**Additional context**
Some JRPG games (such as RPG MAKER VX, if it can be called a game) and some monster capture games (such as Nexomon and Pet Kingdom have similar features)

Lalatianalinda:
I would like to propose an improvement regarding the display of player names during dialogues. Currently, in the base.po file, the variable {name} is referenced to display the player's name. However, I am wondering where this variable {name} actually comes from.

If I had to manually add this reference to each dialogue, it would quickly become complicated, as I would need to modify each dialogue one by one, along with each associated translation.

I also noticed that in the npc.py file, the attribute self.name represents the name of the NPC that is used in the dialogues. Wouldn't it be easier to create a function that automatically adds the player's name at the beginning of each dialogue ? This would significantly simplify the process and ensure consistency in the display of names.

Lalatianalinda:
I'm working on displaying npc names in dialogues.

In your opinion, which option would be the most effective?

Manually modifying each dialogue in base.po to include the npc's name, although this could become complicated and prone to errors.

Or using the next_text function in DialogState to automatically add the player's name at the beginning of each dialogue, which could simplify the code and ensure consistency (but what would be the best approach to dynamically retrieve the name of the speaking character?).

JaskRendix:
Manually modifying each dialogue can get pretty messy and prone to errors, but hey, it's up to the **modder** to decide if they want to go that route. And using the `next_text` function to add the player's name (which is shown with `${{name}}`) isn't what you're looking for since you want the NPC's name.

[...]

PR #2671, opened 2025-03-20
Adding NPC Name at the beginning of Translated Dialogues

Lalatianalinda:
**Description**
This PR modifies the `TranslatedDialogAction` class to automatically add the NPC's name at the beginning of the dialogue. The name is retrieved from the first element of the `raw_parameters` parameter, then translated before being added to the dialogue text.

**Changes Made**
- Modified the `start()` method to extract the NPC's name from the first segment of the `raw_parameters`
- Used `T.translate()` to translate the NPC slug
- Added the NPC's name at the beginning of the first dialogue message in the format `"{npc_name}: {text}"`

**Current Limitations and Future Improvements**
The method of extracting the NPC's name is still basic and has certain limitations:
- The current implementation assumes that the first segment of `raw_parameters` (separated by underscores) is always the NPC slug, which may not be the case for all dialogues
- There is no special handling for nameless NPCs

**Next Steps**
- Improve the detection of the NPC slug
- Add handling for nameless NPCs


JaskRendix:
As suggested here: https://github.com/Tuxemon/Tuxemon/issues/2001#issuecomment-2688264387 is  better to create a new event action, because you need to pass the npc_slug in the event action. The name is the player's name, not the NPC name.

```
    name = "new_translate_dialog_action"
    npc_slug: str
    raw_parameters: str
    avatar: Union[int, str, None] = None
    position: Optional[str] = None
    style: Optional[str] = None

    def start(self) -> None:
        ....
```
so in the map it'll be
`<property name="act1" value="new_translate_dialog_action npc_slug,raw_parameters"/>`

Davidliang11292:
Thank you very much. My original account @DavidLiang1129 has been lost. If you have any questions, please contact my new account (this one).

PR #2674, opened 2025-03-21
MoneyController + fix modify_money + transfer_money

JaskRendix:
PR creates the MoneyController class. It encapsulates the MoneyManager and its related logic and it provides methods for getting and setting the money state, separating money management from the NPC class's core logic. This makes the NPC class less cluttered and allows for easier modification of money-related behavior. It fixes a bug in `modify_money` (still hardcoded on the self.player instead of the character) as well as `transfer_money` (it was wrongly transfer the money to bank instead  to PC)


bitcraft:
looks good.  always a good idea to break related logic into reusable classes.

bitcraft (approved):

PR #2678, opened 2025-03-21
Reload the config after changes ingame

### LINKED ISSUE #2677: Changing the volume of the music is not persistent
When entering a wild encounter after changing the volume of the music in the menu, the volume is effectively reset to whatever value it was when launching the game. The value of the input in the menu doesn't change.

**To Reproduce**
1. Launch the game
2. Open the options menu and set the music volume to 1 if it's already mute, to 0 otherwise
3. Enter a wild encounter in the grass
4. The music plays at the volume you had before changing it

 - OS: Linux Mint 21.3
 - Python Version 3.10.15

**Additional context**
I've seen in the code that there have been recent changes on the way the config works, but I have the same issue before PR 2661.
[EDIT] I don't know if it's relevant, but I run a Spyder and the Cathedral campaign.

Divarrek:
PR fixes #2677 by:

- Making sure the config is actually reloaded after changes (the call to `reload_config` used to update the ConfigParser but not the values that had already been read)
- Changing the direct calls to `current_music.play()` into calls to the `play_music` event in order to have the volume use `client.config.music_volume` instead of falling back to the outdated `prepare.CONFIG.music_volume`

### DIFF
--- a/tuxemon/config.py
+++ b/tuxemon/config.py
@@ -30,71 +30,14 @@ class TuxemonConfig:
 
         # update with customized values
         if config_path:
-            cfg.read(config_path)
+            self.cfg.read(config_path)
 
-        # [display]
-        resolution_x = cfg.getint("display", "resolution_x")
-        resolution_y = cfg.getint("display", "resolution_y")
-        self.resolution = resolution_x, resolution_y
-        self.splash = cfg.getboolean("display", "splash")
-        self.fullscreen = cfg.getboolean("display", "fullscreen")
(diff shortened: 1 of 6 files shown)


JaskRendix (approved):

bitcraft (dismissed):

JaskRendix:
It looks like the style test is having some issues. No worries, you can run the command `tox -eftm` . This should get things back on track.

bitcraft:
looks good, but please check the failing test

kerizane (approved):

PR #2684, opened 2025-03-23
Expanding Shop Features: Reselling

JaskRendix:
Previously, you could only resell items that the shop was already selling. For instance, if the shop sold potions, you could only resell potions. Now, with this change, you can resell items like Tuxeballs or potions, even if the shop doesn’t sell them.

Here's how it works:
- A new parameter, `resale_multiplier`, has been added to the economy system. This is a float value that defines the percentage of the original price at which an item can be resold. By default, this is set to 0.5 (50%) across all JSON files in the economy folder. For example, if a Tuxeball costs 100, you can resell it for 50.
- You can now change the shop's background by specifying a new path in the JSON files.
- A new flag, `resellable`, has been introduced for items. This is set to `false` by default, allowing modders to decide whether certain items can be resold or not. For example, some items can only be bought but not resold.

This is just the beginning, other improvements will follow.

### DIFF
--- a/mods/tuxemon/db/economy/cotton_scoop.json
+++ b/mods/tuxemon/db/economy/cotton_scoop.json
@@ -1,5 +1,7 @@
 {
   "slug": "cotton_scoop",
+  "resale_multiplier": 0.5,
+  "background": "gfx/ui/item/item_menu_bg.png",
   "monsters": [],
   "items": [
     {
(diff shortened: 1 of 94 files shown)


la-xirsoi (dismissed):

Sanglorian:
Thank you, this is a great addition!

kerizane (approved):

PR #2695, opened 2025-03-30
Enabling Cross-Mod Dependencies

JaskRendix:
PR introduces an enhancement to the database configuration system: `mod_dependencies`. This addition aims to improve modularity and resource sharing between mods while eliminating unnecessary data duplication. Key points:
- mods can now declare dependencies on other mods. When a mod is loaded, its dependencies are automatically resolved and their corresponding data tables are made available to ensure integration
- by leveraging dependencies, mods can reuse tables from other mods without duplicating data
- a new `mod_dependencies` property has been added to the `DatabaseConfig` class. It allows mods to list other mods they depend on. Dependencies are resolved recursively, ensuring that all required tables are preloaded before the mod itself is loaded
- "mods", "db", ".json", ".yaml": these hardcoded strings have been moved into `DatabaseConfig` dataclass and it includes "monster" default table
- the `model_map` will be loaded from the  configuration file and this involve specifying the table names and corresponding model classes in the configuration (temporary, the idea is to create a system like PluginDiscovery, maybe even involved it into the ModData loader, but it requires more testing and attempts)
- `reload` method is improved by clearing both `preloaded` and `database` data, integrating mod path handling for data loading, adding logging for traceability, error handling to capture and report issues, and simplifying the workflow by focusing on table-specific reloading

```
{
    "active_mods": ["mod_a", "mod_b"],
    "mod_versions": {
        "mod_a": "1.0.0",
        "mod_b": "1.0.0"
    },
    "mod_activation": {
        "mod_a": true,
        "mod_b": true
    },
    "mod_tables": {
        "mod_a": ["npc", "monster"],
        "mod_b": ["npc"]
    },
    "mod_dependencies": {
        "mod_b": ["mod_a"]  <--- mod_b will get monster from mod_a
    }
}
```

### DIFF
--- a/mods/db_config.json
+++ b/mods/db_config.json
@@ -1,35 +1,64 @@
 {
-    "active_mods": [
-      "tuxemon"
-    ],
-    "mod_versions": {
-      "tuxemon": "1.0.0"
-    },
-    "mod_activation": {
-      "tuxemon": true
-    },
-    "mod_tables": {
-      "tuxemon": [
-        "item",
-        "monster",
-        "npc",
-        "status",
-        "technique",
-        "encounter",
-        "dialogue",
-        "environment",
-        "sounds",
-        "music",
-        "animation",
-        "economy",
-        "element",
-        "taste",
-        "shape",
-        "template",
-        "mission",
-        "terrain",
-        "weather"
-      ]
-    },
-    "mod_table_exclusions": {}
-  }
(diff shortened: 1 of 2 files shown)


Sanglorian:
This sounds so useful! Thank you

la-xirsoi (approved):

PR #2720, opened 2025-04-10
FrameManager + Enum (animation state)

JaskRendix:
PR introduces a new class, `FrameManager` in `surfanim.py` and an `Enum` for the animation state. Now the `SurfaceAnimation` class focuses on animation logic, while the `FrameManager` class is dedicated to frame management and image-related tasks.  The original implementation (animation state: playing, etc.) used string literals for the state, which could lead to typos and other errors.

### DIFF
--- a/tests/tuxemon/test_surfanim.py
+++ b/tests/tuxemon/test_surfanim.py
@@ -3,11 +3,10 @@
 import unittest
 
 import pygame
+from pygame.surface import Surface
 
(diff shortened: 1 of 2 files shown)


la-xirsoi (approved):
Great work! Tuxemon's code base has definitely suffered from Primitive Obsession for a long time and this is a great step in the right direction!

PR #2728, opened 2025-04-12
Enhancing AI Decision-Making (Combat)

### LINKED ISSUE #8: Implement Smarter Combat AI
Right now we have a basic AI that just selects the first technique its active creature has. We need to create various levels of AI that make better decisions.

JaskRendix:
PR adds several improvements to AI decision-making processes (fixes #8 )
- introduced `AIConfigLoader` to load and cache configurations for techniques and items using YAML files
- `ai_opponent.yaml` file is primarily used for double combat scenarios. It evaluates the monsters in the player's party and determines the optimal target for the AI opponent (AI vs Human), based on the parameters defined within the file. It allows the possibility to specify the character's slug (opponent) against which the player is fighting.
- `ai_items.yaml` file enables modders to configure the use of specific items for NPC trainers. The functionality depends on whether the trainers possess items in their inventory. Currently, the only available item is the potion
- `ai_trainers.yaml` file empowers modders to create tailored strategies for specific NPC trainers by defining unique slugs. It also offers flexibility in customizing techniques, enabling modders to expand or limit the set of techniques available to individual characters
- `ai_techniques.yaml` file empowers modders to fine-tune the decision-making process for monsters, whether they are AI opponent trainers or wild monsters. By leveraging a range of customizable parameters, this file allows modders to define how techniques are selected, ensuring that monsters can make more strategic choices based on specific conditions or priorities

No changes, it'll keep picking up the technique randomly.
For now, the file is located in the mod folder, but it’ll be moved into each mod’s subfolder, so each scenario will use different rules.

### DIFF
--- a/mods/ai_items.yaml
+++ b/mods/ai_items.yaml
@@ -1,0 +1,29 @@
+# General Rules for Item Evaluation
+items:
+  item_slug:
+    # The upper threshold for the item's use.
+    # The item will not be used if the current health is ABOVE this percentage of max HP.
+    # Optional: Can be omitted if no upper limit is needed.
+    hp_below: 0.3
+    
+    # The lower threshold for the item's use.
+    # The item will not be used if the current health is BELOW this percentage of max HP.
+    # Optional: Can be omitted if no lower limit is needed.
+    hp_above: 0.8
+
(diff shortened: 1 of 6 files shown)


la-xirsoi (approved):
great work! especially documenting in the yaml files what all the properties are and what they mean!

PR #2765, opened 2025-04-26
Centralize collision methods (eg. `add_collision`, `remove_collision`, etc) in WorldState

JaskRendix:
PR transfers both methods from the Entity class to WorldState, along with the EventAction methods (`add_collision` and `remove_collision`), laying the groundwork for the development of a dedicated Collision class.


la-xirsoi (approved):
beautiful

PR #2767, opened 2025-04-28
Refactor StartState and Event-Driven Game Initialization with `start_game` EventAction

JaskRendix:
PR introduces the following changes to the `StartState` class and integrates a new event action called `start_game` to streamline game initialization.

### DIFF
--- a/tuxemon/event/actions/start_game.py
+++ b/tuxemon/event/actions/start_game.py
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: GPL-3.0
+# Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import final
+
+from tuxemon import prepare
+from tuxemon.event.eventaction import EventAction
+from tuxemon.time_handler import today_ordinal
+
+logger = logging.getLogger(__name__)
+
+
+@final
+@dataclass
+class StartGameAction(EventAction):
+    """
+    Starts a new game session and initializes gameplay components.
+
+    Script usage:
+        .. code-block::
+
+            start_game <map_name>,<mod>
+
+    Script parameters:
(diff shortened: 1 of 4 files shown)


tuxemon/main.py
```diff
@@ -95,10 +95,13 @@ def configure_game_states(
         client.push_state("SplashState", parent=client.state_manager)
         client.push_state("FadeInTransition")
 
-    if config.skip_titlescreen and config.mods and len(config.mods) == 1:
-        destination = f"{prepare.STARTING_MAP}{config.mods[0]}.tmx"
-        map_name = prepare.fetch("maps", destination)
-        client.push_state("WorldState", map_name=map_name)
+    if config.skip_titlescreen and config.mods:
+        if len(config.mods) == 1:
```
la-xirsoi, line 99: what should happen if more than one mod is enabled?

tuxemon/event/actions/start_game.py
la-xirsoi, line 1: Do you think there's a use for this in scripting?

To be clear, I'm not asking for you to remove this or to put it somewhere else, just want to hear your thoughts on it.

la-xirsoi (commented):
Got a couple questions before merging.

tuxemon/main.py
JaskRendix: after #2768, it could open the ModsChoice State, where it’ll decide whether to start a new game, randomly trigger a mod, or go straight to LoadState to continue a saved game.

tuxemon/event/actions/start_game.py
JaskRendix: I haven't really thought about it before, but there's definitely potential for it in scripting. We could find ways to incorporate it effectively depending on the context and requirements.

tuxemon/main.py
la-xirsoi: makes sense!

tuxemon/event/actions/start_game.py
la-xirsoi: I think we should move this out of the event engine if we don't intend for it to be used by scripts.

PR #2768, opened 2025-04-28
`ModsChoice` Class for Selecting Campaign Mods

JaskRendix:
PR introduces a new `ModsChoice` class to handle mod selection dynamically for starting a new game. This feature improves flexibility and player experience, especially when multiple mods are (will be) available. This doesn't change anything for now, but it brings us closer to separate folders for campaigns, which is an important milestone.


tuxemon/states/start/__init__.py
```diff
@@ -128,3 +135,41 @@ def process_event(self, event: PlayerInput) -> Optional[PlayerInput]:
             return None
         else:
             return super().process_event(event)
+
+
+class ModsChoice(PygameMenuState):
```
la-xirsoi, line 140: lets move this class into it's own file in the menu folder

la-xirsoi (changes requested):

la-xirsoi (approved):

PR #2786, opened 2025-05-08
Refactoring `pathfind_to_player` for Flexible Character Movement

JaskRendix:
PR renames `pathfind_to_player` to `char_pathfind_to_char`, eliminating the hardcoded `self.session.player` reference and enabling more flexible entity interactions. For example, it allows directing an NPC toward another without relying solely on pathfinding or manually calculating coordinates.


tuxemon/event/actions/char_pathfind_to_char.py
```diff
@@ -0,0 +1,67 @@
...
+            (e.g. "character_maple").
+        moving_entity: The entity that will move toward the target
+            (e.g. "character_jimmy").
+        direction: Determines approach direction
+            (up, down, left, or right).
+        distance: Number of tiles to maintain from the player
+            (e.g. 2,3,4).
+    """
+
+    name = "char_pathfind_to_char"
```
la-xirsoi: it's a small thing, but i think "char_pathfind_to_char" is a bit wordy.
if you're cool with it, i think "pathfind_to_char" is good (only characters should be pathfinding anyway).
I won't block merging this over it though.

If you wanna leave the name of the action as is, just say so and resolve this comment and I'll merge.

la-xirsoi (changes requested):

tuxemon/event/actions/char_pathfind_to_char.py
JaskRendix: done, "pathfind_to_char" is better

la-xirsoi (approved):
Looks Good! Thanks for the PR :)

PR #2799, opened 2025-05-15
Refactoring Object Instantiation with Classmethods

JaskRendix:
PR improves object instantiation by introducing a dedicated `@classmethod create` for the `Monster`, `Item`, `Technique`, and `Status` classes. It also streamlines the codebase by removing an obsolete test file and renaming a method for consistency.  

Changes:
- removed `test_monster.py`, which was redundant since a functioning test already exists
- renamed `load_from_db` to `load` to align with naming conventions used in other classes


kerizane:
Good, but I would have gone with a classmethod, rather than a function, given that the functions are all dedicated to a single class.

However, it's really not a big deal, so I won't force you to do extra work.

Let me know if you want me to merge it as is.

JaskRendix:
@kerizane done. At first, I made it a class method, then switched it to an external method, but in the end, I went back to making it a class method again.

kerizane (approved):

PR #2802, opened 2025-05-16
fix various

JaskRendix:
PR updates:
- made `_calculate_frame_duration` more flexible by injecting values
- refined `animate_party_hud_in` for better functionality
- fixed a frustrating type hint issue by using `self.update_background(self.graphics.background)`, bringing the total type hints down to 29 (strict type hints #1211)
- added `self.menu_visibility` to `MainCombatMenuState`, so now we can easily tweak which buttons are visible using the new `update_menu_visibility` method
- enhanced `CombatTargetMenuState` to fix a crash caused by a missing `self.window`, while addressing the issue, I also updated the code, removing unused elements for better efficiency
- moved `ignore_collisions` to the `Entity` class because it's more about how the entity interacts with the game world, rather than its movement specifically, it just makes more sense to keep it with the entity's other properties
- moved `move_direction` to the `Mover` class because it's directly related to the entity's movement, and the `Mover` class is all about handling movement, it's a better fit there, and it keeps all the movement-related stuff in one place
- defined property `is_moving` in the `Body` class
- fixed `fishing` bug in `CoreEffect`, previously, `next()` was only considering the first valid direction for fishing, which restricted fishing in certain cases, now, all directions enabling fishing in a tile are considered, making it possible to fish while swimming too
- enhanced `facing_sprite` in `CoreCondition` by using a set
- fixed `pathfind_to_char` bug in `EventAction`, corrected value inversion to ensure the NPC faces the correct direction, added a check to determine if the destination (entity's position) is adjacent to the NPC (up, down, left, or right), and prevented unnecessary pathfinding when the entity is already within reach
- updated `show_result_as_dialog` method in tools.py, it'll accept `Technique` too
- removed `combat_cry` and `faint_cry`, unused attributes of the `Monster` class... there are already `combat_call` and `faint_call`
- extracted the `handle_double_action` and `handle_single_action` methods to ensure that the menus for each monster are opened sequentially, rather than simultaneously, to avoid inconsistencies that occurred when the menus were opened one over another during double battles (double battles)
- removed two `__init__` files from the item and status `effects` folder (everything has been moved to core months ago)
- extracted `prepare_layout` utility method for `CombatAnimations`
- updated the `tuxemon_border._surface` initialization to use `convert_alpha()` to ensure correct handling of transparency and color depth, this change improves compatibility with rendering and maintains visual integrity when processing the surface


JaskRendix:
@kerizane removed all empty lines for better readability and easier file inspection

JaskRendix:
@kerizane, thanks again for merging the other PRs!
If possible, could we start with this one in next round?
It would help clean up a bunch of small things. Appreciate it!

la-xirsoi (approved):

PR #2842, opened 2025-06-07
Enhancements for `Technique` Class: `get_combat_state` and `set_combat_state` methods

JaskRendix:
PR introduces enhancements to the `Technique` class.

Changes:
- added explicit type annotations (`bool`, `str`, etc) to attributes for improved type clarity and static analysis
- cleaned up redundant fields and improved `load()` function logic, `nr_turn` was just a holdover from back when `Status` and `Technique` were lumped together. It really belongs to `Status` (and there is).

### DIFF
--- a/tuxemon/combat.py
+++ b/tuxemon/combat.py
@@ -212,8 +212,7 @@ def get_target_monsters(
     Raises:
         ValueError: If an objective is not a valid TargetType.
     """
-    combat = technique.combat_state
-    assert combat
+    combat = technique.get_combat_state()
     monsters = []
     for objective in targets:
         if objective not in list(TargetType):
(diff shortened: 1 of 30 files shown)


kerizane:
Are these attributes still used, but set somewhere else (e.g. in load?). Thanks, but it's poor practice to have attributes that aren't set within __init__.

If you want to improve this part of the code I would have these attributes all passed in to init. You could convert load to a class factory method.

JaskRendix:
@kerizane Cool, for now I'll just clean up the redundant attributes in the load. After that, I’ll take a closer look at the factory method.

JaskRendix:
@kerizane the factory method will be addressed in a separate PR. This one will now focus on simplifying an inconvenient issue that forces assertions to be scattered across the codebase. That concern is tackled in other PRs related to CombatState too (Item, Status)

la-xirsoi (approved):

PR #2845, opened 2025-06-08
Simplify `lookup` Method by Removing Literal and Adding Type Checking

JaskRendix:
PR introduces the following changes:
- removed the `Literal` type with 20+ elements from the `lookup` method parameter, this was a bottleneck, requiring constant edits whenever a new table type was added
- replaced `TableName = Literal[...]` with a simple `Optional[str]` type, allowing for more flexibility without needing predefined overloads
- eliminated multiple `@overload` definitions for specific table names
- a `lookup` class method was introduced, allowing retrieval of instances using a `slug`, this ensures consistent database access across 16 classes, utilizing `table_name` to reference the correct table, and if an entry isn't found, a `RuntimeError` is raised for clear error handling
- a `ClassVar[str]` attribute was added for `table_name` in each class, maintaining a structured and uniform approach to table references at the class level

### DIFF
--- a/tuxemon/animation_entity.py
+++ b/tuxemon/animation_entity.py
@@ -8,7 +8,7 @@ from collections.abc import Iterable
 from pygame.surface import Surface
 
 from tuxemon import prepare
-from tuxemon.db import db
+from tuxemon.db import AnimationModel, db
 from tuxemon.graphics import create_animation, load_frames_files
 from tuxemon.map_view import AnimationInfo
 
@@ -34,11 +34,7 @@ class AnimationEntity:
 
     def load(self, slug: str) -> None:
         """Loads animation."""
-        try:
-            results = db.lookup(slug, table="animation")
-        except KeyError:
-            raise RuntimeError(f"Animation {slug} not found")
-
+        results = AnimationModel.lookup(slug, db)
         self.slug = results.slug
         self.file = results.file
 
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 from enum import Enum
 from importlib import import_module
 from pathlib import Path
-from typing import Annotated, Any, Literal, Optional, Union, overload
+from typing import Annotated, Any, ClassVar, Literal, Optional, Union, cast
 
 import yaml
 from PIL import Image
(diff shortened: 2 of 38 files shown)


tuxemon/animation_entity.py
```diff
@@ -36,6 +36,10 @@ def load(self, slug: str) -> None:
         """Loads animation."""
         try:
             results = db.lookup(slug, table="animation")
```
kerizane: Bit of a shame to lose the typehints, but I will trust you when you say this is a bottleneck.

What if we made this a class factory method? Then we wouldn't need the type checking.

Something like:
`AnimationModel.lookup(slug, db)`

The table name could be stored as a class attribute.

kerizane (changes requested):

tuxemon/animation_entity.py
JaskRendix: I've updated the method to use `classmethod`, following your advice, but to remove `isinstance`, I had to use `cast()`.  
```
@classmethod
def lookup(cls, slug: str, db: ModData) -> ShapeModel:
    """Retrieve an instance from the database using a slug."""
    try:
        return cast(ShapeModel, db.lookup(slug, table=cls.table_name))
    except EntryNotFoundError:
        raise RuntimeError(f"Animation '{slug}' not found")
```
on the plus side, this approach allows lookup to work with a single string across different parts of the codebase, eliminating scattered `KeyError` handling and `RuntimeError` raising

tuxemon/db.py
```diff
@@ -300,13 +309,22 @@ class AttributesModel(BaseModel):
...
+    table_name: ClassVar[str] = "shape"
     slug: str = Field(
         ..., description="Slug of the shape, used as a unique identifier."
     )
     attributes: AttributesModel = Field(
         ..., description="Statistical attributes of the shape."
     )
 
+    @classmethod
+    def lookup(cls, slug: str, db: ModData) -> ShapeModel:
```
kerizane, line 321: There's probably a way to this this in a generic way on the superclass, but this approach is good enough.

kerizane (approved):

tuxemon/animation_entity.py
kerizane: Looks good, thanks.

PR #2851, opened 2025-06-11
Improve Initialization Flexibility + Path

JaskRendix:
PR enhances the initialization process by introducing a headless mode, making the game more adaptable to different environments.  

Changes:  
- introduced `headless_init()` to allow initialization without relying on `pygame`
- updated `init()` to accept a `platform` argument, enabling selection between `"pygame"` and `"headless"`
- refactored `pygame_init()` to ensure `platform.init()` is called before other initialization steps
- improved error handling by raising `ValueError` for unsupported platforms
- enhanced logging to provide better visibility during initialization
- encapsulated joystick blacklist logic in `_compile_joystick_blacklist()`
- refactored user storage setup into `_setup_user_environment()`, ensuring game directories are created with proper error handling
- added logging to track initialization steps and potential failures
- improved exception handling to log critical errors when directory creation fails, preventing silent failures
- the type hint for `config_path` was changed from `Optional[str]` to `Optional[Path]` (build up on #2812)
- the `try`-`except` block for checking if the file exists was replaced with a simple `if` statement using the `Path.exists()` and the `open()` function was replaced with the `open()` method of the `Path` object

### DIFF
--- a/tuxemon/prepare.py
+++ b/tuxemon/prepare.py
@@ -25,36 +25,53 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# TODO: refact this out when other platforms supported (such as headless)
-PLATFORM = "pygame"
 
-# list of regular expressions to blacklist devices
-joystick_blacklist = [
-    re.compile(r"Microsoft.*Transceiver.*"),
-    re.compile(r".*Synaptics.*", re.I),
-    re.compile(r"Wacom*.", re.I),
-]
+def _compile_joystick_blacklist() -> list[re.Pattern[str]]:
+    """Compiles a list of regex patterns for blacklisting joysticks."""
+    logger.debug("Compiling joystick blacklist patterns.")
+    return [
+        re.compile(r"Microsoft.*Transceiver.*"),
+        re.compile(r".*Synaptics.*", re.I),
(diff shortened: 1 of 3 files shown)


tuxemon/prepare.py
```diff
@@ -25,37 +25,54 @@
...
-
-# Create game savegame dir if missing
-paths.USER_GAME_SAVE_DIR.mkdir(parents=True, exist_ok=True)
-
-# Generate default config
-config.generate_default_config()
-
-# Read "tuxemon.yaml" config from disk, update and write back
-CONFIG = config.TuxemonConfig(paths.USER_CONFIG_PATH.as_posix())
+def _compile_joystick_blacklist() -> list[re.Pattern[str]]:
```
la-xirsoi, line 29: why do it this way? is there some benefit to having it defined in a function?

JaskRendix: @la-xirsoi it's not strictly necessary, but it makes the code more modular and testable. By isolating the blacklist in a function, it becomes easier to mock or extend later, say if we want to add patterns from config or environment variables. Also, logging when it’s compiled helps with debugging in complex environments.

la-xirsoi (approved):

PR #2854, opened 2025-06-11
Extending BaseLookupModel

JaskRendix:
PR follows up on @kerizane’s suggestion in PR #2845: *"There's probably a way to do this in a generic way on the superclass, but this approach is good enough."*  

So far, I’ve implemented `BaseLookupModel` for the models that actually use the `lookup` method, while the others remain on `BaseModel`. My question is: should we extend this approach further? Do we want *all* models to inherit from `BaseLookupModel` and gain the classmethod by default, or keep things as they are and only apply it where needed?  

Let me know your thoughts!

### DIFF
--- a/tuxemon/db.py
+++ b/tuxemon/db.py
@@ -5,12 +5,22 @@ from __future__ import annotations
 import difflib
 import json
 import logging
+from abc import ABC, abstractmethod
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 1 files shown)


la-xirsoi:
i would instead leave the BaseModel alone, for things that aren't loaded from the db, and create a BaseLookupModel that only has the things needed for lookup. Then stuff that needs the lookup can inherit from both.

la-xirsoi (approved):

Issue #2870, opened 2025-06-18
LLM Driven Dialog


ThatGuySam:
**Is your feature request related to a problem? Please describe.**
More an improvement

**Describe the solution you'd like**
Integrations with local LLM apis such as Ollama or Apple’s Foundation Models framework

**Describe alternatives you've considered**
Not sure

**Additional context**
I was watching some of the Apple WWDC videos and they demoed a 8bit game where NPC dialog was driven by the local LLM on device. 

https://youtu.be/6Wgg7DIY29E?si=1dYCfCAYINLAsBo2

While I was watching it I couldn't help but think this would be a super fun feature in a pokemon game to have NPC always say different and perhaps take actions driven by a language model

JaskRendix:
@ThatGuySam thanks for the suggestion

PR #2871, opened 2025-06-18
Add `GeneratedDialogAction` Event for Dynamic LLM-Based NPC Dialogue

JaskRendix:
PR adds a new `GeneratedDialogAction` event that lets NPCs generate dialog dynamically using a local or cloud-based LLM. It’s inspired by issue #2870 and that rad WWDC demo with generative dialog in retro games. I loved the idea and wanted to try it out myself, but real talk, my PC doesn’t have the juice to run beefy models like Mistral locally. So I built it to be flexible.

### DIFF
--- a/docs/handcrafted/action_list.rst
+++ b/docs/handcrafted/action_list.rst
@@ -44,6 +44,7 @@
 .. autoscriptinfoclass:: tuxemon.event.actions.fade_out.FadeOutAction
 .. autoscriptinfoclass:: tuxemon.event.actions.fadeout_music.FadeoutMusicAction
 .. autoscriptinfoclass:: tuxemon.event.actions.format_variable.FormatVariableAction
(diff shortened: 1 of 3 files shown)


ThatGuySam:
I have Ollama with several models on my MacBook. I'd be happy to test it out and provide back some screen recordings if you can give instructions on how to set a custom api url.  

I was thinking it might be fun to have an LLM take all the character dialog and convert them into "actor prompts" like: 
```markdown
You're name is X, you're an Y, and you're trying to accomplish Z task in the scene. 

You must tell {player name} about A
```

Almost, like you're instructing an Actor in real life with the things they should know in the scene

ThatGuySam:
Also, assuming all this is under some kind of AI Mode. 

Not trying to dictate default behavior in a house that's not mine :)

PR #2890, opened 2025-06-23
Add support for Charmed status effect

JaskRendix:
waiting:
- [x] #2889

PR introduces the groundwork for the new **Charmed** status effect, which will be ready to use after the merging of #2889. The effect causes techniques targeting a Charmed opponent to have a 50% chance of failure, offering a strategic new layer of disruption.

At the moment a placeholder, fully transparent icon has been added.

@Sanglorian could you let me know which of these status texts you prefer for when a combatant is affected?

1. `{target} is charmed and might hesitate to act.`
2. ~`{target} is beguiled and may not attack.`~
3. ~`{target} falls under a charm and loses focus.`~
4. ~`{target} is enchanted and could miss their chance.`~

We could use Charmed for some of the techniques tagged with `"love"`, where it could replace current status effects to better fit the theme:
- *Lust* might apply **Charmed** instead of **Exhausted**
- *Cat Calling* could apply **Charmed** instead of **Harpooned**


Sanglorian:
Thanks! I think "{target} is charmed and might hesitate to act." is clearest, so it's the text that I'd go with. 

EDIT: And sorry, forgot to say I think changing Lust and Cat Calling to apply Charmed is a great idea.

la-xirsoi (approved):

PR #2943, opened 2025-07-21
Refactor asset and localization system with mod-aware path resolution, caching, and metadata-driven entry map

### LINKED ISSUE #2949: .mo files are not replaced in .tuxemon folder on Windows compilation if they already exist
**Describe the bug**
I installed Tuxemon on my Windows computer after a long time without it. 

I received an error on trying to run Tuxemon:

Validation failed for 'aardant' in table 'item'

With some help from ChatGPT, I located a folder ".tuxemon" in the same folder as my "Tuxemon" folder.

The .mo file in .tuxemon dated back to 2021. When I deleted it and ran Tuxemon again, it worked.  

I think that means that Python is not building a new mo file if it finds one already existing, even if that file might be out of date.

JaskRendix:
waiting
- [x] #2944 

PR introduces a new asset-loading mechanism.

Here's a breakdown of the changes:
- introduces `fetch_mod_asset_roots()`, this method precomputes all valid mod root directories based on `CONFIG.mods` and standard search paths. This eliminates repeated path-checking during asset fetches
- optimizes `fetch_asset()` with LRU caching, assets are now retrieved via a simple, fast lookup that uses a cache to avoid redundant filesystem operations... the cache size is currently set to 512 for a good balance of memory and efficiency
- adds a population guard, calling `fetch_mod_asset_roots()` multiple times won't re-scan unless `force=True` is explicitly passed. This protects against accidental reinitialization and noisy logging
- optional `force=True` refresh, we can now force a full mod path rebuild in case your mod list changes during runtime
- improves logging clarity, asset discovery and cache hits/misses are clearly logged, aiding debugging and profiling
- replaces older `fetch()` logic, the original fetch function manually scanned multiple folders per call, this logic is now consolidated into a cleaner system with precomputed roots and a centralized entry point
- added `LocaleAudit`, a centralized utility class for analyzing translation coverage, reporting missing domains, validating completeness, and generating structured insights from `LocaleFinder`
- modified `LocaleFinder` to support multiple root directories (`root_dirs: list[Path]`), enabling recursive discovery of `.po` files across all active mod `l18n` folders
- introduced `fetch_all_l18n_roots()` to locate all `l18n` directories within every active mod, surpassing the limitations of `fetch_asset("l18n")` which returns only a single path
- confirmed that `fetch_mod_asset_roots(prepare.CONFIG)` is called prior to `fetch_all_l18n_roots()` to properly populate `_MOD_ASSET_ROOTS`, guaranteeing accurate discovery of localized assets
- refactored various locale-related classes into a dedicated module to prevent `locale.py` from becoming overly large and difficult to navigate
- added `extract_mod_name(map_path)` utility to derive the active mod name from a map path (e.g. `mods/<mod_name>/...`), this ensures that `WorldState` instances are now initialized with a new `mod_name` attribute, enabling per-mod context awareness during gameplay and supporting multi-mod compatibility
- added support for the new `starting_map` attribute in each mod's `mod.yaml`, allowing mods to declare their entry point directly in metadata, this replaces previously hardcoded map names, and is accessed via a new method in `ModData` (`get_mod_attribute`) that safely retrieves structured fields for dynamic map loading
- introduced `GameLauncher` class, it centralizes game session bootstrapping logic using mod metadata, handles map loading, teleportation, session variable setup, and optional state cleanup, and replaces duplicated launch code across menu callbacks and entry flow
- translation management now includes automatic cleanup of stale `.mo` files and smarter recompilation logic that only rebuilds `.mo` when the corresponding `.po` file is more than **one week** newer, this fixes #2949 a bug where outdated translations persisted due to overly eager timestamp checks (or installation), ensuring fresher output without unnecessary rebuilds


kerizane:
Conflicts

JaskRendix:
@kerizane fixed

kerizane (approved):

PR #2948, opened 2025-07-23
Enhance `StatChangeEffect` with Step-Based Scaling and Value Clamping

JaskRendix:
PR implements the idea I raised in #2512, which was later discussed in-depth by @Sanglorian in #2523. At first, I considered creating two separate effects, `stepup` and `stepdown`, but after exploring the options, it made more sense to extend the existing `StatChangeEffect` instead. It keeps the system cleaner, more flexible, and avoids unnecessary duplication.

PR introduces several improvements to the `StatChangeEffect` system used in combat stat manipulation for monsters. These updates make the effect system more flexible, accurate, and safer to use:
- added support for **step-based stat scaling**, allowing effects to modify a stat by a percentage of its base value instead of only using fixed values
- integrated `max_step_limit` to clamp total scaling up or down, preventing extreme stat changes that could break balance
- enhanced randomness with `max_deviation` now working for both value-based and step-based changes
- improved health handling: safely overrides current HP using the `overridetofull` flag and prevents healing above max HP using a `min(new_value, target.hp)` clamp
- safeguarded base stat assignment by validating against `StatType.__members__` to avoid setting invalid or unintended attributes
- added debug-level logging to trace both value-based and step-based mutations, making the system more transparent during runtime
- updated the class docstring to reflect all new attributes and mechanisms in a clearer, more accurate format

```
    {
        "value": float,             # Direct stat adjustment (ignored if using step)
        "step": int,                # Optional step delta (e.g., +2 steps to speed); used in scaling
        "max_deviation": int,       # Optional random deviation applied to step or value
        "max_step_limit": float,    # Max cumulative scaling boundary (e.g., ±0.5 for ±50% total change)
        "scaling_mode": str,        # "linear" uses base*(1 + step), "nonlinear" uses tiered multipliers
        "operation": str,           # Arithmetic operation: "add", "subtract", "divide" (for value logic only)
        "overridetofull": bool      # If True and stat is HP, fully restores current HP to max
    }
```

### DIFF
--- a/tuxemon/core/effects/statchange.py
+++ b/tuxemon/core/effects/statchange.py
@@ -25,23 +25,32 @@ class StatChangeEffect(CoreEffect):
     Applies a stat-altering effect to a target monster during combat.
 
     This effect modifies one or more combat-relevant stats on a monster, either
-    through addition, subtraction, or division, depending on configuration. The affected
-    stats are pulled from the status's stat components (e.g. `status.statmelee`, etc.),
-    and can impact either permanent base stats or temporary health values.
+    through value-based operations (e.g., add/subtract) or step-based scaling,
+    depending on configuration. Step-based changes allow incremental percentage
+    adjustments relative to a stat's base value, with clamping via `max_step_limit`
+    to ensure balance.
 
-    This class supports optional randomness via `max_deviation`, and HP overrides using
-    `overridetofull` which resets current HP to the max HP.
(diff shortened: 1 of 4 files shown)


Sanglorian:
This sounds great! I think having techniques and items adjust stats in predictable but temporary ways will be really useful for expanding tactical options in combat.

kerizane (approved):

PR #2952, opened 2025-07-26
Allowing encounters on Route 2

Sanglorian:
The encounter Events on Route 2 were missing an "11", which I believe was causing the player to not have random encounters in Route 2. I think this should fix it.


JaskRendix (approved):

JaskRendix:
Hey @kerizane, thanks a ton for merging! If you’ve got room for more, could you prioritize @Sanglorian's two PRs next? Would be super helpful!

kerizane (approved):

PR #2968, opened 2025-07-29
Fix XP bar reset and HUD flicker on monster level-up  

### LINKED ISSUE #2966: XP bar doesn't update automatically; when it does update, it flickers
**Describe the bug**
When a monster levels up, its XP bar fills up. But it stays filled up even when the monster fights in later battles. 

Agnite defeated Budaye, and increased to level 7.

Image

Agnite then passed out; I fought the battle again. 

Agnite still shows a full XP bar. 

Image

Interestingly, when you quit the game and reload it, the XP bar has the proper behaviour: 

Image

---

Also, when the XP bar fills up, the HP/XP "block" flickers for a moment. 

Ideally, the XP bar would increase a little more slowly so the player can follow its motion. 

**Expected behavior**
When a monster levels up, the XP bar should refresh - either immediately (showing how far the monster is to the next level) or at least for the next encounter. 

 - Windows 10

JaskRendix:
PR fixes #2966 , the visual behavior of the XP bar after a monster levels up. Instead of staying filled indefinitely, the bar now resets to 0 and animates toward the correct progress for the next level. The animation is slightly slowed so players can better follow the change.

It also sheds light on a flicker seen in the HP/XP block when the XP bar fills. This is tied to the HUD refresh logic. While we can disable the update to avoid flickering, doing so leads to overlapping level numbers (e.g. level 11 and 12 both visible), which would require additional adjustments to HUD rendering and timing. That cleanup is best left for a separate pass.

### DIFF
--- a/tuxemon/states/combat/combat.py
+++ b/tuxemon/states/combat/combat.py
@@ -1075,7 +1075,7 @@ class CombatState(CombatAnimations):
             if owner.isplayer:
                 self.task(partial(self.animate_exp, winner), interval=2.5)
                 self.task(
-                    partial(self.update_hud, owner, False, True), interval=3.2
+                    partial(self.update_hud, owner, False, True), interval=4.0
                 )
 
     def animate_party_status(self) -> None:
--- a/tuxemon/states/combat/combat_animations.py
+++ b/tuxemon/states/combat/combat_animations.py
@@ -269,21 +269,71 @@ class CombatAnimations(Menu[None], ABC):
             transition="out_quint",
         )
 
-    def animate_exp(self, monster: Monster) -> None:
-        target_previous = monster.experience_required()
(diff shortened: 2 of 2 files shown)


Sanglorian:
Thanks for this! I think this will remove a point of confusion for new players

JaskRendix:
https://github.com/user-attachments/assets/2cf6ddd5-4aa2-43d5-bb97-4c7e5f090736

@Sanglorian better now? the flickering’s turning out to be way more work than I expected. If we want to clean up those other quirks too, we’ll probably need to refactor and split `build_hud` off from `CombatAnimation`

Sanglorian:
Perfect, thanks! Although I was on the edge of my seat watching that fight, I thought your Puparmor would get knocked out

la-xirsoi (approved):

PR #2978, opened 2025-07-31
Fix for Grabbed and Stuck Status Not Reverting Technique Stats After Removal

### LINKED ISSUE #2976: Grabbed penalty continues after condition lost
**Describe the bug**

While a monster is Grabbed, the power of their Melee and Reach attacks is halved. 

But this doesn't appear to be reversed if the monster loses the Grabbed condition - as shown here, with a flinching Agnite.

(Unless the halving is reversed, but the display does not update.)

Image

JaskRendix:
PR fixes #2976 a bug where a monster affected by the *Grabbed* condition experiences reduced power for its Melee and Reach techniques, but that reduction was not properly reversed after the status was removed. To fix this, I’ve added logic under the `ON_END` phase to reapply default stats to the monster’s techniques when the status expires or is cleared. This ensures that the technique potency and power values return to normal. I’ve also applied the same fix to the *Stuck* status effect, which shared the exact same issue and behavior. Finally, regarding stat modifiers from status effects, I’ll be adding the relevant lines under PR #2948, since that one already modifies `StatChangeEffect`.

### DIFF
--- a/tuxemon/core/effects/grabbed.py
+++ b/tuxemon/core/effects/grabbed.py
@@ -34,14 +34,20 @@ class GrabbedEffect(CoreEffect):
     def apply_status_target(
         self, session: Session, status: Status, target: Monster
     ) -> StatusEffectResult:
+        if self.divisor == 0:
+            raise ValueError("StuckEffect divisor must be non-zero.")
+
         done: bool = False
         ranges = self.ranges.split(":")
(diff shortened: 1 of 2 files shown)


Sanglorian:
Thanks heaps for this!

la-xirsoi (approved):

Issue #2981, opened 2025-08-01
If you lose to a trainer, you sometimes do not have to fight them again


Sanglorian:
**Describe the bug**
When encountering a trainer out in the world, if you lose to them then they should be there the next time you visit, ready to fight you. 

For example, in Route 2 if you lose to Roddick, the next time you go there he should challenge you again. 

This is how it works for Billie in Route 2 - exactly as intended. If you lose to her, you have to fight her again. 

Although Roddick's code looks identical to Billie's as far as I can tell, for some reason if you lose to him you do not have to fight him again. 

If anyone could shed light on this I'd be very grateful! I've been experimenting and I can't figure out what the difference is between him and Billie. 

**To Reproduce**
1. Only have one, weak tuxemon in your party,
2. Lose to Roddick on Route 2.
3. Return to Roddick. 

**Desktop (please complete the following information):**
 - Windows 10

JaskRendix:
The real issue seems to be with how the variable is assigned. We need to make sure that variable is handled separately and correctly for Roddick so the game knows you haven’t beaten him yet.

I’m testing a few things right now to confirm, but that’s the gist of it. I’ll update once I’ve nailed down a fix. Let me know if you’ve noticed anything else weird with other trainers!

PR #2982, opened 2025-08-01
Support pre/post battle dialogue in `NpcModel` + Beautifier NPC folder JSON

JaskRendix:
PR introduces new dialogue fields to the `NpcModel`:
- `pre_battle`: Slug for dialogue shown before battle initiation  
- `post_battle_win`: Slug for dialogue shown after the NPC wins  
- `post_battle_lose`: Slug for dialogue shown after the NPC loses  
- `post_battle_draw`: Slug for dialogue shown after a draw
- `greeting`: Dialogue shown when first interacting with the NPC  
- `idle`: Ambient or filler dialogue used during repeated interactions  
- `farewell`: Dialogue shown when ending an interaction  

This change is part of a broader effort to reduce the number of hardcoded or map-bound strings, particularly in battle-related events. By centralizing NPC dialogue into the model, we gain:
- dialogue logic is now decoupled from map triggers  
- dialogue can be managed and localized directly from NPC JSON  
- helps prevent unintended dialogue triggers during auto-teleport events (notably in the Spyder campaign), where translated strings are often invoked prematurely (if set_variable isn't used)

This is the first step toward resolving issue #2981, by moving dialogue into the NPC model, we lay the groundwork for more robust and predictable event handling.

### DIFF
--- a/mods/tuxemon/db/npc/37707_female.json
+++ b/mods/tuxemon/db/npc/37707_female.json
@@ -1,10 +1,8 @@
 {
   "slug": "37707_female",
-  "template": 
-    {
-      "sprite_name": "37707_female",
-      "combat_front": "heroine",
-      "slug": "heroine"
-    }
-
+  "template": {
+    "sprite_name": "37707_female",
+    "combat_front": "heroine",
+    "slug": "heroine"
+  }
 }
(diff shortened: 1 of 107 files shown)


Sanglorian:
Thank you! And such quick work too, much appreciated.

kerizane (approved):

PR #2985, opened 2025-08-02
Fix: Flinching Condition Not Removed After Trigger

### LINKED ISSUE #2984: Flinching isn't removed when it triggers
**Describe the bug**
The flinching condition has a 50% chance to cost the monster its turn. When it does so, the monster should lose the flinching condition. 

Currently, if a monster flinches, it still keeps the flinching condition. 

There is code in flinching.py that seems like it should clear the condition:

`            user.status.clear_status(session)`

 - Windows

JaskRendix:
PR fixes #2984 the flinching condition was not being removed after it triggered and caused the monster to lose its turn. The code intended to clear the flinching status was incorrectly targeting `user`, which refers to the monster that applied the condition, rather than `target`, which is the monster affected by it.
```python
# Previous (incorrect)
user.status.clear_status(session)

# Updated (correct)
target.status.clear_status(session)
```
Now, when a monster flinches and loses its turn, the flinching condition is properly cleared as expected.

### DIFF
--- a/tuxemon/core/effects/flinching.py
+++ b/tuxemon/core/effects/flinching.py
@@ -24,7 +24,6 @@ class FlinchingEffect(CoreEffect):
 
     Parameters:
         chance: The chance.
-
     """
 
     name = "flinching"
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thank you, that's fixed it!

la-xirsoi (approved):

PR #2989, opened 2025-08-03
XP formula divides total XP by number of hits

Sanglorian:
This should correct an error in the XP formula, a hold over from when XP used to be apportioned based on how many hits each participant did.

It was dividing XP based on number of hits the knocked-out monster had taken, meaning that a monster knocked out in two hits gave half as much XP as one knocked out in one.

### DIFF
--- a/tests/tuxemon/test_reward.py
+++ b/tests/tuxemon/test_reward.py
@@ -83,12 +83,11 @@ class TestRewardSystem(unittest.TestCase):
         self.assertEqual(money, expected_money)
 
     def test_calculate_experience_default_method(self):
-        hits, _ = self.damage_tracker.count_hits(self.loser, self.winner)
         experience = calculate_experience(
             self.loser, self.winner, self.damage_tracker
         )
         expected_experience = int(
-            (self.loser.total_experience // (self.loser.level * hits))
+            (self.loser.total_experience // (self.loser.level))
             * self.loser.experience_modifier
         )
         self.assertEqual(experience[0], expected_experience)
@@ -117,7 +116,6 @@ class TestRewardSystem(unittest.TestCase):
         total_exp = calculate_experience_base(
             self.loser.total_experience,
             self.loser.level,
-            self.damage_tracker.count_hits(self.loser, self.winner)[0],
             self.loser.experience_modifier,
         )
         participants = self.damage_tracker.get_attackers(self.loser)
@@ -130,11 +128,10 @@ class TestRewardSystem(unittest.TestCase):
         experience = calculate_experience_base(
             self.loser.total_experience,
             self.loser.level,
-            hits,
             self.loser.experience_modifier,
         )
         expected_experience = int(
-            (self.loser.total_experience // (self.loser.level * hits))
+            (self.loser.total_experience // (self.loser.level))
             * self.loser.experience_modifier
         )
         self.assertEqual(experience, expected_experience)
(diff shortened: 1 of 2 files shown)


JaskRendix:
Just a heads-up: they're gonna rack up a ton of experience now. If you want to tone down the XP multiplier, let me know, I’ve got a PR that tweaks those lines.

After running some models, the simplest and most effective way is using this formula: `new_modifier = old_modifier / hits`

The catch is, we have to make an assumption about how many hits were landed. For a typical fight, let’s say around 4 to 5 his, that feels like a reasonable baseline for a "normal" encounter.

Example: `add_monster cardiling,7,spyder_route2_graf,5,10`
```
old: 81.67 xp, 5 multiplier, (3 hits) > 82, I kept float value for testing
old: 61.25 xp, 5 multiplier, (4 hits)
old: 49 xp, 5 multiplier, (5 hits)
old: 40.83 xp, 5 multiplier, (6 hits)
old: 35 xp, 5 multiplier, (7 hits)
old: 30.625 xp, 5 multiplier, (8 hits)
```
```
new: 245 xp, 5 multiplier
new: 98 xp, 2 multiplier
new: 73.5 xp, 1.5 multiplier
new: 49 xp, 1 multiplier
```

JaskRendix:
@Qiangong2 about your campaign, here’s a breakdown of all the battles (monsters), showing the XP output from the old formula and the suggested multiplier for the new one (based on an average of 4 to 8 hits). I've grouped together monsters that share the same level and multiplier to keep things simple. Let me know if this looks good to you or if you’d prefer to bump the XP even higher.

add_monster aardorn,2,omnigrunt,80,10 (80 is the multiplier)
```
old: 4 hits > 160
old: 5 hits > 128
old: 6 hits > 106.67
old: 7 hits > 91.43
old: 8 hits > 80
new: multiplier 40 > 160
```

add_monster rockitten,3,professor,27,40
```
old: 4 hits > 182
old: 5 hits > 145
old: 6 hits > 121
old: 7 hits > 104
old: 8 hits > 91
new: multiplier 20 > 180
```

add_monster chloragon,3,acolyte,80,100
```
old: 4 hits > 540
old: 5 hits > 432
old: 6 hits > 360
old: 7 hits > 308
old: 8 hits > 270
new: multiplier 40 > 360
```

add_monster mystikapi,5,garth,80,100
add_monster mystikapi,5,loch,80,100
add_monster chloragon,5,loch,80,100
```
old: 4 hits > 2500
old: 5 hits > 2000
old: 6 hits > 1666
old: 7 hits > 1428
old: 8 hits > 1250
new: multiplier 70 > 1750
```

add_monster lambert,8,loch,80,100
```
old: 4 hits > 10240
old: 5 hits > 8192
old: 6 hits > 6826
old: 7 hits > 5851
old: 8 hits > 5120
new: multiplier 115 > 7360
```

encounters:
mods/tuxemon/db/encounter/taba_ba_passageway_1.json
mods/tuxemon/db/encounter/taba_ba_passageway_2.json
mods/tuxemon/db/encounter/taba_ba_passageway_3.json

you put 15 as exp_req_mod, I would suggest 3 with the formula

Qiangong2:
@JaskRendix The XP changes look good to me. 

3 sounds about right for the xp mod. The reason it was so high before was because of how difficult it was to level up. The formula change should fix that issue.

la-xirsoi (approved):

Sanglorian:
Piggybacking on this PR -- @JaskRendix and @Qiangong2 , are you comfortable with me changing this formula once more -- to divide XP by the number of participants?

I've been experimenting with it some more and I think this leads to a better play experience because otherwise the player has an incentive to swap through all the monsters on their team so they all get a piece of the XP. 

When it splits between participants, there's a trade-off to switching -- the monster that switches in will get some XP, but the original monster will get less. 

But interested to hear what you think from your play throughs!

PR #2993, opened 2025-08-03
Fix Reward Move List Overwritten + Equal Method (after #2989)

JaskRendix:
waiting
- [x] #2989 

PR refactors the `award_rewards` method in `RewardSystem` to address a bug.

Bug Fix: previously, `rewards_data.moves` was overwritten each time a monster learned new moves and this caused only the **last monster's** learned moves to be retained.
  ```python
  rewards_data.moves = winner.moves.update_moves(...)
  ```
fixed by using `.extend(...)` to accumulate all learned moves:
  ```python
  new_moves = winner.moves.update_moves(...)
  if new_moves:
      rewards_data.moves.extend(new_moves)
  ```
Example: If Monster A learns `["Fireball"]` and Monster B learns `["Ram"]`, only `["Ram"]` was preserved. Now both are included: `["Fireball", "Ram"]`.

Bug Fix: the expression `round(monster_hits / total_hits)` is not a proportional distribution. For example, if a winner dealt 30 hits out of a total of 100, the result would be `round(0.3)`, which is **0**. The winner would get no experience. This revised code calculates the proportion of damage dealt by the winner `(monster_hits / total_hits)` and multiplies it by the base experience, ensuring that each participant receives a fair share based on their contribution.

### DIFF
--- a/tests/tuxemon/test_reward.py
+++ b/tests/tuxemon/test_reward.py
@@ -83,12 +83,11 @@ class TestRewardSystem(unittest.TestCase):
         self.assertEqual(money, expected_money)
 
     def test_calculate_experience_default_method(self):
-        hits, _ = self.damage_tracker.count_hits(self.loser, self.winner)
         experience = calculate_experience(
             self.loser, self.winner, self.damage_tracker
         )
         expected_experience = int(
-            (self.loser.total_experience // (self.loser.level * hits))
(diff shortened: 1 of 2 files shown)


la-xirsoi (approved):

Sanglorian:
Good catch!

PR #2995, opened 2025-08-03
Removing the code that removes the enraged and sniping conditions

Sanglorian:
Currently, the enraged and sniping conditions end in any round in which the monster uses a technique - even the technique to give themselves enraged or sniping in the first place.

I think instead these conditions should last until replaced. 

I believe deleting these two files is all that is needed to ensure that the enraged and sniping conditions survive from one round to another. (I've tested and it's worked so far for me.)


JaskRendix:
ok, but you need to remove the triggers

go here `mods/tuxemon/db/status/sniping.json`
and remove this:
```
    {
      "type": "sniping"
    }
```

go here `mods/tuxemon/db/status/enraged.json`
and remove this:
```
    {
      "type": "enraged"
    }
```

la-xirsoi (changes requested):
Can't just remove the code that implements the status effects, things that reference them will generate errors.

Sanglorian:
Good spotting, thank you both! I've done that now

la-xirsoi (approved):

PR #3009, opened 2025-08-06
Fix weblate conflict

JaskRendix:
PR fixes weblate conflict @kerizane


kerizane:
Hmm, lots of stuff was deleted. Is this correct?

JaskRendix:
@kerizane these are on our end. Once we fix it, it'll sync with what's already been done on Weblate. Technically, nothing needs to be added to the other languages, just the English base .po file.

PR #3026, opened 2025-08-09
Add `BaseCondition` to unify shape, types, tags, and terrain checks

JaskRendix:
waiting:
- [x] #3028

PR introduces a new `BaseCondition` class that replaces the previously separate shape and type conditions with a more flexible, unified condition evaluator.

What this adds:
- `BaseCondition` that can check monster attributes like: `tags`, `types`, `shape` and `terrains`
- support for negation using `!` (e.g., `!fire`)
- support for multiple conditions using `:` (e.g., `fire:!water`)
- `match` flag to control whether **all** conditions must match (`true`) or **any** (`false`) - false default
- runs the JSON beautifier for all items in the `item` folder to keep formatting consistent
- renamed **cat** into **species** (class Monster) as it's called in Tuxepedia
- renamed **category** into **species_name** (class Monster)

Consideration: once other PRs are merged, the `get_dataset` method will be moved into the `Monster` class to centralize attribute access and reduce duplication

Matches if the type is either *water* or *fire* (default match = false).
```
{
  "type": "base",
  "parameters": ["types", "water:fire"],
  "operator": "is"
}
```
Matches if the type is *water* and *not fire* (default match = false).
```
{
  "type": "base",
  "parameters": ["types", "water:!fire"],
  "operator": "is"
}
```
Matches only if the type is *exactly* both *water* and *fire* (match = true).
```
{
  "type": "base",
  "parameters": ["types", "water:fire", "true"],
  "operator": "is"
}
```


Sanglorian:
Very helpful for TMs and so on, thank you!

kerizane (approved):

PR #3030, opened 2025-08-09
Elemental Expansion: From 5 to 13 Types

JaskRendix:
PR shifts from 5 to 13 element types. With hundreds of monsters and techniques already in play, the original 5-type structure limited strategic diversity, too many creatures shared the same elemental identity, and matchups became predictable.

By expanding to 13 types, it dramatically increases the number of possible combinations. Dual-type monsters now have 78 unique pairings, compared to just 10 before. This opens the door to more specialized techniques, clearer monster roles, and deeper tactical choices.

It’s a change that echoes systems like Tuxemon, where type variety is key to long-term strategic depth. This means fewer clones, more distinct builds, and a meta that rewards creativity over repetition. Balance will be key, but the potential for richer gameplay is undeniable.

@Sanglorian


Sanglorian:
Thanks for this @JaskRendix ! I do think this is an important step forward for the project - although I still have a fondness for the original Wu Xing style five elements, I feel like they didn't really reflect the diversity of monsters and techniques that we have.

Sanglorian:
Also, on this note -- are people happy if I rename the technique Venom to Venomous? Then "Venom" can refer exclusively to the type.

la-xirsoi (approved):

PR #3028, opened 2025-08-09
Quality of life changes

Sanglorian:
Collection of changes I've made while playing through the game: 

* A Goth that got in the way of people in the cafe now stays within certain bounds
* In Route 2, the player faces opponents who challenge them (will roll this out elsewhere as I go)
* NPCs face the player when it makes sense
* Updating attribution to have more information on technique sound effects
* Some Leather Town map improvements
* TMs acquired in Spyder now operate based on tag logic (this may need to change to reflect #3026 ?)

* Ignore the tags condition and borders - @JaskRendix has introduced PRs with more enduring and elegant ways of fixing these.

PR #3032, opened 2025-08-10
fix various

JaskRendix:
PR addresses:
- reintroduces the changes for TM items that were previously removed to avoid conflicts with #3028 
- adds `tags` for `Technique`
- updates `BaseCondition` in this way the condition can be used for Techniques too (types + tags)

Detailed:
- `BaseCondition` refactors the original `get_dataset` method into two specialized methods, `get_dataset_monster` and `get_dataset_technique`, to support both `Monster` and `Technique` targets.. it introduces a shared helper method `_test_conditions` to eliminate duplicated logic in `test_with_monster` and `test_with_tech`

### DIFF
--- a/mods/tuxemon/db/item/tm_all_in.json
+++ b/mods/tuxemon/db/item/tm_all_in.json
@@ -1,8 +1,9 @@
 {
   "conditions": [
     {
-      "type": "tag",
+      "type": "base",
       "parameters": [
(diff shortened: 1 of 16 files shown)


Sanglorian:
Thank you!

kerizane (approved):

PR #3033, opened 2025-08-10
Monster Menu Visibility Logic Based on Game State

JaskRendix:
PR updates the logic controlling which options appear in the monster submenu. The goal is to ensure that only relevant actions are presented to the player, improving usability and reducing unnecessary interactions.

Changes:
- **Info**: Always visible. Provides basic stats and details about the selected monster.
- **Techniques**: Visible only if the monster has at least one move/technique.
- **Item**: Visible only if the monster is currently holding an item.
- **Move**: Visible only if the player has more than one monster. It’s unnecessary when only one monster exists.
- **Release**: Visible only if the player has more than one monster. Prevents triggering a release dialog.
- fixes index update logic: previously, releasing Katapill (as shown in the image) triggered an error `[2025-08-10 08:34:15,657] tuxemon.menu.menu - ERROR - Unexpected error in menu event processing: cannot access local variable 'monster' where it is not associated with a value`, this was due to the cursor pointing to a now-invalid index. The fix ensures that when the last monster is deleted, the cursor moves up to a valid entry, preventing the error

Screenshot_2025-08-10_08-15-32
Screenshot_2025-08-10_08-14-52

### DIFF
--- a/tuxemon/states/monster/__init__.py
+++ b/tuxemon/states/monster/__init__.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-from collections.abc import Callable, Generator, Sequence
+from collections.abc import Callable, Generator
 from functools import partial
 from typing import TYPE_CHECKING, Any, Optional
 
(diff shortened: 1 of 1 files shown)


kerizane (approved):

Sanglorian:
Nice one, thanks for coding @JaskRendix and approving @kerizane 

This will clean up the menu and avoid confusion

PR #3040, opened 2025-08-11
save upgrade flow to include technique renaming

JaskRendix:
PR improves the save upgrade system. It also introduces support for renaming techniques across all relevant parts of the save data.

Key changes:
- refactored the save upgrade system to better support versioned upgrades and universal fixes
- introduced `_handle_change_tech_slug(save_data)` to rename technique slugs using the `TECHNIQUE_RENAMES` dictionary, applies to all monsters in the main save data and applies to all monsters stored in monster boxes (preparing change proposed by @Sanglorian in #3030)
- registered `_handle_change_tech_slug` in `apply_universal_fixes(npc_state)` so it runs regardless of save version
- updated `upgrade_save` to apply versioned upgrades via the `VERSION_UPGRADES` dictionary and then apply universal fixes
- replaced the older `upgrade_npc_state` approach with a more modular and extensible upgrade strategy
- clarified the upgrade strategy in the module docstring, including how and when universal fixes are applied

### DIFF
--- a/tuxemon/save_upgrader.py
+++ b/tuxemon/save_upgrader.py
@@ -2,7 +2,8 @@
 # Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
-from collections.abc import Mapping
+import logging
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thank you - really appreciate you fixing this to preserve save files as we change a technique name.

la-xirsoi (approved):

PR #2979, opened 2025-08-01
Fix blurry Tuxemon sprites by converting to indexed PNG format

JaskRendix:
PR fixes #2950, where some Tuxemon sprites appeared blurry while others remained crisp.

What was done:
- all `.png` sprite images in the `battle` folder, including `front`, `back`, and both `menu`, were batch converted to **indexed PNG format** using the following command:
  ```bash
  for img in *.png; do
    convert "$img" -colors 256 -background none -alpha on PNG8:"$img"
  done
  ```
This conversion:
- Reduces color depth to 256 indexed colors
- Preserves transparency
- Matches the crisp rendering seen in GIMP when converting to indexed mode

Why this fixes the issue:
- Indexed PNGs eliminate soft gradients and anti-aliasing artifacts that can cause sprites to appear blurry
- This format is ideal for pixel art and sprite-based games

Visual result:
- Sprites now appear consistently sharp across all battle views

PR #3048, opened 2025-08-12
Fix blurry Trainer sprites by converting to indexed PNG format

JaskRendix:
PR replicates the refactoring performed in #2979, but specifically for trainer sprites.


Sanglorian:
Very nice, thank you!

la-xirsoi (approved):

PR #3067, opened 2025-08-18
Next tranche of quality of life changes 

Sanglorian:
This is a set of small changes, that I've made as they come up playing through the game: 

* Adding in small element icons of 12x12 alongside the 24x24 icons. This gives the option to show the element icons at "full size" rather than shrunk down. 
* Fixing the bit depth for backgrounds so they don't blur. 
* Changing the translation of citypark from Citypark to City Park
* Adds line spacing to translated dialogue
* Giving the fences green shadows when they fall on grass (rather than gray)

The one substantive change is that XP is now divided between participants. So if SHYBULB and TWEESHER both hit SQUABBIT before it is knocked out, the XP that SQUABBIT awards is divided in half - half going to SHYBULB and half to TWEESHER.


JaskRendix:
@kerizane if you're able to, could you start by merging this one? Then try to clear out as many as you can, I've got a few PRs stuck because others are still blocked

kerizane (approved):

Sanglorian:
Thank you @kerizane !

PR #3072, opened 2025-08-19
Synchronization of Monster and Technique Types

JaskRendix:
PR builds upon the implementation in #3030, aligning monster and technique definitions with the updated type system.

@Sanglorian


Sanglorian:
Thank you! I'm very excited for this, I think it's going to add a lot of depth to the game.

la-xirsoi (approved):

PR #3073, opened 2025-08-19
Enhance Translation System with Domain Discovery and Plural Support

JaskRendix:
PR introduces two key improvements to the translation infrastructure, making it more dynamic, scalable, and capable of handling plural forms.

Added `discover_and_load_all_domains()` to `TranslatorManager`:
- scans all available `.po` files using `LocaleFinder`
- loads a translator for each discovered domain using the specified locale
- ensures that all translation domains, including those introduced by mods or external content—are initialized without manual intervention
- this single change allows the system to automatically handle any new `.po` file added (e.g., `monster.po`, `items.po`) without requiring further code modification
- `TranslatorManager` object (`T`) will now be aware of all available translation domains
- developers can use `T.translate("message", domain="...")` throughout the codebase to retrieve translations from the appropriate domain

Updated `TranslatorPo` to Support Plural Translations:
- replaced the cached translation logic with direct access to `gettext.gettext` and `gettext.ngettext`
- introduced `translate_plural()` for handling singular/plural forms based on quantity
- added `has_plural_translation()` to check for the existence of plural translations
- ensures fallback logic still applies, including fallback to the `"base"` domain if needed

### DIFF
--- a/tests/tuxemon/test_translator_po.py
+++ b/tests/tuxemon/test_translator_po.py
@@ -9,7 +9,7 @@ from tuxemon.locale_dir.translator import TranslatorPo
 
 class TestTranslatorPo(unittest.TestCase):
 
-    @patch("gettext.translation")
+    @patch("tuxemon.locale_dir.translator.translation")
     def test_translation_success(self, mock_translation):
         mock_trans = MagicMock()
         mock_trans.gettext.return_value = "Bonjour"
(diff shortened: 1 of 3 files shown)


la-xirsoi:
I'm not sure having all localizations loaded is a good idea. I like the dynamic loading and the discovery, but I think only the active locale should be loaded.

JaskRendix:
@la-xirsoi you're right, trying to load every single localization upfront would be a nightmare for performance and memory.

But if you take a closer look at the `discover_and_load_all_domains` method, it's pretty smart about it. It scans all the domains (like 'base' and any mods), but it only loads translators for the current locale. So it's not pulling in every language file, just the ones it needs right now.

Then, when you switch languages, the `change_language` method kicks in and reloads the translators for all those domains using the new language files. Clean and efficient.

la-xirsoi (approved):

PR #3074, opened 2025-08-20
Add Per-Animation Timing and Controls via `AnimationModel`

JaskRendix:
This change was prompted by a question from @Sanglorian asking whether it was possible to slow down specific animations. Upon investigating, I realized that animation timing was previously set universally across all animations, regardless of how many frames they contained. This made it impossible to fine-tune the speed or behavior of individual animations.

To address this, I’ve updated the `AnimationModel` and `AnimationEntity` classes to support per-animation customization. These changes allow modders and developers to define playback behavior directly in the animation database.

Changes:
- `AnimationModel`, added the following fields: `duration` which controls the duration of each frame (default: `0.09`), `loop` whether the animation should repeat (default: `False`), `rate` which is playback speed multiplier (default: `1.0`), `flip_axes` which are the axes to flip the animation frames (`x`, `y`, `xy`, or none)
- `AnimationEntity`: changed `duration` and `loop` to be optional parameters; if not provided, these values are now loaded from the database via `AnimationModel`; converted `frames` to a list to support `len()` and indexing; applied `rate` and `flip_axes` to the `SurfaceAnimation` object and added a debug log to confirm animation loading and frame count

All animation definitions in `db/animation` can now use the new attributes. For example, the following entry:
```json
{
  "file": "technique",
  "slug": "12_hits_for_separation1"
}
```
can now be extended to:
```json
{
  "file": "technique",
  "slug": "12_hits_for_separation1",
  "duration": 0.12,
  "loop": true,
  "rate": 0.8,
  "flip_axes": "x"
}
```
More work is needed. In particular, Update `MethodAnimationCache` class.


Sanglorian:
Thank you! This will be a great improvement; at the moment more complicated animations just go by in a flash.

la-xirsoi (approved):

PR #3092, opened 2025-09-03
Tuxepedia entry redesign

Sanglorian:
1. Using smaller element icons, but at "full" size
2. Rearranging the position of some elements, and displaying some at larger size
3. Where possible, making elements line up with the "pixels", as if it were 256x144 pixels.

What do people think? I think it neatens up the display, while keeping the information and spirit of the original design -- but welcome any thoughts! 

@JaskRendix , do you know how I might increase line-spacing for the "description"? Sorry, I know you implemented line-spacing generally but I couldn't figure out how to apply it! 

image

### DIFF
--- a/tuxemon/menu/menu.py
+++ b/tuxemon/menu/menu.py
@@ -47,6 +47,7 @@ class FontSettings:
     medium: int = prepare.SCALE * prepare.FONT_SIZE
     big: int = prepare.SCALE * prepare.FONT_SIZE_BIG
     bigger: int = prepare.SCALE * prepare.FONT_SIZE_BIGGER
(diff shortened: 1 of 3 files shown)


JaskRendix:
@Sanglorian for the description, try adding and set `leading=30` where there is `wordwrap=True` to bump up the line spacing a bit. That should give it some extra breathing room.

Sanglorian:
> @Sanglorian for the description, try adding and set `leading=30` where there is `wordwrap=True` to bump up the line spacing a bit. That should give it some extra breathing room.

That's fixed it, thank you!

la-xirsoi (approved):

PR #3101, opened 2025-09-07
Technique preview within combat menu

Sanglorian:
This adds icons to the technique preview within the combat menu. 

It also shows Power based on most of the damage formula: Power * (7+level). 

I like this way of showing it because it allows the user to more easily calculate how much damage will be done. Against an enemy whose Armor or Dodge is equal to the attacker's Melee or Ranged, the Power shown is equal to the damage that will be done. If the attacker's Melee/Ranged is double the target's Armor/Dodge, the damage done will be double, etc. 

I hope this makes it easier for players to make strategic decisions. 

While doing that I made a couple of other tweaks:
* The combat background was being stretched to cover the full height (144 pixels) even though the area where the background is displayed is just 108 pixels tall. 
* That inadvertently left a few white pixels showing because the combat menus have rounded edges -- so now the background's last row will extend to cover the remaining space. 
* The opponent's monster and island were slightly off-kilter; this change makes them line up exactly with the pixels for the background. 

I have been working on some icons that could replace the recharge text, making room for Potency to be displayed as well ... but I wanted to see how this was received first. 

What do folks think? 

image


JaskRendix:
combat_menu isn't anymore inside the folder "combat", you can find it directly in states

Sanglorian:
> combat_menu isn't anymore inside the folder "combat", you can find it directly in states

Ah, thanks! I think it should be fixed now.

la-xirsoi (approved):

PR #3103, opened 2025-09-07
`BaseClient` Detox - No More Peeking into StateManager (1)

JaskRendix:
PR begins cleaning up the `BaseClient` by removing some methods that expose the state stack (via the state manager). The goal is to keep the client lean and avoid breaking encapsulation, Demeter says no 👀.

Gone:
- get_queued_state_by_name
- queue_state
- replace_state
- push_state_with_timeout
- active_states
- current_state
- active_state_names

to be addressed:
- get_state_by_name
- pop_state
- remove_state_by_name
- push_state

### DIFF
--- a/tuxemon/base_client.py
+++ b/tuxemon/base_client.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from collections.abc import Mapping, Sequence
+from collections.abc import Sequence
 from enum import Enum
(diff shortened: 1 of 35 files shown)


la-xirsoi:
you actually had this correct before, this change violates the Law of Demeter because you are reaching into the methods of the member of a member. This makes the abstraction "leaky", making the implementation details of your class members visible, and making any future refactoring more difficult.

PR #3107, opened 2025-09-08
Replace client-based screen size and rect with prepare constants

JaskRendix:
waiting:
- [x] #3101

PR updates screen size and rectangle references to use `prepare.SCREEN_SIZE` and `prepare.SCREEN_RECT` instead of accessing them through `self.client.screen.get_size()` and `get_rect()`.

Changes:
- replaced all instances of `self.client.screen.get_size()` with `prepare.SCREEN_SIZE`
- replaced all instances of `self.client.screen.get_rect()` with `prepare.SCREEN_RECT`
- this avoids passing screen-related values through the client object
- helps distinguish between headless and pygame-based clients more cleanly
- reduces coupling between menu logic and client implementation
- removed `capture_screenshot` method from graphics.py (exactly the same in save.py)

### DIFF
--- a/tests/tuxemon/test_world_transition.py
+++ b/tests/tuxemon/test_world_transition.py
@@ -3,8 +3,6 @@
 from unittest import TestCase
 from unittest.mock import MagicMock, patch
 
-from pygame.surface import Surface
-
 from tuxemon.client import LocalPygameClient
 from tuxemon.movement import MovementManager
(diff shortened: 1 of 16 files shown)


JaskRendix:
`shop_monster` and `shop_item` moved to #3157

la-xirsoi (approved):

PR #3109, opened 2025-09-09
Refactor Status Stat Fields into Unified Dictionary Format

JaskRendix:
PR introduces a structural refactor to how stat-related data is handled in the `StatusModel` and its associated effect logic.

Key changes:
- replaces individual stat fields (`statspeed`, `statarmour`, `statdodge`, `statmelee`, `statranged`, etc.) with a single `stat_modifiers` dictionary keyed by stat slug
- simplifies the model by consolidating all stat modifiers under one field, improving scalability and reducing redundancy
- updates the `StatChangeEffect` method to iterate over `stat_modifiers` instead of accessing each stat field manually
- ensures consistent handling of stat changes, including value-based and step-based operations, within the new dictionary structure
- preserves all existing functionality, including HP overrides, deviation logic and clamping safeguards
- prepares the system for future expansion of stat types without requiring model changes

### DIFF
--- a/mods/tuxemon/db/status/blinded.json
+++ b/mods/tuxemon/db/status/blinded.json
@@ -14,13 +14,15 @@
   "sfx": "sfx_pulse",
   "slug": "blinded",
   "sort": "meta",
-  "statdodge": {
-    "value": 0.5,
-    "operation": "*"
-  },
-  "statspeed": {
-    "value": 0.5,
-    "operation": "*"
+  "stat_modifiers": {
+    "speed": {
+      "value": 0.5,
+      "operation": "*"
+    },
+    "dodge": {
+      "value": 0.5,
+      "operation": "*"
+    }
   },
   "cond_id": 17,
   "use_failure": null,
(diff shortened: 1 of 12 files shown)


Sanglorian:
Thanks Giorgio! This looks good to me. 

I've been taking a look at the non-linear stat step model you've created. I think it could replace some conditions like Hard Shell, Focused and Slowed, plus make techniques more varied. Thanks for setting it up. 

Two thoughts: 

1. Could we print text when the stat is modified, something like: "Aardorn's Melee rose 2 steps" or "Weavifly's Dodge fell 1 step"?
2. Could the formula use a different number to max_step? As I'd like to keep max_step at 6, but make the formula use (3 + step)/3 or 3/(3 + step) instead of (6 + step)/6 or 6/(6 + step).

JaskRendix:
@Sanglorian 

> Could we print text when the stat is modified, something like: "Aardorn's Melee rose 2 steps" or "Weavifly's Dodge fell 1 step"?

I'm going to look into as soon as it's merged

> Could the formula use a different number to max_step? As I'd like to keep max_step at 6, but make the formula use (3 + step)/3 or 3/(3 + step) instead of (6 + step)/6 or 6/(6 + step).

yes
```

  "stat_modifiers": {
    "armour": {
      "value": 1.5,
      "operation": "*",
      "max_step_limit": 3.0 (if this line is missing, then it's default 6.0)
    }
```

la-xirsoi (approved):

PR #3119, opened 2025-09-11
Fix animation playing over wrong monster when technique is shared

### LINKED ISSUE #3115: When both monsters select the same technique, animation plays over both monsters the first time the technique is used
**Describe the bug**
When both monsters in a battle have selected the same technique (for example, Agnite and Aardorn choose Gnaw), when the player's monster uses the technique, the animation plays over both monsters. 

(In an earlier version I said it was based on which went first, that was my mistaken assumption -- when I played against a slower opponent, the animation still played over the player's monster when it was the player monster's action.)

 - Windows

JaskRendix:
PR fixes #3115 

Bug: when both monsters in a battle select the same technique (e.g. Agnite and Aardorn both use *Gnaw*), the animation appears to play over the player's monster regardless of whose turn it is. This happens even when the opponent is slower and the player's monster acts first.

Cause: the animation was being scheduled to play with a short delay (`interval=0.6`), but the kill timer was based on `action_time`, which could be `0.0` if no combat text was generated. As a result, the animation was often killed before it started, making it appear as if it played over the wrong sprite.

Fix: we now ensure that every animation has a minimum visible duration by calculating a `safe_action_time` that accounts for both the animation’s intrinsic duration and the delay before it starts:

```python
safe_action_time = max(action_time, animation.animation.duration + start_delay)
```

This guarantees that the animation remains on screen long enough to play, even when `action_time` is zero. The animation is also positioned over the target sprite, so this fix resolves the visual misplacement.

### DIFF
--- a/tuxemon/states/combat_state.py
+++ b/tuxemon/states/combat_state.py
@@ -748,11 +748,16 @@ class CombatState(CombatAnimations):
         if target_sprite and animation:
             animation.rect.center = target_sprite.rect.center
             assert animation.animation
-            self.task(animation.animation.play, interval=0.6)
+            start_delay = 0.6
+            self.task(animation.animation.play, interval=start_delay)
             self.task(
-                partial(self.sprites.add, animation, layer=50), interval=0.6
+                partial(self.sprites.add, animation, layer=50),
+                interval=start_delay,
             )
-            self.task(animation.kill, interval=action_time)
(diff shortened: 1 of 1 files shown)


Sanglorian:
Thank you!

la-xirsoi (approved):

PR #3120, opened 2025-09-11
Add `ComboDetector` Support to `InputManager` with Timed Input Matching

JaskRendix:
waiting:
- [x] #3118

PR introduces support for detecting button combinations (combos) within the `InputManager` class using the `ComboDetector` system.

Key changes:
- integrated `ComboDetector` into `InputManager` to monitor input history and detect predefined combos
- added `setup_combos()` method to register specific button sequences with associated callbacks
- defined two example combos: `UP, UP, DOWN, A` triggers `on_special_move()` and `LEFT, RIGHT, LEFT, RIGHT, B` triggers `on_super_combo()`
- each combo includes a maximum delay between button presses to ensure timing-sensitive detection
- input history is managed and truncated to prevent memory bloat while maintaining responsiveness
- `PlayerInput` events now include a `timestamp` field, allowing `ComboDetector` to verify not just the button sequence but also the timing between presses
- introduced `ComboManager` to encapsulate combo detection and edge filtering logic, improving modularity and separation of concerns
- added `ButtonEdgeFilter` to suppress repeated input events from held buttons, ensuring clean combo detection
- implemented `AddComboAction` as a scriptable `EventAction`, allowing dynamic registration of combos via symbolic button names (e.g. `LEFT:A:B`) and optional timing constraints


tuxemon/event/actions/add_combo.py
```diff
@@ -0,0 +1,94 @@
...
+        except KeyError as e:
+            logger.warning(f"Unknown button name in combo: {e}")
+            return
+        except Exception as e:
+            logger.warning(f"Invalid combo definition: {self.values} — {e}")
+            return
+
+        def on_combo_triggered() -> None:
+            logger.info(f"Combo '{self.combo_name}' triggered!")
+            # Optional: trigger a named event or script
```
la-xirsoi: i dont see a way to specify this optional callback

tuxemon/event/actions/add_combo.py
```diff
@@ -0,0 +1,94 @@
...
+    "BACK": 16384,
+    "MOUSELEFT": 32768,
+}
+BUTTON_NAME_TO_ID = {k.upper(): v for k, v in BUTTON_NAME_TO_ID.items()}
+BUTTON_ID_TO_NAME = {v: k for k, v in BUTTON_NAME_TO_ID.items()}
+
+
+@final
+@dataclass
+class AddComboAction(EventAction):
```
la-xirsoi, line 39: should we have a way to remove/unregister a combo?

JaskRendix: yes

tuxemon/event/actions/add_combo.py
JaskRendix: eh, initially I thought about open_dialog, so trigger a dialog box, eventually I could remove it (the comment), do you have any suggestions?

la-xirsoi: mentioned in a DM, but adding here for completeness:

"I suggest having it take the name of an event to trigger"

JaskRendix: done

la-xirsoi (approved):

PR #3128, opened 2025-09-13
Refactor `MonsterTakeState` and Introduce Drone-Based Monster Transfer

JaskRendix:
PR refactors the `MonsterTakeState` logic and introduces a more modular and maintainable structure for handling monster actions.

Key changes include:
- extracted monster-related actions (pick, move, release, swap, info, etc.) into a dedicated `MonsterActionHandler` class to simplify state logic and improve code reuse
- updated `MonsterTakeState` to delegate all monster interactions to `MonsterActionHandler`, reducing complexity and improving readability
- added support for swapping a party monster with one from the kennel using a drone-themed item effect (`DroneEffect`)
- introduced localized dialog messages that reference the drone, providing a consistent and immersive experience across pickup, swap and release operations
- improved menu option filtering and state cleanup logic to prevent redundant state stacking and ensure smoother transitions

Players will always be able to access the monster box, even if the kennel is full. To avoid triggering the “kennel full” popup, the only available action in that case will be to `release` a monster. This way, players won’t be forced to drop one from their party just to see what’s inside the box. Added a check to prevent the move action from showing kennels that are already full.

https://github.com/user-attachments/assets/35a6b4c3-95d5-4ad1-b37a-99a960e82c62

Screenshot_2025-09-14_14-02-17


Sanglorian:
Very cool, thank you!

la-xirsoi (approved):

PR #3131, opened 2025-09-14
Add `escape_key` Item and `TeleportItemEffect` for Player Teleportation

JaskRendix:
PR introduces a new usable item, `escape_key`, along with its associated effect class, `TeleportItemEffect`. When activated, this effect teleports the player to a specified map and coordinates. If the target map is `"center"`, the effect instead uses the player's configured faint recovery location.

Key features:
- adds `escape_key` to the item registry  
- implements `TeleportItemEffect` with configurable destination logic  
- supports fallback to player-defined faint teleport coordinates  
- integrates with `event_engine` via `"transition_teleport"` action

```
  "effects": [
    {
      "type": "teleport_item",
      "parameters": [
        "spyder_flower_city.tmx","17","30" or "center" (only parameter)
      ]
    }
  ],
```


Sanglorian:
Oh great, this is a very useful addition - thank you

la-xirsoi (approved):

PR #3139, opened 2025-09-16
Replace `variable_set` with `battle_outcome` in Spyder maps to reduce reliance on game variables

JaskRendix:
PR updates conditional logic in the Spyder maps by replacing:
```xml
<property name="cond1" value="is variable_set hospital1aurora:yes"/>
```
with:
```xml
<property name="cond1" value="is battle_outcome player,won,spyder_hospital1_aurora"/>
```
This change eliminates the need for manually setting `hospital1aurora:yes` via `set_variable`, reducing reliance on global game variables and improving clarity of event logic. It also lays the groundwork for resolving issue #2981, which requires removing the `set_variable` call entirely and relying solely on battle outcomes for conditional branching.


Sanglorian:
Oh great, this is a really useful simplification for map-making. Thank you!

la-xirsoi (approved):

PR #3143, opened 2025-09-18
Passport-style monster information page

Sanglorian:
I've been working on a new way of presenting information on your tuxemon -- a "passport" style display. 

What do people think?

It makes space by removing information that is already in Tuxepedia, like blurb, category and future evolutions. I think this is okay because people can always look up the monster entry in Tuxepedia to see the other information -- but let me know if you think otherwise. 

Currently the labels are part of the background, but if people like the design I will go back and switch them to being translated text so it works for all languages. 

image

I've also been playing around with passport stamps that could show where a mon was captured -- just for fun as we don't record that information currently. 

passport stamps - routes, park


Davidliang11292:
Cool Design!

Sanglorian:
I've made a big structural change here, which is to allow for multiple fonts (in this case two -- the thin Pizel font as well as our default Press Start 2P. There should be a fall back that the localizations with different scripts (Japan and China) use their default fonts, not Pizel.

I'm particularly interested in what people think of this, and whether it's implemented well. I think a thinner font (and, later, a smaller font for certain numbers, like HP values) will expand the game in important ways. 

EDIT: I've also set the labels via the translate function, so they're compatible with other languages instead of being part of the background.

la-xirsoi (dismissed):

Sanglorian:
Thanks @la-xirsoi  -- I fixed the inconsistency that was blocking the merge.

la-xirsoi (approved):

PR #3153, opened 2025-09-20
Fix: Correct Status Resolution in `pre_checking` Phase - 2984

### LINKED ISSUE #2984: Flinching isn't removed when it triggers

JaskRendix:
Fixes #2984. Previously, the `status.use()` call mistakenly passed `target` instead of `monster` as the subject of the status effect. This caused status resolution to behave incorrectly, especially for flinching, since the status was being queried from the wrong entity. The fix ensures that `monster` is correctly passed, allowing status effects to resolve as intended during the `PRE_CHECKING` phase. This change restores proper status behavior and prevents issues like premature status disappearance or incorrect technique overrides.

This made me realize that a refactor of the Status effect system is needed. There are several confusing behaviors and edge cases that warrant a more structured intervention. That’s why this PR is just a temporary patch to address the immediate issue, I'll handle the broader refactor in a separate PR #3154.

### DIFF
--- a/tuxemon/combat/session.py
+++ b/tuxemon/combat/session.py
@@ -514,7 +514,7 @@ class CombatSession:
         status = monster.status.get_current_status()
         if status:
             result_status = status.use(
-                session, target, EffectPhase.PRE_CHECKING
+                session, monster, EffectPhase.PRE_CHECKING
             )
             if result_status.techniques:
                 technique = random.choice(result_status.techniques)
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thank you! Particularly for the quick work

kerizane (approved):

PR #3116, opened 2025-09-11
Removing the tired condition

Sanglorian:
In discussion with @JaskRendix , we agreed the "tired" condition isn't really necessary. This would remove it, but I haven't properly tested it yet so have marked this PR as a draft. 

If accepted, I can jump into the wiki and change the details on the four techniques that currently use the Tired condition. 

EDIT: Also adds a testing map -- will add and expand as we refine the game, and then remove when we get closer to a public launch.

PR #3154, opened 2025-09-20
Refactor `Status.use()` to Resolve Host Internally

JaskRendix:
waiting
- [x] #3109
- [x] #3116
- [x] #3145
- [x] #3146
- [x] #3149 
- [x] #3153

PR updates the `Status.use()` method to internally resolve the correct target using `self.get_host()` instead of requiring it to be passed manually.

- previously, callers had to pass the `target` monster explicitly, which led to confusion and bugs. especially when the wrong entity was passed (#3153)
- the status effect system is designed around each status being bound to a specific monster (`host`), and using `get_host()` ensures that effects are always applied to the correct entity
- this refactor simplifies the method signature and reduces the risk of misuse


Sanglorian:
Really good, thank you!

kerizane (approved):

PR #3164, opened 2025-09-25
Correct a few typos in the README

FlorentCLMichel:


JaskRendix:
@FlorentCLMichel thanks for updating the README! While you're at it, could you also fix the typo on the same line "reccomended" should be "recommended"? Also spotted a couple more in the file:  "systax" → "syntax" and "give your self potions" → "give yourself potions". Would be awesome if you could clean those up too.

FlorentCLMichel:
@JaskRendix Ah, good point! I'll clean these up and see if I can spot more.

kerizane (approved):

PR #3175, opened 2025-09-30
Refactor `NuPhoneContacts` to Support Conditional Dialogue and Dynamic Call Logic

JaskRendix:
PR refactors the `NuPhoneContacts` menu system to support more flexible and context-aware phone call interactions with NPCs. The previous implementation relied solely on map name translation to determine call outcomes. The new version introduces structured call logic based on game state conditions and external data.

Key changes:
- replaced hardcoded map translation logic with a dynamic call system using `PHONE_CALLS_DATA`
- introduced `_check_conditions()` method to evaluate map and variable-based conditions for each NPC
- added `_start_call()` method to handle dialogue selection based on matched conditions
- updated `choice()` to store the selected NPC slug and trigger a confirmation dialog before initiating the call
- improved fallback behavior by using default dialogue when no conditions match
- preserved relationship display (type and strength) in the contact list
- ensured compatibility with existing UI styling and menu layout

### DIFF
--- a/mods/phone_calls.yaml
+++ b/mods/phone_calls.yaml
@@ -1,0 +1,60 @@
+# PHONE_CALLS_DATA structure for NPC interactions via NuPhone.
+# Each NPC (e.g., npc_example) defines how phone calls behave based on game state.
+# The system checks each entry in 'conditional_calls' in order. The first matching
+# condition is used.
+# If no conditions match, the 'default' dialogue is used.
+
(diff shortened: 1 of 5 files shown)


Sanglorian:
Good stuff, thank you! Will be a good way to prompt new missions for players, as well as be a new way to give players information.

kerizane (approved):

PR #3187, opened 2025-10-08
Fix Linux Startup Failure Caused by PR #3185

JaskRendix:
PR fixes the Linux Startup failure introduced after merging PR #3185. On Linux, the game flat-out refused to start because Pydantic threw a fit over the monster sounds config. This issue could’ve been caught instantly by simply starting the game on Linux.

To clean up it and simplify the flow, `MonsterSoundsModel` is no longer optional. Both parameters inside it are now optional, which actually makes sense. With this fix, the game finally launches properly on Linux again.

### DIFF
--- a/tests/tuxemon/test_monster_actions.py
+++ b/tests/tuxemon/test_monster_actions.py
@@ -55,6 +55,7 @@ class TestMonsterActions(unittest.TestCase):
         height=80,
         weight=24,
         catch_rate=100.0,
+        sounds={},
         lower_catch_resistance=0.95,
         upper_catch_resistance=1.25,
     )
@@ -74,6 +75,7 @@ class TestMonsterActions(unittest.TestCase):
         height=45,
         weight=4,
         catch_rate=100.0,
+        sounds={},
         lower_catch_resistance=0.95,
         upper_catch_resistance=1.25,
     )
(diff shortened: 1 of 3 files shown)


Sanglorian:
Thanks for fixing this, @JaskRendix . What was the source of the error? I wouldn't have expected adding call details to monsters that didn't already have them to break anything, so I'm glad you've caught it and fixed it.

JaskRendix:
@Sanglorian the hiccup came from how `MonsterSoundsModel` was originally set up, it expected both `combat_call` and `faint_call` to be present, no exceptions. Basically, it wasn’t cool with having just a `combat_call` or just a `faint_call` on its own. All good now though!

Sanglorian:
> @Sanglorian the hiccup came from how `MonsterSoundsModel` was originally set up, it expected both `combat_call` and `faint_call` to be present, no exceptions. Basically, it wasn’t cool with having just a `combat_call` or just a `faint_call` on its own. All good now though!

Ah makes sense, thanks! 

EDIT: Also I'll go in soon and make sure every monster has both combat_call and faint_call

JaskRendix:
@kerizane here's a fix. If you can merge a few more, like the one about the shop, that’d help me get the rest ready for review.

kerizane (approved):

PR #3194, opened 2025-10-11
Fix Linux Startup Failure Caused by PR #3192 + fix remove after PR #3116

JaskRendix:
PR fixes the Linux startup failure introduced after merging PR #3192. On Linux, the game refused to launch because **a required audio file was missing**, which caused Pydantic to throw an exception. This wasn’t a configuration issue, it was a missing file.

The crash could’ve been caught instantly by simply starting the game on a Linux system.

**sound_glitch_call_faint** doesn't exist
**sound_whistle_faint** doesn't exist
**sound_wood_call_faint** doesn't exist
**sound_bugthing_faint** doesn't exist
**sound_robot1_faint** doesn't exist
**sound_frog_faint** doesn't exist

so I simply removed the _faint and used the main one.

This PR also fixes the broken behavior of the `remove` technique effect introduced in PR #3116. That PR added support for `"positive"` and `"negative"` status categories, but failed to account for how `self.status` is evaluated. Since `self.status` is a string and not a category object, the effect was bypassing `current_status` and skipping the category check. As a result, the technique became useless when targeting `"positive"` or `"negative"` statuses.

This fix restores the intended behavior by:
- retrieving `current_status` from the target monster
- checking whether the status category matches `"positive"` or `"negative"` before removing it
- ensuring that explicit status slugs and `"all"` still work as expected

The effect now behaves correctly across all status types and target configurations.


JaskRendix:
If you're still around, could you please merge this? The issue is blocking all my progress and I can't move forward until it's resolved. @kerizane

la-xirsoi (approved):

Sanglorian:
Sorry @JaskRendix -- I should have tested it better. I'll mark my future PRs as drafts so you have a chance to look at them first.

PR #3201, opened 2025-10-14
Fix #2981: Correct battle tracking and enhance event scripting

### LINKED ISSUE #2981: If you lose to a trainer, you sometimes do not have to fight them again

JaskRendix:
PR fixes #2981 and introduces two key improvements:
* adds a new event condition `BattleOutcomeCountCondition` that checks whether a character has achieved a specific battle outcome a given number of times against an opponent. This is especially useful for handling repeat battles with NPCs without relying on multiple variables (`is battle_outcome_count <fighter>,<outcome>,<opponent>,<count>`)
* updates the `teleport_faint` event action to trigger healing **only after** the character has arrived at the destination map. This resolves the issue where the dialog box would briefly appear and close if the player lost to a trainer who initiated the battle by walking toward them.
* fixes a bug in battle tracking where the opponent slug was incorrectly recorded during losses and draws. This caused outcome-based conditions to fail when checking results against specific NPCs. The fix ensures accurate opponent attribution for all battle outcomes.
* removed duplicate Tru NPC from Dragon's Cave, the Nimrod Boss version will now be used exclusively to avoid redundancy (@Sanglorian)
* swapped Billie’s encounter parties between the Dojo and Route 6, the stronger team now appears at Route 6, balancing progression difficulty (@Sanglorian)
* added a gatekeeper NPC at Nimrod to prevent players from skipping Route A, the Mansion, and the Dojo, this blocker will disappear once the Dojo is completed (@Sanglorian)
* as a result, removed a duplicate `set_variable` from Route6 that was previously added to ensure Dojo completion

### DIFF
--- a/docs/handcrafted/condition_list.rst
+++ b/docs/handcrafted/condition_list.rst
@@ -1,3 +1,4 @@
+.. autoscriptinfoclass:: tuxemon.event.conditions.battle_outcome.BattleOutcomeCountCondition
 .. autoscriptinfoclass:: tuxemon.event.conditions.battle_outcome.BattleOutcomeCondition
 .. autoscriptinfoclass:: tuxemon.event.conditions.bill_exists.BillExistsCondition
 .. autoscriptinfoclass:: tuxemon.event.conditions.bill_is.BillIsCondition
--- a/mods/tuxemon/db/npc/spyder_dragonscave_npcs.json
+++ b/mods/tuxemon/db/npc/spyder_dragonscave_npcs.json
@@ -201,25 +201,6 @@
     }
   },
   {
-    "slug": "spyder_dragonscave_tru",
-    "speech": {
-      "profile": {
-        "default": {
-          "pre_battle": "spyder_dragonscave_tru1",
-          "post_battle_win": null,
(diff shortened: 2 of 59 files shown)


Sanglorian:
Nice, thanks a lot for this! This is an important bug fix for battle logic, as well as fixing those narrative problems.

la-xirsoi (approved):

PR #3206, opened 2025-10-17
Support for Persistent Status Effects After Combat

JaskRendix:
PR introduces support for status effects that can persist beyond the end of combat.

The changes includes:
- added a new base model `StatusBehaviors` class
- added a new boolean field `persists_after_combat` to the `StatusBehaviors` class
- this field is also included in the status JSON definition to allow configuration via data
- the default value for `persists_after_combat` is set to `False`, meaning status effects will continue to be removed after combat unless explicitly marked otherwise (`poisoned` will probably be marked later)
- updated the `end_combat` method to check this flag before removing a status, if `persists_after_combat` is `True`, the status will remain active after combat ends
- removed the old `check_counter_expiry` method from the `Status` class and introduced `is_use_expired` in `Status` to check use-based expiration without performing removal  
- added `check_and_clear_use_expiry` method in `MonsterStatusHandler` to encapsulate both the check and the removal logic and updated combat flow to use `check_and_clear_use_expiry` instead of directly checking or clearing status  
- centralized expiry logic in `MonsterStatusHandler`, keeping `Status` focused on state tracking, reason is separation of concerns

### DIFF
--- a/mods/tuxemon/db/status/blinded.json
+++ b/mods/tuxemon/db/status/blinded.json
@@ -1,6 +1,7 @@
 {
   "animation": null,
   "modifiers": [],
+  "behaviors": {},
   "category": "negative",
   "effects": [
     {
--- a/mods/tuxemon/db/status/burn.json
+++ b/mods/tuxemon/db/status/burn.json
@@ -16,6 +16,7 @@
       "multiplier": 2.0
     }
   ],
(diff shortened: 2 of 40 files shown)


Sanglorian:
Great stuff, thank you!

la-xirsoi (approved):

PR #3226, opened 2025-10-24
Updating techniques to match the wiki

Sanglorian:
There were a handful of techniques that had different rules in GitHub than on the wiki. For a bunch, I thought the GitHub rules made more sense so I updated the wiki. 

These are the ones I thought the wiki made more sense than the GitHub -- but please let me know if you disagree!

* **Energy Beam** – accuracy reduced to `95%`, Potency reduced to `50%`, Power set to `3`, and recharge raised to `4`. Having a Power value makes sure it does damage, matching its effect. 
* **Fire Shield** – Accuracy and Potency raised to `100%` each and power set to `0` to reflect its role as a condition imposer. 
* **Fluff Up** – Potency increased to `100%`, as otherwise I think this would be quite weak.
* **Lust** – Power lowered to `0` as this is listed as a condition imposer
* **Piece of Cake** – Healing Power increased to `0.5`. 0.1 felt quite low to me. 
* **Solar Synthesis** – Light tag added (Plant already on both wiki and on GitHub)

There are also several techniques that had Potency, but since they don't impose conditions, the Potency had no effect so I've set it to 0: Adamantine, Ice Storm, Ruby, Saber, Strangulation and Tux Attack.


HippasusTwo:
Those changes seem good to me.

la-xirsoi (approved):

PR #3228, opened 2025-10-24
Plus and minus icons to indicate warm and cold tastes

Sanglorian:
I didn't quite implement the monster_info passport style display properly, and so the plus and minus icons to indicate which stats a monster's warm and cold tastes affect weren't showing up. This should fix it.


tuxemon/states/monster_info.py
```diff
@@ -381,9 +383,14 @@ def add_menu_items(
         # Helper: find which stat a taste affects
         def get_stat_for_taste(slug: str) -> str | None:
             taste = lookup_tastes.get(slug.lower())
-            if not taste or not taste.modifiers:
+            if not taste:
                 return None
```
JaskRendix, line 387: Everything looks good, but this original version was already spot-on:
```python
if not taste or not taste.modifiers:
    return None
```
Gotta check this, no modifiers means things break. Then set it ready.

Sanglorian: Perfect, thank you!

JaskRendix (approved):

kerizane (approved):

PR #3231, opened 2025-10-24
Enhance `Modifier` System with Conditions, Stacking and Extensibility

JaskRendix:
PR enhance `Modifier` System:
- introduced support for conditional modifiers using a new `CONDITION_REGISTRY` (e.g., `hp_below_50`, `full_hp`)
- added `turns_remaining` logic to allow time-limited modifiers, with automatic decrement (`tick_turns`) and cleanup (`remove_expired_modifiers`)
- implemented `StackingMode` logic to support additive, multiplicative and override stacking behaviors in cumulative calculations
- added `priority` sorting and `max_stacks` enforcement to control modifier application order and limits
- refactored attribute handling into a registry (`ATTRIBUTE_HANDLER_REGISTRY`) with support for custom handlers via `register_handler`
- extracted `preview_multiplier()` as a standalone function to simulate modifier effects without altering internal state
- maintained backward compatibility with existing modifier modes

It'll allow stuff like this (example):
```json
{
  "attribute": "type",
  "values": ["frost"],
  "multiplier": 1.5,
  "priority": 7,
  "stacking": "multiplicative",
  "max_stacks": null,
  "condition_name": "hp_below_50",
  "source": "Frozen Fury",
  "turns_remaining": 2
}
```
Explanations:
- `"type"` this modifier targets tuxemon based on their elemental type
- `["ice"]` it applies specifically to frost-type tuxemon.
- `1.5`  when active, it increases damage dealt to frost-type tuxemon by 50%.
- `7` determines how this modifier ranks among others, higher priority modifiers are considered first when stacking limits apply
- `"multiplicative"` this modifier multiplies the damage rather than adding to it, if multiple modifiers apply, their effects compound
- `null` no limit on how many similar modifiers can stack, all applicable ones will be used
- `"hp_below_50"` this modifier only activates if the target tuxemon has less than 50% of its HP remaining
- `"Frozen Fury"` free text or source (e.g. technique, item, whatever)
- `2` the modifier will expire after 2 turns, making it a temporary effect

### DIFF
--- a/mods/tuxemon/db/status/burn.json
+++ b/mods/tuxemon/db/status/burn.json
@@ -6,14 +6,16 @@
       "values": [
         "fire"
       ],
-      "multiplier": 0.0
+      "multiplier": 0.0,
+      "source": "fire immunity"
     },
     {
       "attribute": "type",
       "values": [
         "frost"
       ],
-      "multiplier": 2.0
+      "multiplier": 2.0,
+      "source": "fire extra damage"
     }
   ],
   "behaviors": {},
--- a/mods/tuxemon/db/status/poison.json
+++ b/mods/tuxemon/db/status/poison.json
@@ -6,7 +6,8 @@
       "values": [
         "venom"
       ],
-      "multiplier": 0.0
+      "multiplier": 0.0,
+      "source": "poison immunity"
     }
   ],
   "behaviors": {},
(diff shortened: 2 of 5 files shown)


Sanglorian:
Oh wow, this sounds like a very powerful tool -- it looks like it could be used for things like terrain and weather effects?

JaskRendix:
Yes, it could be used.

kerizane (approved):

Issue #3229, opened 2025-10-24
Jingles for victory in battle and healing


Sanglorian:
**Is your feature request related to a problem? Please describe.**

After playing Tuxemon on their stream, Rexy Goblina pointed out there's no sound effects for winning a battle or getting your team healed at a Cafe, by a nurse, etc. 

I think these would be cool additions -- and did some scouting on OpenGameArt for sound effects. 

I like [Dizzy Crow's CC0 licensed VictorySmall](https://opengameart.org/content/8-bit-sound-fx) for winning a battle.

[VictorySmall.wav](https://github.com/user-attachments/files/23123589/VictorySmall.wav)

I'm less sure on NenadSimic's Picked Coin Echo 2 for the team healing sound effect, but I do like that it sounds a bit like a bell and a bit like a machine. [CC BY on OpenGameArt](https://opengameart.org/content/picked-coin-echo-2).

What do people think? Would this be feasible to implement, and are these the right sound effects if so?

HippasusTwo:
Not too bad sounding. Could work.
An alternative heal jingle is the first one of these, it's a bit softer sounding, but not too different:
https://opengameart.org/content/3-heal-spells
And a search for victory pulls up a lot of different jingles, if someone wanted to compare them:
https://opengameart.org/art-search?keys=victory
What might be a nice touch is to have a different victory end music/jingle for bosses versus normal battles?

Sanglorian:
> Not too bad sounding. Could work. An alternative heal jingle is the first one of these, it's a bit softer sounding, but not too different: https://opengameart.org/content/3-heal-spells And a search for victory pulls up a lot of different jingles, if someone wanted to compare them: https://opengameart.org/art-search?keys=victory What might be a nice touch is to have a different victory end music/jingle for bosses versus normal battles?

Nice, thanks @HippasusTwo -- I'm happy to go with any tone (I'll leave as is for now but feel free to swap out in a PR), and I agree a different victory jingle could be fun.

PR #3232, opened 2025-10-25
Adding chime when healing

Sanglorian:
This adds a jingle that plays when a player heals their team at an "official" station - the inn, the Cathedral or a cafe. Only for the Spyder campaign. 

So far, I haven't added the jingle when the player is healed either by sleeping in their own bed or directly by an NPC. But I'm very happy to do so if people think it'd be a useful addition -- I just figured that the places where I added feel more like a "machine" is being used. 

Partially addresses #3229


mods/tuxemon/gfx/ui/background/spyder_tumble.png
kerizane, line 1: Why is this changing?

kerizane (changes requested):

mods/tuxemon/gfx/ui/background/spyder_tumble.png
Sanglorian: Ah, while I was testing the new jingles I noticed this was blurry -- a problem we have on Windows where images that aren't indexed are blurry when resized. It should be a sharp image now.

la-xirsoi (approved):

kerizane (approved):

PR #3237, opened 2025-10-26
Fixing condition damage modifiers

Sanglorian:
When a Fire mon like Agnite has the Burn condition, they still take damage -- although it should be set to zero. I believe this fixes it -- pointing to the element slug.


JaskRendix:
No, it doesn't fix it. This is fixed in #3231

Sanglorian:
> No, it doesn't fix it. This is fixed in #3231

Great! I'll close this one

PR #3239, opened 2025-10-30
October quality of life changes

Sanglorian:
Have been working on a bunch of changes, none of them too major I think although they do build up! I'll try to do smaller more frequent QOL pull requests in future. 

**Misc**
* Adding attribution for some files that were missing it
* Monster menu background now the right dimensions
* Slightly lengthened combat animations to 0.1 second per frame (from 0.09).
* Element symbols don't overlap if a technique has two elements (in combat menu)

**Spyder gameplay changes**
* Swapping out TM Cavity for TM Acid. I made a mistake choosing Cavity because, as a technique that's limited to Gliffary, it's not really useful as a TM. 
* Plague text no longer all-caps the name of the tuxemon it affects (matches other combat messages which are normal capitalisation)
* The Move Re-Learner in the Dojo no longer allows you to learn a technique that you couldn't have learned through levelling up yet (e.g. if you are level 15, you can't learn a technique that you'd normally learn at level 20). Although I think I went too far on this as I'd still like techniques learned on evolution to be learnable. 

**Spyder map changes**
* Changing Shopkeeper to use shopkeeper sprites rather than shop assistant - stops it looking like there's two shop assistants with no differences between them. 
* Since we have 13 elements now, referring to the Dojo of the Elements not the Dojo of the Five Elements
* A few spacing changes so text fits in the box when using line spacing. 
* Clarifying why there are two de-evolution potion options 
* Green shadows beneath fences when those fences are over grass (just on a couple of maps as I came across grey shadows)
* Blackboards in the Dojo now teach the lesson of the masters' near them. 
* Dojo battles match their stated aims (as the game mechanics changed, some of these combats were out-of-date)
* "Vivipere" pop up appears when acquiring a Vivipere from the Pet Shop
* You can no longer walk into the Riverboat Captain's space
* Removed a blocking trash bin from Paper Town as it forced all the appearing NPCs to walk around it, and it was empty
* Fixed stuck loop for QQQ in Route 3

Let me know what you think!


mods/tuxemon/maps/spyder_candy_port.yaml
```diff
@@ -3,6 +3,7 @@ events:
     actions:
     - create_npc spyder_captain,37,8
     - char_face spyder_captain,left
+    - add_collision riverboat_captain_block,37,8
```
JaskRendix: why this? creating the npc adds the collision, fixed

la-xirsoi (approved):

PR #3240, opened 2025-10-30
Tweaks to condition and technique effects

Sanglorian:
Some suggested changes to bring conditions and technique effects closer to their descriptions: 

* The money effect takes a formula that already includes the damage multiplier, then applies it again -- I believe only one is needed. 
* If a monster's technique was interrupted due to the Charmed effect, it would still hit if it had perfect accuracy (100%) -- this removes that so 100% accuracy techniques also miss. 
* Nodding Off is now a 50% chance to shake off each turn, or 100% chance after 5 rounds. 

And I've just changed the wording when a technique is lost due to Spyder Bite because I thought it was a bit confusing to say "It missed" just after the sneeze -- sounded like the sneeze missed rather than the original technique.


JaskRendix:
> The money effect takes a formula that already includes the damage multiplier, then applies it again -- I believe only one is needed.

Go ahead and apply it. We’ll just whip up a better formula later (money without damage involved)

> Nodding Off is now a 50% chance to shake off each turn, or 100% chance after 5 rounds.

I'm not a fan, but hey, if you’re into it, at least remove it from `status.py`. No point in keeping dead code around.
```
    def has_reached_duration(self) -> bool:
        """Checks if the status has reached or exceeded its duration."""
        return self._nr_turn >= self.duration > 0
```

Sanglorian:
> > The money effect takes a formula that already includes the damage multiplier, then applies it again -- I believe only one is needed.
> 
> Go ahead and apply it. We’ll just whip up a better formula later (money without damage involved)
> 
> > Nodding Off is now a 50% chance to shake off each turn, or 100% chance after 5 rounds.
> 
> I'm not a fan, but hey, if you’re into it, at least remove it from `status.py`. No point in keeping dead code around.
> 
> ```
>     def has_reached_duration(self) -> bool:
>         """Checks if the status has reached or exceeded its duration."""
>         return self._nr_turn >= self.duration > 0
> ```

Perfect, thanks!

la-xirsoi (approved):

PR #3253, opened 2025-11-04
Battle Mode Integration & Map/Camera System Enhancements

### LINKED ISSUE #2316: Add a challenge mode where you only battle
In a way, yes. The main game seems mostly playable, though last time I tried I ran into enough problems to stop playing the Cathedral campaign. 

It seemed like most of my issues were with the map interactions but the battles worked well, so I was imagining creating a challenge mode which is a series of fights where you have a set of tuxemon and you battle different trainers. You can inspect/read about the enemy tuxemon (levels, moves, etc) before choosing the challenge. Then you don't care if the rest of the game (map, evolutions, all the moves, all the tuxemon, etc) are finished or working.

The benefits of doing this: You can present a complete game that is fun to play, but small enough to be polished. Some people would enjoy the puzzle of winning battles! It would depend on way fewer features working. All the interaction with your character on the map (evolution, fishing rod, walking around, catching new 'mons, etc) wouldn't need to work. Yes, I know that walking around the map is part of the fun and storyline! 

What would this entail: adding a challenge mode to the main menu (I already did this, it was easy). Make a menu screen for selecting which challenge to attempt. Make a screen so you can inspect the challenge (view your tuxemons, view their tuxemons) before attempting it. After selecting a challenge, start the battle with the specific monsters that you have and possibly items. Have a victory screen or lost screen that will kick you back to the select a challenge screen. If you win the challenge, unlock the next challenge. I think 20 challenges would be enough to be satisfying, but we could start with like 5 of course. Also, we could simply reuse encounters with trainers that are already in the game. 

Possible extensions: turn the linear challenge into a branching challenge tree where players can choose which challenge to attempt next. Perhaps you may select four to six monsters to start with, then they all will gain some amount of experience after each battle and your goal is to defeat the final boss/challenge (almost like elite 4 style). There may be perhaps four or eight different battles you'd have to face. Perhaps add terrain types in the branches of the trees that require you to have a specific kind of monster in order to move across it, such as you must have a water monster in order to move across water.

**Additional context**
Code that may be helpful

Random encounter might be easier to start with
tuxemon/event/actions/random_encounter.py:24:    Randomly starts a battle with a monster defined in the "encounter" table

Random battle with npc
tuxemon/event/actions/random_battle.py:26:    Start random battle with a random npc with a determined
I tried to start this action from the cli, but I haven't figured out how to create or load a world state yet....

start battle
tuxemon/event/actions/start_battle.py:22:    Start a battle between two characters and switch to the combat module.

Start a trainer battle from command line!
tuxemon/cli/commands/trainer_battle.py:15:class TrainerBattleCommand(CLICommand):
I thought this would be helpful, but I don't know how to invoke it...

JaskRendix:
PR fixes #2316 and introduces a new Battle option to the main menu, allowing players to engage in combat without starting a full campaign. To support this, several structural improvements have been made:

Battle Mode Basics:
- Players can now start a battle directly from the main menu
- Difficulty levels influence both the number and level of monsters
- No items (e.g., potions) are available yet, but future updates will expand battle mechanics and inventory options

Map System Updates
- Introduced `NullMap`, a safe fallback map used when no campaign map is loaded
- Updated `MapTransition` to accept `None` as a valid `map_name`, automatically loading `NullMap` in that case
- Reflected the same logic in `WorldState`, where `map_name` is now an optional string. The renderer still checks for `None` to determine whether to initialize

Camera Manager Enhancements
- Added `remove_camera()` to allow dynamic removal of cameras
- Added `reset()` to fully clear all cameras and reset the active state

New Event Action: `QuitWorldAction`
- Added a new event action called `QuitWorldAction`
- Allows players to exit the current world state and return to the main menu without quitting the game

https://github.com/user-attachments/assets/5975604e-954d-4f7f-970e-527d611d7aec


Sanglorian:
I'm excited for this, a good way to test the ruleset as a whole.

la-xirsoi (approved):

PR #3255, opened 2025-11-06
IntroState

JaskRendix:
PR adds `IntroState`. After the `Splash` will appear this, then the player presses any key and the `StartState` appears

https://github.com/user-attachments/assets/191760d3-c016-49b7-8099-9d846eda0561


la-xirsoi:
there has gotta be a better way to do this than adding every frame as a separate image file.

JaskRendix:
@la-xirsoi I wish it did, but yeah… Pygame doesn’t do GIFs. Thankfully it’s just the intro, so we only need one.

la-xirsoi (approved):

PR #3265, opened 2025-11-14
Add active effect tracking and staged update support for techniques, items, and statuses

JaskRendix:
Summary of changes:
- added three new lists to the base client to track currently active effects: `self.active_techniques: list[Technique] = []`, `self.active_items: list[Item] = []` and `self.active_statuses: list[Status] = []`
- extended `CoreEffect` with default lifecycle methods: `update(self, session: Session, dt: float)` now exists as a stub, allowing effects to implement staged behavior over time and `is_finished(self) -> bool` now exists as a stub, returning `True` by default, so effects can signal when they are complete
- updated `EffectProcessor` to support staged updates: added `update(session, dt)` which iterates over all managed effects and calls their `update` method and added `is_finished()` which checks all managed effects and returns `False` if any are still running
- applied the same staged update pattern to `FishingEffect`: fishing now progresses through phases (CAST, WAIT, BITE, ENCOUNTER, DONE), this makes it possible to insert animations or timed transitions between phases and as a result, fishing outcomes are no longer immediate; the player experiences the staged sequence before the final result
- applied the same staged update pattern to `DroneEffect` and `BivouacEffect`: both effects now progress through defined phases rather than resolving instantly, this makes it possible to add animations, delays, or dialog to enhance immersion and outcomes are no longer immediate; the player experiences the staged sequence before completion

### DIFF
--- a/mods/fishing.yaml
+++ b/mods/fishing.yaml
@@ -1,27 +1,31 @@
 # Fishing Config Format
 # Each key is the slug of a fishing item (e.g., "fishing_rod", "neptune", "poseidon")
 # Fields:
-#   trigger: float                  # Probability of triggering a fishing encounter (0.0–1.0)
-#   lower_bound: int            # Minimum level of the fished monster
-#   upper_bound: int            # Maximum level of the fished monster
-#   exp_req_mod: float          # Experience modifier for the encounter
-#   stages: list[str]            # Allowed monster stages (e.g., basic, stage1, stage2, standalone)
-#   stage_weights: dict[str, float]   # Optional: weight per stages (e.g., basic: 0.5)
-#   shapes: list[str]            # Allowed monster shapes (e.g., piscine, polliwog)
-#   shape_weights: dict[str, float]   # Weight per shapes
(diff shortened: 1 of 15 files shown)


Sanglorian:
> * applied the same staged update pattern to `FishingEffect`: fishing now progresses through phases (CAST, WAIT, BITE, ENCOUNTER, DONE), this makes it possible to insert animations or timed transitions between phases and as a result, fishing outcomes are no longer immediate; the player experiences the staged sequence before the final result
> * applied the same staged update pattern to `DroneEffect` and `BivouacEffect`: both effects now progress through defined phases rather than resolving instantly, this makes it possible to add animations, delays, or dialog to enhance immersion and outcomes are no longer immediate; the player experiences the staged sequence before completion

This is cool! Did you want me to look to see if there's any suitable fishing animations on OpenGameArt? (Water ripples, for example)

JaskRendix:
> This is cool! Did you want me to look to see if there's any suitable fishing animations on OpenGameArt? (Water ripples, for example)

Later on we can layer in more stuff, but for now I’m picturing quick little sound effects.

la-xirsoi (approved):

PR #3268, opened 2025-11-15
Add `bypasses_selection` flag to `TechBehaviors` + fix double battle loading

JaskRendix:
waiting 
- [x] #3262 

PR introduces a new `bypasses_selection` flag in `TechBehaviors` to simplify and unify the logic for techniques that skip target selection. Previously, the `initialize_items` method contained hard‑coded checks for `aether` type techniques and `meta` sort techniques. This made the targeting rules harder to extend and maintain, especially in the new 2vs2 combat mode where target selection is visible.

It finally sorts out the double battle setup, before, only one monster was showing up, now both load correctly. On top of that, I cleaned up the `CombatMenuState` by dropping a bunch of pointless `hasattr` checks. They were redundant anyway, since those attributes already existed. That little cleanup shaved off several unnecessary lines and makes the class leaner and easier to read.


Sanglorian:
Thanks so much for this! 

Is 2v2 at a point where I could add it into the Spyder campaign, do you think?

JaskRendix:
> Is 2v2 at a point where I could add it into the Spyder campaign, do you think?

@Sanglorian I need to fix still a small issue in 2vs2, then I'll tell you

la-xirsoi (approved):

PR #3270, opened 2025-11-15
Refactoring Environment Music Handling

### LINKED ISSUE #3229: Jingles for victory in battle and healing

JaskRendix:
waiting 
- [x] #3262 

PR fixes #3229. Heads up: we’ll still need a follow‑up PR to add the same attribute to NpcModel, so the music can also change depending on the opponent (override)

Changes:
- replaced `battle_music` string field with a structured `BattleMusicModel`  
- added support for three distinct configurations: battle, victory, and defeat music  
- introduced `MusicProperties` to store both filename and playback volume  
- included validation to ensure referenced music exists in the database  
- updated `EnvironmentModel` to reference `BattleMusicModel` instead of a plain string  
- provides clearer semantics and flexibility for retrieving and playing different outcome tracks


Sanglorian:
Brilliant, thanks for this!

la-xirsoi (approved):

PR #3277, opened 2025-11-22
Introduce asynchronous scheduler with mobile and historical extensions

JaskRendix:
PR replaces the synchronous scheduler with an asynchronous design and adds specialized variants. The synchronous scheduler was limited to short, CPU‑bound tasks and blocked execution during long operations. 

Key changes:
- added `AsyncScheduler` supporting coroutine tasks via `asyncio`
- tasks can be scheduled with delay, repeat, and soft scheduling options
- exceptions inside tasks are caught and logged without stopping the scheduler
- implemented `tick_rate` method to maintain target tick frequency and calculate FPS
- added `MobileAsyncScheduler` with active/inactive state handling and background FPS limiting
- introduced catch‑up logic for tasks after inactivity in mobile mode
- added `HistoricalAsyncScheduler` for deterministic, virtual‑time simulations and replaying events

### DIFF
--- a/tuxemon/clock.py
+++ b/tuxemon/clock.py
@@ -2,414 +2,235 @@
 # Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
 from __future__ import annotations
 
+import asyncio
 import time
-from collections import deque
-from collections.abc import Callable
-from heapq import heapify, heappop, heappush, heappushpop
-from typing import Any, Optional, Union
-
-__all__ = ("ScheduledItem", "Scheduler", "Clock")
+from collections.abc import Awaitable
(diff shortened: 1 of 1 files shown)


tuxemon/clock.py
```diff
@@ -2,414 +2,235 @@
...
-from heapq import heapify, heappop, heappush, heappushpop
-from typing import Any, Optional, Union
+from collections.abc import Awaitable
+from heapq import heapify, heappop, heappush
+from typing import Callable, Optional
 
-__all__ = ("ScheduledItem", "Scheduler", "Clock")
-
-
-class ScheduledItem:
```
la-xirsoi, line 14: I notice you effectively remove this class, but haven't changed any other files.
Are you certain the renamed/removed classes aren't referenced anywhere else?

PR #3280, opened 2025-11-24
Faction: Add Update Loop, Event Bus Integration, Rank Evaluation and Extended Tests

JaskRendix:
Summary of changes:
- added `update` method to `FactionManager` to handle periodic faction updates, reputation decay, and rank changes
- integrated `EventBus` into `Faction` and `FactionManager` to publish events for member joins/removals, relation changes, promotions, and degradations
- introduced `evaluate_rank_change` in `Faction` to unify promotion and degradation logic, reducing duplication and improving clarity
- adjusted `FactionManager.update` to use `evaluate_rank_change` instead of separate promotion/degradation calls
- extended test suites for both `FactionManager` and `Faction`


tuxemon/faction/faction.py
```diff
@@ -16,8 +16,19 @@
 )
 from tuxemon.locale import T
 
+if TYPE_CHECKING:
+    from tuxemon.event.eventbus import EventBus
+
 logger = logging.getLogger(__name__)
 
+FACTION_EVENTS = {
```
la-xirsoi: I think this would work better as an enumerator. We'd get some type-safety and no one has to remember what the valid keys are.

tuxemon/faction/faction.py
```diff
@@ -16,8 +16,19 @@
...
+    "RELATION_CHANGED": "faction_relation_changed",
+    "MEMBER_JOINED": "faction_member_joined",
+    "MEMBER_REMOVED": "faction_member_removed",
+    "PROMOTED": "faction_npc_promoted",
+    "DEGRADED": "faction_npc_degraded",
```
la-xirsoi: It's not very important, but the correct English word for someone losing a rank is "Demoted"

la-xirsoi (changes requested):

tuxemon/faction/faction.py
JaskRendix: Ok

tuxemon/faction/faction.py
JaskRendix: True, thanks for the suggestion

la-xirsoi (approved):

PR #3281, opened 2025-11-25
Refactor `Tuxepedia` into Manager/Data/Reporter Structure

JaskRendix:
Summary of changes
- split responsibilities into three classes: **TuxepediaData** which handles storage and direct access to monster entries, **TuxepediaManager** which provides mutation operations (add, remove, reset) and integrates with the event bus for publishing changes and **TuxepediaReporter** which provides reporting and analytics (most frequent monsters, completeness, distribution)
- introduced event publishing for key actions: monster added, updated, removed, and Tuxepedia reset
- improved reset behavior: option to remove only "seen" monsters or clear all entries, reset now reports removed and remaining counts, plus details of removed monsters
- enhanced status update logic: prevents downgrading from "caught" to "seen" and tracks whether status actually changed


tuxemon/states/pc_kennel.py
```diff
@@ -193,6 +193,7 @@ class MonsterTakeState(PygameMenuState):
     into your current party."""
 
     name: ClassVar[str] = "MonsterTakeState"
+    swap_target: Optional[Monster]
```
la-xirsoi: I don't see where this gets used.

JaskRendix: so at line 229 we’ve got:

```python
swap_target = self.swap_target
if swap_target:
    actions = {
        "swap": lambda: handler.swap(mon, swap_target),
    }
```
mypy threw a fit:  
```
tuxemon/states/pc_kennel.py:228: error: Cannot determine type of "swap_target"  [has-type]
```
the real issue? I hadn’t set up `swap_target` in `__init__` before using it here. Basically, the initializer needs to declare the attribute first, otherwise mypy doesn’t know what type it’s supposed to be. In short: the init has to come before the method, but here it wasn’t.

fixed pc_locker too, same issue, `__init__` after.

la-xirsoi (approved):

PR #3285, opened 2025-11-27
Add `current_map` attribute to Entity/NPC and remove redundant map lookups

### LINKED ISSUE #3323: Enum vs string mismatch
```
  File "/home/giorgio/Tuxemon/tuxemon/combat/session.py", line 434, in update_tuxepedia
    other_player.tuxepedia.add_entry(monster.slug)
  File "/home/giorgio/Tuxemon/tuxemon/tuxepedia.py", line 217, in add_entry
    f"Updated monster {monster_slug}: entry={entry}",
                                            ^^^^^^^
  File "/home/giorgio/Tuxemon/tuxemon/tuxepedia.py", line 65, in __str__
    f"MonsterEntry(status={self.status.name}, "
                           ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'name'
```

JaskRendix:
Fixes #3323

Previously, NPCs didn’t have their own `current_map` attribute and we relied on `session.client.get_map_name()` or similar indirection through `MapManager` to figure out where an NPC was. This worked fine while everything was in memory, but it became awkward once we started saving and restoring NPC state. The NPC itself didn’t know which map it belonged to, so we had to query global state every time.

PR adds a simple `current_map` attribute directly to `Entity` so each character can track its own map slug. With that in place, we can drop methods like `get_event_map` and avoid calling into the client for map names. It makes serialization cleaner, keeps NPC state self‑contained, and ensures that when we save or load NPCs their map assignment is explicit rather than inferred.

Found `pos_update` copy‑pasted in `NPC`. Same code, same effect, same waste of space. So I ripped it out and shoved the logic back into `Entity` where it belongs.  

Also locked down `tile_pos` as private and gave it a proper `set_tile_position` method. Why? Because I’d rather have one obvious entry point than play “guess which random setter silently mangled my state” during debugging.


tests/tuxemon/test_path_controller.py
```diff
@@ -38,6 +38,9 @@ def set_position(self, pos):
         self.tile_pos = pos
         self.position = Vector2(float(pos[0]), float(pos[1]))
 
+    def set_tile_position(self, pos):
+        self.tile_pos = pos
+
```
AndyMender: `set_position` is also used to set `tile_pos`, but it also modifies the `position` attribute. Any risk of redundancy between the methods? Also, perhaps refactoring `set_position` to directly call `set_tile_position()` be also in order?

tuxemon/entity.py
```diff
@@ -226,6 +250,24 @@ def set_position(self, pos: Sequence[float]) -> None:
...
+        self._current_map = map_slug
+
+    def set_tile_position(self, tile_pos: tuple[int, int]) -> None:
+        """
+        Set the entity's tile position in the game world.
+
+        Parameters:
+            tile_pos: Tile position to be set.
+        """
+        self._tile_pos = tile_pos
```
AndyMender: Should these not be decorated with the corresponding `@attribute_name.setter` decorators to match the `tile_pos` and `current_map` properties? In that case the later assignments to attributes would invoke the correct setters automatically.

tuxemon/states/phone_contacts.py
```diff
@@ -172,6 +171,11 @@ def __init__(self, character: NPC) -> None:
 
         self.char = character
 
+        if self.char.current_map:
+            self.current_map = self.char.current_map.split(".")[0]
+        else:
+            self.current_map = "???"
```
AndyMender: It might be good to have a helper function for this? There is a similar scenario in `NuPhoneRadioBase`.

AndyMender (commented):
Hey! I added some minor drive-by comments. If they're not relevant, please feel free to ignore :)

tuxemon/states/phone_contacts.py
JaskRendix: These are two distinct states, and forcing them through a common utility would just create unnecessary coupling. Each file should stay independent and declare its own constant for conversion logic.

tuxemon/entity.py
JaskRendix: You’re right to point that out. In my earlier version I leaned on a `set_tile_position` method, which was just a workaround for not having proper property setters in place.

What I actually needed was to make `tile_pos` a dynamic property derived from `body.position`, so that assignments flow through the physics system instead of bypassing it. That way, later code doesn’t directly mutate a cached `_tile_pos` but instead updates the underlying position, and `tile_pos` reflects it automatically.

So yes, my previous approach was a mistake, and the correct fix was to remove `set_tile_position` entirely and rely on the property mechanism to keep things consistent.

tuxemon/states/phone_contacts.py
```diff
@@ -25,6 +25,8 @@
 
 logger = logging.getLogger(__name__)
 
+UNKNOWN_MAP_SLUG = "unknown_map"
```
la-xirsoi: lets have the unknown map slug defined in a single place, rather than multiple places.
do we have a class where we keep constants?

la-xirsoi (changes requested):
Looks pretty good, just need to move that duplicated constant to a central location that can be referenced by whatever needs it!

tuxemon/states/phone_contacts.py
JaskRendix: yes

la-xirsoi (approved):

PR #3297, opened 2025-12-03
Fix: `TextArea` Animation Robustness and Edge Case Handling + `AlertManager`

JaskRendix:
The changes focus on enforcing the integrity of the iterator protocol (`__iter__`/`__next__`) and cleaning up state management within the `text.setter`.

Iterator tweaks
- We now make sure the iterator is properly set up before you try to use it
- That means no more random crashes when `__next__` gets called too soon

Smarter `__next__`
- If you call `__next__` before any text is loaded, it just shrugs and says "nothing to do" instead of blowing up
- If the text animation is already finished and you call `__next__` again, it calmly says "we’re done here"

Cleaner `text.setter`
- First thing it does is check: “Is this the same text as last time?” If yes, it skips restarting the animation
- If the new text is empty, it wipes the display right away and stops drawing, so you don’t get leftover bits hanging around

AlertManager
- Introduced a new `SplitAlertState` class to encapsulate multi‑line dialog state (`dialog_lines`, `dialog_index`, `dialog_speed`).
- Removed internal fields `_dialog_lines` and `_dialog_index` from `AlertManager`; state is now managed through `SplitAlertState`.
- Extended `AlertEntry` with a `split_state` field so each alert carries its own multi‑line state.
- Updated methods (`_process_next_alert`, `advance_dialog_line`, `_on_line_complete`, `current_message`) to use `SplitAlertState` instead of raw list/index tracking.
- Added `_animate_next_line` helper to simplify animating the current line from a split alert.
- Improved separation of concerns: `AlertManager` now orchestrates alerts, while `SplitAlertState` owns the details of line progression.
- Behavior for multi‑line alerts is clearer: AlertManager waits for external input (via `DialogState`) to advance lines, instead of auto‑advancing internally.

### DIFF
--- a/tests/tuxemon/test_alert_manager.py
+++ b/tests/tuxemon/test_alert_manager.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0
 # Copyright (c) 2014-2025 William Edwards <shadowapex@gmail.com>, Benjamin Bean <superman2k5@gmail.com>
+import time
 import unittest
 
 import pygame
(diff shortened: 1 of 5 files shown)


JaskRendix:
@kerizane this needs to be merged before the others

la-xirsoi (approved):

PR #3302, opened 2025-12-06
Fix Various and Enhance class properties and database defaults

JaskRendix:
Changes:
- renamed `recharge_length` to `cooldown_duration` to better show it means the total number of turns a technique must wait before being reused
- renamed `next_use` to `current_cooldown` to clarify it tracks how many turns remain until the technique is available again
- updated `db.py` to use `default_factory=list` instead of `[]` for safer default list handling
- added a `rarity` attribute to the `Item` class to represent how common or special an item is
- replaced plain attributes like `name` and `description` with properties (and setters where needed, giving more control over how these values are accessed or changed
- introduced `MenuAction(BaseModel)` to replace the previous `dict[str, str]` structure, validation and clearer semantics for menu-related actions
- added `StepEffectType` enum to support flexible step effects (`NONE`, `FLAT_DAMAGE`, `HP_DAMAGE`, `HP_HEAL`, etc.)  
- replaced legacy `step_damage` field with `step_effect_value` (int) and `step_effect_type` (enum) in `StatusModel`
- introduced `_calculate_step_hp_change()` to return negative values for damage and positive values for healing
- renamed `modifiers` (Monster attribute) to `custom_stats` , it makes more clear and avoid confusion with other modifiers (eg. Taste, Status, etc), the class has been renamed `CustomStatBoosts`  since it's saved
- moved the `recharge_moves` in CombatState, so it recharges the opponent too


JaskRendix:
@kerizane this needs to be merged before the others

la-xirsoi (approved):

PR #3325, opened 2025-12-22
Fix Various

JaskRendix:
Changes:
- dropped `lock_controls` / `unlock_controls` in favor of `SinkState`, letting the red dot stay visible during wait states
- removed redundant type annotation on `cls` in `db.py` for cleaner method signatures
- refactored `NpcTemplateModel.sprite_name` validator to correctly validate single‑sheet overworld sprites based on the actual asset directory structure
- added new technique + animation


mods/tuxemon/db/encounter/37707_town_missing.yaml
```diff
@@ -12,13 +12,15 @@ monsters:
...
+  - 100
   monster: d0llf1n
   variables: []
 - encounter_rate: 0.5
   exp_req_mod: 1
   held_items: []
   level_range:
-  - 666
+  - 1
+  - 100
```
la-xirsoi, line 23: i'm curious about this change and the one for d011f1n.
is it required to have both parts of the range?

if the problem is the level being outside of the allowed range, then you missed some others in this file (like f7u1t3ra above).

JaskRendix: Before this change the field was a `Sequence[int]`, so giving just one value didn’t fail. Now that it’s a `tuple[int, int]`, a single value breaks validation and we also enforce the allowed level range. So this fixes the original issue and makes the whole thing properly checked.

la-xirsoi (approved):

PR #3326, opened 2025-12-23
pytest network

JaskRendix:
waiting: 
- [x] #3321 
- [x] #3322

Changes:
- implemented new tests and fixed some small issues


tuxemon/network/manager.py
```diff
@@ -30,9 +30,9 @@ def initialize(self) -> None:
         self.server = TuxemonServer(self.parent)
         self.client = TuxemonClient(self.parent)
 
-    def update(self, time_delta: float) -> None:
+    def update(self, dt: float) -> None:
         if self.client and self.client.listening:
-            self.client.update(time_delta)
+            self.client.update()
```
la-xirsoi, line 35: this change means the dt param is unused, is that intentional?

tuxemon/network/server.py
```diff
@@ -117,41 +129,34 @@ def shutdown(self) -> None:
...
+            self._handle_timeout_disconnection(cuuid)
+
+    def _handle_timeout_disconnection(self, cuuid: str) -> None:
+        """Internal helper to clean up a timed-out client."""
+        logger.info(f"Client Timeout: {cuuid}")
+        event_data = self.event_factory.create_event(
+            EventType.CLIENT_DISCONNECTED, cuuid
+        )
+        self.server.disconnect_client(cuuid)
+        self.notify_client(cuuid, event_data)
```
la-xirsoi, line 157: How can we notify the timed out client if they've already been disconnected?
Will the notification even get through if they're timed out?

JaskRendix: Yep, exactly. The code was trying to ping a client after their socket was already dead. Since the disconnect message is for everyone else, the fix is just flipping the order: broadcast the `CLIENT_DISCONNECTED` event first, then kill the timed‑out client’s socket, then drop them from the registry

tuxemon/network/manager.py
JaskRendix: Yes, that’s intentional. `dt` isn’t used yet, but I’m keeping it in the signature because upcoming work may rely on it.

la-xirsoi (approved):

la-xirsoi (approved):

PR #3328, opened 2025-12-27
Initialize `CoreAssetManager` at client startup and replace direct instantiation with get_assets()

JaskRendix:
Changes:
- added a global initialization function `init_assets()` to control when the CoreAssetManager is created
- added a `get_assets()` accessor to retrieve the shared CoreAssetManager instance throughout the engine
- removed direct instantiation of CoreAssetManager inside Technique/Item/Status and replaced it with calls to `get_assets()`
- ensured the asset manager is initialized during client startup so all systems can safely access it
- prevented repeated or unnecessary creation of CoreAssetManager instances
- improved clarity of asset manager lifecycle and reduced hidden dependencies
- simplified Technique/Item/Status loading and effect/condition parsing by relying on the shared asset manager
- reduced coupling between Technique/Item/Status and CoreAssetManager construction logic
- refactored `CoreManager` to define its own locally bounded `TypeVar`, removing the dependency on `InterfaceValue` from `tuxemon.plugin`
- improved type safety by binding the new local `TypeVar` to `PluginObject`, making `CoreManager` fully self‑describing in its generic constraints
- eliminated the need for `typing.cast` in plugin managers by aligning generics across `CoreManager` and its subclasses 
- simplified imports and improved encapsulation by removing unnecessary type‑hint‑only imports from the plugin module


la-xirsoi (dismissed):

tuxemon/core/asset.py
```diff
@@ -15,6 +15,23 @@
 from tuxemon.db import LogicCondition, ParameterizableRule
 from tuxemon.plugin import PluginObject
 
+_manager: CoreAssetManager | None = None
+
+
+def init_assets() -> CoreAssetManager:
```
la-xirsoi, line 21: i think we can drop init_assets and have get_assets manage initialization in the case that it hasn't been done yet.

la-xirsoi (changes requested):

tuxemon/core/asset.py
JaskRendix: I’d keep `init_assets()`. That’s where all the slow loading happens and doing that work inside `get_assets()` means the first random call could freeze things. The explicit init also catches mistakes instead of hiding them.

la-xirsoi (approved):

PR #3339, opened 2026-01-03
Add item‑granted statuses, status application methods, and fix steps persistence bug

JaskRendix:
Changes:
- added a new field `granted_statuses` to `ItemModel` to allow items to provide one or more status effects to monsters
- added a method `apply_item_statuses` to `MonsterStatusHandler` to apply statuses granted by held items
- added a method `remove_item_statuses` to `MonsterStatusHandler` to remove statuses when an item is unequipped or changed
- removed the old `remove_status` method, which previously duplicated `clear_status` but suppressed `ON_END` effects; this behavior was unnecessary and inconsistent with the rest of the status lifecycle  
- consolidated all status‑removal logic into `clear_status`, ensuring that every status now ends cleanly and predictably by triggering its `ON_END` phase  
- refactored `apply_status` to simplify transition handling, eliminate duplicated branches, and ensure consistent `ON_START`/`ON_END` behavior across all status transitions  
- clarified the semantics of `ResponseStatus.replaced` and `ResponseStatus.removed` so both now follow the same cleanup path, improving maintainability and reducing edge‑case bugs  

Fixed
- fixed a persistence bug in `Status` where the `steps` property could not be restored during save/load because it lacked a setter
- added a setter for `steps` so that status data can be correctly deserialized without raising an exception

Notes
- all existing tests were updated or extended to cover the new item‑status behavior
- new tests ensure that item‑granted statuses are applied, removed, and persisted correctly
- no changes were made to the core combat status logic beyond enabling item‑based status application

### DIFF
--- a/tests/tuxemon/test_evolution.py
+++ b/tests/tuxemon/test_evolution.py
@@ -409,7 +409,7 @@ def test_held_item_conditions(
 ):
     mon, _, _ = setup_evolution
     if held_item_slug is not None:
-        item = MagicMock(slug=held_item_slug)
+        item = MagicMock(slug=held_item_slug, granted_statuses=[])
         mon.equip_item(item)
     evo = MonsterEvolutionItemModel(
         monster_slug="rockat", held_item=evo_item_slug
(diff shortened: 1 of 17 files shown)


HippasusTwo:
Thank you.

la-xirsoi (approved):

PR #3408, opened 2026-02-02
Add tuple comparison support to `TimeIsCondition` + Fix `heal`

JaskRendix:
PR improves the handling of date comparisons in `TimeIsCondition` by replacing string‑based equality checks and ad‑hoc logic with a dedicated tuple‑comparison path. A new `compare_tuple` utility function has been added to provide consistent operator handling for tuple values, mirroring the existing compare function used for numeric comparisons.


tuxemon/core/effects/heal.py
la-xirsoi, line 1: Could you explain this fix?

JaskRendix: Fixed heals were accidentally being calculated as percentages. This change makes fixed heals heal a fixed number and percentage heals heal a percentage, like they were meant to.

JaskRendix: I also removed the `process_event` method from `StartState`, we can just set `self.escape_key_exits = False` instead.

la-xirsoi (approved):

Issue #11, opened 2015-06-06
Implement creature breeding/fusion system


ShadowApex:
The preliminary graphical code has been implemented in core.components.fusion, but we will need to add this as a game mechanic. Players will be able to breed two creatures together that will create a new creature with sprite parts, stats, types, and moves from its parents.

Sanglorian:
More discussion on the gameplay side of this on the forums: https://forum.tuxemon.org/thread.php?pid=2282#p2282

PR #3450, opened 2026-02-25
Improve breeding action with character support and monster filtering + `battle_menu.yaml` change

JaskRendix:
PR enhances the breeding system by allowing the breeding action to target specific characters, such as NPCs or the player, rather than being hardcoded to a single source. It introduces a refined filtering system that automatically restricts the selection menu.

Replaced the **maple_wren** NPC in `battle_menu.yaml` with professor, since professor is a neutral NPC not tied to the Xero campaign. Fixed base.po

This PR also updates the random monster selection logic. The `RandomMonsterAction` now excludes monsters that should already have evolved at the given level, as well as forms that are not yet valid for that level. Two helper methods were added to `MonsterModel` to support this: one checks whether a monster would evolve by a specific level, and the other verifies that the monster’s current form is appropriate for that level.

Renamed test suite into `test_event_action_random_monster`


wiki-me:
Does this that the breeding system is implemented and this [issue](https://github.com/Tuxemon/Tuxemon/issues/11) can be closed? (the last issue for the 0.5 [milestone](https://github.com/Tuxemon/Tuxemon/milestone/3)).

JaskRendix:
@wiki-me Breeding is already implemented. What isn’t implemented yet is fusion, so the issue has been moved to the 0.6 milestone instead of being closed.

la-xirsoi (approved):

PR #3463, opened 2026-03-02
Replace manual intro sprite loading with `AnimationManager`‑based intro animation

JaskRendix:
PR removes the filesystem‑driven intro animation loader and replaces it with the unified `AnimationManager` system. The intro state now requests the `intro` animation by slug, adds it to the sprite group, and updates it through the standard animation update path. Music playback remains unchanged. If the animation cannot be loaded, the state falls back to `StartState`. The update loop is extended to advance animation frames.


la-xirsoi:
This deletes all of the files used in the intro animation but doesn't replace them with anything. Is that correct?

JaskRendix:
@la-xirsoi the new sheet is already uploaded, I was planning to handle it. The file is at `mods/tuxemon/animations/intro/intro.png`, which is the sprite sheet.

la-xirsoi (approved):

PR #3478, opened 2026-03-12
Add legacy campaign "Waves of Chaos"

JaskRendix:
PR adds an older campaign created by @ultidonki. The campaign has been modernized but is not fully tested. Most of the content is intact and playable. Several parts still need updates, mainly map data and references, but the core campaign flow works from start to finish. Some msgids have been removed because the related NPCs no longer exist.


la-xirsoi (approved):

ultidonki:
Thank you for this - I wasn't expecting it!! I'll try and get some playing, maybe fixing some bugs in the future to see if it all works again. :smile:

PR #3510, opened 2026-04-18
Purge legacy Aether element from element‑assignment logic and data structures

### LINKED ISSUE #3508: Orbs and other "element giving techniques" can give the Aether element
**Describe the bug**
Orbs gives the target a random element. This can include "Aether", which is left over from when we had five elements + Aether. We now have thirteen including Normal. 

**To Reproduce**
Unfortunately it's a bit manual, but if you keep using Orbs you should eventually see it. 

**Suggested fix**
Orbs to choose from the set of 13 types.

JaskRendix:
PR removes the obsolete **Aether** element from the codebase. The element was a remnant from the old five‑elements‑plus‑Aether model and is not part of the current set of thirteen elements. Its presence allowed Orbs and other element‑granting mechanics to assign an invalid type. This fixes #3508.


Sanglorian:
Brilliant, thanks!

la-xirsoi (approved):

PR #3511, opened 2026-04-18
Fix Music Transitions, Add Explicit Mute State, and Add Audio Test Coverage

### LINKED ISSUE #3509: Sound cuts out, instead of fading in and out
**Describe the bug**
When changing room or map, or going from map to battle, the original music stops playing immediately and the new music starts. 

Ideally, the old music would keep playing on the new map, fading out, and then the new music would fade in.  

Arguably this comes down to personal preference rather than a bug, but I thought it might qualify as it is listed as fixed in issue #502 https://github.com/Tuxemon/Tuxemon/issues/502

### LINKED ISSUE #3515: Having music muted in Options doesn't mute battle music
**Describe the bug**
Muting music successfully mutes the overworld soundtrack, but doesn't affect the music that plays during battle. 

**Expected behavior**
Battle music should play at the same volume as overworld music.

JaskRendix:
PR restores the expected transition behavior when changing rooms, maps, or entering battle. If a previous track is playing, it now fades out before the new track begins. This fixes #3509

PR adds an explicit mute flag to both the music and sound systems. Music and SFX no longer rely on volume‑0 checks to determine mute state.

Battle music previously ignored the mute setting because mute was inferred from mixer volume. With an explicit mute flag and mute‑aware playback, all music—including battle tracks—now follows the user’s mute setting. This fixes #3515

The Options menu now exposes a single toggle that switches between *Mute Music* and *Unmute Music*, based on the actual mute state rather than mixer volume.

Sound effects now use the same model as music: explicit mute state, logical volume tracking, and consistent application of volume across cached sounds.

PR also adds pytest coverage for both `MusicPlayerState` and `SoundManager`, which were previously untested.

### DIFF
--- a/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
+++ b/mods/tuxemon/l18n/en_US/LC_MESSAGES/base.po
@@ -4787,6 +4787,9 @@ msgstr "Enable"
 msgid "disable"
 msgstr "Disable"
 
+msgid "menu_unmute_music"
+msgstr "Unmute Music"
+
 msgid "menu_mute_music"
 msgstr "Mute Music"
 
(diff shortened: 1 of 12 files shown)


Sanglorian:
Thank you! Just tested this and it's working great.

Sanglorian:
Just tested the mute options and they're working great too -- thanks @JaskRendix !

la-xirsoi (approved):

PR #3512, opened 2026-04-18
Fix XP Overflow, Level‑Up Ordering, and HUD Flow + Refactor CombatState Internals

### LINKED ISSUE #3507: After levelling up, excess XP is lost -- it should carry over
**Describe the bug**
When a monster goes up a level, it doesn't receive any remaining XP. 

For example, a level 9 monster with 50 XP to go until level 10 that receives 75 XP should be left with 25 XP towards level 11. Instead, they start level 10 with 0 XP. 

**Expected behavior**
Overflow XP should apply towards the next level. 

 - OS: Windows

### LINKED ISSUE #3516: Level up stat increase info block obscures dialogue, only shows once even if multiple monsters level up
**Describe the bug**
On level up, a box pops up showing how the monster's stats changed. 

This is really useful! 

But currently it obscures the dialogue box below, and if multiple monsters level up only one displays. Clicking continue clears the stat pop up and the dialogue box. 

**To Reproduce**
1. Fight a battle swapping between two mons, both of which level up

**Expected behavior**
Clicking continue should clear one monster level up block, not all of them. 

Either clicking continue should not clear the dialogue box as well, or the level up block should be higher so it doesn't obscure the dialogue box (or both). 

_Example of dialogue box obscured:_
Image

JaskRendix:
PR fixes the bug where monsters lose leftover XP when leveling up. The XP model already kept the correct `total_experience`, but the reward and UI flow discarded it. The patch exposes the updated XP in `RewardDataEntry` and updates the EXP‑bar animation to use the monster’s real `experience_progress_percent` instead of forcing a reset. Overflow XP now carries into the next level as expected.

Before labeling an issue as mandatory for a release, we need to understand its implications. The XP logic is fixed, but the animation behavior ties into a larger UI refactor, so that part shouldn’t be treated as a release‑blocking requirement. This fixes #3507.

Added a flag to stop the menu from covering the battle UI.
Screenshot_2026-02-03_16-52-55

 The old flow mixed HUD updates with level‑up summary handling, which caused double consumption of summaries, skipped screens, and overlapping XP animations. This refactor splits the responsibilities: `update_hud_and_level_up` now handles only the visual updates for the active monster, while `award_experience_and_money` queues all level‑up summaries in one place. The result is predictable HUD behavior and correct sequencing of every summary screen. This fixes #3516 

PR updates `CombatState` with two internal changes that simplify the code and remove overhead.  The `_decision_queue` now uses a `deque` instead of a list, which gives constant‑time pops from the left and keeps the FIFO behavior intact. Event handler registration is now defined in a single `EVENT_HANDLERS` dict, which makes the mapping between events and methods clear and easier to follow. These changes do not affect combat behavior, and testing shows no regressions.

### DIFF
--- a/tuxemon/combat/reward_system.py
+++ b/tuxemon/combat/reward_system.py
@@ -27,6 +27,7 @@ class RewardDataEntry:
     experience: int
     levels_gained: int = 0
     bond_milestones_crossed: set[int] = field(default_factory=set)
+    total_experience: int = 0
 
 
 @dataclass
@@ -131,6 +132,7 @@ class RewardCalculator:
             experience=awarded_exp,
             levels_gained=levels,
             bond_milestones_crossed=crossed,
+            total_experience=winner.total_experience,
         )
 
     def update_moves_and_messages(
--- a/tuxemon/states/combat_animations.py
+++ b/tuxemon/states/combat_animations.py
@@ -90,6 +90,7 @@ class CombatAnimations(Menu[None], ABC):
         )
         self.background_sprite: Sprite | None = None
         self.monsters_just_leveled_up: dict[str, bool] = {}
+        self.monsters_leftover_xp: dict[str, float] = {}
         env = self.client.environment_manager.get_active_environment()
         if env is None:
             raise RuntimeError(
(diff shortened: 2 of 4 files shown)


Sanglorian:
Thanks for this -- and it's a good point about understanding what we mark as mandatory. With my next bug reports, I won't link them to release 0.5 and leave that for you and @la-xirsoi to decide. 

For this one, I couldn't get the XP overflow to work -- I fought Roddick on Route 2, with two monsters levelling up -- both of them showed as 0 XP afterwards.

JaskRendix:
@Sanglorian this refactor should sort it out. The old flow was eating the level‑up summaries in two spots, which could wipe the leftover XP and make everything show 0. Now the summaries run in one place and the HUD stuff is separate, so overflow XP should stick. My earlier fix in `animate_exp` only worked when one monster leveled up; if two popped at once it could still break. With this flow change it should finally behave.

mods/tuxemon/db/technique/shuriken.yaml
la-xirsoi, line 1: This change seems unrelated to the fix, is it intentional?
You don't have to remove it, just want to make sure it's supposed to go in.

JaskRendix: Yeah, that one’s on me, I tweaked it while testing and forgot to revert.

la-xirsoi (approved):

PR #3513, opened 2026-04-18
Fluff Up animation plays over user

Sanglorian:
Pedantic little change from me - the technique Fluff Up is the user puffing up, but animation plays over the target. This fixes that up.


JaskRendix:
I think you need to pull the latest changes — `fluff_up` is a YAML file now, not JSON anymore.

Sanglorian:
> I think you need to pull the latest changes — `fluff_up` is a YAML file now, not JSON anymore.

My bad, thank you!

PR #3539, opened 2026-05-11
May 2026 Technique changes

Sanglorian:
Includes the previous May PR, plus 
* Surf available to any monster with the Water tag, matching the wiki
* Charmed condition shows a message when it's Charmed that caused an attack to miss, rather than a miss from another cause
* TMs and MMs allow a mon to learn a technique that matches their tags, not just if it's not on their standard learn list
* "+" and "-" signs show next to the stat affected by a monster's tastes
* Cathedral only tells you "All done" after they've used the healing machine


Sanglorian:
Sorry this was part of May-3, which has been merged -- so can close this now.

PR #3541, opened 2026-05-14
fix infinite battle in cafe bug

### LINKED ISSUE #3524: When losing a battle in the Cafe, the battle immediately restarted with healed mons
**Describe the bug**
After healing at the Cotton Cafe but never having gone to the Cotton Cathedral, I started a battle with the trainer in the Cafe. 

I lost the battle, but instead of blacking out and reappearing by the bartender at the Cafe, the battle immediately restarted with my monster healed to full health. 

I'm not sure if the problem is that I'd never visited a Cathedral, or if it was that having lost a battle in the Cafe, I would then reappear in the same Cafe -- and therefore immediately into the same battle. 

 - OS: Windows

HippasusTwo:
adding a single line (unsetting a variable for if you want to battle once you start the battle) fixes the bug outlined here:
fix https://github.com/Tuxemon/Tuxemon/issues/3524


JaskRendix:
just add “fix” in front of #3524 so it links

JaskRendix (approved):

la-xirsoi (approved):

PR #3545, opened 2026-05-15
Fix TextArea not updating when two consecutive dialog lines are identical

### LINKED ISSUE #3528: Combat text gets "stuck" when I swap a monster in
**Describe the bug**

I swapped my Pairagrin for my Agnite. The text "Pairagrin used Whirlwind" got "stuck", still displaying even when I opened up the Fight menu to select my Agnite's technique. 

It might matter that the monster that I swapped out was another Pairagrin -- could there be confusion in the game engine if it's expecting two Pairagrin and there's only one in play? 

Image

EDIT: Sorry as I kept playing I realised it goes beyond this -- when completing the battle, the defeated NPC's dialogue box is empty and there's no way to clear it and keep playing. 

Image

EDIT AGAIN: I've confirmed it's a bug that happens when two monsters with the same name are in battle with each other. 

**To Reproduce**
Not sure, as it doesn't usually happen when I swap one monster for another.

### LINKED ISSUE #3542: Getting Stuck in the Xero Campaign
**Describe the bug**
After the sequence where Aeble announces Omnichannel is going to control Tuxemon. The player parents tell the them to buy potions from the mart before curfew. I then leave Pata Town to the next route and eventually get stopped by an NPC, who tells me it's past curfew and then just says ". . . . . . ." and I cannot proceed the dialogue.

**Expected behavior**
The dialogue would box would end, and I would be able to move again to go to the correct location.

**Screenshots**
Image

 - OS: Windows 10
 - pygame-ce 2.5.7 (SDL 2.32.10, Python 3.14.3)

JaskRendix:
This change removes an early‑return check in `TextArea.text` that skipped updates when the new text matched the previous text. That check caused dialog to break whenever two consecutive lines were identical. The second line never triggered a redraw or animation, so the dialog system treated it as already complete and advanced again. Removing the check forces a proper reset of the text animation every time a new line is shown, even if the content is the same.

fix #3528
fix #3542

### DIFF
--- a/tests/tuxemon/test_text_area.py
+++ b/tests/tuxemon/test_text_area.py
@@ -132,11 +132,11 @@ def test_next_raises_stopiteration_when_not_animated(text_area):
         next(text_area)
 
 
-def test_text_setter_same_value_does_not_restart_animation(text_area):
+def test_text_setter_same_value_restarts_animation(text_area):
     text_area.text = "abc"
     text_area.drawing_text = False
     text_area.text = "abc"
-    assert text_area.drawing_text is False
+    assert text_area.drawing_text is True
 
 
 def test_empty_string(stress_text_area):
--- a/tuxemon/ui/text.py
+++ b/tuxemon/ui/text.py
@@ -147,8 +147,7 @@ class TextArea(Sprite):
 
     @text.setter
     def text(self, value: str) -> None:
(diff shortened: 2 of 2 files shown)


Sanglorian:
Thank you!

la-xirsoi (approved):

PR #3548, opened 2026-05-16
Third May Omnibus PR

Sanglorian:
Minor changes to: 
* Which target a particular technique's animation plays over
* Getting rid of the "Animal" family of techniques as it's too close to Basic to make a distinction between.

---

Doing work on the Tuxepedia app -- to be "pixel perfect" to the 256x144 grid, and to follow Wikipedia's visual design, and adding a few features. More to come on this in the number 4 update for May. 

Here's a preview: 
image

image

image

Not done yet, and of course welcome any thoughts and feedback! 

---

When monsters learn a new move, they now see the technique info screen (like they can access for a monster through the party screen). This gives more info on the new move, helping make a decision about which to keep. There's also a bit of text feedback of which move was remembered and which forgotten. 

---

Finally, addressing a bug where in the technique information menu, the old technique's type symbol would remain even when a new technique had been loaded. This "hide" method fixes that, but open to other solutions of course.


HippasusTwo:
I like the black and white tuxepedia design. If I were to make any changes, I'd make the little side bar (the one saying 'featured, missing') a light blue or some other pale color like silver or brown to provide some slight contrast.

Sanglorian:
> I like the black and white tuxepedia design. If I were to make any changes, I'd make the little side bar (the one saying 'featured, missing') a light blue or some other pale color like silver or brown to provide some slight contrast.

That's a good shout -- maybe the blue-grey from the mock-up?
image

HippasusTwo:
Yeah, that works.

la-xirsoi (approved):

PR #3560, opened 2026-05-21
Allow any monster to use the Daycare

Sanglorian:
Only fully evolved monsters can breed -- but the blocker came a bit too early, and stopped any baby monster from being added to the Daycare at all. This moves that check to the breeding section.


Sanglorian:
Have more a few more changes: 

* In the overworld, moving a roof tile to a higher layer so the wandering Rockitten and Conileaf go under it not over it.
* Babies inherit the better of their parents' IVs -- so it gets easier to breed a "perfect" IV monster as you go along. 
* Bonus XP for being in the Daycare was being rounded down so it never accumulated -- this should fix that.
* Rather than inherit a technique from their father, newborns inherit a move from their non-seed parent -- i.e. they always get a technique from the parent that *doesn't* contribute their body.
* Daycare Centre closes when you withdraw monsters, so the prompt to choose which techniques to keep (if your mon has more than 4) activates as it should.

JaskRendix (approved):

la-xirsoi (approved):

PR #3551, opened 2026-05-17
Including stat increases and conditions in stat calculations in combat

Sanglorian:
Waiting on #3550 

Changes by Claude to make sure stat increases and conditions are included in the stat totals used in combat, instead of just using base stats. I hope this is right -- I don't have the programming skill to say for myself, sorry!

PR #3562, opened 2026-05-23
5th May Omnibus

Sanglorian:
Downstream of #3551 

# Aesthetic

* Changing Frost type symbol to something that fits the style of other types
* Rename Gladiatorbug to Dynastor
* When a condition blocks another condition, let the player know in dialogue ("Focused prevents Rockitten from gaining Slowed")
* Adjusting some dialogue spacing.

# Mechanical

* Boost items use the stat_increase system @JaskRendix programmed
* Damaging techniques apply damage, then apply conditions (if any)
* Players who surf dive into the water if they can.
* Remove bug where an NPC trying to pathfind to an adjacent player would end up facing downward instead of continuing to face the player.
* On level up, keep overflow XP instead of resetting to 0.
* Very heavy monsters are reported in tonnes; very tall in metres.
* Display stat changes in the standard order (HP, Armor, Dodge, Melee, Ranged, Speed)
* Show player icon on the map.

# New features

* Torchlight overlay for times when vision is obscured -- can be expanded to any other png overlay.
* Allow players to set the screen size (experimental -- will remove this in a later PR if I cannot get it to work)
* Technique use logged in console for debugging purposes

# Spyder campaign

* Remove Contacts App and  Bank App from Spyder campaign until I can do them justice
* Moving Swim encounters to the map level so other waterways can have different encounter sets.
* Characters face trainers who ambush them.
* Have added 5 levels to every trainer's monster from Nimrod on -- otherwise, battles too easy.
* Expand Eliane's giveaways to more items and fix up some dialogue overlaps/misses.
* Set wandering at different rates for different NPCs, so it doesn't look like they're moving in unison.
* Cotton Tunnel entrances/exits more intuitive in terms of collision zones.
* Removing Nurse from Dragon's Cave as Benden heals player.
* Greenwash encounters use interior battle terrain
* Reposition Scoop end battles to avoid crash


JaskRendix:
the renaming of gladiatorbug, add it here too
`tuxemon/save_system/save_upgrader.py`
here
`MONSTER_RENAMES: dict[str, str] = {"axylightl": "axolightl", "gladiatorbug": "dynastor"}  # old: new`
so if someone has gladiatorbug in the party, then it doesn't crash

Sanglorian:
Good catch, thanks @JaskRendix

la-xirsoi:
@Sanglorian you've got some merge conflicts here

Sanglorian:
Thanks @la-xirsoi , should be fixed now!

JaskRendix (approved):

la-xirsoi (approved):

PR #3563, opened 2026-05-23
Refining double battles, particularly target selection

Sanglorian:
Follows from #3562 

Improves the targeting system, including moving targeting within effects instead of having it set for the technique as a whole. This allows for a technique that heals the user while damaging an enemy monster.

Also removes the "bypasses_selection" feature of techniques, which I believe is no longer necessary as we have a targeting choice system now.

Also, creates dedicated HUD backgrounds with their own positioning for double battles, so there's more room for monster sprites. Some positioning tweaks to offset one monster from another.

Wording timing changes so the player is asked to choose techniques for one monster, then for the next.


JaskRendix (approved):

la-xirsoi (dismissed):

Sanglorian:
Thank you @la-xirsoi -- have merged development into this so should be good to merge now if you're still happy with it.

la-xirsoi (dismissed):

JaskRendix (approved):

la-xirsoi (approved):

PR #3565, opened 2026-05-23
Interactive Town Map

Sanglorian:
* Shows player location on map
* Map expanded to include routes and side locations
* Cursor more closely tracks the map itself

How the map looks now: 

image


tuxemon/states/phone_map.py
```diff
@@ -23,11 +27,58 @@
...
+}
+
+# Core locations are always navigable; non-core require a tracker visit.
+_CORE: frozenset[str] = frozenset({
+    "leather_town", "cotton_town", "paper_town",
+    "candy_town",   "timber_town", "flower_city",
+    "route1", "route2", "route3", "route4", "route5", "route6",
+    "tunnel", "citypark",
+})
+
```
JaskRendix: if you can move _CORE and _NAV inside the **mods/nu_phone_map.yaml**, then it would be great

Sanglorian: For sure -- have done that now.

Thanks for taking a look!

JaskRendix (approved):

la-xirsoi (approved):

PR #3569, opened 2026-05-24
Recovering battle information from save data

Sanglorian:
When in Dragon's Cave I got this error:

> Error checking condition 'type='battle_outcome_count' parameters=['player', 'won', 'spyder_nimrod_tru', '2'] name='cond20' operator=<Operator.NOT: 'not'> box=BoundingBox(x=10, y=6, width=1, height=1)': 'str' object has no attribute 'value'
> Error checking condition 'type='battle_outcome_count' parameters=['player', 'won', 'spyder_nimrod_tru', '1'] name='cond10' operator=<Operator.IS: 'is'> box=BoundingBox(x=10, y=4, width=1, height=1)'

As I understand it, the problem was with the form in which battle save data was loaded. This didn't matter for yes/no, but it did matter for a count (defeated Tru never, once or twice). 

I believe this fixes it -- I can now trigger the Tru encounter in Dragon's Cave.


JaskRendix:
drop the new `get_state()` into `tuxemon/battle.py`
add those tests into `tests/tuxemon/test_battles_handler.py`

```python
def get_state(self) -> Mapping[str, Any]:
    save_data = {}

    for attr in SIMPLE_PERSISTANCE_ATTRIBUTES:
        value = getattr(self, attr)
        if not value:
            continue

        if attr == "outcome":
            save_data[attr] = value.value
        else:
            save_data[attr] = value

    save_data["instance_id"] = self.instance_id.hex
    return save_data
```

new tests

```python
def test_get_state_serializes_enum():
    battle = Battle()
    battle.outcome = OutputBattle.WON

    state = battle.get_state()

    assert state["outcome"] == OutputBattle.WON.value
    assert not isinstance(state["outcome"], OutputBattle)


def test_battle_round_trip_enum():
    battle = Battle()
    battle.outcome = OutputBattle.LOST

    encoded = battle.get_state()
    decoded = Battle.from_save_data(encoded)

    assert decoded.outcome == OutputBattle.LOST


def test_from_save_data_invalid_enum():
    battle = Battle.from_save_data({"outcome": "not_a_valid_enum"})

    # Should fall back to constructor default (DRAW)
    assert battle.outcome == OutputBattle.DRAW


def test_encode_decode_with_enum_object():
    battle = Battle()
    battle.outcome = OutputBattle.WON

    encoded = encode_battle([battle])
    decoded = decode_battle(encoded)

    assert decoded[0].outcome == OutputBattle.WON
```

Sanglorian:
Thank you @JaskRendix ! How does this look?

JaskRendix (approved):

JaskRendix:
it looks good!

la-xirsoi (approved):

PR #3587, opened 2026-06-12
When an attack targets multiple enemies, it's less damaging

Sanglorian:
It's hard to balance techniques that hit multiple enemies, because in a single battle they're no more powerful than a technique that damages one enemy, and in a double battle they're potentially twice as powerful.

Other games get around this by reducing the damage done by multi-target techniques, only in a double battle. I think that's a good idea -- here, the multiplier is 75%. But I've only added this as a draft to see what others think.


JaskRendix (approved):

la-xirsoi (approved):
a perfectly reasonable balance measure.
another idea, if this still makes them too powerful, is to just split the (base) damage evenly across the targets. (applying the relevant type-modifiers individually, of course)

PR #3595, opened 2026-06-19
Saves/Loads on second page crash

Sanglorian:
When I added #3553 I introduced a bug, sorry!

Load a game that's on the second page of saves (saves 4 through 6). 

Go into the save or load menu. It should start on save 4, 5 or 6 depending on what you have open. But instead it'll start you on save 1, and when you move up or down it crashes. 

These changes should fix that.


la-xirsoi (approved):

Sanglorian:
Thank you @la-xirsoi -- great to get these all merged in.

